PR #14, opened 2013-06-29
Allow using markdown files to be used for detailed descriptions of classes and groups.

API-Beast:
This was possible before, however every markdown file also created a static page which kinda defeated the purpose.


doxygen:
Look like this has been implemented in the meantime, in a more robust way, in [this commit](https://github.com/doxygen/doxygen/commit/2d705fca8ba4047c30f514a6f32806ad761dd269)

PR #18, opened 2013-08-12
add config option NO_WARN_MISSING_PARAMDOC

wilcoxjay:
If \c NO_WARN_MISSING_PARAMS is set to YES, doxygen will not print warnings if only some parameters are documented. If set to the default value NO, those warnings will be printed.


albert-github:
As a reimplementation has been done in #8362 I think this pull request is obsolete.

PR #142, opened 2014-03-18
Implement "double-space line breaks" syntax in Markdown

randomguy3:
Ending a line with two spaces is supposed to create a line break in
Markdown.  This implements that syntax.


BrainStone:
Not a big change and conforms to Markdown. Would be lovely to have it merged!

claudiofantacci:
I've also recently filed a bug report [789354](https://bugzilla.gnome.org/show_bug.cgi?id=789354) about this.
Why hasn't this PR merged after over 3 years?

PR #198, opened 2014-07-13
Bug 645165 - Last member of a group is not documented with "documentation after member" feature

albert-github:
Due to the limited possibilities to detect continuation lines the commands \{ and \} have been documented together with a note about this limitation.


albert-github:
Looks like the problem has been solved in the mean time, so this PR is obsolete.

PR #241, opened 2014-10-15
Support PlantUML in man pages

albert-github:
With this patch it is also possible to use the PlantUML features in man pages. The output is simple ASCII text which is read into the man pages as if it were verbatim text.


AnneTheAgile:
@doxygen , so you mean that with your extra commit the combined patch of yours plus @albert-github 's  could be accepted?
I love PlantUML, I haven't used it in a while so ty for the reminder :).
AnneTheAgile
ps also thank you for migrating to Github! It makes it so easy to work with you.

PR #243, opened 2014-10-19
Support dotfile files with same names from different directories

albert-github:
In case `\dotfile` is used with the name name but from a different directory the result is only 1 output file. This is corrected with this patch.
In the html version the warning in respect to systems without svg lead to problems (`<p> tag inside <p>` tag), this has been made consistent between `\dotfile` and `\dot ...\enddot`


albert-github:
Looks like related to:
Bug 742707 - Doxygen gets confused when two copies of dot file exist

albert-github:
Obsolete by  https://github.com/doxygen/doxygen/pull/11912

PR #244, opened 2014-10-19
Support mscfile files with same names from different directories

albert-github:
In case \mscfile is used with the name name but from a different directory the result is only 1 output file. This is corrected with this patch.
In case the mscgen command failed no error message was given, this is corrected as well.


albert-github:
Obsolete by  https://github.com/doxygen/doxygen/pull/11912

PR #250, opened 2014-11-27
Make number of Rerun loops to generate targets latex directory more flexible

albert-github:
Till now it was only possible to change the Makefile or the make.bat in case the default number of Rerun-loops was insufficient. With this patch the number of loops can also be specified from outside (e.g. make 8 when using make.bat or make latex_count=10  when using Makefile. Advantage is that this value can easier be applied in scripts / other Makefiles.


albert-github:
This proposed pull request is superseded by #9004

PR #274, opened 2015-01-01
Documenting RESULT variable of Fortran FUNCTION

albert-github:
Fortran functions can specify the return value by means of a variable in the RESULT attribute. Analogous to argument variables (@param) it is now possible to document also the return / RESULT variable (@returns).


albert-github:
Suggestion has been implemented.

PR #290, opened 2015-01-11
Bug 739680 - Using HTML entities in PROJECT_NAME

albert-github:
Implementation for LaTeX of special symbols by scanning string in case a & is found.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -6483,7 +6483,7 @@ void filterLatexString(FTextStream &t,const char *str,
         case '&':  // possibility to have a special symbol
                    q = p;
                    cnt = 2; // we have to count & and ; as well
-                   while (*q && *q != ';')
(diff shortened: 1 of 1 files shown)


albert-github:
In principle there is nothing against limiting the search part. Looking only for a-z and A-Z is insufficient, a number of entities contain numbers (e.g. sup2). An alternative would also be to limit the search to e.g. 10 chars.

albert-github:
Would a construct like: 
  while (*q && *q != ';' && ((*q >= 'a' && *q <= 'z') || (*q >= 'A' && *q <= 'Z') || (*q >= '0' && *q <= '9')))
be better than the current proposed:
  while (*q && *q != ';')
Disadvantage of the new construct is we have to do some test for each character.
I can commit this without problems.

doxygen:
Yes, I prefer the character check over the "process until ; or end of string" approach.

albert-github:
Updated as suggested.

PR #294, opened 2015-01-23
Bug 743358 - Wrong inheritance diagrams in html if INLINE_GROUPED_CLASSES is set to YES

albert-github:
Changed fileName to name as the fileName is identical in case of INLINE_GROUPED_CLASSES and thus giving the same image (last) for all occurrences on the page.


albert-github:
Updated patch to replace name by an escaped name for writing to disk.

PR #295, opened 2015-01-29
Add support for emphasising non-ASCII alphanums

grisumbras:
Fixes [Bug 743675 ](https://bugzilla.gnome.org/show_bug.cgi?id=743675).


doxygen:
I've included a simplified version of your patch

PR #298, opened 2015-02-11
hack: work around issue 737472

pabigot:
While I suspect this isn't the correct fix (which requires some sort of configuration option), [bug 737472](https://bugzilla.gnome.org/show_bug.cgi?id=737472) is a fairly big problem for the way I use Doxygen.  The issue report doesn't seem to get much visibility where it is so I'm hoping this might elevate its priority (and provide a workaround until the right fix can be developed).


pabigot:
Closed by #511

PR #308, opened 2015-03-09
runtests.pl: mnn version has dash as separator

groleo:


doxygen:
Numbering format can also be 1.8.9.1-yyy (so with 3 dots!). This is why your fix now breaks the CI build.

groleo:
That's strange. version.py writes versionString[]
either as
  major.minor.revision (src/version.py:56)
or as
  major.minor.revision-mmn (src/version.py:59)
I've checked by running runtest.pl

groleo:
Oh. I'll update the patch then.

PR #326, opened 2015-03-30
add support for github flavored fenced code blocks

halex2005:
Hi!
Here is PR with support for Gihub flavored Fenced Code Blocks with three backquotes characters.
I recently found that this syntax is not supported in doxygen, but all my markdown contains this kind of syntax.
The change is little: in function isFencedCodeBlock() we remember backqoute if fenced code block contains it and then reuse it in comparisons.
Also I changed pre.l and scanner.l to support this kind of fenced code block (I just copied stuff from ~~~ and changed it to ```).

I tested this functionality with compiling on Visual Studio 2008 and it works.

Will you accept this PR?
Thank you for great tool, best regards, Aleksei.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -1366,7 +1366,9 @@ static bool isFencedCodeBlock(const char *data,int size,int refIndent,
   int startTildes=0;
   while (i<size && data[i]==' ') indent++,i++;
   if (indent>=refIndent+4) return FALSE; // part of code block
-  while (i<size && data[i]=='~') startTildes++,i++;
+  char tildaChar='~';
+  if (i<size && data[i]=='`') tildaChar='`';
+  while (i<size && data[i]==tildaChar) startTildes++,i++;
   if (startTildes<3) return FALSE; // not enough tildes
   if (i<size && data[i]=='{') i++; // skip over optional {
   int startLang=i;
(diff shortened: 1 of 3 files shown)


albert-github:
Shouldn't the rules read:
\`\`\`
or
\`\`\`[\`]*
(see e.g. https://michelf.ca/projects/php-markdown/extra/#fenced-code-blocks)

Are there possibilities to specify the used (computer) language as well (see above mentioned reference and https://help.github.com/articles/github-flavored-markdown/)

Shouldn't this extension have some documentation as well?

halex2005:
Yes, I think you right about rules. I modified PR with ```[`]* rules.
About used language - yes, it's supported as for ~~~{.lang}. 
In original code { is ignored in function isFencedCodeBlock(). Extension point "." in language is ignored if exists in writeFencedCodeBlock() function.

This original code support also ~~~lang sytax. So, with backticks in place language also supported.

halex2005:
Should I also modify documentation for fenced code blocks too?

albert-github:
In case there are errors or shortcomings in the original text it is appreciated that this is corrected with a pull request as well.

halex2005:
Should I do something else before this PR will be accepted?

albert-github:
Some documentation in the manual would be useful.

PR #333, opened 2015-04-20
Fix for bug 748208 regression in parsing enums in bitfields

Kasoo:
Fixes [#748208](https://bugzilla.gnome.org/show_bug.cgi?id=748208) which is a regression of [#313527](https://bugzilla.gnome.org/show_bug.cgi?id=313527).

Enums within bitfields are not parsed properly. See bugzilla for test cases and failure examples.


doxygen:
I've pushed a variation on your patch:
https://github.com/doxygen/doxygen/commit/7b2d77abb73288bdd48ca04720fd6779e0bf94c2

PR #338, opened 2015-05-17
Alignment of project name

albert-github:
With this bug fix it is possible to align the project name to another place than the default place "middle" by means of adjusting the style sheet. Until now it was also possible but one had to create an own html header file as well.


albert-github:
Wrong branch (pull request 337 is the correct branch for me)

PR #339, opened 2015-05-18
drop #include <unistd.h>

groleo:
the generated lexers are not interactive, so istty from unistd.h is thus not
needed.

Signed-off-by: Adrian Negreanu <adrian.m.negreanu@intel.com>


groleo:
already in master, as part of the cmake pr

PR #340, opened 2015-05-20
xml: use STRIP_FROM_PATH on @bodyfile attributes.

groleo:


doxygen:
This is already included.

PR #345, opened 2015-05-28
Add USE_READABLE_ANCHORS

alpha0010:
The anchors this option enables are (mostly) human readable.  These are useful to see where a link is going, especially if it is a bald link into the documentation, posted on an external site.  Also this anchor tends to be a bit shorter, which is visually nice.
In the situation that the structure of the documentation changes enough that both the previous md5 anchor and this new anchor are invalidated, the readable name gives a hint of what it had been pointing to, allowing for easier finding (aided by a search engine).

Note: this does add an aliasing warning due to the typecast; let me know if you would prefer I do something different there.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3436,6 +3436,14 @@ remove the intermediate dot files that are used to generate the various graphs.
 ]]>
       </docs>
     </option>
+    <option type='bool' id='USE_READABLE_ANCHORS' defval='0'>
+      <docs>
+<![CDATA[
+ If the \c USE_READABLE_ANCHORS tag is set to \c YES then anchor links (#)
+ will incorporate the method name instead of only the md5 hash.
+]]>
+      </docs>
+    </option>
     <option type='obsolete' id='USE_WINDOWS_ENCODING'/>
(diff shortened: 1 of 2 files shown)


alpha0010:
This is similar to #199, but avoids its issues with overloads.

alpha0010:
Example anchors:
- Specific enum value [#jb-CCProviderStatus:ccpsActive](http://alpha0010.github.io/cb-docs/classcbCodeCompletionPlugin#jb-CCProviderStatus:ccpsActive)
- Function [#pc-GetAutocompList](http://alpha0010.github.io/cb-docs/classcbCodeCompletionPlugin#pc-GetAutocompList)

bbbbbr:
This PR was useful to work from, though needed modifying to work with 1.9.2.
In our case the documentation language is C and we're not worried about overloading, so we're trying out straight function names for the links. It's really nice to have them readable.

Patch attached for anyone who might find it useful.
[doxygen_1.9.2_readable_link_names_patch.diff.zip](https://github.com/user-attachments/files/23764945/doxygen_1.9.2_readable_link_names_patch.diff.zip)

PR #347, opened 2015-06-04
.gitignore: remove entries which are no longer generated

groleo:
un-ignore files or directories we don't generate to
avoid using files from previous builds.

Signed-off-by: Adrian Negreanu <adrian.m.negreanu@intel.com>


albert-github:
Shouldn't /moc and /rcc not have the same treatment as generated_src? As they are generated by QT and thus should be in respective build directories.

groleo:
Yup. They're no longer created

PR #348, opened 2015-06-06
Port Doxygen frontend to Qt 5.

jakepetroules:
Mostly trivial find-and-replace operations. Will make a huge visual improvement especially on OS X due to improvements made to support UI changes in recent versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,9 @@ language: cpp
 compiler:
   - gcc
 before_install:
+    - sudo apt-add-repository -y ppa:ubuntu-sdk-team/ppa
     - sudo apt-get update -qq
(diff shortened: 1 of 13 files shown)


doxygen:
Would be nice to have support for both Qt4 and Qt5, so the result is more portable. I think this is possible.

jakepetroules:
What do you need to support that requires Qt 4?

doxygen:
There are enough Linux distributions that do not yet ship Qt5 (or a user may have opted to only install Qt4). I've just commited a change set that adds support for a Qt5 build of doxywizard with automatic (or forced via the force_qt4 option) fallback to Qt4.

PR #350, opened 2015-06-09
docs,build: specify cmake v3.0 as minimum required version.

groleo:


dirk-thomas:
Is CMake 3.0 really a strict requirement? I was able to build the current version from source using CMake 2.8.12.2 (after changing the version passed to `cmake_minimum_required`).

PR #353, opened 2015-06-11
Fresh "compile from source" instructions

AndreMiras:
This is the compilation instructions using cmake, inspired by Travis CI build as http://www.stack.nl/~dimitri/doxygen/manual/install.html didn't seem up to date.


doxygen:
I've updated the online installation and don't feel like repeating them in the INSTALL file. So I'm closing this pull request.

PR #365, opened 2015-07-12
Fixes bug 485701: Angle brackets (< and >) not escaped in HTML formula alt text

jcrada:
This pull request fixes bug reported in https://bugzilla.gnome.org/show_bug.cgi?id=485701


jcrada:
I have removed the comments, but perhaps should have chosen a more adequate message for git commit.

PR #377, opened 2015-08-02
Adds config options and support for running externa scripts on dot files before rendering them with graphwiz

chintal:
Auto-generated graphs tend to get somewhat messy because of common include files, which create a nest of edges in the graph which make it difficult to see the functionally useful dependency graph. One common example of this is stdint.h includes in embedded projects.

Such applications are probably highly specific, so adding it to the core of Doxygen is probably counterproductive. However, it would be nice to be able to process the doxygen generated .dot files before they are turned into images. Perhaps by means of some kind of plugin? If there were a hook to run a script on a file between the generation of .dot and .png or .svg, that would be sufficient in principle. It would be even better if the script could also somehow obtain contextual information about the graph, but that wouldn't be necessary to make such a hook useful for a large number of cases.

The changes proposed here add options to the config script, which is not a very nice thing. It forces
everyone to update their configuration files. I'm not sure how to get around that problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 .*sw?
 \#*
 .DS_Store
+.kdev4/*
+doxygen.kdev4
 
 *.rej
 *.orig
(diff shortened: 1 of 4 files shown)


doxygen:
Could you tidy up your pull request by combining the commits into a single commit? It is not so interesting to see the development history of this feature.

chintal:
I seem to have unintentionally closed the request in the process of tidying up the commits. I will make another pull request in a short while with the same changes (and hopefully in a cleaner form).

PR #381, opened 2015-08-12
Bug 646002 - htmlonly content appears in generated XML output

albert-github:
All the @\*only comments appear in the xml document. In the other document formats only documentation of the relevant @\*only (in HTML only @htmlonly) appears.
This patch corrects this so only the @xmlonly documentation appears.

### DIFF
--- a/src/xmldocvisitor.cpp
+++ b/src/xmldocvisitor.cpp
@@ -223,32 +223,14 @@ void XmlDocVisitor::visit(DocVerbatim *s)
       m_t << "</verbatim>"; 
       break;
     case DocVerbatim::HtmlOnly: 
-      m_t << "<htmlonly>";
-      filter(s->text());
-      m_t << "</htmlonly>";
-      break;
     case DocVerbatim::RtfOnly: 
-      m_t << "<rtfonly>";
-      filter(s->text());
-      m_t << "</rtfonly>";
-      break;
     case DocVerbatim::ManOnly: 
-      m_t << "<manonly>";
-      filter(s->text());
-      m_t << "</manonly>";
-      break;
     case DocVerbatim::LatexOnly: 
-      m_t << "<latexonly>";
-      filter(s->text());
-      m_t << "</latexonly>";
+    case DocVerbatim::DocbookOnly:
+      /* nothing */ 
       break;
     case DocVerbatim::XmlOnly: 
       m_t << s->text();
-      break;
-    case DocVerbatim::DocbookOnly:
-      m_t << "<docbookonly>";
-      filter(s->text());
-      m_t << "</docbookonly>";
       break;
     case DocVerbatim::Dot:
       visitPreStart(m_t, "dot", s->hasCaption(), this, s->children(), QCString(""), FALSE, DocImage::Html, s->width(), s->height());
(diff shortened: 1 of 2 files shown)


mosra:
@albert-github I just found this by accident -- do you remember what was the rationale for this? I'm treating the XML output as a source for creating HTML files, and thus having `@htmlonly` and others preserved in there makes sense.

The consumers of the XML output can easily ignore these tags, but with this in place there's no way to have raw HTML content embedded in the XML files.

albert-github:
Indeed the workaround with `\htmlinclude` is inconsistent compared to the use of `htmlonly` and actually should be corrected as well.
The documentation states that the `*only` parts are verbatim included in the corresponding output, it is not explicitly mentioned that in the other output formats it is left out (wouldn't make sense to have it there). The XML output is a bit of a "strange" as it is, more or less, not a direct output format that is directly displayed but is used for further processing (same probably, accounts for docbook as well when converting the docbook output to e.g. HTML) and one of the processed formats might be HTML (was a bit hard to formulate, it it the thought that is behind it).

I don't remember the rationale anymore, but seen the fact that the `\*only` parts were removed from the other formats as well I think that it was done because of consistency considerations.
I think there might also have been some thoughts about the validity of the produced code, but as the `\htmlinclude` file contents is just included with a `filter` call, I think it should be possible to include the `\htmlonly` as well.

When it should b done for `\htmlonly` it should also be done when including other `*only` in the XML output. Also the `\*include` should be checked. Special attention has to be paid at the `[block]` with the `\htmlonly`.

@doxygen should this be reverted?

mosra:
Not just in my case with [m.css](https://mcss.mosra.cz/documentation/doxygen/), but I think also Breathe (which feeds Doxygen XML into Sphinx) would benefit from this -- as Sphinx is then able to render [HTML, latex, manpages, ...](https://www.sphinx-doc.org/en/1.8/man/sphinx-build.html) and thus an ability to have e.g. "man only" content propagated to Sphinx's manpage generator.

For the other formats I can understand this as those are "final" -- the HTML output would never get converted to a PDF, for example, and thus keeping those there makes no sense -- but the XML is in all uses I'm aware of treated as an "intermediate", and thus it makes sense to preserve all information in it.

doxygen:
> @doxygen should this be reverted?

Yes, the XML output is meant as a "dump" of doxygen's internals for use by other tools. This makes it different from other output formats which are meant to be used directly, and it is why it makes sense to also include `\htmlonly` and `\htmlinclude` type of information in the XML output.

albert-github:
@doxygen you or me? (I have no problems doing it, but 2 people would be a bit much ...).

doxygen:
@albert-github if you have time please do. I’m currently setting up my new laptop, so it will take some time before I have everything up and running again.

albert-github:
OK I'll do it (probably today or tomorrow, but no promise).
I know the feeling about setting up a new system, I did it also before Christmas and still I find small issues / inconveniences ...

albert-github:
I've just pushed a proposed patch, pull request #7481, to revert this fix.

albert-github:
The pull request #7481 has been integrated in master on github.

PR #383, opened 2015-08-16
Make list of default extensions consistent with language mapping list

albert-github:
In the forum a discrepancy was noted between config.l and config.xml (http://doxygen.10944.n7.nabble.com/FILE-PATTERNS-one-custom-plus-all-defaults-td7308.html). The config.xml list was based on the function initDefaultExtensionMapping in util.cpp and the list in the function Config::check in config.l. Now both routines use the same list.

### DIFF
--- a/src/config.l
+++ b/src/config.l
@@ -1426,71 +1426,7 @@ void Config::check()
     }
   }
 
-  // add default pattern if needed
-  QStrList &filePatternList = Config_getList("FILE_PATTERNS");
-  if (filePatternList.isEmpty())
-  {
-    filePatternList.append("*.c");
-    filePatternList.append("*.cc"); 
-    filePatternList.append("*.cxx");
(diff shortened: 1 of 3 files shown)


clayt0nk:
Without this patch, doxygen core's capacity to process javascript is effectively concealed: the config file comments say that js is processed by default, but that is not what the code does. So anyone who tries to run doxygen against some javascript with default settings, ie. FILE_PATTERNS blank, will get zero results and assume that doxygen just does not work. (We did just that, and wasted considerable time testing plugins.) In fact, doxygen seems to work just fine on javascript, I just built this patch from source and tested against 100+ github Javascript. The result was lots of output. Please commit this patch, or something with similar effect.

WPSlacker:
I've been working with clayt0nk, and this 'config default setting' seems to be a subtle mistake with a large effect on Doxygen's results.  Please commit this patch.

PR #394, opened 2015-09-05
Compilation error Windows (cannot allocate an array of constant size 0)

albert-github:
On windows with Visual Studio 2013 the following error is given (3 times):
Error    14    error C2466: cannot allocate an array of constant size 0    ...\generated_src\resources.cpp    28198    1    _doxygen
In this patch the size of the array declaration for windows is set to 1 in case there are no elements


albert-github:
Pull request is obsolete as the templates contain now a placeholder September 8, e58fb0a46f73f37d16859f43fda1eb5ba4a15c5d)

PR #398, opened 2015-09-29
Problem with latex and bookmarks

albert-github:
Regression to the commit 02a0c353a8947290a3191aead59db08dc84766ce (Started with generating LateX output via the template engine) a lot of message like "removing `math shift' on input line" appeared.
The change also included a change (unintentionally ?) of the bookmarks in the treeview of the PDF document, so it shows the complete signature of a function.
The signature can also contain asterisks which are translated into $\ast\ resulting in the mentioned problem (see e.g. http://tex.stackexchange.com/questions/249107/how-to-use-texorpdfstring-characters-disappear) and the asterisks are not show either.
In principle this would be nice but has big disadvantages (long bookmarks) when there are a number of arguments with a function.

This patch reverts back to only showing the name of the function.

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -1567,7 +1567,7 @@ void LatexGenerator::startMemberDoc(const char *clname,
   t << "\\" << levelLab[level]; 
 
   t << "[{";
-  t << latexEscapeIndexChars(title,insideTabbing);
+  t << latexEscapeIndexChars(memname,insideTabbing);
   t << "}]";
   t << "{\\setlength{\\rightskip}{0pt plus 5cm}";
   disableLinks=TRUE;


doxygen:
I've tried to fix this issue in this commit using \texorpdfstring:
https://github.com/doxygen/doxygen/commit/499ecaedf1ee1222535be27f77050ee633595685
Please let me know if it works for you.

albert-github:
In my, small, example the special character does appear, so this problem is solved with the update 499ecae for this reason I'll close my pull request.
For me there are still some 'not nice things' (but I think this is another discussion):
The extension does make sense for overloaded functions as one can distinguish them from each other. Disadvantage is (especially with larger type names and / or  a number of arguments) the index will get cluttered and the overview will be come unreadable. When looking at the HTML output, in the TREEVIEW, the arguments are not added, so this is a bit inconsistent.

PR #406, opened 2015-10-28
Fix Windows compatibility

alex85k:
I was not able to build the sources on Visual Studio 2015 without these small modifications.
  - "somestring"SOME_MACRO causes the syntax error
  - bundled iconv seems to be incompatible with VS2015 and -DICONV_DIR was overridden

The final line in my buildscript looks like
```
cmake .. -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=%VARIANT% -DCMAKE_INSTALL_PREFIX=%PREFIX1% -DICONV_DIR=%PREFIX% -DICONV_LIBRARY=%PREFIX%/lib/iconv.lib
```

The building batch file is in my collection:
https://github.com/alex85k/winbuilds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,9 @@ if (${CMAKE_SYSTEM} MATCHES "Darwin")
 endif()
 
 if (WIN32)
-    set(ICONV_DIR "${CMAKE_SOURCE_DIR}/winbuild")
+    if(NOT ICONV_DIR)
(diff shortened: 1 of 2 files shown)


alex85k:
Thank you!
Replaced the repository in scripts wil official one: https://github.com/alex85k/winbuilds/blob/master/build_doxygen.bat

If you are interested in continuos integration using Appveyor.com (like Travis CI but with MSVC compiler), let me know. It is very easy to create automatic building script like https://github.com/alex85k/sqlite3-cmake/blob/master/appveyor.yml

PR #411, opened 2015-11-15
Bug 758132 - Brief description of non documented functions

albert-github:
In case a function or an other item is not documented and EXTRACT_ALL has not been specified it should not appear in the list with brief description of functions etc.


doxygen:
In this case I think using Definition::hasDocumentation() would be more appropriate. It checks for brief or details and extract all.

albert-github:
Based on above comment, pull request is withdrawn and a new pull request 415 has been submitted.

PR #412, opened 2015-11-18
Add WARN_AS_ERROR option to stop execution at first warning (equivalent of compilers' -Werror option)

bagage:


bagage:
Any comment @doxygen? Should I post patch on ML for further discussions?

doxygen:
I'm hesitating to include this option. If people enable it and a doxygen update produces a new warning, I expect annoyed users and more support questions and bug reports. So please indeed involve the ML to see if there is really a demand for this.

bagage:
Updated according to latest ML answers:

1. change exit code to 2 when at least one warning occurred during execution
2. add option TREAT_WARNINGS_AS_ERROR which immediately stop build (exit code 1, as errors already do). Also, change "warning:" logs to "error:" in this situation.
3. remove errors which are not true ones and convert them to either real error or warning depending of it (eg. PROJECT_LOGO should be a warning)?

I did 1. and 2. However 3. is harder because it's error specific.

PR #413, opened 2015-11-19
Support flex-2.6.0

heirecka:
The version checks only considered YY_FLEX_SUBMINOR_VERSION and did not
take YY_FLEX_MINOR_VERSION into account, which made them fail with
flex-2.6.0.

I've successfully tested this patch with flex-2.5.39 and 2.6.0. Additionally I generated apidocs for kdelibs with the latter without problems.

### DIFF
--- a/src/code.l
+++ b/src/code.l
@@ -3700,7 +3700,7 @@ void codeFreeScanner()
 extern "C" { // some bogus code to keep the compiler happy
   void codeYYdummy() { yy_flex_realloc(0,0); } 
 }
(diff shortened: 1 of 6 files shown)


pullmoll:
IMHO the version checks should also check for `YY_FLEX_MAJOR_VERSION==2` or they will fail again once a flex major version bump occurs.

doxygen:
I've merged this pull request manually and added the check on YY_FLEX_MAJOR_VERSION

PR #414, opened 2015-11-21
Fix flex version checks for flex>=2.6.0

pullmoll:
The current way to check for the flex version fail with flex>=2.6.0


pullmoll:
Oops. I didn't see someone else already had a pull request for this.

PR #424, opened 2015-12-12
Correct / set types for python variables

albert-github:
The bool python type was not handled.
In case of the initialization by means of a method call a variables was was aid to be a tuple (e.g. env= os.environ.copy())
Methods with initialization with defaults were shown without arguments in the brief description (e.g. def create_dir(dir, create=True):)
Arguments with values in method calls were seen as variables variables in case of continuation lines (e.g.
```
    parser.add_argument("--https_proxy", nargs="?", help="Proxy to be used for https requests",
                        action="store", const="myproxy.domain.com:8080")
```
  in this case action was seen as variable).

### DIFF
--- a/src/pyscanner.l
+++ b/src/pyscanner.l
@@ -113,6 +113,9 @@ static int              g_braceCount;
 static bool             g_lexInit = FALSE;
 static bool             g_packageCommentAllowed;
 
+static bool             g_start_init = FALSE;
+static int              g_search_count = 0;
+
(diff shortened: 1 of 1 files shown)


albert-github:
See also:
Bug 504309 - Python - Lines starting with named arguments are parsed as variable names

albert-github:
See also:
Bug 556683 - python class member definitions not displayed correctly

PR #439, opened 2016-01-08
Master

majdzadeh:
Solving Persian title issue:
When the generated documentation is in Persian (Farsi), the title of the
index page should be reversed. In other words the title should be
generated using theTranslator->trDocumentation()+projPrefix instead of
projPrefix+theTranslator->trDocumentation().

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -1640,7 +1640,7 @@ static void writeAnnotatedClassList(OutputList &ol)
 static QCString letterToLabel(uint startLetter)
 {
   char s[11]; // max 0x12345678 + '\0'
-  if (startLetter>0x20 && startLetter<=0x7f) // printable ASCII character
+  if (isId(startLetter)) // printable ASCII character
   {
     s[0]=(char)startLetter;
     s[1]=0;
@@ -1901,7 +1901,7 @@ static void writeAlphabeticalClassList(OutputList &ol)
     }
   }
 
-  ol.writeString("<table style=\"margin: 10px; white-space: nowrap;\" align=\"center\" width=\"95%\" border=\"0\" cellspacing=\"0\" cellpadding=\"0\">\n");
+  ol.writeString("<table class=\"classindex\">\n");
   // generate table
   for (i=0;i<=maxRows;i++) // foreach table row
   {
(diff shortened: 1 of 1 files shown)


albert-github:
I think there might be more languages that have a similar problem. I therefore think the line in `index.cpp`:
`ol.parseText(projPrefix+theTranslator->trDocumentation());`
should be replaced by
`ol.parseText(theTranslator->trDocumentation(projPrefix));`
and the translators should be adapted to handle this problem.

doxygen:
A less involved alternative would be to add an `isRightToLeft()` method to Translator which returns false by default and can be overloaded by specific languages

albert-github:
I'm not sure if this would completely solve the issue as as far as I can see the '+' operation would remain unless I interpret the proposed solution wrongly.

doxygen:
The solution would then be something like:

		if (theTranslator->isRightToLeft())
		{
			ol.parseText(theTranslator->trDocumentation()+projPrefix);
		}
		else
		{
			ol.parseText(projPrefix+theTranslator->trDocumentation());
		}

So the logic is still explicit in the code, but not bound to one specific language.

albert-github:
I think this solution is better than the proposed, language specific, solution. This solution is certainly less intrusive as my proposed solution and will probably work, certainly in this case as the `trDocumentation` routine is just used at 1 place.
I did a quick search for parseText and a + sign, except this occurrence I found only:
`classdef.cpp:  ol.parseText(displayName()+" "+theTranslator->trMemberList());`
and further a number of occurrences where a space was involved. For this `clasdef.cpp` will the `isRightToLeft` function work as well for all the languages where it works for the `index.cpp` occurrence? As I'm not a language expert I cannot answer this question, but when "exporting" the problem to the translation routine I think it is better handled. I'm more ore less saying doxygen should not create sentences by putting strings together by means of a + sign.

majdzadeh:
Albert, Dmitri,
Thanks for your responses and lots of thanks for considering the issue.

majdzadeh:
By the way, should I form another poll request containing the proposed change?

doxygen:
Yes, a new pull request would be nice

albert-github:
I just created a new proposed pull request #10769 that incorporates the suggestions done in this proposed pull request.
This proposed pull request is therefore obsolete when the proposed pull request #10769 is integrated.

albert-github:
Obsolete as #10769 has been integrated into master.

PR #441, opened 2016-01-13
PDF generation stops when image with caption is included in a table.

albert-github:
In case an image is included in a table and this image has a caption the generation of a PDF stops with the message:
! Misplaced \noalign.
\caption ->\noalign
                    \bgroup \@ifnextchar [{\egroup \LT@c@ption \@firstofone ...
l.45 \end{longtabu}

This problem has in general been described in: http://tex.stackexchange.com/questions/85919/adding-a-caption-to-a-graphic-inside-a-longtable

In this patch the suggestion from this reference is implemented by defining \doxyfigcaption and using this where  a caption is required.

### DIFF
--- a/doc/manual.sty
+++ b/doc/manual.sty
@@ -31,3 +31,10 @@
 \fancyfoot[CO]{\fancyplain{}{}}
 \fancyfoot[RO]{\fancyplain{}{}}
 
+% Define caption that is also suitable in a table
+\makeatletter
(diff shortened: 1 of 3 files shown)


doxygen:
Looks like this breaks the creation of the user manual.

albert-github:
Unfortunately I didn't get the message from Travis CI.

PR #442, opened 2016-01-16
fix ninja build error

sizeofvoid:
error: build.ninja:1548: bad $-escape (literal $ must be written as $$)


albert-github:
This patch breaks the intention of the original patch https://github.com/doxygen/doxygen/pull/419. The intention of the original patch is to create the possibility to specify the LEX_FLAGS during the make phase of build process (very useful for debugging purposes), the $(LEX_FLAGS) is evaluated by make during runtime and thus can be used for separate, changed / touched .l files. The ${LEX_FLAGS} is evaluated by CMake and is thus valid for all .l files and will result in a rebuild of all .l files as CMake has to be rerun to change the value. When using with -d this leads to an enormous amount of extra (unwanted) output.
The suggestion message as given by ninja (literal $ must be written as $$) works on Cygwin, but does not work on Windows.

PR #451, opened 2016-02-06
Bug 761471 - Do not allow ligatures in log output

albert-github:
Replacing QTextEdit by QTextBrowser.
Need to have 1 \<pre\> block as otherwise an empty line appears between 2 blocks as data is read in parts.

### DIFF
--- a/addon/doxywizard/doxywizard.cpp
+++ b/addon/doxywizard/doxywizard.cpp
@@ -10,7 +10,7 @@
 #include <QVBoxLayout>
 #include <QLineEdit>
 #include <QLabel>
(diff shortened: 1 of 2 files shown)


doxygen:
I would expect that QTextBrowser is much slower than QTextEditor and that this is in turn slower than QPlainTextEdit, so what if a large amount of output is produced?

albert-github:
In case of large amounts of output this might be slower, I didn't do measurements. I looked at the functionality in respect to ligatures as reported in the bug and the problem I saw myself on RHEL6 (text in italics)

PR #452, opened 2016-02-10
Add an additional option for EXTRACT_ALL

sloriot:
When EXTRACT_ALL is set to YES, the detailed description for all members is always generated, even if empty. This PR introduces a new tag (`EXTRACT_ALL_NO_DETAILED_IF_EMPTY`) that allows to control that behavior. The default is such that the former behavior is the same as before the introduction of this option.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -770,6 +770,18 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+  </group>
+  <group name='Build' docs='Build related configuration options'>
+    <option type='bool' id='EXTRACT_ALL_NO_DETAILED_IF_EMPTY' defval='0'>
+      <docs>
+<![CDATA[
+ When the \c EXTRACT_ALL tag is set to \c YES and a member or class
+ as no documentation, no detailed section will be produced if
+ the \c EXTRACT_ALL_NO_DETAILED_IF_EMPTY tag is set to \c YES.
+ This tag has no effect if the \c EXTRACT_ALL tag is set to \c NO.
+]]>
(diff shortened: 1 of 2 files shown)


lrineau:
@doxygen Is there something blocking the integration of this pull-request in Doxygen?

As far as I know, this PR is the last thing that prevents the [CGAL project](https://github.com/CGAL) to use the latest version of Doxygen (we have stuck to a modified version of 1.8.4 for).

Once this bug is fixed, our plan is to use the documentation testsuite of the CGAL project as a testsuite for Doxygen `master` also.

sloriot:
@doxygen is there something you don't like about this PR? I can rework it if there is something I can do to improve it.

doxygen:
What I don't like is that this is yet another option to tweak the output. People are already scared away by all the options, and I'd like to avoid that we end up with things like EXTRACT_ALL_BUT_NOT_FOR_ENUMS_WHOSE_VALUE_IS_GREATER_THAN_10 (I'm exaggerating but I hope you see my point) . 

I'm working on a template based output system where you could tweak the output without modifying doxygen.

sloriot:
I understand. We'll try to find a workaround. Thanks.

sloriot:
@doxygen what is the current status of your template based output?

sloriot:
This PR is more than 10 years old now and we are still relying on it with a local fork we maintain (option renamed `NO_ADDITIONAL_DETAILS` in the meantime). I do get your point about the increasing number of options but without this patch our doc is around 10% bigger on the disk, and makes some html pages much longer. Is there any way we could get it officially in the release?

PR #459, opened 2016-03-17
Bug 763720 - Exclusion of a new line at the end of source code file causing nesting of HTML code for function documentation

albert-github:
In case a code fragment does not end with a newline add it explicitly, otherwise following descriptive element might get into the same box.

### DIFF
--- a/src/definition.cpp
+++ b/src/definition.cpp
@@ -878,7 +878,7 @@ bool readCodeFragment(const char *fileName,
     }
   }
   result = transcodeCharacterStringToUTF8(result);
-  if (result.right(1) != "\n") result += "\n";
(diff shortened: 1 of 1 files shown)


src/definition.cpp
```diff
@@ -878,6 +878,7 @@ bool readCodeFragment(const char *fileName,
     }
   }
   result = transcodeCharacterStringToUTF8(result);
+  if (result.right(1) != "\n") result += "\n";
```
doxygen: `result.right(1)` is pretty inefficient for lange strings. Better use something like `!result.isEmpty() && result.at(result.length()-1)!='\n'`

PR #460, opened 2016-03-27
throw object instead of pointer

Godin:


doxygen:
You've changed code that is generated by JavaCC, so the fix should be done in the code generator instead (see https://java.net/projects/javacc/sources/svn/content/trunk/src/main/resources/templates/cpp/TokenManagerBoilerPlateMethods.template). Not clear what the change fixes and under which circumstances though.

PR #464, opened 2016-04-04
lex-debug: fix broken build

groleo:
build/generated_src/scanner.cpp:6:23: error: ‘scannerYY_flex_debug’ was not declared in this scope
 #define yy_flex_debug scannerYY_flex_debug

Per-file lex-debugging:
 add '%option debug' in the .l file.

All-files lex-debugging:
 rm -rf build/*
 LEX_FLAGS=-d make -C build/


albert-github:
In Bug 759932 - 1.8.10 Build Fails in Scanner Generation (YY_BUF_SIZE redefined) on RHEL 5 I sated that the mentioned problems came from:
Problem here is the flex version. RHEL5 comes, by default, with an old version of flex (2.5.4) and doxygen needs a newer version.

Also the links mentioned in this bug report indicate that a higher version than 2.5.4. should be used, so I'm not sure if the version mentioned in this fix is correct.

groleo:
The version check is for the major.minor part. The double definition can be worked around either by using a different flex skl file or adding ifdef guards in the generated files

PR #469, opened 2016-04-13
Document support for nested Compiled HTML Help indices.

c-lipka:
I couldn't find any mention of this topic in the docs, so I turned to the source code, to find that nested indices are indeed implemented in Compiled HTML Help output, but poorly documented even in the source code.


c-lipka:
Actually, ditch this. I mistook this for an undocumented feature, while I'm now realizing that it is currently nothing more than an artifact of the implementation of other features.

PR #474, opened 2016-04-27
Bug 765001 - Bad character escaping scheme in HTML anchor generation.

c-lipka:
HTML Help requires colons in anchor names to be escaped as `%3A` in the .hhk file. This fix achieves exactly that.

The same file is changed as in #473, so whatever pull request is applied second may be reported as conflicting. Changes are to different parts of the file though, so merging should be straightforward.

### DIFF
--- a/src/htmlhelp.cpp
+++ b/src/htmlhelp.cpp
@@ -134,17 +134,8 @@ static QCString field2URL(const IndexField *f,bool checkReversed)
   QCString result = f->url + Doxygen::htmlFileExtension;
   if (!f->anchor.isEmpty() && (!checkReversed || f->reversed)) 
   {
-    result+="#";
     // HTML Help needs colons in link anchors to be escaped in the .hhk file.
-    int prev = 0;
-    int next;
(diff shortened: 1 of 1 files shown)


c-lipka:
Travis CI check failed somewhere in the apt-get step to install texlive; maybe someone can re-trigger the build.

src/htmlhelp.cpp
```diff
@@ -134,7 +134,17 @@ static QCString field2URL(const IndexField *f,bool checkReversed)
   QCString result = f->url + Doxygen::htmlFileExtension;
   if (!f->anchor.isEmpty() && (!checkReversed || f->reversed)) 
   {
-    result+="#"+f->anchor;  
+    result+="#";
```
doxygen: I would expect this could be written much shorter using substitute,e.g.: `result+="#"+substitute(f->anchor,":","%3A");`

c-lipka: Sounds reasonable; I'm not too familiar with the Doxygen code, so wasn't aware this function even existed. I'll test it and update the pull request accordingly.

PR #476, opened 2016-05-01
Added mechanism to have `\addindex` entries link to start of section …

c-lipka:
…in HTML output.

Setting `HTML_INDEX_ANCHORS` to `NO` will suppress the generation of dedicated anchors for `\addindex` entries in the HTML output, and instead make the entries point to the start of the respective section.

This makes sure the user can immediately see the section heading, rather than having to scroll backwards, which is particularly annoying when an index is supposed to refers to an entire section anyway. (While the issue can be worked around by placing `\addindex` commands _before_ the corresponding `\section`, `\subsection` or Markdown heading, this feels the wrong way round to me.)

The mechanism will also support any potential future change allowing "twin" index entries linking to different locations in HTML Help output, by enabling HTML Help's disambiguation popup to display sensible information.


doxygen:
I don't really see why a user would always want to start reading at the beginning of the section. If there is a long description the thing that was indexed may even be out of view. which could be really confusing.

c-lipka:
While I do agree that this change is only moderately useful at present, it will become important once the HTML Help index generation is extended to supports "twin" index entries linking to entirely different places in the documentation: In such cases, the HTML help viewer will show only a single entry in the index, which when clicked will pop up a disambiguation dialog listing the sections referred to by the "twin" index entry; however, for this to work in a useful manner, the index entry MUST link to an anchor that is also referred to from the TOC, because that's where the disambiguation window gets the titles of the sections linked to.

PR #477, opened 2016-05-09
Bug 751984 - Use UTC timezone when displaying QDateTimes parsed from SOURCE_DATE_EPOCH

infinity0:
According to the `SOURCE_DATE_EPOCH` spec [1], "Formatting MUST be deferred until runtime if an end user should observe the value in their own locale or timezone."

However `setTime_t` uses `localtime`, so the output is in the timezone of the build machine running doxygen, and not the timezone of the reader consuming the final output.

To adhere to the spec, the easiest option is to add a `setTimeUtc_t` function that is the same as `setTime_t`, but only uses `gmtime` instead of first trying `localtime`.

[1] https://reproducible-builds.org/specs/source-date-epoch/

### DIFF
--- a/qtools/qdatetime.cpp
+++ b/qtools/qdatetime.cpp
@@ -1158,6 +1158,29 @@ void QDateTime::setTime_t( uint secsSince1Jan1970UTC )
 
 
 /*!
+  Sets the UTC date and time given the number of seconds that have passed
+  since 00:00:00 on January 1, 1970, Coordinated Universal Time (UTC).
+
+  Note that Microsoft Windows supports only a limited range of values for
+  \a secsSince1Jan1970UTC.
+*/
+
+void QDateTime::setTimeUtc_t( uint secsSince1Jan1970UTC )
+{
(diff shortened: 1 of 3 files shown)


doxygen:
I don't understand the change. When doxygen produces output, it should use local time right? i.e. the time is put into the generated output, so how else can users ever observe the time in their own locale?

infinity0:
If the user-doing-the-building sets SOURCE_DATE_EPOCH it means they want reproducible builds, so the build-user can't output this time in their locale. (Reproducible builds by definition means that the build should be independent of the build-user.)

A more complex alternative (to this PR, that still preserves locale) is to allow *reader-users* to observe `SOURCE_DATE_EPOCH` in their own locale, which means adding some javascript to parse this timestamp into the reader's locale. This would allow *builders* to produce the same reproducible output.

doxygen:
Thanks for the clarification. Indeed I agree that viewing the timestamp in the user's local could better be facilitated by javascript.

infinity0:
Great, thanks!

PR #480, opened 2016-05-20
Unify handling of extra packages in formula.cpp and latexgen.cpp

iamthad:
This allows formula generation to handle package arguments correctly.


doxygen:
I've moved the duplicated code into util.cpp, See https://github.com/doxygen/doxygen/commit/9a8d723babdb982daa3f0ddec6098d40d22d8142

PR #484, opened 2016-06-22
Modified (readability) layout for member title in HTML and LaTex

PkLab:
Output from Doxygen is great but visual impact for members heading is a bit complex in special case if member has a lot of params and returns complex/template type.

In order to improve overall readability for HTML and LaTeX/PDF, in this pull request:
 - we use shortest member name in member title heading
 - we have moved full member definition just below the heading
 - we have adapted HTML and LaTeX templates

Changes should not have any effects on RTF and Man generators

Below some screen shoot 

html

pdf

### DIFF
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -1551,10 +1551,14 @@ void HtmlGenerator::endMemberDocList()
   DBG_HTML(t << "<!-- endMemberDocList -->" << endl;)
 }
 
-void HtmlGenerator::startMemberDoc(const char *,const char *,const char *,const char *,bool) 
-{ 
+void HtmlGenerator::startMemberDoc( const char *clName, const char *memName,
+                                    const char *anchor, const char *title, bool showInline)
+{
   DBG_HTML(t << "<!-- startMemberDoc -->" << endl;)
- 
+  t << "\n<h2 class=\"memtitle\">" << title << " "
+    << "<a href=\"#" << anchor << "\" class=\"permantlink\"" "title=\"Permalink to this headline\">&#9854;</a>"
+    << "</h2>"
+    << endl;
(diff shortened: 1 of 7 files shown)


PkLab:
@doxygen might be any chance to be merged for this PR ?

doxygen:
@PkLab I think it an improvement in readability indeed. The only problem I see is with overloaded functions/methods; they would then all get the same heading, which may be confusing.

PkLab:
You are right, but you already doing this in the treeview.

Sure is just a point of view, but use of detailed function definition as is, generates unreadable header especially if the function returns complex type and/or uses complex params/template.

In this PR, overloaded functions/methods get the same headers but have different anchors. At the end, readers know about overloading. Eyes immediately catch function name while implementation details are right after.

In case, some _[overload]_ flag might be added to the function name. For example, after [this line](https://github.com/PkLab/doxygen/blob/short_member_title/src/memberdef.cpp#L2611) you might something like this:
```cpp
    MemberName *mn = Doxygen::memberNameSDict->find(title);
    if (mn && mn->count() > 3)
        title += "() [overload]";
```

StevenPuttemans:
As an extensive user of the OpenCV docs, based on doxygen, I would love to see this feature added to get more clarity in the visual representation of the docs. 

How difficult would it be to check for overloads and do something like
* myFunction
* myFunction [overload 1]
* myFunction [overload 2]

And so on?

That would solve the issue here and explain people explicitly that it handles overload functions. The numbering could even be removed. I guess adding the tag [overload] can simply be done by looking at the `@overload` command inside doxygen?

StevenPuttemans:
Ow and about 10 seconds later I see that @PkLab suggest this himself :+1:

PkLab:
Nice to see the PR merged and congrats for excellent solution that numbers overload functions.

PR #489, opened 2016-07-19
Fix SHOW_NAMESPACES=NO to correctly hide all namespaces

tsparber:
This change will exclude the generated namespace documentation from the resulting outputfiles.

Currently, even if this option is enabled the namespace documentation is included.
The generation of namespace indices is inhibited which also leads to an error when generating a LaTeX documentation (namespaces.tex not found), except when the indices are hidden (`LATEX_HIDE_INDICES=YES`).

This also fixes [Bug 728649](https://bugzilla.gnome.org/show_bug.cgi?id=728649).

The implementation is similar to [`FileDef::isLinkableInProject()`](https://github.com/doxygen/doxygen/blob/77c505b0445c74d1933edceb9a4d2f5fa972d581/src/filedef.cpp#L1888).
A minimal working example can be found in this [Gist](https://gist.github.com/tsparber/215e0829a155bc2174e3c1cabf9e2ed0).

### DIFF
--- a/src/namespacedef.cpp
+++ b/src/namespacedef.cpp
@@ -1114,9 +1114,10 @@ bool NamespaceDef::isLinkableInProject() const
   int i = name().findRev("::");
   if (i==-1) i=0; else i+=2;
   static bool extractAnonNs = Config_getBool(EXTRACT_ANON_NSPACES);
+  static bool showNamespaces = Config_getBool(SHOW_NAMESPACES);
   if (extractAnonNs &&                             // extract anonymous ns
-      name().mid(i,20)=="anonymous_namespace{"     // correct prefix
-     )                                             // not disabled by config
+      name().mid(i,20)=="anonymous_namespace{" &&  // correct prefix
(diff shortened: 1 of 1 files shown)


tsparber:
Upon further investigation, this change was already part of doxygen before the v1.8.1 release.
(See https://github.com/doxygen/doxygen/commit/17ee30b327bf4f07f2411a50a46f5467456beacf#diff-ae4e6b89b9db052ae648d93faec149f6R968).
Is there any reason this was removed?
Is there another way to exclude the namespaces?

(The commit is updated to use the same syntax also before the removal)

albert-github:
This also fixes [Bug 778521 - VHDL: invalid RTF output when SHOW_NAMESPACES = NO](https://bugzilla.gnome.org/show_bug.cgi?id=778521)

albert-github:
I checked the given gist and the example from #6164 and the problem does not occur in the current version of doxygen (1.8.20) anymore.
It looks like it has been solved in the mean time (probably version 1.8.16)

PR #494, opened 2016-07-29
Fixes for handling of PUBLIC/PRIVATE in Fortran

melven:
Before, access statements `public :: myType` in modules were ignored as
well as public/private access specifications in type definitions.

Also added a test case for this which assures the correct behavior for
some cases...


melven:
There seems still to be a problem with module variables - I'll have a look at it...

albert-github:
Melvin,

I think it would be better not to mix all the different subjects in one bug commit / lists of commits in one branch. For integration and (when necessary) bug finding it would be better that each subject gets its own pull request.

melven:
I can try it - but the tests build upon results of the previous commits... And Fortran support currently is quite broken (at least applied to larger modern projects)...

PR #495, opened 2016-08-01
Also map .f95, .f03 and .f08 file types to Fortran

melven:
These are for Fortran 95, Fortran 2003 and Fortran 2008, respectively.


albert-github:
I think useful default extensions, but these extensions should already work with EXTENSION_MAPPING
Note: the table in the paragraph "Step 1: Creating a configuration file" (starting.doc) has to be extended as well.

melven:
Of course, but that's the mapping GCC also uses - so I thought these should be handled automatically...

PR #496, opened 2016-08-01
Fortran private public statements

melven:
Before access statements `public :: myType` in modules were ignored as
well as public/private access specifications in type definitions.

Also added a test case for this which assures the correct behavior for
some cases...


albert-github:
Had a quick look at the results. There are some good improvements but still there are some omissions. I ran the file Strings.f90 (see attachment, had to rename it to a .txt file) from the FF08Depands package of Ian Harvey (see: http://www.megms.com.au/fortran-tools.htm) and used a default Doxyfile (doxygen -g).
In case of the of an interface with multiple members only the first one is not shown (correct, as it is private) but the others (also private) are shown.


[Strings.txt](https://github.com/doxygen/doxygen/files/424614/Strings.txt)

melven:
I have also further fixes for interfaces; there are 3 variants (generic, specific abstract). For the generic case a new compound is defined in doxygen (this is effectively function overloading - so one could also generate multiple functions with the appropriate arguments by copying...)

In the Strings.F90 file there are only generic interfaces. Inside the interface compound all functions should be marked as public and static (which I have already done in another commit).
So I would add these changes as another pull request once this one and #495 are merged
(as they depend on these).

Or did I misunderstand the desired/current behaviour here?

albert-github:
I'm not sure about the static. the functions should be marked as they are defined (in this case the routines mentioned inside the interface should be PRIVATE). The problem of showing the different possibilities i.e. overloads should be handled in a different way.

I don't have a good solution for this yet but I think that in case a a routine is PRIVATE the overloaded routine should be seen with the interface name (as this is the name to be used outside the module), so for Append we should see the possibilities:
- subroutine 	append_(vector, item)
- subroutine 	append(vector, item)
- subroutine 	append(vector, items, mask)

In case of PUBLIC routines we should see:
- subroutine 	append_(vector, item)
- subroutine 	append(vector, item)
- subroutine 	append(vector, items, mask)
- subroutine 	append_ (vector, item)
- subroutine 	append_scalar (vector, item)
- subroutine 	append_vector (vector, items, mask)

I don't think the changes depend of #495 as this can be accomplished by the EXTENSION_MAPPING. Regarding other changes that are directly related to the PRIVATE ? PUBLIC handling I think they can be part of this pull request.

melven:
So for now I just tried to make the current variant of handling generic interfaces consistent, e.g. everything inside is public. I really don't think you can make it private as this would hide something you can call from the outside.

I would prefer the variant of copying routines with the generic interface name instead of a compound for generic interfaces (looks more the way it is used outside the module!). I could change this and make another merge request...

albert-github:
Regarding the PUBLIC / PRIVATE we have to search for a solution, it does contain valuable information for the user as in case of PRIVATE he cannot call `append_scalar` but he has to call `append`. The suggestion about the copy is reasonable, but a bit contrary to what will be done for routines that don't have a generic interface as in this case the full compounds lists are shown.

I looked at the namespace lists / module lists and here the `append_` is missing.

melven:
I added the Strings.f90 as another test; the xml output looks good to me: append_ is private in the module, append is public and in the interface append you have the required entries including append_ ...

melven:
Ok I missed some problems, see the commit above.
(Verifying details by looking at XML in a text editor is quite cumbersome...)

albert-github:
Does not look to bad (besides the PUBLIC / PRIVATE issue we addressed before which has to be dealt with later on).
I found a small problem with an interface in a routine.  Say we have the attached file (renamed to .txt) and the standard Doxyfile with EXTRACT_ALL set to YES than we should not see the a reference to `routine_interface` on any page (except in call graphs etc.) see e.g. page html/interfaceroutine__interface.html



[use_tst.txt](https://github.com/doxygen/doxygen/files/426742/use_tst.txt)

albert-github:
In the file `use__tst_8f.html` I still see the `routine_interface` mentioned, this should not be the case as it is not a subroutine defined in this file.

melven:
So I tried to fix this, but I don't really understand, what the code is supposed to do here.

Nested procedures are also extracted:
```
SUBROUTINE use_tst
CONTAINS
SUBROUTINE inner
END SUBROUTINE inner
END SUBROUTINE use_tst
```
The code tries to detect interface dummy arguments, but I don't even know what this is supposed to do (compiler think its ok though):
```
SUBROUTINE use_tst(routine_interface)
INTERFACE
  SUBROUTINE routine_interface()
  END SUBROUTINE routine_interface
END INTERFACE

END SUBROUTINE use_tst
```

albert-github:
The first example signals a routine that is only accessible from the code of `use_tst`, it is more or less a private routine for `use_tst`. 
In the second example the interface signals the definition of the routine that is used as a argument to the routine `use_tst2`. In older versions one had to write `external routine_interface` but in that case no checks on arguments were possible. Well know examples are functions to order values and defining the `compare` routine as argument.

melven:
1. example: it's clear to me that this is a private routine of `use_tst` - I wanted to point out that doxygen extracts these (not marked as private)... I can have a look at this.

2. I played around with this a bit - I didnt use this short variant, just the equivalent abstract interface definition:
```
module test
  implicit none
  public :: compareInterf, sort1
  abstract interface
    logical function compareInterf(a,b)
      integer, intent(in) :: a, b
    end function
  end interface
contains
  subroutine sort1(iarray, compareFcn)
    integer, intent(inout) :: iarray(1:)
    procedure(compareInterf) :: compareFcn
  end subroutine
end module
```

For doxygen this means that interfaces matching an argument name must be copied/used as argument definition. All other interfaces inside procedures can be ignored except fort the call tree.

By the way: how/where is the call tree determined?

melven:
I'm still not really sure what the desired behaviour is for stuff inside procedures.

In the version above all interfaces that don't refer to dummy arguments are thrown away. Interfaces for dummy arguments are made public (but might not be distinguishable from other interfaces except for the scope as the type of the interface is not shown in the xml).

Procedures defined inside other procedures' "contains" section are marked as private...

melven:
Any thoughts on this?

I have other patches for type bound procedures and inheritance, but these require somehow working private/public statements...

albert-github:
I did a quick test with the attached file (renamed to aa,,txt). Unfortunately only the `program tst` and `SUBROUTINE use_tst2` show up in the documentation (EXTRACT_ALL has been set to YES in the Doxyfile)
In the 1.8.11 and the current github master branch all routines do show.


[aa.txt](https://github.com/doxygen/doxygen/files/441059/aa.txt)

melven:
Thanks for the good feedback and effort here. I fixed the problem in the commit above...

albert-github:
When running the attached example I get the message:
`aa.f90:3: warning: Internal inconsistency: scope for class use_tst1::routine_interface not found!`
and when looking in the output I see:
  `subroutine use_tst1( use_tst1::routine_interface routine_interface) `
in my opinion, here the `use_tst1` should not be present.


[aa1.txt](https://github.com/doxygen/doxygen/files/441213/aa1.txt)

melven:
That was the point above that was not clear to me: `routine_interface` must get shown in the documentation, so you can see the type of the argument to `use_tst1`. But it is not accessible outside of `use_tst1` - if it is put in global scope (or in a sourrounding module scope) - it looks like some global or module interface definition...

albert-github:
This is a bit of a nasty problem.
I had a look at how such a thing is handled in C, well know routines are here the sorting routines where one of the arguments is a compare routine.
`void qsort (void *base, size_t num, size_t size, int(*compar)(const void *, const void *))`

When we translate this to our, show in previous messages, FORTRAN problem, I think this should look like:
  `subroutine use_tst1(subroutine routine_interface(integer i ))`
or, probably better as it is more consistent with the past:
  `external use_tst1(subroutine routine_interface(integer i ))`

melven:
Any documentation of the interface (and its types) will be thrown away this way - still sounds like the best variant...

albert-github:
The documentation of the `routine_interface` should appear with `use_tst1` but like you concluded and I see as well ("This is a bit of a nasty problem") we will have to see how this will work out.

melven:
I tried to make it look like you suggested.

Another strange thing: In the xml-output the type of an argument is just set to the name of the argument (shown correctly in html). This makes it difficult to test this behaviour. Any ideas where/why the real type information gets lost in xml?

albert-github:
Took a bit time to find some time.
in the new version I see:
`subroutine use_tst1( external subroutine(integer i) routine_interface) `
but I expected:
`subroutine use_tst1( external subroutine routine_interface(integer i)) `
from example file `aa1.txt`.

melven:
How should I do this? I only set the type and the name appropriately and it is shown as `<type> <name>`...

By the way in C I obtain `int use_tst1  ( void(*)(int i)  routine_interface )` for similar code.

albert-github:
Intuitively I would have expected the version I wrote for FORTRAN, but looking at the C "equivalent" you are right. When looking at a more complex situation, see bb.txt I see the line (just took the argument):
`external real(prec) function, dimension(size(y))(real(prec), intent(in) xf)real(prec), dimension(:), intent(in) yf, )  	func, `
it is a bit unreadable and also not correct as after the `xf` there is a `)`, this should be a `,` and after the `yf` a `,`, this should not be present. Problem with FORTRAN is also that a `,` is used as separator for arguments but also for attributes. A possible layout would be, similar to a `normal` function but now embedded as an argument :
```
external real(prec) function, dimension(size(y))(real(prec), intent(in)                 xf,
                                                 real(prec), dimension(:), intent(in)   yf
                                                )                                           func, 
```
I don't know how to accomplish this at the moment (might also for C functions and especially template functions be of an advantage).

I think that for the time being, when the `,` and `)` problem has been solved this should be OK.
(I still will do some more tests in the near future).

[bb.txt](https://github.com/doxygen/doxygen/files/453528/bb.txt)

melven:
Any news on this?

(Sorry to bother you if you are too occupied, I just have further bug fixes for other Fortran problems...)

albert-github:
Currently I'm quite occupied, maybe end of next week I can have a look.

janusw:
Ping! What's the status here? Looks like this has not been merged yet ...

melven:
I rebased these changes onto the current master and started with a smaller pull request in #9510

PR #512, opened 2016-08-31
make compound.xsd deterministic again

gsauthof:
Some validating XML parsers (e.g. the Qt5 one) complain about
ambiguity and duplicate elements in the compound.xsd.

For example:

    Error XSDError in file:xml/compound.xsd, at line 781, column 20:
      Complex type docCaptionType has non-deterministic content.
    Exception: XSD xml/compound.xsd is invalid

And after this is fixed:

    Error XSDError in file:xml/compound.xsd, at line 683, column 20:
      Complex type docParaType has duplicated element image in
      its content model.
    Exception: XSD xml/compound.xsd is invalid

The first is due to those elements being generated from a table where the
`bareName` is not unique. Resulting in a choice element of `docTitleCmdGroup`
containing the element `trademark` twice. This change filters
out the duplicates.

The second is due to the choice element of `docCmdGroup` referencing
`docTitleCmdGroup` - which defines element `image` - while also locally
defining the `image` element. This change remove the redundant local
definition.

I noticed this while using [doxy2man](https://github.com/gsauthof/doxy2man) which validates the xml generated by doxygen, by default.

### DIFF
--- a/src/htmlentity.cpp
+++ b/src/htmlentity.cpp
@@ -17,6 +17,8 @@
 #include "message.h"
 #include "ftextstream.h"
 
+#include <set>
+
 //! Number of doxygen commands mapped as if it were HTML entities
 static const int g_numberHtmlMappedCmds = 11;
 
@@ -476,12 +478,20 @@ DocSymbol::SymType HtmlEntityMapper::name2sym(const QCString &symName) const
 
 void HtmlEntityMapper::writeXMLSchema(FTextStream &t)
 {
+  std::set<QCString> seen;
   for (int i=0;i<g_numHtmlEntities - g_numberHtmlMappedCmds;i++)
   {
     QCString bareName = g_htmlEntities[i].xml;
     if (!bareName.isEmpty() && bareName.at(0)=='<' && bareName.right(2)=="/>")
     {
       bareName = bareName.mid(1,bareName.length()-3); // strip < and />
+      std::pair<std::set<QCString>::iterator, bool> p = seen.insert(bareName);
+      if (!p.second)
+        continue;
+      // otherwise the image element (with docEmptyType) would clash
+      // with image (docImageType) in the group definition docCmdGroup
+      if (bareName == "image")
+        continue;
       t << "      <xsd:element name=\"" << bareName << "\" type=\"docEmptyType\" />\n";
     }
   }
--- a/templates/xml/compound.xsd
+++ b/templates/xml/compound.xsd
@@ -374,6 +374,13 @@
  
   <xsd:group name="docTitleCmdGroup">
     <xsd:choice>
+      <xsd:group ref="docTitleCmdGroupWithoutImage"/>
+      <xsd:element name="image" type="docEmptyType" />
+    </xsd:choice>
+  </xsd:group>
+
+  <xsd:group name="docTitleCmdGroupWithoutImage">
+    <xsd:choice>
       <xsd:element name="ulink" type="docURLLink" />
       <xsd:element name="bold" type="docMarkupType" />
       <xsd:element name="emphasis" type="docMarkupType" />
@@ -402,7 +409,7 @@
 
   <xsd:group name="docCmdGroup">
     <xsd:choice>
-      <xsd:group ref="docTitleCmdGroup"/>
+      <xsd:group ref="docTitleCmdGroupWithoutImage"/>
       <xsd:element name="linebreak" type="docEmptyType" />
       <xsd:element name="hruler" type="docEmptyType" />
       <xsd:element name="preformatted" type="docMarkupType" />


albert-github:
Regarding the trademark I can see that it is double in the xsd file due to the fact that doxygen has 2 definitions: `&tm;` and `&trade;` . One is for historical reasons the other is the HTML-Entity definition.

Regarding the "image" I think the solution is not correct as one should represent the symbol for the imaginary part of a complex number and the other an image.
As I'm not familiar with xsd (and don't have, as far as I know, a validating XML parser at hand) I don't know a solution but seen the code in `visitPreStart` in `xmldocvisitor.cpp` I can imagine something like type="docImageType" for the image and type="docEmptyType" fir the imaginary part of a complex number.

gsauthof:
@albert-github, you are right, the image element ambiguity need to be resolved differently. I had overlooked the different types of the image element.

The issue is that the image element is visible twice in one group defintion, once with type `docImageType` and with `docEmptyType`.

Since the `docImageType` is a superset of `docEmptyType` (the attributes child elements are optional this can be resolved via referencing the `docTitleCmdGroup` without the image element - from the `docCmdGroup`.

I've rebased my PR to fix this.

A diff between the different `compound.xsd` thus looks like this:


    --- compound_orig.xsd	2016-08-31 19:42:35.468809226 +0200
    +++ xml/compound.xsd	2016-08-31 19:42:42.262843686 +0200
    @@ -374,6 +374,13 @@
      
       <xsd:group name="docTitleCmdGroup">
         <xsd:choice>
    +      <xsd:group ref="docTitleCmdGroupWithoutImage"/>
    +      <xsd:element name="image" type="docEmptyType" />
    +    </xsd:choice>
    +  </xsd:group>
    +
    +  <xsd:group name="docTitleCmdGroupWithoutImage">
    +    <xsd:choice>
           <xsd:element name="ulink" type="docURLLink" />
           <xsd:element name="bold" type="docMarkupType" />
           <xsd:element name="emphasis" type="docMarkupType" />
    @@ -548,7 +555,6 @@
           <xsd:element name="oline" type="docEmptyType" />
           <xsd:element name="frasl" type="docEmptyType" />
           <xsd:element name="weierp" type="docEmptyType" />
    -      <xsd:element name="image" type="docEmptyType" />
           <xsd:element name="real" type="docEmptyType" />
           <xsd:element name="trademark" type="docEmptyType" />
           <xsd:element name="alefsym" type="docEmptyType" />
    @@ -640,7 +646,6 @@
           <xsd:element name="lsaquo" type="docEmptyType" />
           <xsd:element name="rsaquo" type="docEmptyType" />
           <xsd:element name="euro" type="docEmptyType" />
    -      <xsd:element name="trademark" type="docEmptyType" />
         </xsd:choice>
       </xsd:group>
     
    @@ -650,7 +655,7 @@
     
       <xsd:group name="docCmdGroup">
         <xsd:choice>
    -      <xsd:group ref="docTitleCmdGroup"/>
    +      <xsd:group ref="docTitleCmdGroupWithoutImage"/>
            <xsd:element name="linebreak" type="docEmptyType" />
             <xsd:element name="hruler" type="docEmptyType" />
             <xsd:element name="preformatted" type="docMarkupType" />

albert-github:
Must my limited knowledge of xsd but I don't get it yet as in the resulting code there would still be `<image/>` for both cases.
I just looked at the `\image` command and here the possibility for `xml` is not yet supported. Looking at the `docbook` possibility (used command: `\image docbook doxygen_logo.gif`) the result in the (xml) file is:
```
    <figure>
        <title></title>
        <mediaobject>
            <imageobject>
                <imagedata width="50%" align="center" valign="middle" scalefit="1" fileref="doxygen_logo.gif"></imagedata>
            </imageobject>

        </mediaobject>
    </figure>
```
so here no `<image/>` but a complete "tree".

Looking at another image in the output (call graph between a `func1 `and `func2`, the output for `docbook` is:
```
    <figure>
        <title>Dependency diagram</title>
        <mediaobject>
            <imageobject>
                <imagedata width="50%" align="center" valign="middle" scalefit="1" fileref="aa_8c__incl.png"></imagedata>
            </imageobject>
        </mediaobject>
    </figure>
```
and for `xml`:
```
        <references refid="aa_8c_1ad791fbbbff0895f1f19676a1bf633267" compoundref="aa_8c" startline="19" endline="21">func2</references>
```

I looked at another "image" command the `\msc` bu this is in the `xml `given as `<msc>` and `</msc>` and for `docbook `as:
```
    <figure>
        <title></title>
        <mediaobject>
            <imageobject>
                <imagedata width="50%" align="center" valign="middle" scalefit="1" fileref="inline_mscgraph_1"></imagedata>
            </imageobject>

        </mediaobject>
    </figure>
```

Sofar my investigations.

- Looks to me that the `<image>` is not used in the `xml` version as the command `\image` is not supported for `xml` so we would have the freedom to choose how an image is shown and the imaginary part symbol can use the `<image>` tag. @doxygen is this correct?
- `xml` should be implemented for the `\image` command, in my opinion similar as it is done for `docbook`, @doxygen what is your opinion?
- `\msc` (and probably some other graphical formats) should also be implemented as done for `docbook`, @doxygen what is your opinion?`

doxygen:
@albert-github Doxygen's XML output generator does support images (see `XmlDocVisitor::visitPre(DocImage *)`), same holds for msc graphs.

It will actually output something like `<image type="html"...>` for a `\image html ...` command. 

Remember that XML is meant as a "dump" of doxygen's internals, not as a direct document output format (such as docbook, which happens to use the same XML notation). So also a `\htmlonly...\endhtmlonly` block will also appear in the XML output.

If the schema validator has a problem with the "imaginary" symbol, I think it is better to use a `<imaginary/>` tag in the XML output. I don't expact it will be used much in practice, so I don't think anyone will have a problem with this change.

albert-github:
- The image tag is indeed also used for e.g. the `\image html` command (so here we do have our exception): `
<image type="html" name="doxygen_logo.gif"></image>
`
-  The `xml` format is "meant as a "dump" of doxygen's internals, not as a direct document output format",  but in practice users use it as a basis for further processing (seen questions / remarks on the format currently and in the past). Seen it intentions compatibility is not really an issue , so changing the version for the imaginary part symbol to `<imaginary/>` is not a big deal, same hods for changing the output for the `\image` command to something more resembling the `docbook` output, but as it is a "dump" format changing the imaginary part is more consistent and intuitive.
- Seen the fact that the `xml` format is meant as a "dump" format the `\msc` command is correct.

doxygen:
I've just pushed the following fix https://github.com/doxygen/doxygen/commit/13946338f9e5fcdb6dca7f8e4c2429ca000455fb

gsauthof:
Not having to care about compatibility simplifies things.

With your change the Qt5 validating parser also is able to validate the XML with the generated XSD.

I am thus closing this pull request.

PR #513, opened 2016-09-01
Bug 766059 - DOT_PATH not expanded

albert-github:
The original problem in this bug request was that forward slashes were used in the PATH instead of backslashes.
In the command shell this does not result in problems, but when calling a program from within doxygen this leads to the problem that the executable cannot be found.
In this patch the PATH variable is, just for the process, changed in such a way that ther are back slashes instead of forward slashes (Windows only, not on Cygwin).

### DIFF
--- a/src/portable.cpp
+++ b/src/portable.cpp
@@ -448,4 +448,26 @@ bool portable_isAbsolutePath(const char *fileName)
   return false;
 }
 
-
+/**
+ * Correct a possible wrong PATH variable
+ *
+ * This routine was inspired by the cause for bug 766059 was that in the Windows path there were forward slahes.
+ */
+void portable_correct_path(void)
+{
+#if defined(_WIN32) && !defined(__CYGWIN__)
+  const char *p = portable_getenv("PATH");
+  char *q = (char *)malloc(strlen(p) + 1);
+  strcpy(q, p);
+  bool found = false;
+  for (int i = 0 ; i < strlen(q); i++)
+  {
+    if (q[i] == '/')
+    {
+      q[i] = '\\';
+      found = true;
+    }
+  }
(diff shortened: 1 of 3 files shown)


src/portable.cpp
```diff
@@ -448,4 +448,26 @@ bool portable_isAbsolutePath(const char *fileName)
...
+  for (int i = 0 ; i < strlen(q); i++)
+  {
+    if (q[i] == '/')
+    {
+      q[i] = '\\';
+      found = true;
+    }
+  }
+  if (found) portable_setenv("PATH",q);
+#endif
```
doxygen: I'd expect that q needs to be freed at this point to avoid a memory leak.

src/portable.cpp
```diff
@@ -448,4 +448,26 @@ bool portable_isAbsolutePath(const char *fileName)
...
+  bool found = false;
+  for (int i = 0 ; i < strlen(q); i++)
+  {
+    if (q[i] == '/')
+    {
+      q[i] = '\\';
+      found = true;
+    }
+  }
+  if (found) portable_setenv("PATH",q);
```
doxygen: I'd expect that q needs to be freed at this point to avoid a memory leak.

doxygen:
Looks like  a workaround for a Windows bug/limitation (e.g. handling forward slashes while evaluating PATH), rather than a doxygen bug, but if it helps people I'll accept it.

albert-github:
- It is indeed a workaround for a Windows bug/limitation but it helps people that have forward slashes in their PATH (for whatever reason automatic by a program or by hand as was the case for the submitter of the bug report).
- regarding the memory leak of `q` this is indeed the case when the value of `q` is copied, but I'm no sure if all routines do such a copy. The size of the memory leak is relatively small.

PR #514, opened 2016-09-02
Bug 770660 - Code snippet always shows line numbers from 1

albert-github:
This patch makes the handling of the \snippet and other commands consistent between the different languages (no line numbers anymore with python) and also introduces analogous to \includelineno  the command \snippetlineno.

Some non relevant changes:
- *code.l Calculation of the end line was incorrect, in case of a snippet the end line was the number of lines of the snippet and not reltive to the start line.
- *code.l made consistent over the different laguages, enabling exBlock and inlineFragment
- testing/indexpage.xml in test 14 the \snippet command was used with python and giving line numbers, linenumbers are now gone (consistency)

### DIFF
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -179,6 +179,7 @@ documentation:
 \refitem cmdskipline \\skipline
 \refitem cmdsnippet \\snippet
 \refitem cmdsnippetdoc \\snippetdoc
+\refitem cmdsnippetlineno \\snippetlineno
 \refitem cmdstartuml \\startuml
 \refitem cmdstruct \\struct
 \refitem cmdsubpage \\subpage
@@ -2183,7 +2184,7 @@ Commands for displaying examples
   This command works the same way as \ref cmdinclude "\\include", but will add line
   numbers to the included file.
 
-  \sa section \ref cmdinclude "\\include".
+  \sa sections \ref cmdinclude "\\include" and \ref cmdsnippetlineno "\\snippetlineno".
 
 <hr>
 \section cmdincludedoc \\includedoc <file-name>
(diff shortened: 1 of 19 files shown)


albert-github:
Looks like travis has some problems getting the texlive distribution.

doxygen:
It sometimes times out too soon indeed. 
You can manually start the run again.
I just did; let's see if it works this time.

albert-github:
Terminated again.
- how can I restart the travis process ?
- is it possible to increase the "timeout" time (I noticed before that the tex part takes quite long)

doxygen:
> On 02 Sep 2016, at 15:49 , albert-github <notifications@github.com> wrote:
> 
> Terminated again.
> 
> 	• how can I restart the travis process ?

There is a restart button (circle with arrow) on the right hand side of the screen below the more options menu (it's easy to overlook)

> 	• is it possible to increase the "timeout" time (I noticed before that the tex part takes quite long)

Not easily, I've now split it in to smaller commands, one for each latex package, and made the biggest one
more verbose, so travis know that things didn't get stuck.

Seems to help.

doxygen:
I've squashed the 3 commits into 1, and pushed that https://github.com/doxygen/doxygen/commit/9ae1af9b8679a0f14cb568d1db3afcc6e3ba40a6. Please double check if all changes are in correctly.

albert-github:
It was OK to squash the 3 commits as the second commit undid the first and the third placed the commit in a branch (unfortunately I didn't know a better way to do it)
As far as I can see it is all squashed OK.

albert-github:
@doxygen Regarding the travis restart.
- I don't see the mentioned restart button, probably due to the fact that I'm not signed in to travis as there is a button: "Sign in with GitHub"
- clicking the "Sign in with GitHub" results in a page with the question about authorization:
travis_auth
I'm not sure if I should grant permissions (as there are write permissions as well!)

PR #515, opened 2016-09-02
Sorting in LaTeX index and missing \@ in index

albert-github:
Sorting in latex index and missing \@ in index

When looking at the index of the doxygen (pdf) manual we see that \_\_init\_\_ is on a bit strange place (between "\" items) and that the item "\@" is missing.

Made handling of the index consistent by using latex... routines and adjusting the label routine in respect to the @.

### DIFF
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -179,6 +179,7 @@ documentation:
 \refitem cmdskipline \\skipline
 \refitem cmdsnippet \\snippet
 \refitem cmdsnippetdoc \\snippetdoc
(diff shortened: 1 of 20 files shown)


doxygen:
I've cherry-picked the commit containing the actual change and push the result https://github.com/doxygen/doxygen/commit/558958dca5e5cb50bd33f3344cab151aba9ce076. Please try to provide a clean pull request without other commits to keep the change history clean.

albert-github:
Cherry pick is OK, problem was the, wrong reverted, commit with pull request #514. I hope that new commits will be OK from now on.

PR #525, opened 2016-09-23
code.l: Fixed cross referencing issue when using array.

groleo:


src/code.l
```diff
@@ -2566,7 +2566,8 @@ RAWEND    ")"[^ \t\(\)\\]{0,16}\"
 					  generateClassOrGlobalLink(*g_code,yytext);
   					  g_name+=yytext; 
 					}
-<Body>{SCOPENAME}/{BN}*[:;,)\]]		{ // "int var;" or "var, var2" or "debug(f) macro" , or int var : 5;
+<Body>{SCOPENAME}/{BN}*[:;,)\[]		{ // "int var;" or "var, var2" or "debug(f) macro" , or "int var : 5;"
```
doxygen: You removed `]` from the pattern, but I think this breaks highting symbolic constant `b` in this code `a[b]`

groleo:
Does the test in the image covers the cases you mention ?

https://pbs.twimg.com/media/CtJOmNJWcAAZEdZ.jpg:large

PR #526, opened 2016-09-23
xmlgen:in code highlight, translate the Copyright sign to its code point

groleo:


src/xmlgen.cpp
```diff
@@ -132,6 +132,7 @@ inline void writeXMLCodeString(FTextStream &t,const char *s, int &col)
       case '&':  t << "&amp;"; col++;  break;
       case '\'': t << "&apos;"; col++; break;
       case '"':  t << "&quot;"; col++; break;
+      case 0xA9:  t << "&#169;"; col++; break;
```
doxygen: This change is not needed in my opinion. The XML output is already UTF-8 encoded, so any non-ascii character should appear as such, and are handled by the default case.

groleo:
I noticed the error while loading the XML files with lxml. Beside the .xml file encoding which is utf-8, i've also used a XMLParsed with encoding="utf-8", but I still got errors:

    for event, elem in context:
  File "src/lxml/iterparse.pxi", line 208, in lxml.etree.iterparse.__next__ (src/lxml/lxml.etree.c:144363)
lxml.etree.XMLSyntaxError: Input is not proper UTF-8, indicate encoding !
Bytes: 0xA9 0x3C 0x73 0x70, line 797, column 74

doxygen:
If you get the message that the input encoding is not UTF-8 it probably means that your source code is also not UTF-8 encoded, and as a result you need to set INPUT_ENCODING correctly in your doxygen config file.

groleo:
How does INPUT_ENCODING works when some files are UTF-8 encoded while other are ISO-8859 ?

How I see it is that doxygen generates an XML file with UTF-8 encoding, w/o making sure it really is UTF-8, nor does it print an error in which it would let the user know something is wrong.

doxygen:
INPUT_ENCODING assumes all files have the same encoding. It is not really possible to reliably detect the encoding used in a given file (without knowing how the result should look like)

groleo:
So is it correct to say that, by default, doxygen assumes the inputs are utf8, but it doesn't check that ?
 And I guess the same goes when INPUT_ENCODING is non-empty: doxygen won't actually check the encoding.

doxygen:
Indeed if INPUT_ENCODING is empty or the default value UTF-8, then the input is assumed to be UTF-8 encoded. This is not checked indeed.

PR #530, opened 2016-09-25
For manpages remove trailing dash when no brief description

artur-kink:
Currently, if a definition does not have a brief description then the manpage will show:

>NAME
>Definition -

The change will show the manpage as

>NAME
>Definition

Also cleaned up manpage output for a group. Currently it seems to duplicate both the definition name and the brief description, leading to a big mess in the NAME section.

### DIFF
--- a/src/filedef.cpp
+++ b/src/filedef.cpp
@@ -373,9 +373,10 @@ void FileDef::writeBriefDescription(OutputList &ol)
     if (rootNode && !rootNode->isEmpty())
     {
       ol.startParagraph();
+      ol.pushGeneratorState();
       ol.disableAllBut(OutputGenerator::Man);
       ol.writeString(" - ");
-      ol.enableAll();
+      ol.popGeneratorState();
       ol.writeDoc(rootNode,this,0);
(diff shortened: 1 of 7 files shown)


src/filedef.cpp
```diff
@@ -365,14 +365,17 @@ void FileDef::writeDetailedDescription(OutputList &ol,const QCString &title)
...
   {
     DocRoot *rootNode = validatingParseDoc(briefFile(),briefLine(),this,0,
                        briefDescription(),TRUE,FALSE,0,TRUE,FALSE);
 
     if (rootNode && !rootNode->isEmpty())
     {
       ol.startParagraph();
+      ol.disableAllBut(OutputGenerator::Man);
+      ol.writeString(" - ");
+      ol.enableAll();
```
doxygen: This is not correct. You need to use ol.pushGeneratorState() and ol.popGeneratorState() to avoid that some output format that was already disabled, gets enabled.

artur-kink: Thanks for the review, I have made then changes suggested.

PR #535, opened 2016-10-05
Add mscgen images to index.qhp

kanje:
Include mscgen images to QtHelp *.qch output.


anderswallander:
Thanks alot! I could use this change as a pattern for fixing same missing functionality for plantuml into index.qhp. See https://github.com/doxygen/doxygen/pull/538.

PR #536, opened 2016-10-05
Add support for using PlantUML to generate class diagrams

jmk0:
Current implementation of Doxygen uses Graphviz to generate UML class diagrams.  I like the look of PlantUML better, and added support for allowing the use of that tool as an alternative to Graphviz.


cheoljoo:
hpp2plantuml is alternative  to generate PlantUML class diagrams.

PR #544, opened 2016-11-11
fixed 2 compilation warnings:

deem0n:
```
vhdlparser/TokenMgrError.cc:101:16: warning: comparison of constant 255 with expression of type 'char' is always true
      [-Wtautological-constant-out-of-range-compare]
        if (ch < 0xff) {
            ~~ ^ ~~~~
```
```
vhdlparser/VhdlParserTokenManager.cc:3032:24: warning: comparison of constant 128 with expression of type 'char' is always true
      [-Wtautological-constant-out-of-range-compare]
      else if (curChar < 128)
               ~~~~~~~ ^ ~~~
```

### DIFF
--- a/vhdlparser/TokenMgrError.cc
+++ b/vhdlparser/TokenMgrError.cc
@@ -98,7 +98,7 @@ JAVACC_SIMPLE_STRING addUnicodeEscapes(JAVACC_STRING_TYPE str) {
         retval.append("\\\\");
         continue;
       default:
-        if (ch < 0xff) {
+        if ((unsigned char)ch < 0xff) {
           retval += ch;
           continue;
         }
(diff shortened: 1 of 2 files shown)


doxygen:
The files you modified are generated by JavaCC. So this needs to be fixed in the JavaCC code generator instead, i.e. somewhere here: https://java.net/projects/javacc/sources/svn/show/trunk/src/main/resources/templates/cpp?rev=792

deem0n:
Hi, patching Java looks like overkill for me ;-)

Actually, we have

```c
#ifndef JAVACC_CHAR_TYPE
#define JAVACC_CHAR_TYPE char
#endif
```

So it should be possible to use cmake add_definition(...) to define JAVACC_CHAR_TYPE. Would it be fine ?

PR #545, opened 2016-11-13
Bug 774138 - Please add HTML classes to "Definition at..." & "Referenced by..." for CSS

albert-github:
Added class= to html output for "Definition at..." resulting in p.definition in the css file and for "Referenced by .. " and "References ..." resulting in p.definition in css file.
(also corrected some error messages).


tonyelewis:
Thanks very much for your work on this. I'm happy for [774138](https://bugzilla.gnome.org/show_bug.cgi?id=774138) (on bugzilla.gnome.org) to be closed.

PR #546, opened 2016-11-17
Add HTML_REVERSE_TITLE option to reverse page and project titles

eepp:
See the option's help for details.

For some reason, this does _not_ work as expected. I don't know why, but the new option (`HTML_REVERSE_TITLE`) in my `Doxyfile` seems to be ignored completely: the behaviour is to always pick the first line amongst `<!--BEGIN HTML_REVERSE_TITLE-->` and `<!--BEGIN !HTML_REVERSE_TITLE-->`.

Any idea?


albert-github:
A bit of a late and quick review ...
- what is the objection, except for ease of use, to have an extra setting for this and not using a custom header.html?
- the major problem probably is that `HTML_REVERSE_TITLE` has to be added to the list of select blocks in `htmlgen.cpp` as well, now both title lines will be in the resulting file and only the first `<title>` is picked by the browser.

PR #548, opened 2016-11-21
Better module pages: autogroup include only one level of entites (Fix Issue #774432)

yuanchen-zhu:
Preivously, compound entities autogrouped using `\{` and `\}` will have all their descendents entities included in the group as well. This makes the generated module page very noisy.

This patch maintains a stack of QCString, corresponding to the name of the autogroup for the compound entities currently being recursively parsed. Then upon `initGroupInfo`, an autogroup is added to the entry's list of groups only if the autogroup is different from its parent compound entity's autogroup.

Also, namespaces are ignored when deciding if an entry's parent is already included in the autogroup. This is a more correct behavior since namespaces are open and can be extended at other places. When a
namespace is included in an autogroup via `\{` and `\}`, it really the content of the namespace defined within those braces that should be included in the module page, and not the namespace itself, which can
contain other entities defined elsewhere.

For example, consider the following snippets

```cpp
/// \defgroup autogroup
/// \{
/// desc of x
namespace x {
    /// desc of y
    namespace y {
        /// desc of a
        struct a {
            /// desc of a::b
            struct b{};

            /// desc of a::c
            int c;
        };
    }
}
/// \}
```

Then the module page for `autogroup` will include `x`, `x::y`, `x::y::a`, but not `x::y::a::b` or `x::y::a::c`, whereas the default Doxygen behavior is to include the latter two member entities. If `a` is very complex, and there are a lot of other complex classes defined, then the module page of `autogroup` becomes super noisy, polluted by all the implementation details of classes in `x::y`.

### DIFF
--- a/src/commentscan.h
+++ b/src/commentscan.h
@@ -76,8 +76,8 @@ void groupEnterFile(const char *file,int line);
 void groupLeaveFile(const char *file,int line);
 void groupLeaveCompound(const char *file,int line,const char *name);
 void groupEnterCompound(const char *file,int line,const char *name);
-void groupPushAutoGroup();
-void groupPopAutoGroup();
+void groupPushAutoGroup(Entry *e);
+void groupPopAutoGroup(Entry *e);
 
 void openGroup(Entry *e,const char *file,int line);
 void closeGroup(Entry *,const char *file,int line,bool foundInline=FALSE);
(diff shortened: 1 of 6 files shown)


yuanchen-zhu:
Any chance of getting this in? Without this, module page with autogroup is basically not usable.

PR #551, opened 2016-12-05
It is more efficient to seek for a single character rather than a str…

orbitcowboy:
…ing.


nuskorpios:
It seems that you accidentally changed the file `vhdlparser/Makefile` to one generated for your system...

PR #552, opened 2016-12-12
Fix: replace deprecated commands with \textXY{ equivalents in LaTeX generator

jhunold:
Fixes LaTeX build errors with latest Ubuntu packages.


t-b:
Why does `\bfseries` not work anymore? And you are aware that changing `\bfseries` to `\textbf{}` might create new problems as the text is now an argument to a command, and that might create problems with fragile arguments etc.

PR #556, opened 2017-01-16
Bug 777266: Environment default value

simonnagl:
Solution for https://bugzilla.gnome.org/show_bug.cgi?id=777266

Please tell me if you prefer patches at bugzilla.

Now we only try to replace valid variable names.
I also removed the support for variables named for e.g. PROGRAMFILES(X86). Had this a special meaning?

Do I need to write a test case for this? If yes can you point me to a similar test?

Thank you for reviewing this in advance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@
 
 /doxygen_docs
 /doxygen.tag
+
+build/*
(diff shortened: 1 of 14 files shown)


doxygen:
Windows allows environment variable _names_ with braces! So this was added as a specific fix for this bug report https://bugzilla.gnome.org/show_bug.cgi?id=547759 (the second to last comment)

simonnagl:
I added the support for variable names with braces and dots. Please recheck.

simonnagl:
Added regression tests for environment variables with default substitution.
Testing the real replacement would need another <identifier> which may look the like this:

- env:    'argument' is a line that which sets an environment variable passed to Doxygen.

simonnagl:
Added also regression tests for environment variable substitution.

PR #562, opened 2017-04-06
Make use of clang compilation database in clang assisted parsing

dvj:
See: https://clang.llvm.org/docs/JSONCompilationDatabase.html

This uses clang-tooling to read in a compilation database so that clang assisted parsing can use the actual compilation arguments. This is mostly meant as a replacement to using `CLANG_OPTIONS`, which is either very difficult or impossible to get right, depending on project complexity.

Dimitri expressed interest in this http://clang-developers.42468.n3.nabble.com/Doxygen-1-8-4-uses-libclang-to-improve-parsing-tp4032204p4032209.html

The code changes are fairly straight-forward, but to make this build easily, the changes require clang 3.9+. Clang switched to a CMake build system in 3.8 and got the kinks worked out by 3.9. In particular, this relies on `find_package(Clang)` to get the link lines correct. In this respect, it may be worth adding another option (`use_clang_tooling`?) to support only this new code. On macOS, this entails using `libc++` over `libstdc++`.

As written, this ignores the options in `CLANG_OPTIONS` if the database is found from `CLANG_COMPILATION_DATABASE_PATH` and the currently parsed file is in the database.

Also note that I didn't indent a code block in `clangparser.cpp` to make this PR easier to read. 

To make use of this feature in an end-user project, if using CMake, it's as simple as adding:
```    set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE INTERNAL "")```
to your top level `CMakeLists.txt` to generate the compilation database, and then adding
```    CLANG_COMPILATION_DATABASE_PATH=<path to cmake build dir>```
to your doxygen file.

### CODE REVIEW

**Comment by ycdtosa** [comment]
i hope this one gets merged...
CLANG_OPTIONS is a pain to use.

File: src/config.xml [hunk 1]
```diff
@@ -1637,6 +1637,20 @@ to disable this feature.
 ]]>
       </docs>
     </option>
+    <option type='string' id='CLANG_COMPILATION_DATABASE_PATH' setting='USE_LIBCLANG' defval='0'>
+      <docs>
+<![CDATA[
+ If clang assisted parsing is enabled you can provide the clang parser with the
+ path to the <a href="http://clang.llvm.org/docs/HowToSetupToolingForLLVM.html">
+ compilation database</a> used when the files were built. This is equivilent to
```
lrineau: typo on `equivilent`

PR #564, opened 2017-04-28
fix typo

3nids:


3nids:
I think this PR is not repsonsible of the build timeout on Travis.
I think it can be merged safely.

PR #566, opened 2017-05-09
fixes vhdl literal bug

olagrottvik:
A small change in vhdlparser.jj (and updated code with javacc) that fixes bug reported in [bug 782375 ](https://bugzilla.gnome.org/show_bug.cgi?id=782375)

valid vhdl literals like:

32X"FF_FF"
8b"1111_----"
are reported as errors. This pull request addresses this bug.


olagrottvik:
added new commit to address bug with doxygen reporting syntax error when encountering decimal base identifier

PR #568, opened 2017-05-24
Add links behind nav entries "Namespaces" and "Files" (matching "Classes")

kossebau:
When layout file has "namespacelist" and/or "filelist" set to visible=no,
the files "namespaces.html" resp. "files.html" are still generated,
but not reachable.

So same fix as done for "classlist" should be used,
cmp. 700a9ac3c177fdef25b9b00ed6bb5e0ea963d236

Also gives a consistent result & at least in generated QCH files fixes the error
message currently shown when clicking on "Namespaces" or "Files" section titles in the navigation tree.

Using lne->baseFile() and thus the string from the central db instead of using a hardcoded duplicated filename feels better, and thus first patch also adapts the code for "LayoutNavEntry::Classes".

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -4151,7 +4151,7 @@ static void writeIndexHierarchyEntries(OutputList &ol,const QList<LayoutNavEntry
         case LayoutNavEntry::Classes: 
           if (annotatedClasses>0 && addToIndex)
           {
-            Doxygen::indexList->addContentsItem(TRUE,lne->title(),0,"annotated",0); 
+            Doxygen::indexList->addContentsItem(TRUE,lne->title(),0,lne->baseFile(),0);
             Doxygen::indexList->incContentsDepth();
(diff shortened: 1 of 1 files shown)


kossebau:
Rebased the commits to latest master, but again Travis dies during
> Get:7 http://ppa.launchpad.net/texlive-backports/ppa/ubuntu precise/main amd64 texlive-latex-extra-doc all 2012.20120611-1~ubuntu12.04.1 [293 MB]

with

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
> Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

Meh. This spoils the otherwise nice idea of CI, if one gets flagged Red only because CI is failing elsewhere :/

All I can say is that locally the patch builds fine and also works as expected for mw :)

PR #569, opened 2017-05-28
Bug 783134 - LaTeX output for \tparam block fails to compile when it contains a \code block

albert-github:
General problem regarding having a code / verbatim section inside a table.
Besides handling of the $ some other characters need special handling as well as the \n.


mminutoli:
@albert-github It works perfectly for me.

PR #573, opened 2017-06-13
Bug 783768 - Add QHP_NO_CONSTRUCTOR_KEYWORDS option

kossebau:
Allows to avoid clashes between keywords for classes themselves and
for their constructors methods

Not sure if this option should be limited to QCH generation alone and not be a generic option. Currently I only work with QCH generation, so cannot judge how useful this option would be for other output formats.

Fixes https://bugzilla.gnome.org/show_bug.cgi?id=783768

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -2171,6 +2171,15 @@ The \c DOCSET_PUBLISHER_NAME tag identifies the documentation publisher.
 ]]>
       </docs>
     </option>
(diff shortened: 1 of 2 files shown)


albert-github:
An more generic implementation has been made (without an extra setting required).

I think that #9258 supersedes this proposed pull request.

albert-github:
The more generic patch #9258 has been integrated, I think this proposed pull request can be closed.

PR #574, opened 2017-06-21
build: fix the way lang_cfg.h is generated

brgl:
The following pull request fixes a build issue we've been experiencing
with buildroot recently.

This header is generated by running cmake/lang_cfg.cmake and
redirecting its stderr. If any warning is emitted by this script, it
ends up in the generated header and breaks the build.

To avoid such problems: pass the path to the header as an argument to
the cmake script and use the cmake 'file' command instead of 'message'.

We can't even use message(STATUS...) as - although it prints to stdout
as opposed to other types of messages - it prepends all output with
a double hyphen.


brgl:
Looks to me as if the build failures are not related to the patch.

PR #577, opened 2017-06-28
new Doxyfile option: MODIFY_SPECIAL_CHARS

joenio:
this solves problems with Filename Special Chars on doxyparse and
allows doxygen to not escape special chars and space.

special chars escaped by default:

  _-:/<>*&|.!,{}?^+%()+=$\@

Signed-off-by: Igor Ribeiro Barbosa Duarte <igor.ribeiro.duarte@gmail.com>


joenio:
hi @doxygen, it is possible to give some feedback about this merge request? thanks.

joenio:
sorry for the insistence, but in the Analizo (www.analizo.org) project we maintain a fork of doxygen called doxyparse, we are planning to stop keeping this fork and start to contribute directly to doxygen, once that doxyparse was already in doxygen as an addon, but we need to include some features in doxygen, this PR is the first and more essential feature we need on the doxygen side.

we have some others contributions to doxygen and more specifically to doxyparse, but we need this PR first, if possible.

the doxyparse is just an addon that modify the default output of doxygen, instead output in a human-readable format, it is intented to produce a machine-readable output, today a kind of YAML format.

doxyparse has been used in many software engeneering research regards on software metrics, quality metrics and so on, it is a nice scientific tool and we want to keep it in a way more close to you guys, the doxygen's maintainer team.

thanks for your attention and for the good work on this wornderful tool.

joenio:
trying again to get some feedback here, guys, it's possible to review this PR with some feedback? thanks in advance!

albert-github:
- Where in doxyparse does the original version give problems?
- Wouldn't it be a possibility to translate the string back in doxyparse?

joenio:
#### Where in doxyparse does the original version give problems?

In doxyparse we map each source code file parsed by doxygen, then we generate a output related to each source code file and its metadata collected by doxygen, eg. functions, methods, variables, call between funcions or methods. Without this patch the doxyparse doesn't know how to differ between source code files that has in fact named with _01, _02, _03 and so on, and those files that are renamed by doxygen translating "," to "_00" or "{" to "_02" etc.

for example, imagine a C++ project with the files:

* src/main.cpp
* src/dog{.cpp
* src/dog_01.cpp

after doxygen translate "{" to "_01" I don't know anymore the differs between `src/dog{.cpp` and `src/dog_01.cpp` because both is going to be identified by doxygen as `src/dog_01.cpp`.

joenio:
#### Wouldn't it be a possibility to translate the string back in doxyparse?

It's not possible because there is no way to differs among files that in fact has name containing "_01", "_02", and so on from files that was translated by doxygen.

joenio:
#### is there a place where I can find the doxyparse clone as used by the Analizo project?

yes, we keep a repository here https://github.com/analizo/doxyparse

the master branch has the changes we are doing on doxyparse before submit pull-request to doxygen upstream (this repository here).

joenio:
ping

joenio:
Hello @albert-github can you help me here? Thanks, sorry again, but I'm trying for more than 2 years to approve this PR. There is some work I need to do to help you?

albert-github:
**Where in doxyparse does the original version give problems?**
**Wouldn't it be a possibility to translate the string back in doxyparse?**

Strange as the `{` is translated to `_02` and `_02` is translated to `__02` so they are in my opinion distinguishable or do you have an example where both `{` and `_02` result in `_02` (or something similar)?

It would also be useful to have a small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

joenio:
there is no way to diferentiate when doxygen escape resulting on "__02" from when the filename is in fact "__02", how can I infer if the resulting "__02" was caused by doxygen escaping from when the real filename is in fact "__02"?

by the way, I revert that fix on doxyparse side and doing a workaround to deal with that, the workaround is a weak solution but it will works for a time... closing this pull request for now.

thanks for the feedback.

albert-github:
I don't understand the part about the `__02` you wrote.
In case doxygen detects an underscore in the input it replaces it with a double underscore, and in case the input contains 2 consecutive underscores each underscore is replaced by 2 underscores in total thus a resulting 4 underscores.

PR #586, opened 2017-08-04
Marks JS as freely licensed

zachwick:
Marking the resulting JS in this way ensures that visitors to the
resulting HTML documentation do not have to choose between running
non-free JS and experiencing the HTML documentation as it was
intended.

The JS was already freely licensed, so this change just
ensures that the appropriate labelling occurs so that GNU LibreJS [0]
parses the JS correctly as freely licensed.

[0] https://www.gnu.org/software/librejs/

Signed-off-by: zachwick <zach@zachwick.com>


albert-github:
Added a regression in case of xhtml (pull request #594)

PR #588, opened 2017-08-17
preprocess all Fortran files

janusw:
Currently doxygen only preprocesses Fortran files with upper-case file extensions (such as .F and .F90). This seems reasonable, since most Fortran compilers also stick to the convention of preprocessing only upper-case files.

However, basically all compilers have command-line flags for running the preprocessor also on lower-case files (such as .f and .f90). I don't see such a possibility for doxygen, therefore this commit is the only reasonable way I found to have doxygen preprocess .f90 files.

### DIFF
--- a/src/fortranscanner.l
+++ b/src/fortranscanner.l
@@ -2618,8 +2618,9 @@ void FortranLanguageScanner::parseCode(CodeOutputInterface & codeOutIntf,
 
 bool FortranLanguageScanner::needsPreprocessing(const QCString &extension)
 {
-  return extension!=extension.lower(); // use preprocessor only for upper case extensions
+  return true;  // use preprocessor for all extensions!
 }
+
 void FortranLanguageScanner::resetCodeParserState()
 {
   ::resetFortranCodeParserState();


albert-github:
I don't think it is a good idea to set an optional flag from compilers as a default for doxygen.

janusw:
Then it would be great to have a configuration option in doxygen that control which file extensions should be preprocessed. Something like "PREPROCESS_FILE_PATTERNS"?

janusw:
In any case, I don't think it's a huge problem to preprocess all Fortran files. The performance penalty shouldn't be too bad (in particular since normal f90 files will not have any preprocessor directives). In C++, you also preprocess **all** files.

albert-github:
In C++ (etc.) the preprocessing is part of the language (when I'm not mistaken).
In Fortran it is a compiler extension by means of fpp or cpp.
Regarding the performance penalty I think that on small projects this might be acceptable but on larger projects it might be significantly.

janusw:
> In C++ (etc.) the preprocessing is part of the language (when I'm not mistaken).

True. In C++ it's more tightly integrated with the language itself (via #includes etc), while in Fortran preprocessing is indeed 'optional', but still quite common.

> Regarding the performance penalty I think that on small projects this might be acceptable but on larger projects it might be significantly.

For Fortran projects without any preprocessor directives, one can still disable preprocessing globally to avoid any performance hit.

Certainly a configuration option is a more solid solution (I guess I'll open a feature request on bugzilla for that), but until that becomes a reality, this little hack is a useful workaround for me.

Feel free to close the pull request if you're not willing to merge it.

janusw:
> Certainly a configuration option is a more solid solution (I guess I'll open a feature request on bugzilla for that)

See https://bugzilla.gnome.org/show_bug.cgi?id=786417

PR #589, opened 2017-08-17
Bug 786409 - parsing error in Fortran file with preprocessing

albert-github:
Problem with ' (single quote) in Fortran. A string in Fortran can be between single or double quotes.


janusw:
I just tested this branch on a large Fortran project, and it works like a charm. It fixes two problems of the same kind that I was seeing earlier. I'd love to see this merged. Thanks!

PR #590, opened 2017-08-18
Bug 786409 - parsing error in Fortran file with preprocessing

albert-github:
the // is in Fortran the indication for string concatenation and not for comment and should thus be ignored during, Fortran, preprocessing


janusw:
Just tried it, works really well. Fixes all problems I've seen. :thumbsup:

PR #591, opened 2017-08-22
Add language type attribute to programlisting tag

shi-yan:
The current programlisting tag doesn't have any attribute to indicate
the code's language. This makes code highlighting customization
difficult.

For example, the current doxygen generated document doesn't hightlight
javascript code block. If one wants to customize the code block by
parsing the xml files and uses 3rd party syntax highlighter, there is
no way to tell from the xml file what language the code snippet is
using.

This change introduces an attribute "lang" to the programlisting tag.

### DIFF
--- a/src/xmldocvisitor.cpp
+++ b/src/xmldocvisitor.cpp
@@ -211,7 +211,7 @@ void XmlDocVisitor::visit(DocVerbatim *s)
   switch(s->type())
   {
     case DocVerbatim::Code: // fall though
-      m_t << "<programlisting lang=\"" << lang << "\">"; 
+      m_t << "<programlisting language=\"" << lang << "\">"; 
       Doxygen::parserManager->getParser(lang)
                             ->parseCode(m_ci,s->context(),s->text(),langExt,
                                         s->isExample(),s->exampleFile());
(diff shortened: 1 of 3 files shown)


shi-yan:
I'm implementing a customized doxygen document by parsing its xml results.

I want to use a 3rd party syntax highlighter, such as highlight.js.

for this, I need to know the programming language used by each code block. but the programlisting tag doesn't have any attribute to tell me that.

This change adds a new attribute "lang" to the programlisting tag to indicate the programming language.

albert-github:
Why using the abbreviation lang= as attribute and not the full language=?

shi-yan:
I don't have any opinion regarding attribute naming.

changed to "language"

doxygen:
Next to adding the attribute to the XML output, please also add it to the XML schema file (`templates/xml/compound.xsd`) section `listingType`, so the XML output can be validated.

shi-yan:
updated compound.xsd

shi-yan:
I have fixed the unit test failure caused by my change.
Can this be approved now? I have further changes coming to improve the markdown parser.

doxygen:
I've changed  `lang` to `langToString(langExt)` so it shows the language name instead of the file extension

PR #594, opened 2017-08-25
Marks JS as freely licensed (regression pull request 586)

albert-github:
In case of xhtml the & has to be escaped


albert-github:
Small side note, the files in the directory search still keep the extension html although the HTML_FILE_EXTENSION has been set to xhtml.

PR #595, opened 2017-08-25
Bug 776870 - XML Parsing Error for operator<< methods when outputing to XHTML

albert-github:
The < sign in the title has to be escaped, especially in case of xhtml.

NOTE: this bugfix also requires pull request #594 in case #586 has been incorporated.


nmellado:
Yes please.
Also true for `operator<()`.

PR #597, opened 2017-08-30
Use -fvisibility=hidden with GCC/Clang

evelikov:
Allows us to hide all the internal symbols from the final binary.
As a result it becomes ~10% smaller.

Signed-off-by: Emil Velikov <emil.l.velikov@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,13 @@ if (WIN32)
     endif()
 endif()
 
+if(POLICY CMP0063)
+  cmake_policy(SET CMP0063 NEW)
+endif()
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
+
 if ("${CMAKE_GENERATOR}" MATCHES "Ninja") 
   set(LEX_FLAGS )
   set(YACC_FLAGS )


CMakeLists.txt
```diff
@@ -53,6 +53,14 @@ if (${CMAKE_SYSTEM} MATCHES "Darwin")
...
 endif()
 
+if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
+endif()
+
+if (${CMAKE_C_COMPILER_ID} STREQUAL "Clang" OR ${CMAKE_C_COMPILER_ID} STREQUAL "GNU")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fvisibility=hidden")
+endif()
+
```
lrineau: Setting `CMAKE_CXX_VISIBILITY_PRESET` to `hidden` should have the same effect.

evelikov: Did not know about those, thanks. I cannot see which version of cmake introduced them so I'm a bit cautious about using them.
That said, if @doxygen or others prefer that route I can respin.

lrineau: That was already availlable in CMake 3.0.0.

evelikov: [Here](https://github.com/doxygen/doxygen/blob/master/CMakeLists.txt#L14) it says that doxygen requires 2.8.12.

lrineau: That variable [was already in 2.8.12 as well][ref].

[ref]: https://cmake.org/cmake/help/v2.8.12/cmake.html#variable:CMAKE_LANG_VISIBILITY_PRESET

doxygen: In that case I prefer the simpler solution suggested by @lrineau

evelikov:
Updated the patch and force pushed to the same branch.

CMakeLists.txt
```diff
@@ -64,6 +64,11 @@ if (WIN32)
     endif()
 endif()
 
+cmake_policy(SET CMP0063 NEW)
```
lrineau: You cannot set that policy unconditionnaly. It was [introduced in CMake 3.3[doc], and that triggered a CMake in [travis]. To be compatible with all versions of CMake, you can write:
```cmake
if(POLICY CMP0063)
  cmake_policy(SET CMP0063 NEW)
endif()
```

[doc]: https://cmake.org/cmake/help/v3.3/policy/CMP0063.html
[travis]: https://travis-ci.org/doxygen/doxygen/jobs/271596022#L771

evelikov: Thanks, addressed with v3. Currently running through the Travis checks.

evelikov:
Humble ping - is there anything I can help to get this moving?

bkmcfarland:
unsubscribe

PR #601, opened 2017-09-09
Correction display of backtick in LaTeX

albert-github:
The backtick in LaTeX, running text, was not show as a backtick but more as a \&lsquo; (left single quote).
This is corrected in the LaTex string filter, consequently some quotes were not "nice" anymore in the documentation, they have been replaced by pairs of \&lsquo; / \&rsquo;


doxygen:
Hi Albert, I made an alternative solution (see commit https://github.com/doxygen/doxygen/commit/beaa386ca97341e66ad673660c808993240df637). Here the documentation does not need to be changed, since single quoted text is detected as part of markdown processing. Can you check if it works for you or if you see any problems?

albert-github:
The implementation in util.cpp is identical to the implementation proposed in this pull request so OK.

The potential problem I see is the fact that an automatic conversion takes place from \`...'  to \&lsquo;...\&rsquo; (&lsquo;...&rsquo;). I can imagine that some users might want to have the freedom to specify the quotes themselves which is taken a bit from them now (although I think the solution gives nicer results than when having \`...' in the output text).

The change in install.doc was a real bug, a quote was missing (it went a bit under in this pull request, new pul request required).

doxygen:
I've included the fix in install.doc now. The `...' conversion is coupled to having Markdown support enabled. I think it is in line with the spirit of Markdown; i.e. writing readable comments and have them translated into nice looking output.

PR #602, opened 2017-09-09
Physical newlines in ALIASES configuration tags.

albert-github:
Some commands read input till the end of the physical line. In case these commands are used in an alias the rest of the line is lost / gives not the required results.
This patch creates the possibility to have physical newlines in ALIASES.
See also: https://stackoverflow.com/questions/46050789/doxygen-alias-with-multiple-commands

### DIFF
--- a/doc/custcmd.doc
+++ b/doc/custcmd.doc
@@ -40,7 +40,9 @@ The simplest form of an alias is a simple substitution of the form
  put the command `\sideeffect` (or `@sideeffect`) in the documentation, which 
  will result in a user-defined paragraph with heading <b>Side Effects:</b>.
 
-Note that you can put `\n`'s in the value part of an alias to insert newlines.
+Note that you can put `\n`'s in the value part of an alias to insert newlines
+(in the resulting output). You can put `^^` in the value part of an alias to
+insert a newline as if a physical newline was in the original file.
 
 Also note that you can redefine existing special commands if you wish.
 
--- a/src/config.xml
+++ b/src/config.xml
@@ -532,7 +532,10 @@ Go to the <a href="commands.html">next</a> section or return to the
  will allow you to
  put the command \c \\sideeffect (or \c \@sideeffect) in the documentation, which 
  will result in a user-defined paragraph with heading "Side Effects:".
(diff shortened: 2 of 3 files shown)


t-b:
@albert-github I'm also using ALIASES so this enhancement is really appreciated. I'm not sure about the syntax though. Why `^^`?

albert-github:
The choice for ^^ is more or less arbitrarily. When looking at regular expressions the $ signals the end of a line and ^ the start of a line.  So a choice could be $^ this can give some misunderstanding with environment variables, same accounts for e.g. $$ so therefore I chose ^^.

rafaeltmc:
This does not work with section and subsection

albert-github:
@rafaeltmc 

Best to open a new issue for the problem you face (with a reference to this this issue).
In the new issue can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?
Please also specify the doxygen version used.

rafaeltmc:
I actually just made it work by using "\\_linebr" instead of "^^". Not sure if I should still open an issue.

albert-github:
Yes better to open an new issue, `\_linebr` should not be used (as it is an internal command)

Inujel:
I successfully used ^^ in ALIASES in 1.8.14, but it's not working anymore in 1.8.15.

I use the following alias:
ALIASES               +=   "_test{3}=@defgroup TEST-\1 \2^^\details \testid{\3}"

Now a "@_linebr" and the \details part appear verbatim in the group name:
ObjectClasses.StructureDefinition[DCF_Device](0)@_linebr\details...

Is there another way around? 
Our team is heavily using multiline aliases.

albert-github:
What happens when you place spaces around the "^^", otherwise please create a new issue with attached a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

Inujel:
It's not working either. The @_linebr remains unparsed, with spaces around,
in the group name.
Ok, I'll post a new issue.

Le mer. 12 juin 2019 à 10:36, albert-github <notifications@github.com> a
écrit :

PR #603, opened 2017-09-19
Suppression warnings about deprecated directive

lrineau:
Before this PR, the warnings was, with 3.0.4:
```
constexp.y:46.1-13: warning: deprecated directive, use ‘%name-prefix’ [-Wdeprecated]
 %name-prefix="constexpYY"
 ^^^^^^^^^^^^^
[...]/src/constexp.y:46.1-13: warning: deprecated directive, use ‘%name-prefix’ [-Wdeprecated]
 %name-prefix="constexpYY"
 ^^^^^^^^^^^^^
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -63,7 +63,7 @@ set_source_files_properties(${GENERATED_SRC}/configoptions.cpp PROPERTIES GENERA
 
 # ce_parse.h
 add_custom_command(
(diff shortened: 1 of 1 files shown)


lrineau:
My build machine has recently been upgraded from CentOS 7.3 to 7.4, and `biso`n was upgraded from 2.7 to 3.0, hence the new warning. Commit e12ec76f044b07d4e4fe167c93103be2879abaca has re-introduced the deprecated directive, probably to be compatible with older `bison` versions. But, if Doxygen uses deprecated directives on purpose, this PR adds `-Wno-deprecated` to the `bison` commands, to remove the warning.

PR #612, opened 2017-10-16
SHOW_USED_FILES_SRC to exclude source files displayed

SamXion:
SHOW_USED_FILES usually lists source and header files (for C++) but this is not always desired because implementation details may not be of interest in a code documentation (this is what we need here at work for our code docs).
    
Introducing SHOW_USED_FILES_SRC configuration option allows to exclude source files from being listed if it is set to NO. By default it is set to YES which resembles the way it was before (i.e. both source and header files are displayed) to remain compatible with old configurations.

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -800,7 +800,7 @@ void ClassDef::findSectionsInDocumentation()
 // add a file name to the used files set
 void ClassDef::insertUsedFile(FileDef *fd)
 {
-  if (fd==0) return;
+  if (fd == 0 || (!Config_getBool(SHOW_USED_FILES_SRC) && fd->isSource())) return;
   if (m_impl->files.find(fd)==-1) m_impl->files.append(fd);
   if (m_impl->templateInstances)
   {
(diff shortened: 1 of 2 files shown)


albert-github:
Regarding the config.xml file. A number of things hat to be improved here (by head) so also some automatic checking can be done (!):
- text for the default is automatically set 
- the dependency can be made explicitly by means of depends='SHOW_USED_FILES'

SamXion:
Thanks @albert-github for the review and suggestions concerning the ```config.xml```. I will update this accordingly!

About ```isSource()``` I am very new to doxygen's code so I am not sure if this is a pointer to use something else instead or if I should probably mention this in the setting's description to make sure people know about its 'limitation'? :) I'll check out ```util.cpp``` in any case.

SamXion:
Ok, hmm, I updated the ```config.xml``` but I couldn't find anything that lets me set the text automatically (if ```defval=''``` is not the one). Can you point me to how that works (the code's ```Argument``` struct does not help either)?

Thanks

PR #614, opened 2017-10-20
fix impossible range checks for chars

DerDakon:
The default type for char is a signed char on most platforms, which is limited to -128..127, so these checks are always true. Explicitely cast to unsigned char to make the checks work as intended.

Spotted by the Clang in TravisCI builds.


doxygen:
The files you changes are actually generated by javacc, so the fix should be made in the code generator of javacc, otherwise it will be lost next time the generator is run.

DerDakon:
Where are the input files of that generator?

doxygen:
the source of the generated files can be found in this file `vhdlparser/vhdlparser.jj`

PR #617, opened 2017-10-23
Provide dot path to plantuml

wich:
If `HAVE_DOT = YES` and `DOT_PATH` is configured, pass the path to the dot executable to plantuml with the `-graphvizdot` option.

Rationale is that plantuml does not look for dot on the path, but assumes dot is at `/usr/bin/dot`. The changes allow dot to be in a non-standard location which is configured in the `Doxyfile`.


albert-github:
Is it guaranteed that the `dotpath `has a directory separator as last character (as `dot` is directly appended to it)?

wich:
Yes, it is guaranteed that it has a directory separator as last character, this is taken care of in `Config::checkAndCorrect` in `configimpl.l`. See also `dot.cpp` where `dot` is also directly appended to the config string.

PR #630, opened 2017-12-07
Changed the navtree collapsed list icon to be more symmetric to the expanded list icon

faheel:
1. The right arrow symbol, `►` (Unicode `&#9658;` "BLACK RIGHT-POINTING POINTER"), used in the navtree to show that a list item is collapsed is thinner at the base and not symmetric to the down arrow symbol, `▼` (Unicode `&#9660;` "BLACK DOWN-POINTING TRIANGLE"), which makes it look a little weird.

    The new symbol, `▶` (Unicode `&#9654;` "BLACK RIGHT-POINTING TRIANGLE"), is symmetric to the down arrow symbol. Here's how it looks in the rendered docs (the one beside "constructors" is how it currently looks, and the one beside "functions" is how it's going to look with the changes):
    
   screenshot-2017-12-22 bigint bigint class reference

1. Made a small typo fix to README.md: changed `github` to `GitHub`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ There are three mailing lists:
 Source Code
 ----------------------------------
 In May 2013, Doxygen moved from 
-subversion to git hosted at github
+subversion to git hosted at GitHub
 * https://github.com/doxygen/doxygen
 
 Enjoy,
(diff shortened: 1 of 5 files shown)


faheel:
@doxygen Your views on this?

PR #632, opened 2017-12-16
Fix the perl script of doc/arch.doc to toggle debug information for a given flex file

mehw:
Found some typos:
- 'CMakefiles' where the 'f' is lowercase instead than uppercase.
- A missing '\n' characater.
- A missing ';' at end of line.

Also the perl regexp to toggle the flex debug flag in build.make searches the non existing string '-P${file}YY'.

### DIFF
--- a/doc/arch.doc
+++ b/doc/arch.doc
@@ -237,7 +237,7 @@ else {
 
 # touch the file
 $now = time;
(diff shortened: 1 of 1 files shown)


mehw:
Thanks for merging!

PR #633, opened 2017-12-18
Add warning suppression pattern

NobodyOnSE:
The motivation for this is stated in https://stackoverflow.com/questions/47876081/disable-warnings-from-tags-files

This is a first try. Currently the balance between writing the code and having performance is less on the performance side: Each warning checks against the pattern list. For large lists and many warnings, this is unfeasible.

A better approach would be to set a per file flag whether to output warnings and to avoid calling the warn functions on them altogether. This should reduce the number of useless patternMatch calls drastically.

For the time being, not using the feature does not seem to cost much:
Parsing my project which adds Qt's `.tags` resulted in following (average of 3) numbers
- Vanilla Doxygen (no suppression pattern support): 4.8s, 1443 warnings
- 0 Suppression Patterns : 4.6s, 1443 warnings
- 1 Suppression Pattern (*.tags): 4.8s, 325 warnings
- 2 Suppression Patterns(*.tags nomatch): 4.5s, 325 warnings
- 3 Suppression Patterns (*.tags nomatch dummy-pattern): 4.8s, 325 warnings

These limited performance tests are inconclusive about the impact of the change. Can you recommend some larger projects to test this with?

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -1201,6 +1201,13 @@ FILE_VERSION_INFO = "cleartool desc -fmt \%Vn"
 ]]>
       </docs>
     </option>
+    <option type='list' id='SUPPRESS_WARNINGS_ON_PATTERNS'>
+      <docs>
+<![CDATA[
+Suppress all warnings (\c WARNINGS , \c WARN_IF_UNDOCUMENTED , \c WARN_IF_DOC_ERROR , \c WARN_NO_PARAMDOC) from files that match one of the given patterns.
+]]>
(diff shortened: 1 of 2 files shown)


doxygen:
Can you check if this has not been already solved by this commit? https://github.com/doxygen/doxygen/commit/aac84d5624b96d8937ff543ab8724c269b8726ab

NobodyOnSE:
Yes, the warning output for the `.tags` files is gone with this commit.

PR #636, opened 2018-01-02
-Fixed: some of remained work to support RTL in HTML output

ahoogol:


albert-github:
Due to the fix:
```
ul {
  overflow: hidden; /*Fixed: list item bullets overlap floating elements*/
}
```
The drop down lists in the main menu disappeared and the scroll bar in the treeview disappeared too. My proposed fix is:
```
#side-nav ul {
  overflow: visible; /* reset ul rule for scroll bar in GENERATE_TREEVIEW window */
}

#main-nav ul {
  overflow: visible; /* reset ul rule for the navigation bar drop down menus */
}
```

I created pull request #754 for this.

I don't know where the original problem occurred that required the original fix if `ul`

ahoogol:
Thanks @albert-github for reporting this issue and sorry for side effects of my changeset.
The original problem was occurred when TOC went floating left in RTL languages, and a UL list appeared in the start of document body. Then in the HTML, the bullets of the list overlaped the floated TOC. I googled that time and found [this solution](https://stackoverflow.com/a/710264/638463).
It was better to assign style to that situation and not to general.

albert-github:
@ahoogol Can you check whether or not my fix doesn't interfere with the problem you saw? or attach a small example showing the problem (I'm not used to RTL languages).

ahoogol:
@albert-github Look at this sample i uploaded here:
[SampleOverlappedBullets.zip](https://github.com/doxygen/doxygen/files/2065299/SampleOverlappedBullets.zip)

ahoogol:
@albert-github, in the treeview the RTL is enforced by inserting LTR and RTL marks in front of titles. but the treeview is on left side always regardless of RTL is enforced or not.

albert-github:
@ahoogol  So this is desired behavior ? or should it also be right alignd but sometjing that doesd not happen (yet)? In the in page TOC (\tableofcontents command in a page with subpages / sections) is right aligned.

ahoogol:
@albert-github, I think this is desired behaviour because all of doxygen pages is not in RTL direction (class and file documentations) so if the treeview go to right and left in navigating successive pages, i think this would not be attractive. I think it will be better to define a new config variable to specify explicity the treeview alignment for the all pages.

ahoogol:
@albert-github, can you attach a example? i dont know your means exactly

albert-github:
Basically your example, only one extra file and I just "played" a bit with the options in the configuration files (especially the OUTPUT* options).

[example.zip](https://github.com/doxygen/doxygen/files/2072082/example.zip)

ahoogol:
@albert-github, first lets us to distinguish between two generated text type: 
1. interface text: that is generated by doxygen translation according to `LANGUAGE` config.
2. user text: that is written by user in files included in `INPUT` config.

My purpose of creating `OUTPUT_TEXT_DIRECTION` was to set style to user text not to interface text. because we (our persian developers) are more comfortable with english interface than translated interface and we always left the `LANGUAGE` to default english.

albert-github:
@ahoogol I do understand, I do prefer also the  English interface texts (I don't always understand the settings from my native language). I understand the reason for the `OUTPUT_TEXT_DIRECTION` so that some documentation can be written in the native language. The later is of course possible when all developers understand that language. I've been working with different nationalities and code suppliers and found some texts in the different languages in the code, also in languages that I or some of my direct colleagues didn't understand (and it was a bit of a nightmare to translate it when we needed to modify the code though Google translate could help us most f the time). I personally prefer to write all comments in English but this my own preference.

My point about the `OUTPUT_LANGUAGE` (although you don't use it) is that Persian is a RTL language and when one sets the `OUTPUT_LANGUAGE` to Persian I think all texts should appear RTL. This is, as far as I can see, now not the case. Am I right here that this is a bug?

Note: please edit `LANGUAGE`  to `OUTPUT_LANGUAGE` as this is the name in the Doxyfile (I nearly got caught off-guard by it).

ahoogol:
Yes, you are right. this is a bug. when one sets the `OUTPUT_LANGUAGE` to Persian, all **interface** text should appear RTL. although the **user** text should be affected by `OUTPUT_TEXT_DIRECTION ` still.

PR #641, opened 2018-01-20
Use Latax syntax rather than Tex syntax for setting lengths

nocnokneo:
Replace occurances of old Tex syntax like:

    \myvar=1mm

with standard Latex syntax:

    \setlength{\myvar}{1mm}

This syntax is more correct and easier for downstream tools like latexdiff to
deal with.


t-b:
@nocnokneo hbadness and tolerance are not lengths. See e.g. https://en.wikibooks.org/wiki/TeX/hbadness and https://en.wikibooks.org/wiki/TeX/tolerance.

nocnokneo:
Whoops, yes, you're right. Fixed.

PR #652, opened 2018-02-07
Fix <see langword="..." /> producing empty text

Cyberboss:
Follow up to my original PR: #613 

Uses the same italicizing `paramref` does


doxygen:
I see Albert made a similar change which I already merged. Can you check if anything is missing, and if so, resolve the merge conflict?

PR #653, opened 2018-02-10
Make it possible to list namespace members in file scope for XML output

mosra:
For better consistency with the HTML output, where each file documentation lists (and links to) all members of given namespace. This also makes it possible to be consistent with the HTML output in case a namespace is not documented and thus all its member detailed docs should be put into corresponding file docs instead.

In order to be backwards compatible and avoid breaking stuff for existing users of the XML output, this is controlled by a new `XML_NAMESPACE_MEMBERS_IN_FILE_SCOPE` configuration option that defaults
to `NO`.

Note that this, unlike the HTML output, will put the whole detailed docs into the file scope instead of just listing them. It's up to the user of the XML output to filter and deduplicate this information. It can be done for example by comparing member ID prefixes with compound ID -- íf different, the detailed docs are already somewhere else.

Cc: @Leandros

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -2881,6 +2881,15 @@ or
 ]]>
       </docs>
     </option>
+    <option type='bool' id='XML_NAMESPACE_MEMBERS_IN_FILE_SCOPE' defval='0' depends='GENERATE_XML'>
+      <docs>
+<![CDATA[
+ If the \c XML_NAMESPACE_MEMBERS_IN_FILE_SCOPE tag is set to \c YES, doxygen
+ will include namespace members in file scope as well, matching the HTML
+ output.
+]]>
+      </docs>
+    </option>
   </group>
   <group name='Docbook' docs='Configuration options related to the DOCBOOK output'>
     <option type='bool' id='GENERATE_DOCBOOK' defval='0'>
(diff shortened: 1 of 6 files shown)


mosra:
@doxygen / @albert-github any chance this could get merged? It's been almost a year since.

albert-github:
Are these the quick member links in HTML ? If so why not use the same type of construct as used in the routine `writeQuickMemberLinks` of `namespacedef.cpp` instead of adding a new configuration tag?

Can you add a small example of how it looks in HTML / an image of the relevant part and the corresponding source code?

mosra:
Um, no, this is not about any quick member links at all. Smallest possible example and expected XML output is already in the test files part of this PR. 

To give you images: If I use the test files to generate HTML, it will look like this for the `*.h` file:

image

with each of these member links pointing to the namespace documentation, which then looks like this:

image

- - -

The thing is, the above information for the `*.h` file **is not exposed to the XML output at all**, because, if you look at the actual code, for some reason there's this:

```cpp
    // namespace members are also inserted in the file scope, but 
    // to prevent this duplication in the XML output, we filter those here.
    if (d->definitionType()!=Definition::TypeFile || md->getNamespaceDef()==0)
```

So this PR adds that, under a new default-to-off configuration option (because I assume just adding these would cause problems for existing tools like Breathe which do *not* expect duplicate members).

doxygen:
I'll rename the option to `XML_NS_MEMB_FILE_SCOPE` (to meet maximum length limitations for proper alignment in the config file) and move the tests to 077 and 078.

mosra:
Excellent, thank you a lot! :)

PR #654, opened 2018-02-12
Consistency for "group" commands

albert-github:
When a non-recognized character appears in a group-command id this is echoed to the console as the default (lex-)rule will be used.
This patch ignores these characters, furthermore in case of a \ in the description this is shown in the output as well.

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -1517,14 +1517,16 @@ RCSTAG    "$"{ID}":"[^\n$]+"$"
   					  if (*yytext=='\n') yyLineNr++;
 					  BEGIN( Comment );
   					}
+<GroupDocArg1>.				{ // ignore other stuff
+  					}
 <GroupDocArg2>"\\"{B}*"\n"		{ // line continuation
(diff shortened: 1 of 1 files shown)


albert-github:
For an example:
```
/** @file
  * some details
  */


/**
  * @defgroup c_q"uot\lists  \\par C \c Quoted Lists title
  * \"C Quoted Lists details
  * @{
  */
  int C_SEL_NONE        = 0; /**< "NONE, not present */
/**
  * @}
  */


/**
  * a function
  * @param par  "description of par"
  * @param par2 description of par
  */
void fie(int par, int par2);
```
 In version i.8.14 (and earlier this gives:
```
...
\\\\Building group list...
...
and with the patch the \\\\ is gone
```

albert-github:
Alternative implementation could be (analogous to section labels):
```
<SectionLabel>.                         { // invalid character for section label
                                          warn(yyFileName,yyLineNr,
                                              "Invalid or missing section label"
                                              );
                                          BEGIN(Comment);
                                        }
```

PR #661, opened 2018-02-20
Bug 793644 - Long functionnames with links are not split in latex output when the line is to short.

albert-github:
Regression on Bug 778730 - doxygen build fails
Stakexchange TEX article: When I prevent hyphenation using an \mbox, the box gets pushed into the right margin (https://tex.stackexchange.com/questions/53364/when-i-prevent-hyphenation-using-an-mbox-the-box-gets-pushed-into-the-right-ma)

Implemented \sloppy preventing the overflow.


albert-github:
The solution does not work for all situations and furthermore in the refman.tex file there is also a remark about \sloppy being obsolete and some code from \sloppy used directly.

PR #684, opened 2018-03-27
Bug 794567 - Doxygen crash when using \code{.markdown}...\endcode in VHDL source

albert-github:
The handling of the code command in comment is handled when entire comment is handled.
Handling was incomplete only the backslash version was checked (and crashed).

### DIFF
--- a/src/vhdljjparser.cpp
+++ b/src/vhdljjparser.cpp
@@ -80,7 +80,6 @@ static struct
   int iDocLine;
 } str_doc;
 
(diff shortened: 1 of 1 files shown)


doxygen:
I see only code being removed in the commit. Not being too familiar with the VDHL code myself, I just want to double check if this was this indeed the intention.

albert-github:
Yes, deleting the code was intentionally as the handling of the comment is done already more general (consistency) and the handling in the VHDL part was incomplete and resulted in the mentioned crash.

PR #697, opened 2018-04-09
Test 32 reference to bell signal (XHTML, LaTeX)

albert-github:
The handling of ^G (bell signal, ASCII value 7) in XHTML and LaTeX was not done properly (error message / crash).


doxygen:
@albert-github I've made it a bit more generic by escaping all ASCII control characters, not just the bell character.

albert-github:
Should not be a problem. I'm always a bit afraid of the characters that have a specific meaning, like 
- Horizontal Tab
- Line Feed
- Vertical Tab
- Form Feed
- Carriage Return

PR #698, opened 2018-04-09
Improvement LaTeX output

albert-github:
For a number of languages the output in LaTeX has been improved so the regular documentation can be generated (automatically) in LaTeX / PDF as well.


albert-github:
Fixes also:
- [Bug 774459 - LaTeX Korean package should be changed from {hfont} to {kotex}  ](https://bugzilla.gnome.org/show_bug.cgi?id=774459)
- [Bug 789976 - PDFLATEX_CMD_NAME option missing](https://bugzilla.gnome.org/show_bug.cgi?id=789976)

PR #703, opened 2018-04-11
Path for external commands on windows

albert-github:
In case a path in the name of an executable contains one or more '/' characters they are replaced by a '\\'.
- inspired by the cause for bug 766059
- https://stackoverflow.com/questions/49750869/what-is-the-qhg-location-path-relative-to-for-doxygen

### DIFF
--- a/src/portable.cpp
+++ b/src/portable.cpp
@@ -196,7 +196,6 @@ int portable_system(const char *command,const char *args,bool commandHasConsole)
     }
   }
 #endif
-  fprintf(stderr,"AME ==> we should not get here\n");
   return 1; // we should never get here
 
 }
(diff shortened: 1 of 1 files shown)


albert-github:
Related issue:
- https://bugzilla.gnome.org/show_bug.cgi?id=792846
- https://stackoverflow.com/questions/48394181/is-it-possible-to-wrap-a-doxygen-filter-command-in-quotes

doxygen:
@albert-github I've moved the substitution code to a separate function, and optimized it a bit. Didn't test it on Windows yet. Can you check if my corrected code still works?

albert-github:
@doxygen Unfortunately, doesn't compile:
```
...\src\portable.cpp(39) : error C3861: 'substitute': identifier not found
...\src\portable.cpp(466) : error C3861: 'substitute': identifier not found
```

included `util.h` in `portable.cpp` it compiles.

I cannot find my relevant test environment for this bug but the code check looks OK (except for the problem with `portable.cpp` of course).

doxygen:
Added the missing include.

PR #706, opened 2018-04-14
Include "empty" directories documenting themselves.

c-lipka:
Non-code directories can now be documented via a `.dox` (or similar) file in the directory itself.

Previously, directories would only be included in the documentation (e.g. "File list" tree or modules' "Directories" section) if they contained any _code_ input files (as opposed to _documentation_ input files, e.g. `.dox` or `.md` files), preventing the use of `.dox` files to document otherwise empty or non-code directories in a repository.

With this pull request, directories are instead included if they contain _any_ input files, provided at least one of them is a code file _or_ the directory is documented using a corresponding `\dir` special command. (The `\dir` command may reside anywhere, but the primary use case is for it to reside in a `.dox` file in the directory itself.)

The exact mechanism is that `buildDirectories()` will now add the directory of _any_ input file to the tree, regardless of code vs. documentation status. (Note that this also adds _undocumented_ documentation-only directories, which need to be filtered out further downstream, but such mechanisms seem to already be in place.)

### DIFF
--- a/src/dirdef.cpp
+++ b/src/dirdef.cpp
@@ -930,14 +930,14 @@ void buildDirectories()
     for (;(fd=fni.current());++fni)
     {
       //printf("buildDirectories %s\n",fd->name().data());
-      if (fd->getReference().isEmpty() && !fd->isDocumentationFile())
+      if (fd->getReference().isEmpty())
       {
         DirDef *dir;
         if ((dir=Doxygen::directories->find(fd->getPath()))==0) // new directory
         {
           dir = DirDef::mergeDirectoryInTree(fd->getPath());
         }
-        if (dir) dir->addFile(fd);
+        if (dir && !fd->isDocumentationFile()) dir->addFile(fd);
       }
       else
       {


abathur:
@c-lipka Do you have a broader perspective on the relationship between files, directories, and documentation input files? I ran into something a little odd, but I'm not sure if it's a bug or not. You didn't cause this--just asking in case you have context since the previous blame for these lines is 14 years old.

I noticed that a markdown file containing a page compound (a documentation file, I guess) will get documented as a "file", but it won't get the usual directory -> innerFile -> file relation that source code files have. 

I could imagine this being intentional, to keep documentation-only-files, by themselves, from appearing as both pages and files?

abathur:
@c-lipka No worries! I know that feeling. :)

AlexeyKryshen:
Hi,

What should I do now if I don't want my /docs folder with *.dox files to appear in files list in generated documentation ? My generated documentation now contains useless empty "docs" folder under files list. And I want to avoid it. Any suggestions ?

By the way, I am surprised your pull request was accepted. It is silently change the well-established Doxygen behavior and IMHO it is very inconvenient change. *.dox files are *.dox files, they are more technical details of documenting, not the ones to be documented by themselves. AFAIK, most projects have /docs folder with *.dox files to customize generated docs and they are not considered as a part of documentation for a good reason. Imagine the C library for example. If I generate doc for library users, they are interesting only in library API, and I don't see any single reason to include my /docs folder into the documentation. Indeed, what could be a reason to do so? 

At least such behavior change should be guarded by config switch, don't you think so ? It must be disabled by default and enabled explicitly via Doxygen config switch. It was not a bug to exclude *.dox only folder from file list. I am sure it was done for a reason (to exclude *.dox from generated documentation).

Best regards, Alexey

abathur:
Alexey,

I'm not sure if my reply will be helpful since I'm not a maintainer and I don't have any experience with this feature--I only noticed your comment because I previously posted in this thread.

> What should I do now if I don't want my /docs folder with *.dox files to appear in files list in generated documentation ? My generated documentation now contains useless empty "docs" folder under files list. And I want to avoid it. Any suggestions ?

1. Go back to using Doxygen 1.8.14 or whichever version you were using before updating. If this version worked fine before, it should still work fine, now.
2. Clone Doxygen, revert the commit associated with this pull request, rebuild Doxygen, and enjoy life.
3. according to the initial post by @c-lipka your directory should only be in the generated documentation if one of two things are true:
    - "at least one of them [files in the doc/ directory] is a code file"
    - "the directory [itself] is documented using a corresponding \dir special command. (The \dir command may reside anywhere, but the primary use case is for it to reside in a .dox file in the directory itself.)"

    You could check both of these things (particularly, ensuring your directory isn't documented with a special command) and see if your directory is properly excluded.
4. If neither of those things are true in your case, then it's possible there is a bug in the pull request. You could submit a bug report with enough information for someone else to reproduce the issue (possibly by creating a repo on github with a minimal test case), which will make it easier to fix.

> By the way ... Best regards, Alexey

This isn't helpful. Open-source software like Doxygen improves when real people with busy lives manage to find some spare time, energy, and motivation and choose to donate them to the community. It's great to report when a change causes you trouble--that's how we know what to improve--but it's not okay to waste people's time and energy with demotivating attacks on their contributions.

AlexeyKryshen:
Hi,

I found at least one case not handled properly by this PR (and which lead to observed wrong behavior actually): this PR is not well suited for configuration with enabled EXTRACT_ALL option.

Well, technically things going right as "expected" in some sense:
- EXTRACT_ALL cause all things to be added to generated docs, regardless of is there documentation provided for the entity or not (see `dirdef.cpp : generateDirDocs(OutputList &ol)` and `definition.cpp : Definition::hasDocumentation() const`)
- changes introduced by this PR unconditionally add unwanted dirs to list of entities (particularity, to list of dirs)

The problem is: previously Doxygen for a good reason provide a behavior that *.dox-only folders are not added to the list of entities to documentation (it was handled exactly by code this PR changed). Unfortunately, in a case of EXTRACT_ALL, changes introduced by this PR, broke the expected old behavior and provide no way to workaround it (but I may be wrong here: it author know how to workaround it, it would be nice to hear a words from him).

Best regards, Alexey

c-lipka:
> - The changes this PR brings to Doxygen look suspicious. Indeed, there are only two line of code changed and only effect I could imagine from static code analysis of that change is than any directory from INPUT tag of Doxyfile (and subdirs, if recursive) will be added to documentation if it contains at least any of input file (being it source code, or Doxygen accompanying *.dox files).

Yes, that is a very acute observation - and your surprise about the simplicity of this change, and the absence of an immediate test whether the directory is commented or not, has been shared by the author from the moment they created it. But their tests indicated that it worked, so they conjectured the pre-existence of such a filter downstream.

Still seeing this as a risk, the author of the pull request decided to explicitly mentioned this in the pull request description:

>> Note that this also adds undocumented documentation-only directories, which need to be filtered out further downstream, but such mechanisms seem to already be in place.
>
> What mechanisms author means ? Why not to express it explicitly ? Does author really sure they are exists ? (see below)

The author means _some_ mechanism inside Doxygen that they observed to exist, but - due to the complexity of the code, and the high percentage of meta-programmed code preventing the efficient use of common debugging tools - never managed to positively identify.

> I found at least one case not handled properly by this PR (and which lead to observed wrong behavior actually): this PR is not well suited for configuration with enabled EXTRACT_ALL option.

Well, I guess you just managed to identify a loophole in the filter mechanism (and maybe even the mechanism itself) the author of the PR was seeing at work.

My initial instinct would be to look at ways to fix the behaviour specifically for the `EXTRACT_ALL` case, maybe by having that mechanism specifically not kick in for directories; but I guess that might open up other cans of worms, so I concede that it may indeed the most prudent thing to add a configuration switch, to specify whether non-source directories can or cannot be documented. The `EXTRACT_ALL` mechanism could continue to invariably include all document-able entities, and it would be controlled by that other new config switches whether doc-only directories would qualify as document-able or not.

If you or someone else would like to pick up that glove, I would be happy to let you; if not, I'll look into this as soon as I find sufficient time (which may take a while though; mind you, I'm not working on the Doxygen codebase on any regular basis, so I'd first have to find my way back; for someone already "in the flow", adding a config switch should be a matter of minutes).

c-lipka:
> I really don't want to be offensive in any way,

Then... maybe don't?

It's actually quite simple: Focus on what _is_ wrong (and how it can be fixed), not how it _went_ wrong. Keep any assumptions on the latter matter (whether substantiated or otherwise) to yourself.

(Unless of course you are a quality manager; but then the question "what _went_ wrong [with this product]" would be the wrong one, except as a first step towards answering the question "what _is_ wrong (and how can it be fixed) [with our production process]".)

AlexeyKryshen:
> ... I concede that it may indeed the most prudent thing to add a configuration switch ...

Let's hold for a while. Currently I explore Doxygen code in attempt to find a way to fix it without a configuration switch and preserve EXTRACT_ALL old behavior. It is my first dive into Doxygen code, but surprisingly at higher level it well-structured and well-designed, so maybe solution will not be too complex.

I would suggest to move our further discussion to [my issue report](https://github.com/doxygen/doxygen/issues/6803). It seems to be right place to discuss possible ways ti fix the issue. Don't you mind ?

PR #712, opened 2018-05-02
Problem with TEST_FLAGS when using cmake for Ninja

albert-github:
The command sequence:
`    cmake -G Ninja -D CMAKE_BUILD_TYPE=Release ..`
`    cmake --build . -- -j 3 -v || echo error && cd .. && exit 1`
results in the error message:
`    ninja: error: build.ninja:2384: bad $-escape (literal $ must be written as $$)`

In respect to `LEX_FLAGS` and `YACC_FLAGS` (the later still required some quotes), similar problems appeared in the past and this was solved by setting the (local) `LEX_FLAGS` and `YACC_FLAGS` to the empty string. A similar solution didn't work directly for the `TEST_FLAGS` so here the code has been duplicated once with the flags and once (for Ninja) without the flags.


albert-github:
The problem regarding adding it to .travis.yml is not up to me.
The moment the tests are more geared towards functionality of doxygen and not towards the build systems.  The build systems are changing less frequently and also another important build (system) is not tested: Windows with nmake. Also the different compiler versions and flavors of *nix are not tested.

albert-github:
Made obsolete by means of pull request 722

PR #717, opened 2018-05-10
Bug 769414 - PHP: New array syntax not supported when parsing initial value

albert-github:
Implemented possibility string initializations by means of '[' ... ']' construct (besides the already existing 'array(' ... ')').


albert-github:
Related issue #6041

PR #722, opened 2018-05-13
Problem with TEST_FLAGS when using CMake for Visual Studio

tusharpm:
I tried to run the tests on Windows using ctest (and the custom `tests` target), and got the following output:
```
    Start 1: suite
1/1 Test #1: suite ............................***Failed    0.10 sec
usage: runtests.py [-h] [--updateref] [--doxygen [DOXYGEN]]
                   [--xmllint [XMLLINT]] [--id IDS [IDS ...]] [--all]
                   [--inputdir [INPUTDIR]] [--outputdir [OUTPUTDIR]]
                   [--noredir] [--xml] [--xhtml] [--pdf] [--keep]
runtests.py: error: unrecognized arguments: $(TEST_FLAGS)


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.10 sec

The following tests FAILED:
          1 - suite (Failed)
Errors while running CTest
```
I went through #712 which solves this problem for Ninja.
I'm trying to solve it for _all_ generators by moving the logic for `TEST_FLAGS` to `runtests.py`.


tusharpm:
To get the details and output of tests, one can use `ctest --verbose`; however, one must use it sparingly as it prints a lot.
For granular testing with ctest, the individual tests need to be added in CMake (or a separate CTestTestfile.cmake needs to be generated by some means).

PR #724, opened 2018-05-14
Bug 796094 - Backslashes in default values confuse the parser (and cause params to be ignored)

albert-github:
Corrected assignment / counting.


Mekk:
I briefly tested the fix, looks like it resolves the issue in both the example case I reported, and in true samples I use in actual code.

PR #736, opened 2018-05-21
Enable in page table of contents for LaTeX

albert-github:
currently the \\tableofcontents command is only supported for HTML.
In this patch:
- enable in page table of contents for LaTeX: \\tableofcontents['{'[option[:level]][,option[:level]]*'}']
     where option can be 'HTML',  'LaTeX' or 'XML'
- add the possibility to have maximum levels in the in page table of contents (possible per output type and per in page table of contents. Default is show all)

(side effect: possibility to have options with, nearly, all commands.)

### DIFF
--- a/doc/autolink.doc
+++ b/doc/autolink.doc
@@ -16,7 +16,7 @@
  */
 /*! \page autolink Automatic link generation
 
-  \tableofcontents
+  \tableofcontents{html,latex}
 
   Most documentation systems have special `see also' sections where links
   to other pieces of documentation can be inserted.
--- a/doc/changelog.doc
+++ b/doc/changelog.doc
@@ -1,5 +1,5 @@
 /** \page changelog Changelog
-\tableofcontents
+\tableofcontents{html,latex}
 \section log_1_8 1.8 Series
 
 \subsection log_1_8_14 Release 1.8.14
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -2016,15 +2016,19 @@ Make sure you have first read \ref intro "the introduction".
 \endverbatim
 
 <hr>
-\section cmdtableofcontents \\tableofcontents['{'[option][,option]*'}']
+\section cmdtableofcontents \\tableofcontents['{'[option[:level]][,option[:level]]*'}']
 
   \addindex \\tableofcontents
   Creates a table of contents at the top of a page, listing all
-  sections and subsections in the page. The option can be `HTML` or `LaTeX`.
-  In case no option is specified \c \\tableofcontents acts as if just the
-  option `HTML` was specified. In case of multiple \c \\tableofcontents
-  commands in a page the option(s) will be used additional to the already
-  specified option(s).
+  sections and subsections in the page. The `option` can be `HTML` or `LaTeX`
+  or `XML`. When a `level` is specified this means the maximum nesting level
+  that is shown. The value of `level` should be in the range 1..5, values outside
+  this range are considered to be 5. In case no `level` is specified `level` is
+  set to 5 (show all)
+  In case no `option. is specified \c \\tableofcontents acts as if just the
+  `option` `HTML` was specified. In case of multiple \c \\tableofcontents
+  commands in a page the `option`(s) will be used additional to the already
+  specified `option`(s), but only the last `level` of an `option` is valid.
 
   \warning This command only works inside related page documentation and
            \e not in other documentation blocks and only has effect in the
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -2016,15 +2016,19 @@ Make sure you have first read \ref intro "the introduction".
 \endverbatim
 
 <hr>
-\section cmdtableofcontents \\tableofcontents
+\section cmdtableofcontents \\tableofcontents['{'[option][,option]*'}']
 
   \addindex \\tableofcontents
   Creates a table of contents at the top of a page, listing all
-  sections and subsections in the page. 
+  sections and subsections in the page. The option can be `HTML` or `LaTeX`.
+  In case no option is specified \c \\tableofcontents acts as if just the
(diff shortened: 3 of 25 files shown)


mosra:
In #625 I implemented emitting the `\tableofcontents` placeholder for the XML output, but without any "type" attribute, just an empty tag. 

Now, the `image` command is exported to XML like this:

```xml
<image type="html" name="image.png"></image>
```

Would you consider adding similar attribute (`types`?) to the XML `<tableofcontents/>` element as well to make it consistent with `<image>`? Thank you! :+1:

albert-github:
I think this should be possible as well, the only thing is that it has to be defined how the XML-tags should look like, especially in the `xsd` file. As soon as I have a `xsd` definition I can have a look at the implementation.

mosra:
One idea (`<tableofcontents>` having `<include type="html" />` children):

```patch
diff --git a/templates/xml/compound.xsd b/templates/xml/compound.xsd
index b4356acd..d00da253 100644
--- a/templates/xml/compound.xsd
+++ b/templates/xml/compound.xsd
@@ -29,7 +29,7 @@
       <xsd:element name="innergroup" type="refType" minOccurs="0" maxOccurs="unbounded" />
       <xsd:element name="templateparamlist" type="templateparamlistType" minOccurs="0" />
       <xsd:element name="sectiondef" type="sectiondefType" minOccurs="0" maxOccurs="unbounded" />
-      <xsd:element name="tableofcontents" minOccurs="0" maxOccurs="1" />
+      <xsd:element name="tableofcontents" type="tableofcontentsType" minOccurs="0" maxOccurs="1" />
       <xsd:element name="briefdescription" type="descriptionType" minOccurs="0" />
       <xsd:element name="detaileddescription" type="descriptionType" minOccurs="0" />
       <xsd:element name="inheritancegraph" type="graphType" minOccurs="0" />
@@ -120,6 +120,16 @@
     <xsd:attribute name="kind" type="DoxSectionKind" />
   </xsd:complexType>
 
+  <xsd:complexType name="tableofcontentsType">
+    <xsd:sequence>
+      <xsd:element name="include" type="tableofcontentsKindType" minOccurs="0" maxOccurs="unbounded" />
+    </xsd:sequence>
+  </xsd:complexType>
+
+  <xsd:complexType name="tableofcontentsKindType">
+      <xsd:attribute name="type" type="DoxOutputTypeKind" />
+  </xsd:complexType>
+
   <xsd:complexType name="memberdefType">
     <xsd:sequence>
       <xsd:element name="templateparamlist" type="templateparamlistType" minOccurs="0" />
@@ -547,7 +557,7 @@
 
   <xsd:complexType name="docImageType" mixed="true">
     <xsd:group ref="docTitleCmdGroup" minOccurs="0" maxOccurs="unbounded" />
-    <xsd:attribute name="type" type="DoxImageKind" /> 
+    <xsd:attribute name="type" type="DoxOutputTypeKind" />
     <xsd:attribute name="name" type="xsd:string" /> 
     <xsd:attribute name="width" type="xsd:string" /> 
     <xsd:attribute name="height" type="xsd:string" /> 
@@ -837,7 +847,7 @@
     </xsd:restriction>
   </xsd:simpleType>
 
-  <xsd:simpleType name="DoxImageKind">
+  <xsd:simpleType name="DoxOutputTypeKind">
     <xsd:restriction base="xsd:string">
       <xsd:enumeration value="html" />
       <xsd:enumeration value="latex" />
```

However, thinking more about this, I think it might be more flexible to have multiple `\tableofcontents` commands, each with just one "output type", like below:

```
\tableofcontents{html}
\tableofcontents{latex}
```

This could *later* (no, I'm *not* suggesting to implement this as part of this PR ;)) allow additional options for example to limit depth of the TOC box, differently for each output type, for example:

```cpp 
/* Include only one level of headers for HTML output but two in the LaTeX output */
\tableofcontents{html,1}
\tableofcontents{latex,2}
```

(In that case the XML would be just e.g. `<tableofcontents type="html"/ >`, no nested sequence of tags.)

albert-github:
Thanks for the thoughts.

The already proposed extension of the implementation of the `\tableofcontents ` command is `\tableofcontents['{'[option][,option]*'}']` it is also to use multiple `\tableofcontents ` commands on a page (the the `\tableofcontents ` are concatenated). In this implementation the 'level' mechanism would be a bit harder to implement and I'm not sure if it would be nice from a consistency point of view, but it is an interesting thought.

I'll see what I can achieve with the suggested `xsd` change.

albert-github:
Updated the pull request in respect to XML and also implemented the possibility for maximum levels shown in the in page table of contents.

albert-github:
Restructuring is not a bad idea.
I  Have to check more thoroughly but I think the code in types.h:
```
    bool isHtmlEnabled()  const { return (m_mask & Html)!=0;  }
    bool isLatexEnabled() const { return (m_mask & Latex)!=0; }
    bool isXmlEnabled()   const { return (m_mask & Xml)!=0;   }
```
is not correct and should read:
```
    bool isHtmlEnabled()  const { return (m_mask & (1 << Html))!=0;  }
    bool isLatexEnabled() const { return (m_mask & (1 << Latex))!=0; }
    bool isXmlEnabled()   const { return (m_mask & (1 << Xml))!=0;   }
```

albert-github:
Also in `commands.doc` is a small problem (not restructure problem, but one in the original patch) around line 2040 it reads:
> In case no \`option.

this has to be
> In case no \`option\`

doxygen:
Indeed, see https://github.com/doxygen/doxygen/commit/b6c724e8ab5caeef1e3d065e4c2fa05740d87859 for the fixes

doxygen:
and https://github.com/doxygen/doxygen/commit/ae25f5c41545e17b37d67ab7dabf422f0eb757aa for the documentation fix

albert-github:
Looks OK now, except for the `.` after \`option\`

PR #743, opened 2018-05-26
Bug 796355 - LaTeX: Class scrbook Error: undefined old font command `\tt'

albert-github:
- Replaced \tt by \texttt
- Replaced the left over \bf by \textbf
and taking care that { } still match.

### DIFF
--- a/src/latexdocvisitor.cpp
+++ b/src/latexdocvisitor.cpp
@@ -218,7 +218,7 @@ void LatexDocVisitor::visit(DocURL *u)
     if (u->isEmail()) m_t << "mailto:";
     m_t << u->url() << "}";
   }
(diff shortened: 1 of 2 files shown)


t-b:
@albert-github Why not use `\bfseries` and friends instead? Using \text*{} has the disadvantage that everything inside the {} must not be fragile. See [1] and [2].

[1]: https://tex.stackexchange.com/a/139592
[2]: https://tex.stackexchange.com/a/4747

albert-github:
Analogous to the previous changes for \bf

the pairs \texttt and \ttfamily as well as \textbf and \bfseries are used, reason to me is unknown why. Maybe this should be made more consistent.

PR #755, opened 2018-06-04
Bug 796494 - Collapsed treeview arrow displays as emoji in Microsoft Edge

albert-github:
The right  arrow (\&#9654;) is in Microsoft Edge displayed as an Emoji it displays as a white arrow inside a blue square with a black border instead of a monochrome arrow. Unicode contains another right arrow (\&#9258;) which displays correctly (also in Chrome, Opera, FireFox, Internet Explorer; in some even better).

The Unicode naming is a bit confusing:
- for the down arrow we only have:
  - \&#9660;  'BLACK DOWN-POINTING TRIANGLE'
- for the right arrow we have:
  - \&#9654; 'BLACK RIGHT-POINTING TRIANGLE'
  - \&#9658; 'BLACK RIGHT-POINTING POINTER'

so it looks logical to use both Triangles but this is not the case.

### DIFF
--- a/src/ftvhelp.cpp
+++ b/src/ftvhelp.cpp
@@ -257,7 +257,7 @@ void FTVHelp::generateIndent(FTextStream &t, FTVNode *n,bool opened)
   while (p) { indent++; p=p->parent; }
   if (n->isDir)
   {
-    QCString dir = opened ? "&#9660;" : "&#9654;";
+    QCString dir = opened ? "&#9660;" : "&#9658;";
     t << "<span style=\"width:" << (indent*16) << "px;display:inline-block;\">&#160;</span>"
       << "<span id=\"arr_" << generateIndentLabel(n,0) << "\" class=\"arrow\" ";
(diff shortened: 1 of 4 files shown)


lrineau:
You should check that you have not reverted https://github.com/doxygen/doxygen/pull/630.

albert-github:
@lrineau 
This looks like to revert #630 indeed, but from the tests I did the result did look like to be OK (for the browsers I did check). Secondary point was the under Microsoft Edge the wrong symbol was  shown.

I used the following file to have a look at the (initial) results, I used it as aa.h (.h extension not accepted by github in attachments)
[aa.txt](https://github.com/doxygen/doxygen/files/2076455/aa.txt)

Do you have an indication that the current versions of the arrows don't look OK in the different browsers?

lrineau:
With Firefox 60.0, I confirm that the look of the triangles is now like the screenshot of the first message of #630: the two triangles (down-pointing and right-pointing) look different.

PR #761, opened 2018-06-10
Bug 395343 - caller graph can be improved by having caller on left and callee on right

albert-github:
Caller graph will now have the called function on the right hand side.


doxygen:
I've slightly changed the code such that lrRank and rank are now combined. See https://github.com/doxygen/doxygen/commit/cbca2bdcd77ce0ab6b907cf2d4bab65340a57f00. Please check if I didn't break anything.

albert-github:
According to the, small, test case I have it looks OK.

PR #765, opened 2018-06-12
Bug 779407 - plantuml:an unwanted newline is generated after @startuml

albert-github:
Added special plantuml commands (don't like white space / newlines in front).


doxygen:
I decided to address the root cause instead; i.e. the fact that doxygen added a newline that the user didn't type: see https://github.com/doxygen/doxygen/commit/032d8fd0cd8bd33448186081c54168889e25d55e

albert-github:
Looks all OK (why didn't I think about this :-( ))

PR #768, opened 2018-06-18
Bug 784752 - Spurious error message when a conditional block contains a defgroup

albert-github:
Some commands (a.o. \defgroup) terminate the comment scanner and afterwards the comment scanner is "continued" but the state is not remembered but should have been (at least in respect to the "\if" stack).
To be in a defined stated the state is explicitly reset when a new file (block) is started).


albert-github:
Seen the tests with the original issue I think the problem has been solved in the mean time

PR #771, opened 2018-06-18
Bug 734308 - Error message when using memberof in a C macro

albert-github:
Made a better warning message, a 'define' is a global setting and cannot be made a member of a struct / class.

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -671,6 +671,10 @@ void ClassDef::internalInsertMember(MemberDef *md,
               case MemberType_Variable:
                 addMemberToList(MemberListType_variableMembers,md,FALSE);
                 break;
+              case MemberType_Define:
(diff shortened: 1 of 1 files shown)


fakenmc:
Hi, thanks for the bug fix, but defines still can't be related to a class :/

albert-github:
@fakenmc In case you think it is a bug:
- Please open a new issue
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.
- Please also specify the doxygen version used.

fakenmc:
Will do, thanks!

fakenmc:
I have rechecked this, the \relates tag does indeed work in Doxygen 1.8.15 for relating function-like macros to classes.
Thanks again.

PR #773, opened 2018-06-19
Bug 796621 - @cond does not stop at @endcond Fortran

albert-github:
The backslash (\) is in Fortran not an escape character but was handled as such, this has been corrected.
Handling of the different Fortran string types ('...' and "...") made more consistent.


albert-github:
Also solves: Bug 596872 - Fortran: Backslashes in literals are always treated as special chars

PR #780, opened 2018-06-28
Bug 691689 - Line numbers for examples

albert-github:
\examplelineno to have the possibility to specify per example whether or not line numbers are shown.


doxygen:
Wouldn't it be nicer to have a `{lineno}` option to `\example` instead of a new command?

doxygen:
I've now made the snippetdoc, snippetlineno, includedoc, and includelineno obsolete (they still work but are no longer recommended to use), and added the lineno and doc options to \snippet, and \include, and lineno option to \example. 

Note that I also introduced QCStringList (next to QStringList) to avoid conversions from QCString to QString and back.

This is the commit 0b4b3698b29436b299d4e4a315d610bc1ab98acb

PR #783, opened 2018-06-30
Enable possibility of CLANG for Cygwin

albert-github:
- During compilation there were some problems with strdup (POSIX), replaced by qstrdup
- During compiling / linking there were some problems with killpg which was recognized, added standard definition.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -13,6 +13,10 @@
  *
  */
 
+#if !defined(_WIN32) || defined(__CYGWIN__)
+#define _DEFAULT_SOURCE 1
+#endif
+
 #include <locale.h>
 
 #include <qfileinfo.h>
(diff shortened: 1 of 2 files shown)


src/doxygen.cpp
```diff
@@ -113,6 +113,10 @@ extern void initResources();
 #if !defined(_WIN32) || defined(__CYGWIN__)
 #include <signal.h>
 #define HAS_SIGNALS
+extern "C"
+{
+  int killpg(pid_t pgrp, int sig);
```
doxygen: I don't really like forward declaring of standard functions. Can you check if putting `#define _DEFAULT_SOURCE` before the `#include <signal.h>` also works? Looking at the header, this seems to trigger that the killpg prototype becomes declared.

albert-github: I agree that having the forward declaration of standard functions is not the nicest way but I thought it would be the least dangerous one in this case. I added the `_DEFAULT_SOURCE` define at the beginning of the file (should be before the first invocation of `features.h` otherwise it has not the desired effect).

I updated the pull request.

PR #784, opened 2018-07-01
Tooltip was twice 'HTML escaped'

albert-github:
The tooltip was already 'HTML escaped' except for some '<' and '>' characters but was again 'HTML escaped' resulting in e.g. that a single apostrophe (') was translated to `&#39;` and again to `&amp;#39;` resulting in `&#39;` in the tooltip.
Only '<' and '>' are now 'HTML escaped again'

### DIFF
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -514,6 +514,27 @@ void HtmlCodeGenerator::docify(const char *str)
   }
 }
 
+void HtmlCodeGenerator::docify_tt(const char *str)
+{
+  m_t << getHtmlDirEmbedingChar(getTextDirByConfig(str));
+
+  if (str && m_streamSet)
+  {
+    const char *p=str;
+    char c;
+    while (*p)
+    {
+      c=*p++;
+      switch(c)
+      {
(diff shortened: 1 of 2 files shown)


doxygen:
Hi Albert, do you have an example of a twice escaped HTML tooltip, as I don't see it like that in the code?

I tried the following with `SOURCE_BROWSER = YES`:

```
/** @brief A class with > and < and &. */
class Test {};
```

albert-github:
Hi Dimitri,

The description is so that only `<` and `>` have to be escaped again and not a character like `&`.

The example: [pull_784_.zip](https://github.com/doxygen/doxygen/files/2169460/pull_784_.zip)

It is based on: https://stackoverflow.com/questions/43852075/doxygen-parsing-ampersands-for-ascii-chars
Just hover over the functions in the code.

doxygen:
Looks like the problem is in TextDocVisitor::filter; this produces the unneeded escapes. So I've fixed this instead.

albert-github:
Checked against the original problem as well (#378) and this looks still OK as well.

Issue #4529, opened 2018-07-02
HTML tags <u> and </u> not supported (Origin: bugzilla #667993)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.7.6.1* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[TestDocs.zip](https://sourceforge.net/projects/doxygen/files/attachments/bug667993_205341_TestDocs.zip/download/download)_ (ID 205341)

On 2012-01-16 10:08:13 +0000, Bernd Giesen wrote:

    Created attachment 205341
    Sample project for "HTML tags <u> and </u> not supported"
    
    Unlike the <b></b> and <i></i> the <u></u> tags for underlining comment text
    are not supported in doxygen comment blocks. See the "Detailed class
    description" in the attached sample project (Test.chm)

albert-github:
I've just pushed a proposed patch #6384

albert-github:
Code has been integrated in master on github.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.16.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6384, opened 2018-07-07
Bug 667993 - HTML tags <u> and </u> not supported

albert-github:
Added underline possibility and strike through possibility for the different output formats insofar it is possible (other similar possibilities are not always possible for all output formats either).


albert-github:
Related issue #4529

Issue #3677, opened 2018-07-02
<![CDATA[ is not handled inside C# comments (Origin: bugzilla #610436)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.6.2* on platform *Other*
Assigned to: Dimitri van Heesch

On 2010-02-19 09:33:37 +0000, tkrotoff@gmail.com wrote:

    Example of a C# valid comment:
    
    /// <code>
    /// <![CDATA[
    ///     <<
    ///     >>
    /// ]]>
    /// </code>
    
    <<
    >> is the piece of text we want to show inside a code tag.
    Since < and > and reserved characters by XML, we have to "escape" them using
    <![CDATA[ otherwise Visual Studio 2008 will complain.
    
    Problem with Doxygen 1.6.2 is that the generated HTML documentation contains
    <![CDATA[ and should not.

albert-github:
I've just pushed a proposed patch, pull request #6398.

albert-github:
Code has been integrated in master on github.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.15.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6398, opened 2018-07-19
Bug 610436 - <![CDATA[ is not handled inside C# comments

albert-github:
Added handling of "\<![CDATA[ ... ]]>" i.e. XML CDATA sections in an analogous way top HTML comment but in this case the text is retained (as normal doxygen comment)  and the special XML characters `<`,`>` and `&` are taken as if they were escaped.


albert-github:
Related issue #3677

PR #6409, opened 2018-07-26
Bug 635155 - Bad protection type in xml output for private union

albert-github:
The protection for a.o. union was reset to "public" but but it should stay (analogous to e.g. enum) on the protection of the section (left Object-C untouched).
In the HTML documentation the "protected" and "private" labels were not considered at all in case of e.g. union


albert-github:
A bit more extended test case:  [aa.cpp.txt](https://github.com/doxygen/doxygen/files/2231990/aa.cpp.txt)

PR #6413, opened 2018-07-27
Issue 6411: CSS for Markdown tables do not use HTML_COLORSTYLE_HUE, HTML_COLORSTYLE_SAT config variables

albert-github:
Corrected a number of inconsistencies in the doxygen css file (e.g. a markdown table looked different from a HTML table, see header when changing colors) by using the same color placeholder.

### DIFF
--- a/templates/html/doxygen.css
+++ b/templates/html/doxygen.css
@@ -134,12 +134,12 @@ a.qindex {
 a.qindexHL {
 	font-weight: bold;
 	background-color: ##AA;
-	color: #ffffff;
+	color: ##ff;
 	border: 1px double ##98;
 }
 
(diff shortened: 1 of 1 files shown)


albert-github:
The proposed patch is to have a more consistent set of colors. Doxygen uses a placeholder for some colors in the template doxygen.css (starting with ##) and these values are filled by means of the `HTML_COLORSTYLE_...` settings. In the template doxygen.css there are some more colors used, these are listed in the attached file: [remaining_colors.txt](https://github.com/doxygen/doxygen/files/2235853/remaining_colors.txt). In the 'placeholder' list there are no placeholders for the referenced colors (they probably should not be changed when the 'color theme'  changes.

PR #6420, opened 2018-08-02
Making environment variables available in all `\...only` parts

albert-github:
In the documentation it was stated that environment variables were available in `\htmlonly`, `\latexonly` and `rtfonly`, but actually it was only available in `\htmlonly`.
- made environment variables available in all `\...only` parts as-is, so actual text from environment variable.
- adjusting documentation


albert-github:
Related to pull related to #6419

PR #6430, opened 2018-08-06
 'self' keyword in PHP documentation

albert-github:
In PHP the word `self` was not color coded and the functions were not referenced in the CALL / CALLER graphs, REFERENCES / REFERENCED lists.


albert-github:
Wrong mentioning of issue #6430 above

Issue #6175, opened 2018-07-02
plantuml:an unwanted newline is generated after @startuml (Origin: bugzilla #779407)


doxygen:
status NEW severity *minor* in component *build* for *---*
Reported in version *1.8.13* on platform *Other*
Assigned to: Dimitri van Heesch

On 2017-03-01 09:55:52 +0000, swmatisa@gmail.com wrote:

    When the following valid comment with PlantUML is added in a file:
    /** 
    @startuml
    version
    @enduml
    */
    
    An unwanted newline is generated after @startuml. The content of generated
    inline_umlgraph_1.pu is:
    @startuml
    
    version
    @enduml
    
    
    This is an invalid plantUML file because "version" must be at the fist line
    after @startuml
    
    The correct content of generated inline_umlgraph_1.pu is:
    @startuml
    version
    @enduml

On 2018-06-12 17:01:02 +0000, albert wrote:

    Added the following special plantuml commands:
    author
    checkversion
    licence
    license
    sudoku [<seed>]
    testdot
    version
    
    I've just pushed a proposed patch to github (pull request 765,
    https://github.com/doxygen/doxygen/pull/765).

albert-github:
Alternative implementation has been integrated in master.

albert-github:
See also regression pull request #6452

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.15.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6452, opened 2018-08-19
Bug fix (very tiny) for plantuml

cheoljoo:
# What is the problem
-  Most of @startuml are running well. But , the following plantuml is not processed. 
```
 /**
 * @brief Clear the rule.
 * @startuml
 * participant CReceiver
 * opt MsgFromSender()
 *  alt cmd=IGN_STATUS_IN && value == 0
 *     alt
 *            CReceiver ->> CHmi : Draw_HMI sendMsg() ->  sendMessage(HMI,MSGIDN,...)
 *      else
 *          note right CReceiver : Draw Directly
 *      end
 *  else
 *      note right CReceiver : mbBlockDraw = false
 *  end
 * end
 * @enduml
 */
```

# debugging method
- change into DOT_CLEANUP = NO in Doxyfile   to remain the plantuml file for debugging.
- the following context is result of remaining plantuml file.
```
@startuml
  participant CReceiver
  opt MsgFromSender()
   alt cmd=IGN_STATUS_IN && value == 0
      alt
             CReceiver ->> CHmi : Draw_HMI sendMsg() ->  sendMessage(HMI,MSGIDN,...)
       else
           note right CReceiver : Draw Directly
       end
   else
       note right CReceiver : mbBlockDraw = false
   end
  end@enduml
```

# Solution
- add return in front of @enduml

### DIFF
--- a/src/plantuml.cpp
+++ b/src/plantuml.cpp
@@ -47,7 +47,7 @@ QCString writePlantUMLSource(const QCString &outDir,const QCString &fileName,con
   }
   QCString text = "@startuml\n";
   text+=content;
-  text+="@enduml\n";
+  text+="\n@enduml\n";
   file.writeBlock( text, text.length() );
   file.close();
   return baseName;


albert-github:
Regression due to commit 032d8fd0cd8bd33448186081c54168889e25d55e (pull request #765) where all leading and trailing returns were removed.

cheoljoo:
Dear albert.

#765 and #6175 is remove the trailing returns after @startuml.
#6175 does not generate the error. 
But , this patch is different.  If you do not apply this patch , **we can not see the plantuml pictures** because plantuml.jar can not generate the png file.  
plantuml.jar does not support the "end@enduml" .
So we should add the return between end and @enduml.

Could you try to run with upper example again?

With BestRegards.
Charles.

albert-github:
Dear Charles,

I just mentioned that through the fix in #765 / #6175 the problem occurred, the leading (necessary) and trailing (problematic) were removed. Your fix fixes the problem regarding the required return at the end.
(In version 1.8.14 everything was working except for commands like 'version').

PR #6465, opened 2018-08-24
Dramatically reduce the time when rerun the doxygen with plantuml comments (Imporvement of performance)

cheoljoo:
# What is the problem
- When we make doxygen output , each plantuml comment spends a lot of time (more than 1 second).
	- If doxygen generates 6 plantuml , it takes 18.124 sec (98%). But total time is 18.498 sec (100%).
	- Generating plantuml on java occupied most of time.
- It regenerate all plantuml output even though we change or add just one plantuml comments  or we change the doxygen comments.
	- Reuse the pre-generated plantuml output!!

# Main Idea
- Goal
	- Reduce the running time of doxygen from making plantuml result.
- Method
	- Reuse the result(png or svg ...) of unchanged plantuml comments.

```
9.689 sec: Total elapsed time: 9.689 seconds
(of which 9.340 seconds waiting for external tools to finish)
```
- According to upper timeline ,  generating images with java is a big burden.
- So I try to remove java running process.   I guess we can reduce the java running time if we reuse already generated images.

# What to do
	- PLANTUML_CACHE = YES
	- PLANTUML_CACHE_DEBUG_PRINT = YES			<- It has dependency on PLANTUML_CACHE
- We reduced the time to run the java process for making the plantuml result (png, svg ...).
- Changed Configuration (Doxyfile)
	- HAVE_DOT               = YES
	- DOT_IMAGE_FORMAT       = png
		- generate  png  file from plantuml
	- GENERATE_DOCBOOK       = YES
		- generate  png  file from plantuml
	- GENERATE_PERLMOD       = YES
		- make a perl module
- results
	- Original : original doxygen makes html and docbook plantuml png separately. (my example has 3 plantuml comments in source code.)
	- First Trial [^1] : When we use modified version , html is running first and then reuse the html's plantuml png results for generating docbook.
		- reduce the time to make a docbook plantuml result.
	- Second Trial [^2] : Change One comment : When we change one of three plantuml comments, we can reuse the previous compiled doxygen result.
		- reduce the time to make a plantuml for unchanged plantuml.
	- Change output type [^3] : When we change output format png into svg, we can resue the previous compiled doxygen result about same contents..
		- DOT_IMAGE_FORMAT       = png  -> svg
		- reduce the time to make a plantuml for unchanged plantuml with the same format..

|        Test Type          |  Original  |  First [^1] | Change One [^2] | Change output type [^3] |
|---------------------------|------------|-------------|-----------------|-------------------------|
| Elapsed Time              | 18.498 sec | 9.689 sec   |  3.153 sec      |  9.744 sec              |
| Guessing Time (Only HTML) | 9.498 sec  | 9.498 sec   |  3.153 sec      |  9.744 sec              |

- Analysis from results
	- Although you use only html type ,  you can get the benefit from second trials.

[^1] : First - When we use modified version , html is running first and then reuse the html's plantuml png results for generating docbook.
[^2] : Change One - When we change one of three plantuml comments, we can reuse the previous compiled doxygen result.
[^3] : Change output type - When we change output format png into svg, we can reuse the previous compiled doxygen result about same contents..

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3491,6 +3491,22 @@ to be found in the default search path.
 ]]>
       </docs>
     </option>
+    <option type='bool' id='PLANTUML_CACHE' defval='0'>
+      <docs>
+<![CDATA[
+If the \c PLANTUML_CACHE tag is set to \c YES, doxygen will
+create the intermediate plantuml files (*.cache.pu & others) that are used to improve the performance.
+]]>
+      </docs>
+    </option>
+    <option type='bool' id='PLANTUML_CACHE_DEBUG_PRINT' defval='0' depends='PLANTUML_CACHE'>
+      <docs>
+<![CDATA[
+If the \c PLANTUML_CACHE_DEBUG_PRINT tag is set to \c YES, doxygen will
+print how to reuse this cache and how to improve the performance.
+]]>
+      </docs>
+    </option>
     <option type='int' id='DOT_GRAPH_MAX_NODES' minval='0' maxval='10000' defval='50' depends='HAVE_DOT'>
       <docs>
 <![CDATA[
--- a/src/plantuml.cpp
+++ b/src/plantuml.cpp
@@ -19,8 +19,13 @@
 #include "doxygen.h"
 #include "index.h"
 #include "message.h"
+#include "util.h"
 
 #include <qdir.h>
+#include <string>
+#include <map>
+
+#include <dirent.h>
 
 static const int maxCmdLine = 40960;
 
@@ -53,11 +58,31 @@ QCString writePlantUMLSource(const QCString &outDir,const QCString &fileName,con
   return baseName;
 }
 
+
+struct comparer
+{
+    public:
+    bool operator()(const std::string x, const std::string y) const
+    {
+         return x.compare(y)<0;
+    }
+};
+
+extern void restoreCacheFromFile(std::string *path);
+static std::map<std::string, std::string, comparer> *puMap = nullptr;	//* plantuml Map < plantuml_contents , plantuml.[format] binary name> 
+static std::map<std::string, int, comparer> *puDir = nullptr;		//* directory Map
+
 void generatePlantUMLOutput(const char *baseName,const char *outDir,PlantUMLOutputFormat format)
 {
   static QCString plantumlJarPath = Config_getString(PLANTUML_JAR_PATH);
   static QCString plantumlConfigFile = Config_getString(PLANTUML_CFG_FILE);
   static QCString dotPath = Config_getString(DOT_PATH);
+
+  if(Config_getBool(PLANTUML_CACHE_DEBUG_PRINT)){
+	  printf("config PLANTUML_CACHE : %d\n",Config_getBool(PLANTUML_CACHE));
+    printf("config PLANTUML_CACHE_DEBUG_PRINT : %d\n",Config_getBool(PLANTUML_CACHE_DEBUG_PRINT));
+    printf("config DOT_CLEANUP : %d\n",Config_getBool(DOT_CLEANUP));
+  }
 
   QCString pumlExe = "java";
   QCString pumlArgs = "";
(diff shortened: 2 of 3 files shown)


albert-github:
The caching of the information in one way or the other is a good idea.
When generating the the `dot ` images the `.dot` files are written and based on this `.dot` files a `MD5` sum is calculated. With subsequent calculations the `.dot` file is written as well as the `MD5` sum is calculated. when the `MD5` sums are identical the image is not regenerated. Furthermore with `dot` images these are generated in multiple threads.
Another possibility would be to collect all plantuml requests in one file and process this file. Plantuml will generated a new image for each original plantuml request. Something similar is done for formulas. This gives less overhead as only 1 process is started.

I think one of these techniques would be sufficient.


Regarding the code (small review):

- in case of debug you use printf statements, I think that here there better options would be
  - create an -d flag (e.g. -d plantuml) that gives the information, so no need for using the configuration option
 - usage of printf
   - printf should not be used,all output is steered over calls to routines as defined in message.cpp, 
 - usage of std::string / std::map
    - in doxygen the string manipulation is. mostly, done by means of Qt classes (e.g. classes QCString, QString). For consistency these classes should be used.
    - map manipulation is also done by means of Qt classes )dictionary and list classes)
 - dirent / opendir /readdir
    - when  browsing through directories the routines line `readFileOrDirectory` should probably be used or the techniques used in this routine

cheoljoo:
I uploaded new version in #6467

cheoljoo:
I uploaded new version in #6467

trec789:
Hello cheoljoo,

I believe I am having some issues with this change.
When using Doxygen 1.8.16 with html output the plantuml images won't update. I know that it is the intention but it causes the following problems:
- if I remove the html files in the output folder (for whatever reason) the plantuml images are not recreated.
- if I change the PLANTUML_CFG_FILE config param in Doxyfile, e.g. for changing the diagrams color, the plantuml images won't be updated.

The issues seems to be related to Doxygen 1.8.16 only. Doxygen 1.8.15 works fine since it always re-create the images.
I tried to add PLANTUML_CACHE = NO without success.

My system:
Windows 10
Doxygen 1.8.16
plantuml.jar 1.2019.11
Graphviz 2.38.0

albert-github:
Do you mean that you removed the individual html files or the complete html directory ?

A workaround will be to remove also the file `inline_umlgraph_cache_all.pu`, probably it is in the directory where you started doxygen.

albert-github:
Problems like noted with the `PLANTUML_CFG_FILE` will occur when:
- `PLANTUML_JAR_PATH`, the jar file is replaced for another / new one (name does not have to change though!)
- `PLANTUML_INCLUDE_PATH` , when the content changes or the contents of the file. This is quite difficult to detect as an include can include again files from other places.

**Note** this is not the pull request that landed into release but the pull request #6723 did. The setting `PLANTUML_CACHE` does not exist here anymore

albert-github:
@trecmar please open a new issue with the mentioned problems as otherwise the problems might get lost. **Don't reopen** this pull request (or the #6723 pul request) for this purpose.

trecmar:
Hi,

Regarding the first email:
- I removed the individual html files only, not the html dir.
- Removing the inline_umlgraph_cache_all.pu worked. Thank you for the tip.

Second email:
- I agree that change the configuration causes a change in the output, but is difficult to detect.
- I would suggest adding some kind of Doxygen command line flag to force a “clean build” or “rebuild all” type of thing.

Third email:
- I will open a new issue as suggested and add this info.

Thanks,
Marc

cheoljoo:
sorry @trecmar  for my late response.  albert gave the answer.  thanks @albert-github 
when you remove html files in html direcotry , you should remove inline_umlgraph_cache_all.pu file also.
because doxygen will decide after comparison with inline_umlgraph_cache_all.pu whether doxygen create the plantuml png files or not.

and I will try to add some configure options in Doxyfile. but doxygen board members decide that this is default.   because we can find the bug easily when users use this option as default.

Thanks all.  cheoljoo.

PR #6467, opened 2018-08-27
Imporvement of performance : Reduce the Java running count

cheoljoo:
# What is it
- #6465 is the first reqeust.  So I followed up the albert advice.
	- in case of debug you use printf statements, I think that here there better options would be
		- create an -d flag (e.g. -d plantuml) that gives the information, so no need for using the configuration option
		- Follow up : create -d plantuml [debug.h, debug.cpp]
	- usage of printf
		- printf should not be used,all output is steered over calls to routines as defined in message.cpp,
		- Follow up : use Debug::Print(Debug::Plantuml, ...    and msg()
	- usage of std::string / std::map
		- in doxygen the string manipulation is. mostly, done by means of Qt classes (e.g. classes QCString, QString). For consistency these classes should be used.
		- map manipulation is also done by means of Qt classes (dictionary and list classes)
		- Follow up : use QDict QList QCString [plantuml.h, plantuml.cpp]
	- dirent / opendir /readdir
		- when browsing through directories the routines line readFileOrDirectory should probably be used or the techniques used in this routine
		- Follow up : I will follow when I try to reduce redundancy of creating plantuml image.

- Keep the seperate pu files  because of easy debugging.

- Run java minimally
	- Test Case : 4 plantuml
	- Original Run without PLANTUML_RUN_JAVA_ONCE
		- 8.2 sec = 4 times * ( 1.6(java vm) + 0.1 * 1(load multiple file) + process each plantuml (0.35) *1 ) <- prediction
	- New with PLANTUML_RUN_JAVA_ONCE
		- 3.499 sec = 1.6(java vm) + 0.1 * 4(load multiple file) + process each plantuml (0.35) *4 <- prediction
	- Creating Java Virtual Machine has a big portion.

- Result
	- Improving Performance : 8.2 seconds -> 3.499 seconds  (in case of 4 plantuml)
	- Performance in theory  :  1 to 4 times faster 
 

# Configuration
	- PLANTUML_RUN_JAVA_ONCE = YES

# Debugging
- doxygen -d plantuml
- doxygen -d time -d plantuml

# Tested with following options
- source code of 3 plantuml
- PLANTUML_RUN_JAVA_ONCE = YES  or    NO
- DOT_IMAGE_FORMAT       = png  or    svg
- DOT_CLEANUP            = YES  or    NO
- -d plantuml  -d time

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3491,6 +3491,14 @@ to be found in the default search path.
 ]]>
       </docs>
     </option>
+    <option type='bool' id='PLANTUML_RUN_JAVA_ONCE' defval='0'>
+      <docs>
+<![CDATA[
+If the \c PLANTUML_RUN_JAVA_ONCE tag is set to \c YES, doxygen will run java once for improving the performance.
+This will run once at the end of process like dot graphics processing.
+]]>
+      </docs>
+    </option>
     <option type='int' id='DOT_GRAPH_MAX_NODES' minval='0' maxval='10000' defval='50' depends='HAVE_DOT'>
       <docs>
 <![CDATA[
--- a/src/debug.cpp
+++ b/src/debug.cpp
@@ -48,6 +48,7 @@ static LabelMap s_labels[] =
   { "markdown",     Debug::Markdown     },
   { "filteroutput", Debug::FilterOutput },
   { "lex",          Debug::Lex },
+  { "plantuml",     Debug::Plantuml },
   { 0,             (Debug::DebugMask)0  }
 };
 
--- a/src/debug.h
+++ b/src/debug.h
@@ -37,7 +37,8 @@ class Debug
                      ExtCmd       = 0x00000400,
                      Markdown     = 0x00000800,
                      FilterOutput = 0x00001000,
-                     Lex          = 0x00002000
+                     Lex          = 0x00002000,
+                     Plantuml     = 0x00004000
                    };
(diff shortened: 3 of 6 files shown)


albert-github:
I ran a simple test too, 40 pu files (distributed over 2 directories), and went from 44 to 18 sec. so also a considerable improvement.

In the debug statements the `__PRETTY_FUNCTION__` is a GCC extension (see e.g. https://stackoverflow.com/questions/4384765/whats-the-difference-between-pretty-function-function-func) and should not be used. On native Windows (I use the 2013 version i.e. 18.00.31101) and the `__PRETTY_FUNCTION__` is unknown. I also had problems with `__FUNC__` and `__func__` so best is just not to use it or use `__FILE__` the debug will give enough information in that way too.

- In the `PLANTUML_RUN_JAVA_ONCE ` all the `.pu` files are given to the `plantuml.jar` with their full path (see `-d extcmd` when running). I think this will lead to problems when there are quite a few files ("command line" overflow). Seen this merging the files into one file like done for formulas. A small test with one file revealed that it didn't bring a lot / nothing time wise, but here the problem with the "command line" overflow shouldn't / can't happen.
- I think it still would make sense to implement the 'md5' possibility for, especially, the non `PLANTUML_RUN_JAVA_ONCE ` version, i.e. if the file already exists don't try to recreate the image.

PR #6472, opened 2018-08-29
Improvement of performance when use plantuml : Reduce the Java running count (3rd trial)

cheoljoo:
Improvement of performance : Reduce the Java running count (3rd trial)

# History
- 1st : #6465 
- 2nd : #6467 

# What is it
- The following is comments from ( #6467 ) second trial. 
	- In the debug statements the __PRETTY_FUNCTION__ is a GCC extension. But this project can run in windows.
		- I got it.  I removed __PRETTY_FUNCTION__.
	- In the PLANTUML_RUN_JAVA_ONCE all the .pu files are given to the plantuml.jar with their full path (see -d extcmd when running). I think this will lead to problems when there are quite a few files ("command line" overflow). Seen this merging the files into one file like done for formulas. A small test with one file revealed that it didn't bring a lot / nothing time wise, but here the problem with the "command line" overflow shouldn't / can't happen.
		- Yes. The command line has overflow problem.  So I change into one file.
	- I think it still would make sense to implement the 'md5' possibility for, especially, the non PLANTUML_RUN_JAVA_ONCE version, i.e. if the file already exists don't try to recreate the image.
		- This is a next item.
			- PlantumlManager::instance()->insert() : this function will be changed to skip image.

# Configuration
	- PLANTUML_RUN_JAVA_ONCE = YES


cheoljoo:
I will release the new pull request.

PR #6474, opened 2018-08-30
WIP: Init Windows AppVeyor CI

Croydon:
This adds AppVeyor configuration to tests doxygen builds in the following build matrix:

[VS Studio 2015, 2017] x [Release, Debug] x [32bit, 64bit]

To enable AppVeyor for this repository, you need to go to https://ci.appveyor.com/projects, login with your GitHub account and add the repository as a project.

The result will look like this: https://ci.appveyor.com/project/Croydon/doxygen/build/1.0.6

Voila, continuous testing for Windows 😄

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,9 +22,7 @@ init:
 install:
   - ps: Invoke-WebRequest https://github.com/lexxmark/winflexbison/releases/download/v2.5.15/win_flex_bison-2.5.15.zip -OutFile flex.zip
   - 7z x flex.zip -oC:\deps\flex
-  - rename C:\deps\flex\win_bison.exe bison.exe
-  - rename C:\deps\flex\win_flex.exe flex.exe
-  - ps: Invoke-WebRequest https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs923/gs923w32.exe -OutFile gswin32c.exe
+  - ps: Invoke-WebRequest https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs924/gs924w32.exe -OutFile gswin32c.exe
   - gswin32c /S /D=C:\deps\ghostscript
   - ps: choco install -y miktex
   - refreshenv
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,36 @@
+image: Visual Studio 2017
+
+configuration:
+  - Release
+  - Debug
+
+platform:
+  - x64
+  - Win32
+
+environment:
+  # VS VERSION IN CMAKE STYLE
+  matrix:
+    - VSVERSION: "15 2017"
+    - VSVERSION: "14 2015"
+
+init:
+  - cmake --version
+
+install:
+  - ps: Invoke-WebRequest https://github.com/lexxmark/winflexbison/releases/download/v2.5.15/win_flex_bison-2.5.15.zip -OutFile flex.zip
+  - 7z x flex.zip -oC:\deps\flex
+  - rename C:\deps\flex\win_bison.exe bison.exe
+  - rename C:\deps\flex\win_flex.exe flex.exe
+  - set "PATH=%PATH%;C:\deps\flex"
+
+before_build:
+  - if "%platform%"=="Win32" ( set "CMAKE_GENERATOR_NAME=Visual Studio %VSVERSION%" )
+  - if "%platform%"=="x64" ( set "CMAKE_GENERATOR_NAME=Visual Studio %VSVERSION% Win64")
+  - mkdir build
+  - cd build
+  - cmake .. -G "%CMAKE_GENERATOR_NAME%
+
+build:
+  project: "build\\PACKAGE.vcxproj"
+  parallel: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,21 +16,35 @@ environment:
 
 init:
   - cmake --version
+  - perl --version
+  - msbuild /version
 
 install:
   - ps: Invoke-WebRequest https://github.com/lexxmark/winflexbison/releases/download/v2.5.15/win_flex_bison-2.5.15.zip -OutFile flex.zip
   - 7z x flex.zip -oC:\deps\flex
   - rename C:\deps\flex\win_bison.exe bison.exe
   - rename C:\deps\flex\win_flex.exe flex.exe
-  - set "PATH=%PATH%;C:\deps\flex"
+  - ps: Invoke-WebRequest https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs923/gs923w32.exe -OutFile gswin32c.exe
+  - gswin32c /S /D=C:\deps\ghostscript
+  - ps: choco install -y miktex
+  - refreshenv
+  - pip install conan
+  - conan install libxml2/2.9.8@bincrafters/stable -g virtualrunenv
+  - activate_run.bat
+  - set "PATH=%PATH%;C:\deps\ghostscript\bin;C:\deps\flex"
 
 before_build:
   - if "%platform%"=="Win32" ( set "CMAKE_GENERATOR_NAME=Visual Studio %VSVERSION%" )
   - if "%platform%"=="x64" ( set "CMAKE_GENERATOR_NAME=Visual Studio %VSVERSION% Win64")
   - mkdir build
   - cd build
-  - cmake .. -G "%CMAKE_GENERATOR_NAME%
+  - cmake -G "%CMAKE_GENERATOR_NAME%" ..
 
 build:
   project: "build\\PACKAGE.vcxproj"
   parallel: true
+
+test_script:
+  - msbuild "testing\tests.vcxproj" /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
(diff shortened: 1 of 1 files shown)


albert-github:
Looks a **nice** enhancement to me.

@Croydon  Any reason why to build the Release and Debug version?

Croydon:
> @Croydon Any reason why to build the Release and Debug version?

I have encountered several times in the past that only one version breaks, so testing both seems more reliable. Any reason not to do this?

albert-github:
@Croydon  No reason not to do it but wanted to have it clear why it is necessary.

(I also added 2 extra questions to the original question).

appveyor.yml
```diff
@@ -30,6 +30,8 @@ before_build:
   - mkdir build
   - cd build
   - cmake .. -G "%CMAKE_GENERATOR_NAME%
+  - cmake .. -G "%CMAKE_GENERATOR_NAME% --target tests
+  - cmake .. -G "%CMAKE_GENERATOR_NAME% --target docs
```
lrineau: Have you tested that commit?

1. I find strange that it is in the `before_build:` section.
2. `--target` has no sense for CMake unless `--build <dir>` is also used.

albert-github:
In the lines like `- cmake .. -G "%CMAKE_GENERATOR_NAME%` aren't we missing the closing double quote  so it would read `- cmake .. -G "%CMAKE_GENERATOR_NAME%"`

Croydon:
@albert-github Already fixed.

Please give me a few minutes to test and get it this right before reviewing it again 😄

Croydon:
The tests seem to have several dependencies https://ci.appveyor.com/project/Croydon/doxygen/build/job/agcjyrcy8hm3ihin#L1065

Since I'm not familiar with Doxygen's build system/requirements, could somebody point me to a complete list of requirements?

The installation page is talking about LaTex and Ghostscript, but where is xmllint coming from?

albert-github:
`xmllint` is used to compare the results from the generation of the `xml` output as used by the tests. The output of `xmllint` is given to the `diff` command to see if there are differences or not between the official version and the new version. In case of differences the submitter of the change can see if this is expected (and update the test) or not (check / correct the new code ).

Croydon:
I was rather asking from a packaging perspective. Seems like xmllint is part of libxml2.

albert-github:
Looks like still missing `perl` (and `bibtex`), found the log files.

Croydon:
The tests are looking for a file `bib4404.bbl`. I can't see this file within the current portable version of MiKTeX. What version does it need?

Croydon:
`xmllint.exe` is in PATH. See https://ci.appveyor.com/project/Croydon/doxygen/build/1.0.18/job/200ucg1jjhwo8dgu#L58 & https://ci.appveyor.com/project/Croydon/doxygen/build/1.0.21/job/g3uutalwgwws3daa#L130

Still xmllint fails to be executed https://ci.appveyor.com/project/Croydon/doxygen/build/1.0.21/job/g3uutalwgwws3daa#L1048

Any idea why @albert-github?

albert-github:
Very hard to say, some ideas, directly after the `PATH` statement:
  - `xmllint --help` 
  - dependency walker ( or similar) on `xmllint`
  - You probably looked into the `runtests.py` on the `testing` directory line 155 - 160 as well, maybe some print statements might help here to see what happens. A possibility here is also to see if any other command can run with `os.popen`.

(when I run the tests on my system, I run it with `nmake` and `-G "NMake Makefiles"`)

albert-github:
What happens when you rename the `zlib.dll` form the used directory to `zlib1.dll`?
(at my system xmllint from libxml2 complains about missing `zlib`.dll`)

Croydon:
I noticed the same locally and moving a zlib1.dll file locally into the same directory fixes it for me.

However, in the CI it doesn't seem to help.

albert-github:
You wrote "moving a zlib1.dll file locally into the same directory fixes it for me". Is this the same `zlib.dll` / `zlib1.dll` as the one obtained with `nuget`? 
 - Can we at any place download the used `zlib1.dll` / the `nuget` file? 
 - is it possible to "push" in a doxygen independent way the `zlib1.dll` that you used or is it necessary to add a directory for AppVeyor CI with this `dll`? (the extra directory is not something very nice though, but for test purposes acceptable, at least I think it is with proper description where it comes from etc. and correct license. Maybe even build it from https://zlib.net/).

Croydon:
It's downloadable from here https://www.nuget.org/api/v2/package/grpc.dependencies.zlib.redist/1.2.8.10

I just tried the `build\native\bin\v140\Win32\Release\dynamic\cdecl\zlib.dll` locally (after renaming it to zlib1.dll) and it worked. CI still complains with exactly this file.

albert-github:
Same results on this side, I tried the `cdecl` version and it worked, the `stdcall` failed with a missing entry point. Very strange that it fails.
I think it might be worthwhile to drop a line to the people of App Veyor CI or to the forum for some suggestions.

Croydon:
So I dropped this outdated, unmaintained xmllint Windows binaries and the nuget stuff and I am using now a clean solution by using [Conan](http://conan.io) to get the xmllint dependency.

And the tests work fine now 😄

albert-github:
Double thumbs up for the perseverance, good that the tests work now.

Regarding the documentation build, might it be that there is a problem with the 32-bit version. Looks like the 32-bit version (slightly outdated but not that bad) is used for Win32 and x86. 
When I run I get(as said before I use nmake):
```
...
[100%] Generating Latex and HTML documentation.
This is pdfTeX, Version 3.14159265-2.6-1.40.19 (MiKTeX 2.9.6745 64-bit)
entering extended mode
(_formulas.tex
LaTeX2e <2018-04-01> patch level 5

[100%] Built target run_doxygen
Scanning dependencies of target doxygen_pdf
...
```
Have you tried  to run locally with vcproj as well? ( I did that in the past but during development I find nmake and on Cygwin make more convenient, maybe a bit old fashioned)

Croydon:
Both 32bit and 64bit Doxygen builds are getting tested and Windows 32/64bit respectively, so I don't think that it has something to do with the 32bit version.

I can try to use the command line installer version instead of the portable one.

albert-github:
Might have to do something with the needed (LaTeX) packages, I'm not sure if all that are needed have been loaded by default (don't think so), so MikTeX should load them on the fly (was an idea I got from some questions on stack).
I searched on stack, resulting in the following URL: https://stackexchange.com/search?q=appveyor+miktex
maybe these articles can give some hints as well, otherwise it might be useful to post a question there.

albert-github:
Added remark to previous comment.

albert-github:
I saw that some more progress has been made!

One question is the renaming necessary in:
```
  - ps: Invoke-WebRequest https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs923/gs923w32.exe -OutFile gswin32c.exe
  - gswin32c /S /D=C:\deps\ghostscript

```
isn't it possible to have the more (in my opinion) clear:
```
  - ps: Invoke-WebRequest https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs923/gs923w32.exe -OutFile gs923w32.exe
  - gs923w32 /S /D=C:\deps\ghostscript
```

Croydon:
I have done the renaming so that it doesn't need to be changed at multiple places when Ghostscript is getting updated in the future. And I have decided to use the `gswin32c` name since this is the name of the binary we need.

Croydon:
The remaining issue(s) seems like a problem of doxygen itself. I might not be able to help with this further.

albert-github:
Good results sofar. A bit strange that some builds succeed and some don't, looks a bit like a race condition.
Might be that in the doxygen CMake configuration a dependency is missing ( I don't see a dependency  regarding the "configvalues.h").
In the file appveyor.yml I see the line:
`  parallel: true`
this makes in principle sense but might hurt here, would it be possible to remove this?

albert-github:
Looks like the debug version takes a very long time, a bit strange seen one of the previous results where debug version ran through quite fast (build 1.0.59) (but this was with the 2015 compiler.
Might have a number of reasons:
  - debug version of 2017 is significantly slower than the 2015 version (I don't have data for this)
  - the systems are slower due to a higher total load(1.059 ran during the night from Sunday to Monday)

To get some data I think best is to leave the job running and (when necessary) kill just the sub job. Based on the data a decision can be made. The slow build cannot come, in my opinion, from the `parallel: false` as this is only relevant for the build itself.

Croydon:
It doesn't run slow, it completely freezes at the test step. I have already restated this build because I was thinking that this is very odd, same result.

It makes no sense to me, why it should work with parallelization, but not without.

Croydon:
I'm sure that it has nothing to do with being slow. It absolutely freezes at the test step at around 9 minutes and stays there for the remaining 51 minutes.

albert-github:
I fully agree that is very likely that it freezes somewhere, but it would be good to have the 'split' tests as in that case we might have an indication where to start to search.

Croydon:
Agreed, but I'm also not sure how to do that yet.

Croydon:
This pull request wasn't ready yet to get merged. Two of the builds still keep failing.

Croydon:
Also it needs activation on AppVeyor as described in the pull request description

PR #6475, opened 2018-08-30
Sorting of index in case of LaTeX

albert-github:
In case of LaTeX the sorting was so that lowercase came after uppercase e.g.: `username` came after `useSsl`, although the index should be case insensitive.
The problem was that the sort key was filtered in such a way that a.o. uppercase symbols were preceded by `\+` for hyphenation. The key doesn't need this hyphenation (as there is a separate field for the display name).
The `\+` has been filtered out now.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -6865,11 +6865,37 @@ void filterLatexString(FTextStream &t,const char *str,
   }
 }
 
+static void reFilterLatexString(FTextStream &t,const char *str)
+{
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github Instead of filter and then refiltering I noticed that enabling insideTabbing will also give the desired escaping directly.

albert-github:
Looks indeed that the new filter is not necessary and that the insideTabbing enabling works.

PR #6482, opened 2018-09-03
update to the experimental sqlite3 generator

abathur:
This PR expands the breadth, depth, and quality of the experimental sqlite3 generator. The databases it generates are _approaching_ parity with the XML output. 

Overview here; see next 2 comments for: _notes on review_, and _loose threads / open questions_.

(WIP: This is close, but I hope to clean up a few loose threads here. The diff is big; I realize this may take a while. I don't mind jumping through hoops.)

### Overview of changes by file

- *sqlite3gen.cpp*: Refine schema to simplify common queries. Fix duplicate/junk data. Collect most kinds, fields, and relations.

- *xmlgen.cpp/h*: Declare XMLCodeGenerator in xmlgen.h so sqlite3gen can render brief/detailed/inbodydescription fields as XML.

- *addon/doxypysql/search.py*: Update per schema changes; support Python 3.

### DIFF
--- a/src/sqlite3gen.cpp
+++ b/src/sqlite3gen.cpp
@@ -23,11 +23,15 @@
 #include "qtbc.h"
 #include "sqlite3gen.h"
 #include "doxygen.h"
+#include "xmlgen.h"
+#include "xmldocvisitor.h"
 #include "config.h"
 #include "util.h"
+#include "outputlist.h"
 #include "docparser.h"
 #include "language.h"
 
+#include "version.h"
 #include "dot.h"
 #include "arguments.h"
 #include "classlist.h"
@@ -35,65 +39,129 @@
 #include "namespacedef.h"
 #include "filename.h"
 #include "groupdef.h"
+#include "membername.h"
+#include "memberdef.h"
 #include "pagedef.h"
 #include "dirdef.h"
+#include "section.h"
 
+#include <sys/stat.h>
 #include <qdir.h>
 #include <string.h>
 #include <sqlite3.h>
 
-//#define DBG_CTX(x) printf x
-#define DBG_CTX(x) do { } while(0)
+// enable to show general debug messages
+// #define SQLITE3_DEBUG
 
-const char * schema_queries[][2] = {
+// enable to print all executed SQL statements.
+// I recommend using the smallest possible input list.
+// #define SQLITE3_DEBUG_SQL
+
+# ifdef SQLITE3_DEBUG
+#  define DBG_CTX(x) printf x
+# else // SQLITE3_DEBUG
+#  define DBG_CTX(x) do { } while(0)
+# endif
+
+# ifdef SQLITE3_DEBUG_SQL
+// used by sqlite3_trace in generateSqlite3()
+static void sqlLog(void *dbName, const char *sql){
+  msg("SQL: '%s'\n", sql);
+}
+# endif
+
+const char * table_schema[][2] = {
+  /* TABLES */
+  { "meta",
+    "CREATE TABLE IF NOT EXISTS meta (\n"
+      "\t-- Information about this db and how it was generated.\n"
+      "\t-- Doxygen info\n"
+      "\tdoxygen_version    TEXT PRIMARY KEY NOT NULL,\n"
+      /*
+      Doxygen's version is likely to rollover much faster than the schema, and
+      at least until it becomes a core output format, we might want to make
+      fairly large schema changes even on minor iterations for Doxygen itself.
+      If these tools just track a predefined semver schema version that can
+      iterate independently, it *might* not be as hard to keep them in sync?
+      */
+      "\tschema_version     TEXT NOT NULL, -- Schema-specific semver\n"
+      "\t-- run info\n"
+      "\tgenerated_at       TEXT NOT NULL,\n"
+      "\tgenerated_on       TEXT NOT NULL,\n"
+      "\t-- project info\n"
+      "\tproject_name       TEXT NOT NULL,\n"
+      "\tproject_number     TEXT,\n"
+      "\tproject_brief      TEXT\n"
+    ");"
+  },
   { "includes",
     "CREATE TABLE IF NOT EXISTS includes (\n"
       "\t-- #include relations.\n"
       "\trowid        INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
       "\tlocal        INTEGER NOT NULL,\n"
-      "\tid_src       INTEGER NOT NULL,  -- File id of the includer.\n"
-      "\tid_dst       INTEGER NOT NULL   -- File id of the includee.\n"
-      ");\n"
-    "CREATE UNIQUE INDEX idx_includes ON includes\n"
-      "\t(local, id_src, id_dst);"
+      "\tsrc_id       INTEGER NOT NULL REFERENCES path, -- File id of the includer.\n"
+      "\tdst_id       INTEGER NOT NULL REFERENCES path, -- File id of the includee.\n"
+      /*
+      In theory we could include name here to be informationally equivalent
+      with the XML, but I don't see an obvious use for it.
+      */
+      "\tUNIQUE(local, src_id, dst_id) ON CONFLICT IGNORE\n"
+    ");"
   },
-  { "innerclass",
-    "CREATE TABLE IF NOT EXISTS innerclass (\n"
+  { "contains",
+    "CREATE TABLE IF NOT EXISTS contains (\n"
+      "\t-- inner/outer relations (file, namespace, dir, class, group, page)\n"
       "\trowid        INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
-      "\trefid        INTEGER NOT NULL,\n"
-      "\tprot         INTEGER NOT NULL,\n"
+      "\tinner_rowid  INTEGER NOT NULL REFERENCES compounddef,\n"
+      "\touter_rowid  INTEGER NOT NULL REFERENCES compounddef\n"
+    ");"
+  },
+  /* TODO: Path can also share rowids with refid/compounddef/def. (It could
+   *       even collapse into that table...)
+   *
+   * I took a first swing at this by changing insertPath() to:
+   * - accept a FileDef
+   * - make its own call to insertRefid
+   * - return a refid struct.
+   *
+   * I rolled this back when I had trouble getting a FileDef for all types
+   * (PageDef in particular).
+   *
+   * Note: all colums referencing path would need an update.
+   */
+  { "path",
+    "CREATE TABLE IF NOT EXISTS path (\n"
+      "\t-- Paths of source files and includes.\n"
+      "\trowid        INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
+      "\ttype         INTEGER NOT NULL, -- 1:file 2:dir\n"
+      "\tlocal        INTEGER NOT NULL,\n"
+      "\tfound        INTEGER NOT NULL,\n"
       "\tname         TEXT NOT NULL\n"
-      ");"
+    ");"
   },
-  { "files",
-    "CREATE TABLE IF NOT EXISTS files (\n"
-      "\t-- Names of source files and includes.\n"
-      "\tname         TEXT PRIMARY KEY NOT NULL\n"
-      ");"
-  },
-  { "refids",
-    "CREATE TABLE IF NOT EXISTS refids (\n"
-      "\trefid        TEXT PRIMARY KEY NOT NULL\n"
-      ");"
+  { "refid",
+    "CREATE TABLE IF NOT EXISTS refid (\n"
+      "\t-- Distinct refid for all documented entities.\n"
+      "\trowid        INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
+      "\trefid        TEXT NOT NULL UNIQUE\n"
+    ");"
   },
   { "xrefs",
     "CREATE TABLE IF NOT EXISTS xrefs (\n"
-      "\t-- Cross reference relation.\n"
+      "\t-- Cross-reference relation\n"
+      "\t-- (combines xml <referencedby> and <references> nodes).\n"
       "\trowid        INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
-      "\trefid_src    INTEGER NOT NULL, -- referrer id.\n"
-      "\trefid_dst    INTEGER NOT NULL, -- referee id.\n"
-      "\tid_file      INTEGER NOT NULL, -- file where the reference is happening.\n"
-      "\tline         INTEGER NOT NULL, -- line where the reference is happening.\n"
-      "\tcolumn       INTEGER NOT NULL  -- column where the reference is happening.\n"
-      ");\n"
-    "CREATE UNIQUE INDEX idx_xrefs ON xrefs\n"
-      "\t(refid_src, refid_dst, id_file, line, column);"
+      "\tsrc_rowid    INTEGER NOT NULL REFERENCES refid, -- referrer id.\n"
+      "\tdst_rowid    INTEGER NOT NULL REFERENCES refid, -- referee id.\n"
+      "\tcontext      TEXT NOT NULL, -- inline, argument, initializer\n"
+      "\t-- Just need to know they link; ignore duplicates.\n"
+      "\tUNIQUE(src_rowid, dst_rowid, context) ON CONFLICT IGNORE\n"
+    ");\n"
   },
   { "memberdef",
     "CREATE TABLE IF NOT EXISTS memberdef (\n"
       "\t-- All processed identifiers.\n"
-      "\trowid                INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
-      "\trefid                INTEGER NOT NULL,  -- see the refids table\n"
+      "\trowid                INTEGER PRIMARY KEY NOT NULL,\n"
       "\tname                 TEXT NOT NULL,\n"
       "\tdefinition           TEXT,\n"
       "\ttype                 TEXT,\n"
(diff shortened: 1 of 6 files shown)


abathur:
### Notes on review

Since the diff is big, I've taken a few steps to simplify review:

- Squashed several dozen commits down to 2 and placed comments on the diff to highlight specific issues/questions.
- Compiled the following notes to help you triage.

#### xmlgen.cpp/h
Since I don't have a C++ background and so many people depend on xmlgen, my biggest overall concern is verifying how I split XMLCodeGenerator.

#### addon/doxypysql/search.py
I performed some manual tests to demonstrate rough before/after equivalency for this client and put them up in another branch. Here's a condensed, annotated version of this process: https://github.com/abathur/doxygen/blob/dpysql/dpy_before_after_compare.md

#### sqlite3gen.cpp
My main concerns here are:

- general C++ string/pointer/memory hygiene

- that I'm using the right Doxygen APIs and idioms

Since I'm not sure if anyone is comfortable reviewing the large SQL share of this, it may help to demonstrate that it works using an external project.

To guide and verify the schema and implementation, I built a substantive Python [client](https://github.com/abathur/doxy_db) for the database. The client's tests evaluate both the equivalence of the SQL/XML outputs, and the client's ability to extract this data.

The client's CI runs:

- build my fork of Doxygen
- run Doxygen on its own `examples/` directory ([examples.conf](https://github.com/abathur/doxy_db/blob/master/examples.conf)) to generate XML and SQL output
- run `xsltproc combine.xslt index.xml > all.xml` on the XML output
- run the client's test suite

Here are some specific lines in a successful run:
- Doxygen build begins: https://travis-ci.org/abathur/doxy_db/builds/423772385#L653
- Doxygen run begins: https://travis-ci.org/abathur/doxy_db/builds/423772385#L1033
- Doxygen SQLite3gen begins: https://travis-ci.org/abathur/doxy_db/builds/423772385#L1372
- Doxy_db tests begin: https://travis-ci.org/abathur/doxy_db/builds/423772385#L1533

As a whole, this CI run demonstrates that sqlite3gen and the resulting databases can make a good foundation for reasonably complex tools. For this review, the most relevant `t_xml_equiv.py` tests begin here: https://travis-ci.org/abathur/doxy_db/builds/423772385#L1580

Since `examples/*` doesn't yet have everything I'd like to eventually test, I built these tests to skip when none of the target entity are in the XML. You can see https://travis-ci.org/abathur/doxy_db/builds/423772385#L1666 for a summary of which were skipped and why.

abathur:
### Loose threads / open questions

- If/when the basics are approved, I'd like to document the schema changes from a few angles (new db user, old->new db user, internal/contributor?) What format should this take? One (sectioned?) or multiple `.md` files as with `doxygen.md`? 

- I've added a "schema version", but I haven't declared a constant outside of the code yet. Curious where you think this should live for now. Left inline comments on this.

- I've used `doxygen/examples/*` as a test corpus, but I'm curious about the long-term possibility of Doxygen having some form of minimal comprehensive official test corpus. I could imagine having a script that constructs it from a subset of files in `testing/` and `examples/`. Ideally, it:

  - includes every possible compound and member
  - includes all possible entity relations
  - either integrates with or is drawn from Doxygen's tests in a way that ensures it can stay updated
  - minimizes overlapping coverage to keep runtime down

src/sqlite3gen.cpp
```diff
@@ -23,77 +23,145 @@
...
       "\t-- #include relations.\n"
       "\trowid        INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
       "\tlocal        INTEGER NOT NULL,\n"
-      "\tid_src       INTEGER NOT NULL,  -- File id of the includer.\n"
-      "\tid_dst       INTEGER NOT NULL   -- File id of the includee.\n"
-      ");\n"
-    "CREATE UNIQUE INDEX idx_includes ON includes\n"
-      "\t(local, id_src, id_dst);"
+      "\tsrc_id       INTEGER NOT NULL REFERENCES path, -- File id of the includer.\n"
+      "\tdst_id       INTEGER NOT NULL REFERENCES path, -- File id of the includee.\n"
```
abathur: First of many column/variable renames that fit the same pattern (from `<common>_<specific>` to `<specific>_<common>`). This emphasizes the important part of the name, making the code, schema, queries, logs, and query results all easier to scan/read.

src/sqlite3gen.cpp
```diff
@@ -23,77 +23,145 @@
...
+      /*
+      In theory we could include name here to be informationally equivalent
```
abathur: "name" here refers to the inner-text of the includes node. Left it out because the _best_ case is that it duplicates what is in `path.name`, and the worst-case is that it's ambiguous and tricks people into mis-using this association table.

src/sqlite3gen.cpp
```diff
@@ -23,77 +23,145 @@
...
+      with the XML, but I don't see an obvious use for it.
+      */
+      "\tUNIQUE(local, src_id, dst_id) ON CONFLICT IGNORE\n"
+    ");"
   },
-  { "innerclass",
-    "CREATE TABLE IF NOT EXISTS innerclass (\n"
+  { "contains",
+    "CREATE TABLE IF NOT EXISTS contains (\n"
```
abathur: The "contains" table combines all of the inner* tables, making them *much* easier to reason about. We can still query for equivalent results by joining against compounddef and filtering by compound kinds.

src/sqlite3gen.cpp
```diff
@@ -107,7 +175,7 @@ const char * schema_queries[][2] = {
       "\tstatic               INTEGER DEFAULT 0, -- 0:no 1:yes\n"
       "\tconst                INTEGER DEFAULT 0, -- 0:no 1:yes\n"
       "\texplicit             INTEGER DEFAULT 0, -- 0:no 1:yes\n"
-      "\tinline               INTEGER DEFAULT 0, -- 0:no 1:yes\n"
+      "\tinline               INTEGER DEFAULT 0, -- 0:no 1:yes 2:both (set after encountering inline and not-inline)\n"
```
abathur: 'both' is a local neologism so that the generator can know when it has seen both parts of a memberdef with separate declaration and definition.

src/sqlite3gen.cpp
```diff
@@ -138,55 +206,71 @@ const char * schema_queries[][2] = {
...
+  { "compoundref",
+    "CREATE TABLE IF NOT EXISTS compoundref (\n"
+      "\t-- Inheritance relation.\n"
+      "\trowid          INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
+      "\tbase_rowid     INTEGER NOT NULL REFERENCES compounddef,\n"
+      "\tderived_rowid  INTEGER NOT NULL REFERENCES compounddef,\n"
+      "\tprot           INTEGER NOT NULL,\n"
+      "\tvirt           INTEGER NOT NULL,\n"
+      "\tUNIQUE(base_rowid, derived_rowid)\n"
+    ");"
```
abathur: `compoundref` folds in both inheritance relation tables (`basecompoundref` and `derivedcompoundref`) into single bi-directional association table.

src/sqlite3gen.cpp
```diff
@@ -224,221 +505,321 @@ struct SqlStmt {
...
+    ":inbodydescription"
+  ")"
+  ,NULL
+};
+/*
+We have a slightly different need than the XML here. The XML can have two
+memberdef nodes with the same refid to document the declaration and the
+definition. This doesn't play very nice with a referential model. It isn't a
+big issue if only one is documented, but in case both are, we'll fall back on
+this kludge to combine them in a single row...
```
abathur: Not sure common, but there's a chance both a declaration and a definition have their own documenting comments. This isn't a big problem in XML, but in SQL we need to combine the entries for referential integrity. There are a few ways to proceed:

- have separate columns for saving declaration/definition descriptions
- intentionally ignore one of the descriptions (i.e., only take the definition)
- implicitly ignore one (i.e., keep only the first-set or last-set)
- use some heuristic to combine the descriptions

For now, at least, I've chosen the latter. If encountered, the description fields will end up looking something like `Declaration: <decl_desc> Definition: <def_desc>`. I may need to polish this a little if we retain this practice.

src/sqlite3gen.cpp
```diff
@@ -520,72 +901,118 @@ static int step(SqlStmt &s,bool getRowId=FALSE, bool select=FALSE)
...
   return rowid;
 }
 
-static int insertRefid(const char *refid)
+static void recordMetadata()
 {
-  int rowid=-1;
-  if (refid==0) return rowid;
+  bindTextParameter(meta_insert,":doxygen_version",versionString);
+  bindTextParameter(meta_insert,":schema_version","0.2.0"); //TODO: this should be a constant somewhere; not sure where
```
abathur: Not sure where you'd prefer a constant with this schema version (or if you'd prefer I leave it as a literal for now).

src/sqlite3gen.cpp
```diff
@@ -23,77 +23,145 @@
...
-      "\trefid_dst    INTEGER NOT NULL, -- referee id.\n"
-      "\tid_file      INTEGER NOT NULL, -- file where the reference is happening.\n"
-      "\tline         INTEGER NOT NULL, -- line where the reference is happening.\n"
-      "\tcolumn       INTEGER NOT NULL  -- column where the reference is happening.\n"
-      ");\n"
-    "CREATE UNIQUE INDEX idx_xrefs ON xrefs\n"
-      "\t(refid_src, refid_dst, id_file, line, column);"
+      "\tsrc_rowid    INTEGER NOT NULL REFERENCES refid, -- referrer id.\n"
+      "\tdst_rowid    INTEGER NOT NULL REFERENCES refid, -- referee id.\n"
+      "\tcontext      TEXT NOT NULL, -- inline, argument, initializer\n"
```
abathur: Local neologism. Attempting to validate the sqlite3 output against the XML output revealed that there were 3 different contexts that resulted in the documentation of an xref (inline, argument, and initializer).

src/sqlite3gen.cpp
```diff
@@ -600,8 +1027,8 @@ static void insertMemberFunctionParams(int id_memberdef, const MemberDef *md, co
 
       if (!a->attrib.isEmpty())
       {
-        bindTextParameter(params_select,":attributes",a->attrib.data());
-        bindTextParameter(params_insert,":attributes",a->attrib.data());
```
abathur: Early on I superstitously added many .data() calls while trying to fix junk data in columns. Once I had that working, I cut out the ones that seemed unnecessary (including some that were already present).

src/sqlite3gen.cpp
```diff
@@ -871,15 +1419,132 @@ static void writeTemplateList(const ClassDef *cd)
...
+        def->docFile(),
+        def->docLine()
+      ),
+      FALSE
+    );
+}
 ////////////////////////////////////////////
 
+/* (updated Sep 01 2018)
+DoxMemberKind and DoxCompoundKind (compound.xsd) gave me some
```
abathur: I had to collect this just to reason through things as I worked. I think it's worth keeping around for a bit in case the core concern ('kind' overlap for member/compound types generalized in the def table) materializes. Decided to flag it in case you think it has broader value that would justify moving it somewhere more permanent?

src/sqlite3gen.cpp
```diff
@@ -893,18 +1558,102 @@ static void generateSqlite3ForMember(const MemberDef *md, const Definition *def)
   // enum values are written as part of the enum
   if (md->memberType()==MemberType_EnumValue) return;
   if (md->isHidden()) return;
-  //if (md->name().at(0)=='@') return; // anonymous member
 
-  // group members are only visible in their group
-  //if (def->definitionType()!=Definition::TypeGroup && md->getGroupDef()) return;
```
abathur: These conditionals were already commented out in the XML they got copied in from, so I cut them here.

src/sqlite3gen.cpp
```diff
@@ -1110,172 +1872,201 @@ static void generateSqlite3ForMember(const MemberDef *md, const Definition *def)
   if (mdict!=0)
   {
     MemberSDict::IteratorDict mdi(*mdict);
-    MemberDef *rmd;
+    const MemberDef *rmd;
```
abathur: This change (and others adding const in same pattern) are an attempt to parrot @groleo's change in #604. Won't pretend to fully understand; may not be correct.

src/sqlite3gen.cpp
```diff
@@ -1110,172 +1872,201 @@ static void generateSqlite3ForMember(const MemberDef *md, const Definition *def)
...
 
   if (cd->isReference())        return; // skip external references.
   if (cd->isHidden())           return; // skip hidden classes.
   if (cd->name().find('@')!=-1) return; // skip anonymous compounds.
   if (cd->templateMaster()!=0)  return; // skip generated template instances.
 
-  msg("Generating Sqlite3 output for class %s\n",cd->name().data());
+  struct Refid refid = insertRefid(cd->getOutputFileBase());
+
+  // can omit a class that already has a refid
```
abathur: Unless there are conditions under which we may encounter the class refid before parsing the class?

Also, this same basic issue (but for pages) recurs on line 2470.

src/sqlite3gen.cpp
```diff
@@ -1463,14 +2526,29 @@ static sqlite3* openDbConnection()
...
     return NULL;
   }
-  rc = sqlite3_open_v2(outputDirectory+"/doxygen_sqlite3.db", &db, SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE, 0);
+
+
+  if (stat (outputDirectory+"/doxygen_sqlite3.db", &buf) == 0)
+  {
+    msg("doxygen_sqlite3.db already exists! aborting sqlite3 output generation!\n");
+    msg("If you wish to re-generate the database, remove or archive the existing copy first.\n");
+    return NULL;
```
abathur: Not sure if this is how you'd prefer I check file existence. I couldn't find a clear consensus approach and don't know enough to evaluate the tradeoffs between methods I found.

src/sqlite3gen.cpp
```diff
@@ -138,55 +206,71 @@ const char * schema_queries[][2] = {
...
+  { "member",
+    "CREATE TABLE IF NOT EXISTS member (\n"
+      "\t-- Memberdef <-> containing compound relation.\n"
+      "\t-- Similar to XML listofallmembers.\n"
+      "\trowid            INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,\n"
+      "\tscope_rowid      INTEGER NOT NULL REFERENCES compounddef,\n"
+      "\tmemberdef_rowid  INTEGER NOT NULL REFERENCES memberdef,\n"
+      "\tprot             INTEGER NOT NULL,\n"
+      "\tvirt             INTEGER NOT NULL,\n"
+      "\tUNIQUE(scope_rowid, memberdef_rowid)\n"
```
abathur: The XML `<member>` node can have inner `<scope>` and `<name>` nodes. I couldn't find any `<scope>` that was anything but the current containing scope, nor a `<name>` that wasn't just the name of the memberdef at the corresponding refid.

I've left these out of SQL because they smell like they're just convenience features in the XML. I'll need to rethink this If they *can* actually differ.

doxygen:
When I'm trying with a default Doxyfile and a simple test.cpp

```
int main()
{
  return 0;
}
```
I run into the following error:

```
Generating Sqlite3 output for file test.cpp
Initializing DB schema (views)...
failed to execute query: CREATE VIEW IF NOT EXISTS def (
	-- Combined summary of all -def types for easier joins.
	rowid,
	refid,
	kind,
	name,
	summary)
as SELECT 
	refid.rowid,
	refid.refid,
	memberdef.kind,
	memberdef.name,
	memberdef.briefdescription 
FROM refid 
JOIN memberdef ON refid.rowid=memberdef.rowid 
UNION ALL 
SELECT 
	refid.rowid,
	refid.refid,
	compounddef.kind,
	compounddef.name,
	CASE 
		WHEN briefdescription IS NOT NULL 
		THEN briefdescription 
		ELSE title 
	END summary
FROM refid 
JOIN compounddef ON refid.rowid=compounddef.rowid;
	near "(": syntax error
lookup cache used 0/65536 hits=0 misses=0
finished...
```

Sqlite version 3.8.10.2.

abathur:
@doxygen when I rebuild against sqlite3 3.8.10.2 I do get a database with a malformed schema (but didn't see the exact query error message you pasted).

I rebuilt against 3.9.0 and that _appears_ to work (I haven't extensively tested the output database,  but it has the right tables and they appear to be populated).

While checking the tables, I noticed a small problem with the schema for a convenience view that I added fairly late and don't have under test yet. I'll push a fix for it shortly.

doxygen:
@abathur 
> I rebuilt against 3.9.0 and that _appears_ to work (I haven't extensively tested the output database, but it has the right tables and they appear to be populated).
> 

Ok, then a version check needs to be added to CMakefiles.txt, e.g. something like this:

```
f (sqlite3)
    find_package(SQLite3 REQUIRED)
    if("${SQLITE_VERSION}" VERSION_LESS "3.9.0") 
       message(SEND_ERROR "Doxygen requires at least sqlite3 version 3.9.0")
    endif()
endif()
```

abathur:
@doxygen The current `cmake/FindSqlite3.cmake` script doesn't extract and define either the version string or number. 

I'm not familiar with how to do this, but I'll take a look later today and give it a try unless you suggest otherwise before then.

abathur:
@doxygen I've pushed a draft of this version check. Log snippets:

```
vagrant@ubuntu-bionic:/home/doxygen/build_sql3_test$ cmake -Duse_sqlite3=ON ..
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
-- Found Sqlite3: /usr/local/lib/libsqlite3.so (found version "3.8.10.2")
CMake Error at CMakeLists.txt:91 (message):
  Doxygen requires at least sqlite3 version 3.9.0 (installed: 3.8.10.2)


-- Configuring incomplete, errors occurred!
See also "/home/doxygen/build_sql3_test/CMakeFiles/CMakeOutput.log".
See also "/home/doxygen/build_sql3_test/CMakeFiles/CMakeError.log".
```

```
vagrant@ubuntu-bionic:/home/doxygen/build_sql3_test$ cmake -Duse_sqlite3=ON ..
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
-- Found Sqlite3: /usr/local/lib/libsqlite3.so (found version "3.9.0")
-- Configuring done
-- Generating done
-- Build files have been written to: /home/doxygen/build_sql3_test
```

PR #6491, opened 2018-09-08
Plantuml parse error

albert-github:
Don't replace comment just starting with `{B}*"*"+` with spaces but discard it.

Regression due to move of markdown parser to other place.


albert-github:
Looks like this patch also could have some side effects, created a new proposed patch https://github.com/doxygen/doxygen/pull/10999

PR #6492, opened 2018-09-08
Preprocessor crash in case of a `@` at end of `#if` construct

albert-github:
Example:
```
#if @VTKMANTA_BUILD_SHARED_LIBS_CONFIG@
# define @PROJECT_NAME@_SHARED
#endif
```

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -1172,6 +1172,11 @@ QCString removeIdsAndMarkers(const char *s)
 	{
 	  // skip
 	}
+	else if (*(p+1)=='\000')
+	{
+	  result+=c;
+	  return result;
+	}
 	p+=2;
       }
       else if (isdigit(c)) // number


doxygen:
I don't see a crash (anymore), but it is also not valid code. The GNU compiler says:

```cpp
test.cpp:1:5: error: invalid token at start of a preprocessor expression
#if @VTKMANTA_BUILD_SHARED_LIBS_CONFIG@
    ^
```

albert-github:
Indeed it is not a valid preprocessor expression, but might be accidentally left and doxygen should not crash.

There have been a number of fixes regarding the `@` sign in `pre.l`  and the fix from March 16, 2019 (9fdeef7b5a1018872c7f07b8d03e374a3da840a0 at sign (`@`) not handled correctly in preprocessor) fixed the issue.

Issue #6473, opened 2018-08-30
Plantuml parse error


cheoljoo:
# environment
- macbook
  - $  uname -a
  - Darwin cheoljoolees-MacBook-Pro.local 17.3.0 Darwin Kernel Version 17.3.0: Thu Nov  9 18:09:22 PST 2017; root:xnu-4570.31.3~1/RELEASE_X86_64 x86_64
  - doxygen 1.8.15  has the following error. (developing version)
- linux ubuntu 16.04
  - doxygen 1.8.14 did not have error.  the result was good.
  - doxygen 1.8.15 has the following error. 

# Issue
- I attached 2 files. One (fail.cpp) has the following error msg and the other (succeed.cpp) is success.
- difference between two files.
  - Just add * in starting point of line

## Error Message
```
/Users/cheoljoo/Code/LG/test_doxygen/fail.cpp:12: warning: explicit link request to 'FFFFFF' could not be resolved
/Users/cheoljoo/Code/LG/test_doxygen/fail.cpp:13: warning: explicit link request to 'ADD1B2' could not be resolved
/Users/cheoljoo/Code/LG/test_doxygen/fail.cpp:14: warning: explicit link request to 'FFFFFF' could not be resolved
/Users/cheoljoo/Code/LG/test_doxygen/fail.cpp:21: warning: unexpected command enduml
```

## Attachment
The file with the `fail.cpp` and the `succeed.cpp`: [err.zip](https://github.com/doxygen/doxygen/files/2334649/err.zip)

albert-github:
The problem lies in the way the file is parsed and the usage of the markdown parser. The markdown parser moved its place for consistency in the 1.8.15 version.

 - it is a bit strange to have 2 types of comment style mixed (with `*` and without).

__Workarounds:__

 - disable `MARKDOWN_SUPPORT` but this will, probably, impact results in other parts of the output as well
 - add, in this case, 1 space in front of all lines without the * in the comment. In fact the add some many spaces as there are before the `*` + 1 (so here 0 + 1)

albert-github:
I've just pushed a proposed patch, pull request #6491

albert-github:
I don't think the previous proposed patch #6491 is correct (also might change other parts).
Replaced proposed patch by proposed patch [#10999](https://github.com/doxygen/doxygen/pull/10999)

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

[...]

PR #6495, opened 2018-09-10
Plantuml parse error  : Alternatives

cheoljoo:
solve issue #6473 : I think this is bug

Thanks. ( I got the idea from #6491 )
When I am debugging with -d markdown , writeCodeBlock returns wrong.
```
======== Markdown =========
---- input -------

    @brief  Receive & Process the message from HMI
    @startuml
---- output -----
@verbatim
@brief  Receive & Process the message from HMI
@startuml
@endverbatim
```
**The output should not include @startuml in @verbatim and @endverbatim.**

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -1433,7 +1433,7 @@ static bool isFencedCodeBlock(const char *data,int size,int refIndent,
 
 static bool isCodeBlock(const char *data,int offset,int size,int &indent)
 {
-  //printf("<isCodeBlock(offset=%d,size=%d,indent=%d)\n",offset,size,indent);
+  printf("<isCodeBlock(offset=%d,size=%d,indent=%d)\n",offset,size,indent);
   // determine the indent of this line
   int i=0;
   int indent0=0;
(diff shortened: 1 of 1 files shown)


cheoljoo:
@albert-github  , Would you like to explain what is my mistake in CI?
Because I do not have windows developing environment.
I think that this patch is the first trial to apply new CI. (appveyor/pr)  But , I do not know any error in log of CI.
If you share CI's testing cases, I can do pull request without failure.

albert-github:
No mistake at all, there seems to be a small problem when using the debug version of the 2017 Windows compiler in respect to the tests. For the time being, the results from debug with the Windows 2017 compiler, can be ignored.

albert-github:
@cheoljoo Test script has been adjusted, so from now on problems have to be taken into account again.

albert-github:
I did some tests with the original test case
- 1.8.14: no verbatim
- 1.8.15: verbatim present (would be consistent with the date of this PR being as the PR is based on the development version leading to version 1.8.15).
- 1.11.0: the verbatim was present (now in the form of `iverbatim`)
- 1.12.0 (the current version): no verbatim

Looks like that the counting of the initial comment is not correct here and thus the initial 4 spaces are seen as an indication of the start of a verbatim section.


As far as I can tell this PR just, unconditionally, outputs some information to stdout, this is a bad idea in production code.

In the mean time a fix for the original issue #6473, proposed PR https://github.com/doxygen/doxygen/pull/10999 has been integrated into the current doxygen version (1.12.0) and here we don't see the `@verbatim` anymore

So my suggestion is to close this issue

PR #6503, opened 2018-09-12
Fixing problem with possible not initialized variable (endless loop in VS 2017 debug)

albert-github:
The not initialized variable `l` caused an endless loop in the VS 2017 debug version, variable should not have been used.


Croydon:
If this is really fixing VS debug, could you re-enable tests and docs in appveyor?

PR #6510, opened 2018-09-15
Difference between standard and used Doxyfile

albert-github:
Writes a the differences between the current configuration and the template configuration.

Useful when communicating e.g. in forums.

PR #6512, opened 2018-09-15
AppVeyor: Fix MikTex version and re-enable tests

Croydon:
Now, where the VS2017 debug bug is fixed we can re-enable the debug builds in AppVeyor.

Also all AppVeyor builds did fail, since the MikTex version in chocolatey is outdated and download mirrors started to remove outdated versions, which leds to a 404 error.

So I switched over to install it (again) directly from the MikTex website, which comes with the bonus that we always install the latest version.

Docs are still not working, because it seems to need stuff which is only available in the full download and this simply takes to much time over the installer as it downloads and install every part individually, so we can't take this way or we will hit regularly the 1h limit and have failures again.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ image: Visual Studio 2017
 
 configuration:
   - Release
-#  - Debug
+  - Debug
 
 platform:
   - x64
@@ -24,7 +24,11 @@ install:
   - 7z x flex.zip -oC:\deps\flex
   - ps: Invoke-WebRequest https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs924/gs924w32.exe -OutFile gswin32c.exe
   - gswin32c /S /D=C:\deps\ghostscript
-  - ps: choco install -y miktex
+  - ps: if($env:platform -eq "x64") { Invoke-WebRequest https://miktex.org/download/win/miktexsetup-x64.zip -OutFile miktexsetup.zip }
+  - ps: if($env:platform -eq "Win32") { Invoke-WebRequest https://miktex.org/download/win/miktexsetup-x86.zip -OutFile miktexsetup.zip }
+  - 7z x miktexsetup.zip -oC:\tmpmiktex
+  - C:\tmpmiktex\miktexsetup --local-package-repository=C:\temp\miktex --package-set=basic download
+  - C:\tmpmiktex\miktexsetup --verbose --local-package-repository=C:\temp\miktex --package-set=basic install
   - refreshenv
   - pip install conan
   - conan install libxml2/2.9.8@bincrafters/stable -g virtualrunenv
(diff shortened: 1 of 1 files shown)


Croydon:
//cc @albert-github https://github.com/doxygen/doxygen/pull/6510#issuecomment-421546460

albert-github:
Good to have the tests working as well!

In the appveyor.yml there is at the end the line:

```
  - cmake -G "%CMAKE_GENERATOR_NAME%" -D build_doc=ON ..
```
as long as the documentation is not build this line is not necessary, furthermore why isn't the option integrated in the first `cmake` call  (in `before_build` section)?

albert-github:
Some thoughts.
---

MikTeX has the possibility to install the packages on the fly from either a local directory or directly from a mirror. I use a local directory and use the "Ask" possibility (so I know what will be installed).  I don't have timing information about the different steps.
I could imagine that the basic installation is done as it is done now and that the full set of packages (around 3600 packages, around 3GB data before install) is downloaded and extracted. During the build the required packages could be installed on the fly. Other possibility would be to install the required packages on the fly from the internet.


Questions:
---
 - how long would the download + extract take?
 - would it be possible to install the packages upon request from the local repository? and what would be the required time?
 - would it be possible to install the packages upon request from a mirror repository?  and what would be the required time?

Croydon:
> MikTeX has the possibility to install the packages on the fly from either a local directory or directly from a mirror. 

How can I make use of this?

>  furthermore why isn't the option integrated in the first cmake call (in before_build section)?

Very fair question. I can try to integrate it in the first step, but initially this led to very weird errors, and I proceeded in the context of make it work than make it better :)

albert-github:
1) automatic MikTeX install, in the GUI it is a setting, but on the command line I don't know (I didn't see an option for it, I saw --remote-package-repository=URL but this is only the place where to find the files), maybe best to submit a question in the LaTeX forum (https://tex.stackexchange.com/).

2) very understandable. I think just using `cmake -G "%CMAKE_GENERATOR_NAME%" -D build_doc=ON ..` instead the `cmake -G "%CMAKE_GENERATOR_NAME%" ..` should be sufficient.

Croydon:
And seems like a ulem package is required https://github.com/doxygen/doxygen/commit/a3ef054f5d2af360fa7541e01a6de102d1c7c147

albert-github:
The list of programs looks complete to me, it is just the list of packages that is missing.

Due to a problem I had I had to reinstall MikTeX so, by chance, I have a clean installation of MikTeX just based on the doxygen tests and documentation.
On the directory `C:\Program Files\MiKTeX 2.9\source` I found the used sources (package names might be a slightly different without the '-src' part) and it looks to me that this is quite complete, see attachment MikTex.tx (below)

Some remarks:
  - when having a fixed list a new package has to be added as well, maintenance problem
  - a better solution would be that the files can be installed in the fly.


Attachment:  [MikTex.txt](https://github.com/doxygen/doxygen/files/2387670/MikTex.txt)

PR #6537, opened 2018-10-03
Bug 750331 - Detailed file documentation is being rolled into the author field

albert-github:
See to it that the previous command is properly closed when starting a brief section.


albert-github:
Problem looks like to be solved in the mean time (tested against doxygen 1.9.2).

Issue #6240, opened 2018-07-02
regression with ALIASES or image latex interpretation (Origin: bugzilla #786576)


doxygen:
status NEW severity *normal* in component *build* for *---*
Reported in version *1.8.13* on platform *Other*
Assigned to: Dimitri van Heesch

On 2017-08-21 17:11:58 +0000, divide@divideconcept.net wrote:

    I'm using a few ALIASES in my Doxygen file as shortcut for including images
    in specific formats both in HTML and Latex output:
    
    ALIASES                = "icon{1}=\image latex \"\1\" height=0.5cm \htmlonly
    <img src=\"\1\" style=\"max-height:20px;\"/> \endhtmlonly" \
                             "screen{3}=\image latex \"\1\" width=\3cm \htmlonly
    <img src=\"\1\" style=\"max-width:\2px;\"/> \endhtmlonly" \
                             "screen{4}=\image latex \"\1\" \"\4\" width=\3cm
    \htmlonly <img src=\"\1\" style=\"max-width:\2px;\"/><div
    class=\"caption\">\4</div> \endhtmlonly"
    
    
    It was working fine in Doxygen 1.8.11. Since I upgraded to 1.8.13, it
    doesn't work anymore. I had to change them to:
    
    ALIASES                = "icon{1}=\image latex \"\1\" height=0.5cm &nbsp;
    \htmlonly <img src=\"\1\" style=\"max-height:20px;\"/> \endhtmlonly" \
                             "screen{3}=\image latex \"\1\" width=\3cm &nbsp;
    \htmlonly <img src=\"\1\" style=\"max-width:\2px;\"/> \endhtmlonly" \
                             "screen{4}=\image latex \"\1\" \"\4\" width=\3cm
    &nbsp; \htmlonly <img src=\"\1\" style=\"max-width:\2px;\"/><div
    class=\"caption\">\4</div> \endhtmlonly"
    
    That is, including an extra and unnecessary &nbsp; in between my \image
    latex and my \htmlonly commands. Otherwise when building I get tons of
    errors related to that (misinterpretation of the \image latex command).

albert-github:
I've just pushed a proposed patch, pull request #9477

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.5.
[...]

PR #6548, opened 2018-10-11
issue_6547 Call graph missing due to ALIASES

albert-github:
Replace the `^^` in an alias not with an physical newline but with a doxygen synthetic newline.
Shown problem is due to the fact that line matching didn't work anymore due to the fact that a physical newline was inserted (ever better seen when enabling `INLINE_SOURCES`)

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -9889,7 +9889,7 @@ static void escapeAliases()
     while ((in=value.find("^^",p))!=-1)
     {
       newValue+=value.mid(p,in-p);
-      newValue+="\n";
+      newValue+="\\_linebr";
       p=in+2;
     }
     newValue+=value.mid(p,value.length()-p);


lrineau:
The CGAL Doxygen testsuite shows hundreds of warnings, after this PR.
```
    201 warning: end of comment block while expecting command </div>
    414 warning: Illegal command image as part of a \image
    414 warning: Unknown option '_linebr' after \image command, expected 'width' or 'height'
```
The 414 warnings about images is because of aliases, like this one:
```
                         "cgalFigureBegin{2}=\anchor fig__\1 ^^ \image html \2 ^^ \image latex \2 \"\" width=15cm ^^ \htmlonly <div class=\"cgal_figure_caption\"> \endhtmlonly \latexonly ^^ \endlatexonly ^^ \ref fig__\1" \
                         "cgalFigureBegin{3}=\anchor fig__\1 ^^  <CENTER> <TABLE border=0> <TR> <TD> \image html \2 ^^ \image latex \2 \"\" width=7.5cm ^^ </TD> <TD> \image html \3 ^^ \image latex \3 \"\" width=7.5cm ^^ </TD> </TR> </TABLE> </CENTER> \htmlonly <div class=\"cgal_figure_caption\"> \endhtmlonly ^^ \latexonly ^^ \endlatexonly \ref fig__\1" \
```

Can you tell us why this kind of alias cannot be defined? It is probable that this PR has a bug.

lrineau:
I have verified that the warnings disappear once the sole commit https://github.com/doxygen/doxygen/pull/6548/commits/0077ab48f682cca9caf99b2d92a8448d1d66cb98 of this PR is reverted.

albert-github:
Looks like the same / similar problem as with issue #6240

lrineau:
That looks similar but it is not. The warnings we have are clearly related to this PR:
  - the text of the warnings say "Unknown option '_linebr' after \image command",
  - the warnings disappear if I revert the PR.

albert-github:
Both give a warning about the first command after an command on the same line. I could even reproduce the problem with the following:
```
/** \file */

/** the fie
  * \image latex "b_and_a.png" "The caption" width="2cm" &nbsp; \htmlonly <img src="b_and_a.png" style="max-width:100px;"/><div class="caption">div caption</div> \endhtmlonly
  *
  * \image latex "b_and_a.png" "The caption" width="2cm" \htmlonly <img src="b_and_a.png" style="max-width:100px;"/><div class="caption">div caption</div> \endhtmlonly
  */
void fie(void);
```

It looks all to have to do with the change in commentscan.l:

```
//if (i>0) addOutput(QCString(yytext).left(i)); // removed for bug 689341
```

but I also see in the 1.8.14 version a problem with the example of the bug 689341 (change is from December 26, 2012 for version 1.8.3).

lrineau:
[**@albert-github**](https://github.com/albert-github) commented on [Oct 15, 2018, 5:25 PM GMT+2](https://github.com/doxygen/doxygen/pull/6548#issuecomment-429899413 "2018-10-15T15:25:43Z - Replied by Github Reply Comments"):
> [@lrineau](https://github.com/lrineau) Can you try in doxygen.cpp to replace:
> 
>     newValue+="\\_linebr";
> 
> by
> 
>     newValue+="@_linebr";
> 
> and report back (looks like the idea is correct but the backslashes give some problems).

I am surprised, but it does not change anything. :-(

albert-github:
Indeed strange as at my place it works.
There are 2 places with `\\_linebr` I assume you got the correct one (line 9892 in the `    while ((in=value.find("^^",p))!=-1)` block, but ...

albert-github:
@lrineau I found the CGAL repository (currently cloning). What is the fastest way to get the error messages?

sloriot:
Thanks for the help.
You have to call `cmake` on using `Documentation/doc/CMakeLists.txt`. The variable to specify the doxygen executable is `DOXYGEN_EXECUTABLE` (`-DDOXYGEN_EXECUTABLE=/XX/YY/bin/doxygen`).
Then build a first time the target `doc` (`make doc`) and then a second time. The first time you'll get a lot of warning about missing ref in tag files not already generated but you should also get the other warning.

lrineau:
@albert-github The result of our daily testsuite are there:
https://cgal.geometryfactory.com/CGAL/Manual_doxygen_test/

We have added the column with "Doxygen-master" (fetched and compiled daily before the tests), in order to catch regressions like this one.

albert-github:
I used for the test
- the executable:
  - master version
  -  changed line to `newValue+="@_linebr";` as indicated above
- repository cloned on October 15 early evening.

General warning message
---
During the non modified runs I got the message:

```
warning: ignoring unsupported tag `EXTRACT_ALL_NO_DETAILED_IF_EMPTY                                        =' at line 504, file /cygdrive/d/Programs/CGAL/cgal/build/doc_dxy/BaseDoxyfile
```

Looks to me like something is not correct in the generation of the `BaseDoxyfile`

Running the procedure:
---
- cmake -DDOXYGEN_EXECUTABLE=/cygdrive/d/Programs/Doxygen/fork/doxygen/build_cygwin/bin/doxygen.exe ../Documentation/doc/
- make doc 
  results in the expected messages about `\ref` and missing tag files 
- make doc
  I only get 1 message:
  ```
  /cygdrive/d/Programs/CGAL/cgal/Three/include/CGAL/Three/Triangle_container.h:93: warning: Member getShrinkFactor() (function) of class CGAL::Three::Triangle_container is not documented.
  ```
- changed in /cygdrive/d/Programs/CGAL/cgal/build/doc_dxy/BaseDoxyfile
  `EXTRACT_ALL_NO_DETAILED_IF_EMPTY                                        = YES`
    into
  `EXTRACT_ALL                                        = YES`
  mow I only got the message about the not documented member.
- reverted the `newValue+="@_linebr";` to `newValue+="\\_linebr";`
  I get the errors as indicated about _linebr.

So I think the change to `newValue+="@_linebr";` solves the problem.

lrineau:
That is strange that my result were different.

But I am very happy if the solution is that simple. Thanks @albert-github for the time you spend on Doxygen.

albert-github:
I'm always happy with constructive discussions and error reports.
I'll create a pull request for this change.

albert-github:
I've just pushed an updated proposed patch, pull request #6558

PR #6551, opened 2018-10-12
Fix travis mac build

albert-github:
Attempt to fix mac build of travis in respect to:
- Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/usr/local/man'

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
 
 before_script:
     - if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
-        mkdir -p /usr/local/man
+        mkdir -p /usr/local/man;
         pip install --quiet conan;
         export HOMEBREW_NO_AUTO_UPDATE=1;
         brew install ghostscript;
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,8 @@ jobs:
 
 before_script:
     - if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
-        mkdir -p /usr/local/man;
+        sudo mkdir -p /usr/local/man;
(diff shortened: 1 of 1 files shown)


albert-github:
@Croydon as you might have seen as well the last few days there are some problems with the Mac build. I worte to the people of travis-ci and they suggested to create the `/usr/local/man` directory, but this didn't help.
Do you have any ideas?

Note that the appveyor builds didn't run either (looked like a conan error as well), but in the mean time they work again.

Croydon:
The Windows CI errors are due to Bintray service problems. This should get better as well, they are improving their monitoring to noticing errors faster etc.

Croydon:
Yes, MikTex is unreliable, that is why I asked you once what parts exactly doxygen is requiring from it in the hope that we might be able to split it up and get this dependencies in another way. Maybe I can create Conan packages for it then. However, MikTex completely is just a gigantic monster and extremely hard to package.

albert-github:
MikTeX is indeed very large the problem with a fixed conan package is that as soon as a new LaTeX package is needed by doxygen a new conan package has to be created (and of course when one of the components changes). A conan package would be nice though as this would, I think, increase the build speed considerably but there has to be a good procedure to update in case a new package is needed.

albert-github:
The Mac build just ran through and completed correctly.
Hopefully the real conan / tgdm problem will be fixed soon.

Croydon:
> MikTex is unreliable, that is why I asked you once what parts exactly doxygen is requiring

@albert-github Do you remember where I did ask you this? I actually can't find it anymore 🙈

albert-github:
@Croydon the pull request was: #6512

PR #6562, opened 2018-10-17
Add commands to handle referenced by relation and references relation

albert-github:
Analogous to call graph and caller graph this patch provides an implementation for the referenced by relation and references relation.
Providing the commands:
- referencedbyrelation
- hidereferencedbyrelation
- referencesrelation
- hidereferencesrelation

Motivation is that some lists can get extremely large and also there is now more symmetry between the textual and graphical out.


albert-github:
This issue was also mentioned by Peter Barnes in 2014 in the doxygen users list

> We have some functions and macros that are so widely used the "Referenced by" list is pages long. (Think logging, assertions, …) 
> 
> How can I selectively suppress this list for those objects, both macros and functions.
> 
> Basically I want just those objects to behave as if REFERENCED_BY_RELATION=NO, while its YES in the config file.
> 
> For the macros I tried adding them to PREDEFINED in our doxygen config file but that didn't help.
> 
> (I searched the list archives and StackOverflow but didn't find anything relevant.)
> 
> Oh, this is a C++ project, if that's relevant (but the problem seems general).

doxygen:
@albert-github I like the idea, but find the commands rather long and thus error prone to type. What do you think about using `\showrefby`, `\hiderefby`, `\showrefs`, `\hiderefs` instead?

albert-github:
I agree, I'll create an updated pul request.

PR #6567, opened 2018-10-21
Handling Fortran functions in call graphs

albert-github:
When functions are used they have to be declared and were seen as local variables even when the 'external' keyword had been applied.
Functions are now not seen anymore as local variables as soon as the 'external' keyword has been applied.


albert-github:
Partly based on https://stackoverflow.com/questions/37720759/doxygens-call-caller-graph-references-missing-for-fortran-functions

Example: 
[example.zip](https://github.com/doxygen/doxygen/files/2499037/example.zip)

PR #6569, opened 2018-10-22
Wrong separator in index for a.o. Python, C#

albert-github:
In e.g. the LaTeX output the separator for Pyton, C# is given as '::', this should be '.' .
With this fix the problems in the index are gone.


albert-github:
Source code example to show problem: [example.zip](https://github.com/doxygen/doxygen/files/2501010/example.zip)

PR #6578, opened 2018-10-27
Multiple addindex commands in HTML with same name

albert-github:
It is possible to specify an addindex multiple times. In HTML this resulted in
- a 'a' tag with with the same 'name' attribute
- in CHM only one occurrence was shown.

this has been corrected.
(see e.g. the index of 'flex' in the CHM manual generated ffrom master).

### DIFF
--- a/src/htmldocvisitor.cpp
+++ b/src/htmldocvisitor.cpp
@@ -67,7 +67,12 @@ static const char *contexts[10] =
 static QCString convertIndexWordToAnchor(const QString &word)
 {
   static char hex[] = "0123456789abcdef";
+  static int cnt = 0;
   QCString result="a";
+  QCString cntStr;
+  result += cntStr.setNum(cnt);
+  result += "_";
+  cnt++;
   const char *str = word.data();
   unsigned char c;
   if (str)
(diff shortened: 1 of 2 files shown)


src/htmldocvisitor.cpp
```diff
@@ -67,7 +67,12 @@ static const char *contexts[10] =
 static QCString convertIndexWordToAnchor(const QString &word)
 {
   static char hex[] = "0123456789abcdef";
+  static int cnt = 0;
```
doxygen: @albert-github The issue I see with using a static global counter is that it makes the anchors instable, i.e. adding a \addindex somewhere will change other anchors as well. This makes it difficult to bookmark an anchor for later use. It think it a better to only add a number when there is a anchor clash.

albert-github:
I agree that this change will make these anchors from the `\addindex` command only in HTML instable. In HTML we need these extra anchors as a mechanism to jump to another place. These anchors are only intended for use in HTML in e.g. LaTeX its own mechanism is used and no anchor is generated.

To have a persistent anchor the `\anchor` command should be used.

Furthermore the documentation says about the `\addindex` command:
> 24.97 \addindex (text)
>
> This command adds (text) to the LaTeX index.

(this is not completely up to date as also docbook handles the `\addindex` command and the CHM generation as well).

So I don't really see a problem.

PR #6587, opened 2018-10-31
Multiple `\xreflist` in one page with same key

albert-github:
In case there are multiple `\xreflist` commands with the same key only the first item has a correct linking possibility, the subsequent items don't have it.
Resulting in warnings in e.g. docbook and LaTeX output

### DIFF
--- a/src/reflist.cpp
+++ b/src/reflist.cpp
@@ -148,8 +148,6 @@ void RefList::generatePage()
   for (it.toFirst();(item=it.current());++it)
   {
     doc += " <dt>";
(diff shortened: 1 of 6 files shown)


albert-github:
As an example (with default Doxyfile):

```
/** \file */

/** @page test TEST Page
 *  \tableofcontents
 *
 *  this is some description
 *  \xrefitem custom_change "This is a custom change" "" a mark in description (this is only for testing purposes)
 *
 *  \xrefitem custom_change "This is a custom change" "" a mark in description (second this is only for testing purposes)
 *  */
```

PR #6588, opened 2018-10-31
[C++] Add support for std::shared_ptr

Sleepyowl:
Adds support for std::shared_ptr so it is recognized as a smart pointer


Sleepyowl:
@KineticTheory done. It's up to maintainers now.

PR #6592, opened 2018-11-02
Bug 795568 - "unexpected token TK_EOF as the argument of ref" when target starts with a digit

albert-github:
Enable possibility to reference to file names as well (some filenames ere supported as side effect of other rules).


albert-github:
The example to solve the issue:  [example.zip](https://github.com/doxygen/doxygen/files/2542314/example.zip)

Issue #6517, opened 2018-09-16
Emoji support


reavessm:
Doxygen does a great job creating webpages based on markdown files, could it also render the emojis in the files?

I have some files where I would like `:smiley:` to show up as :smiley:, for example.

albert-github:
Is there anywhere a, free, translation table between the `:smiley:` syntax and the corresponding Unicode character `&#x1F600;` ? 
Analogous also for other formats, especially LaTeX?

There is a nice Unicode list (https://unicode.org/emoji/charts/full-emoji-list.html) with all Emoji characters.
There are also tables with the `:smiley:` syntax (e.g. https://www.webpagefx.com/tools/emoji-cheat-sheet/ or https://gist.github.com/rxaviers/7360908)

albert-github:
That is the same list as the first list I mentioned in my comment.

I'm looking for a table that translates the e.g. `:smiley:` to `&#x1F600` or `U+1F600`

albert-github:
That table (https://api.github.com/emojis) looks very interesting, it is not 100% complete (but that is probably a minor detail. missing at least Scotland, England, Wales but they are easily added).
There are a few culprits in the table like `:man-wrestling:` pointing to https://assets-cdn.github.com/images/icons/emoji/unicode/1f93c-2642.png?v8 so it looks lake 2 codes but in fact it are 4: U+1F93C U+200D U+2642 U+FE0F but comparing this table with the https://unicode.org/emoji/charts/full-emoji-list.html table should do the trick.
This gives a good starting point, translation of this table should not be a big issue for me.

reavessm:
Yeah there definitely are some holes, but here is a json with the unicodes from the table
[emojis.json.txt](https://github.com/doxygen/doxygen/files/2389003/emojis.json.txt)

t-b:
Aren't the github names just the unicode "CLDR Short Name" names all lower case and with `_` instead of spaces? Standardzing on some arbitrary github names does not sound like a good solution.

lrineau:
[**@t-b**](https://github.com/t-b) commented on [Sep 17, 2018, 4:48 PM GMT+2](https://github.com/doxygen/doxygen/issues/6517#issuecomment-422047156 "2018-09-17T14:48:35Z - Replied by Github Reply Comments"):
> Aren't the github names just the unicode "CLDR Short Name" names all lower case and with `_` instead of spaces? Standardzing on some arbitrary github names does not sound like a good solution.

[...]

PR #6608, opened 2018-11-07
Create command for escaped equal sign

albert-github:
This command writes an equal sign (`=`) to the output. This  character sequence has to be escaped in some cases, because it is used in Markdown header processing.


albert-github:
When integrating pay attention to #6517 as here also a new command is added (multiple numbers in cmdmapper.h).

PR #6609, opened 2018-11-09
\todo in table cells breaks table and todo-list

albert-github:
A table (sub-)command should also end a `\xrefitem` or similar commands (`\todo`, `\bug` etc.)


albert-github:
With doxygen 1.13.2 I still see the problem, with doxygen 1.14.0 the problem is gone.


So issue is obsolete as I think the problem has been solved in the mean time.

PR #6614, opened 2018-11-12
issue 6612 Issue with handling of emoji

albert-github:
Regression on adding emoji.
- in case of multiple levels of `::` (e.g. `U::V::W`) this resulted in the fact that `:V:` was seen as an emoji and handled as such
- in a citelist it is possible to have `pages = {104:1--104:8},`, `number = {4},` and `volume = {28},` but this is translated into `28(4):104:1&ndash;104:8` giving a false positive on `:104:`. the citelist has been excluded from emoji handling.

### DIFF
--- a/src/doctokenizer.l
+++ b/src/doctokenizer.l
@@ -38,6 +38,7 @@
 #include "definition.h"
 #include "doxygen.h"
 #include "portable.h"
+#include "cite.h"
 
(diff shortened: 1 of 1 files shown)


lrineau:
Great. Thanks for your reactivity, both of you! (@albert-github and @doxygen).

PR #6622, opened 2018-11-17
Adding perlmod tree nodes  documentation to LaTeX / PDF

albert-github:
Adding the perlmod tree nodes documentation in an identical way to HTML and LaTeX / PDF.
A requirement for this is that LaTeX supports more than the default 4 levels in the 'itemize' environment, therefore the 'itemize' environment has been extended to 12 levels (and for consistency purposes also the 'enumerate' environment). This is all handled in the 'doxygen.sty' file.


albert-github:
In case the pull request #8807 has been merged beforehand the following has to be changed:
- `latexdocvisitor.h`
  change the line::
  ```
  const int latex_maxIndentLevels = 5;
  ```
  into:
  ```
  const int latex_maxIndentLevels = 13;
  ```
- `latexdocvisitor.cpp`
  change the line
  ```
  m_t << "\n\\setcounter{enum"<< integerToRoman(m_indentLevel,false).data() << "}{" << val - 1 << "}";
  ```
  into"
  ```
  m_t << "\n\\setcounter{DoxyEnumerate"<< integerToRoman(m_indentLevel,false).data() << "}{" << val - 1 << "}";
  ```

Issue #6632, opened 2018-11-21
References to Objective C protocols by name broken in 1.8.12


mstorsjo:
Since 1.8.12 (in particular, since 00ee930a1d73e11885197102c54fd4c8141127da), direct refererences to objective C protocols fail to be resolved.

This can be observed with the following test header:

```
@protocol MyProtocol <NSObject>
/**
 * Some callback. 
 */
- (void) callbackMethod;
@end

@interface MyClass
/**
 * See #MyProtocol::callbackMethod.
 */
- (void) directMethod;
@end
```

With 1.8.11 and commits up to the referenced one, this builds fine and produces a hyperlink, rendered like this:
> See callbackMethod (MyProtocol-p).

With 1.8.12 and later, it fails to produce a hyperlink and gives a warning:
```
header.h:9: warning: explicit link request to 'MyProtocol::callbackMethod' could not be resolved
```

mstorsjo:
This seems to be related to #3406, although the existing setup worked fine before 1.8.12.

mstorsjo:
Thanks! As I'm unfamiliar with the doxygen internals I can't comment much about it, but I can at least confirm that it does indeed fix the full original issue I saw in addition to the reduced testcase.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.15.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

mstorsjo:
I can confirm that this issue indeed is fixed in 1.8.15.

PR #6636, opened 2018-11-23
issue #6632 References to Objective C protocols by name broken in 1.8.12

albert-github:
The issue is the change in the spacing for protocols.

In 1.8.11 the option `-d functions` gave (a.o): `member callbackMethod of class MyProtocol-p`
In 1.8.12 the result is: `member callbackMethod of class MyProtocol -p`
(Note the extra space).

### DIFF
--- a/src/reflist.cpp
+++ b/src/reflist.cpp
@@ -162,7 +162,10 @@ void RefList::generatePage()
       if (item->scope->name() != "<globalScope>")
       {
         doc += "\\_setscope ";
-        doc += item->scope->name();
+	if (item->scope->name().right(3)==" -p")
+	  doc += item->scope->name().left(item->scope->name().length()-3) + "-p";
+        else
+          doc += item->scope->name();
         doc += " ";
       }
     }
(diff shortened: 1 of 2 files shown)


src/reflist.cpp
```diff
@@ -162,7 +162,10 @@ void RefList::generatePage()
       if (item->scope->name() != "<globalScope>")
       {
         doc += "\\_setscope ";
-        doc += item->scope->name();
+	if (item->scope->name().right(3)==" -p")
```
doxygen: I think we should find the cause of the space rather that dealing with it, because the -p is added only internally to avoid naming conflicts (class and protocol with the same name is allowed).

albert-github: I do agree, couldn't find it at first glance and the routine (as indicated in the question) that added the space is a quite often used routine. So I was a bit reluctant to do it. I'll have another look at it.

doxygen: I already found the problem and fixed it as part of this pull request.

albert-github: OK thanks.

PR #6638, opened 2018-11-23
Doxygen creates empty image titles for Docbook output

albert-github:
The title of an image needs to be at a specific place in docbook. The current place was incorrect

(based on https://stackoverflow.com/questions/53448764/doxygen-creates-empty-image-titles-for-docbook-output)

### DIFF
--- a/src/docbookgen.cpp
+++ b/src/docbookgen.cpp
@@ -904,7 +904,8 @@ DB_GEN_C
 void DocbookGenerator::endClassDiagram(const ClassDiagram &d, const char *fileName,const char *)
 {
 DB_GEN_C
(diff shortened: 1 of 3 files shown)


doxygen:
I refactored the code a bit, so please take a look. In particular I noticed that in visitPreStart() an empty string was passed where now `height` is assigned to depth. Not sure if that was a bug or a deliberate choice (if it was intended then please fix and add a comment).

```
  if (!height.isEmpty())
  {
    t << " depth=\"" << convertToDocBook(height) << "\"";
  }
```

albert-github:
I don't think the empty string was the intention.
I do see a strange thing in case of a specified `height` but no specified `width`, in that case the `width` is set to `50%`.
I'll have a look at this (and create when necessary a new pul request).

PR #6641, opened 2018-11-24
Incorrect number of start / end paragraph tags for xhtml with image command

albert-github:
In case of `\image` a force closed paragraph is not always force opened again.
Problem can be seen with the default doxygen test 31 (`[031_image.dox]: test the \image command`).


doxygen:
I didn't like the need to keep state for forceEndParagraph() / forceStartParagraph() so I solved it differently. Please check if it still has the correct effect.

albert-github:
Don't like those state variables either, but didn't see a good solution as the force routines were already headache routines in the past. First test looks OK, will do some more thorough check later on.

Issue #6657, opened 2018-12-03
"QGDict::hashAsciiKey: Invalid null key" when using anonymous union/struct


reylau:
I am using Doxygen V1.8.14. The following code generates three "QGDict::hashAsciiKey: Invalid null key" errors:

```
/************************************************************************************************************************************************************************************/
/*                                                                                                                                                                                  */
/* @defgroup   Sft        Software Timers.                                                                                                                                          */
/* @brief                 Provides software timers to compensate for the shortage of hardware timers.                                                                               */
/*                                                                                                                                                                                  */
/************************************************************************************************************************************************************************************/

#include <stdint.h>

/** @defgroup SftInt      Module Interface.
 *  @ingroup  Sft
 *  @brief                This describes public module objects, i.e. objects which may be used by the rest of the FW.
 */

/** @defgroup SftICtVar1  1.129. Objects for Managing the Soft Timers.
 *  @ingroup  SftInt
 *  @{
 */

/** @brief Empty comment */
static volatile struct
   {
   union
      {
      struct
         {
         uint8_t : 1;
         }
      structFlag;
      }
   unionRun;

   struct
      {
      uint8_t ui8ConfigurationComplete;
      uint8_t ui8ConfigurationWordReceptionTimeout;
      uint8_t ui8ConfigurationAcknowledgmentRepetitionPeriod;
      }
   structCounter;
   }
            module_vstructSoftTimers;
/** @} */
```

Notice that the union has been simplified and contained useful members in the original code (which runs OK). If I am re-writing the code as follows, I do not get any error:

```
/************************************************************************************************************************************************************************************/
/*                                                                                                                                                                                  */
/* @defgroup   Sft        Software Timers.                                                                                                                                          */
/* @brief                 Provides software timers to compensate for the shortage of hardware timers.                                                                               */
/*                                                                                                                                                                                  */
/************************************************************************************************************************************************************************************/

#include <stdint.h>

/** @defgroup SftInt      Module Interface.
 *  @ingroup  Sft
 *  @brief                This describes public module objects, i.e. objects which may be used by the rest of the FW.
 */

/** @defgroup SftICtVar1  1.129. Objects for Managing the Soft Timers.
 *  @ingroup  SftInt
 *  @{
 */

/** @brief Empty comment */
union unionRun
   {
   struct
      {
      uint8_t : 1;
      }
   structFlag;
   };

/** @brief Empty comment */
struct structCounter
   {
[...]

PR #6659, opened 2018-12-04
issue #6657 "QGDict::hashAsciiKey: Invalid null key" when using anonymous union/struct

albert-github:
Protect findAt against empty string.


doxygen:
Made the fix a little simpler by using `i>0` as guard, then `scope` cannot be empty.

PR #6662, opened 2018-12-06
CI: Updates to Qt and Winflexbison

Croydon:
* I have created a winflexbison Conan recipe, so we can use this now to install winflexbison and don't need to do it "manual" anymore
  * Also updating winflexbison 2.5.15 -> 2.5.16
  * updating Qt for macOS from 5.11.1 to 5.11.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ before_script:
         conan remote add bincrafters https://api.bintray.com/conan/bincrafters/public-conan;
         printf "[requires]
         libxml2/2.9.8@bincrafters/stable
-        Qt/5.11.1@bincrafters/stable" >> conanfile.txt;
+        qt/5.11.3@bincrafters/stable" >> conanfile.txt;
         conan install . -g virtualrunenv;
         source activate_run.sh;
       fi;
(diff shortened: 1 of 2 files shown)


doxygen:
I see appveyor failed to build with:

```
win_bison: C:\Users\appveyor\.conan\data\winflexbison\2.5.16\bincrafters\stable\package\6cc50b139b9c3d27b3e9042d5f5372d327b3a9f7\bin\data/m4sugar/m4sugar.m4: cannot open: No such file or directory
```

so the bison build/install does not seen to be correct.

Croydon:
The winflexbison recipe didn't copy correctly the `data` directory, so it's missing. A fix is in our CI pipeline, but may still take 12-24h until it's done, since some huge other recipes are scheduled before that. When it's done I will rebase this pull request to re-trigger the doxygen CI.

Croydon:
Background: The upstream winflexbison project had the `data` directory and its content twice. The one path our Conan recipe copied it from got removed in the 2.5.16 release, but I forget to adapt our recipe in this regard, hence this error.

Croydon:
Okay. AppVeyor is now fixed.

However, for some reason Travis started to fail. The very same build did succeed in my repository however...

https://travis-ci.org/doxygen/doxygen/builds/465508189

and also this build here did succeed as well 
https://travis-ci.org/doxygen/doxygen/builds/465511715

So no idea why the status check here says failing..

albert-github:
I restarted the failing (mac) run and it passed now (probably a download glitch on travis).

PR #6664, opened 2018-12-06
XHTML problem with name attribute with VHDL name attribute

albert-github:
When running xhtml checker on the doxygen mux example we get:
```
Syntax of value for attribute name of a is not valid
Document mux/xhtml/classmux__using__with.xhtml does not validate
```
This is due to a space in the name tag, substituting the appropriate code solves the problem (double calling of the routine has no effect).

### DIFF
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -1547,7 +1547,7 @@ void HtmlGenerator::startMemberHeader(const char *anchor, int typ)
   t << "<tr class=\"heading\"><td colspan=\"" << typ << "\"><h2 class=\"groupheader\">";
   if (anchor)
   {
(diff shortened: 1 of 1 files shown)


doxygen:
Double calling of convertToId() does have an effect:

Example
```
printf("v1=%s\n",convertToId("Hello World").data());
printf("v2=%s\n",convertToId(convertToId("Hello World")).data());
printf("v3=%s\n",convertToId(convertToId(convertToId("Hello World"))).data());
```
gives
```
v1=Hello_20World
v2=Hello_5F20World
v3=Hello_5F5F20World
```
So I think this change needs some more thought.

albert-github:
I overlooked the `_`, I ;ll have to have a look.

albert-github:
Superseded by #6671

PR #6672, opened 2018-12-10
Update parsing extension table in documentation

albert-github:
The text underneath the table in the documentation regarding the use of extensions was not correct.
Also added the C / C++ category and made the table a bit better readable.


albert-github:
Reported by : https://stackoverflow.com/questions/53704753/problem-with-doxygen-and-templates-in-tpp-file

PR #6676, opened 2018-12-12
Don't create anchors for anonymous groups (\name)

albert-github:
An anchor with the value `[NOHEADER]` was created for anonymous `\name` commands, resulting in the possibility multiple times the same anchor on a page.
Furthermore there were several tests on `[NOHEADER]` although the header was never set to `[NOHEADER]` (except locally when calculating the anchor), test has been changed to test on empty headers.


albert-github:
Made obsolete by #9216

PR #6681, opened 2018-12-14
Adjustment of xhtml1-transitional.dtd

albert-github:
Adding attributes:
- async
- doxygen

to be able to validate doxygen generated xhtml documents


lrineau:
Maybe you should rename the file, so that it is clear, without reading your comments, that it is a modified version.

albert-github:
Problem is that in that case we also have to adjust our generated files and at that moment they don't validate anymore. Now they do validate although giving a warning on `doxygen=` and `async=`.

PR #6683, opened 2018-12-14
Upgrade to jQuery 1.7.2 to get rid of security scan violations.

lewijw:
Please accept this contribution to upgrade to jQuery 1.7.2.  Our security scans are flagging a problem with jQuery 1.7.1.


lewijw:
The jQuery version was obtained via:
curl https://code.jquery.com/jquery-1.7.2.js > /doxygen/jquery/jquery-1.7.2.js

lewijw:
It is the minimum change needed to get the security scans to pass.

lewijw:
Following up on the question from @albert-github concerning why a later version of jQuery was not used, I have tried later versions.  Unfortunately, even going to the next version past 1.7.2 did not work.  With version 1.8.0, the home page of our documentation did not render correctly.

PR #6698, opened 2018-12-20
Print emoji text in case of unknown emoji

albert-github:
In case an unknown emoji was detected the 'value' was not printed to the output, this is corrected.

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -2779,6 +2779,7 @@ QCString DocLink::parse(bool isJavaLink,bool isXmlLink)
         case TK_EMOJI: 
           warn_doc_error(g_fileName,doctokenizerYYlineno,"Unsupported emoji '%s' found",
               qPrint(g_token->name));
+          m_children.append(new DocWord(this,g_token->name));
           break;
         case TK_HTMLTAG:
           if (g_token->name!="see" || !isXmlLink)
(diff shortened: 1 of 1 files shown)


albert-github:
Emoji are, in e.g. HTML represented by a sequence of unicode characters, in e.g.LaTeX there is not that support and we have to use png files ( :-( ).
In texts emoji are in general represented by `:text:` and later on converted to the real characters, for HTML doxygen does the same and for e.g. LaTeX converts it to a call to the png image.
In the doxygen (master version) documentation there is a chapter about emoji.
Hope this helped a little bit.

Croydon:
> In texts emoji are in general represented by :text: and later on converted to the real characters,

This is more like a help to represent emojis and in this case I understand that it needs extra handling.

I could also directly enter the unicode symbols and it should not need any specific handling. If I'm using this it likely means that I'm also positive that my system and the system of my colleagues have support of new enough unicode versions to handle it. So it also doesn't need to be replaced by an image version as it should be up to the font/system how they are getting displayed.

albert-github:
Did you try it in LaTeX? as there I didn't get it to work with the unicode definition unless using the images, if you have a solution for this I'm interested.

When one wants to use an emoji it is quite hard to get the right code and especially for complex ones (multiple unicode characters for one emojii) .

ob-ivan:
I'm using Doxygen 1.8.15 and I still get false positives when I write documentation about time and refer to specific times in the form of `02:59:59` (one second before 3am). Doxygen detects the middle `:59:` as an emoji and renders my time to `0259`. I'm not sure why it doesn't print the source code as this PR seems to implement.

Should I file a new issue for that?

albert-github:
@ob-ivan 

Strange I tried a s,all test like:
```
/// \file
/// the test 2:59:59 the rest
```
and see with 1.8.15 the expected result (i.e. `2.59:59`).

So yes please file a new issue.

Issue #6691, opened 2018-12-19
Multiple issues with emoji matching


mosra:
Similar to #6612, related to emoji support (#6517) added in #6534 by @albert-github. I have multiple false positives on emoji being detected in [Magnum](https://magnum.graphics) documentation. One example is this:

```cpp

/**
…

Aspect ratio    | CSS class (landscape/portrait)
--------------- | ------------------------------
1:1             | `.aspect-1-1` 
4:3 (1.33:1)    | `.aspect-4-3` (default) / `.aspect-3-4`
3:2 (1.5:1)     | `.aspect-3-2` / `.aspect-2-3`
16:9 (1.78:1)   | `.aspect-16-9` / `.aspect-9-16`
2:1             | `.aspect-2-1` / `.aspect-1-2`

…
*/
```

producing absolutely unwanted and totally bogus warnings such as:

```
platforms-html5.dox:293: warning: Unsupported emoji ':3 (1.33:' found
platforms-html5.dox:297: warning: Unsupported emoji ':2 (1.5:' found
platforms-html5.dox:301: warning: Unsupported emoji ':9 (1.78:' found
```

Another is in code snippets like

```cpp
/**
… `Platform::*Application::viewportEvent(const Vector2i&)` …
*/
```

where the asterisk is a wildcard for multiple types such as `Sdl2Application`, `GlfwApplication` etc., producing 

```
changelog.dox:661: warning: Unsupported emoji ':*Application:' found
```

Some suggestions about how to fix this:

- match emoji names as `[a-zA-Z0-9_-]` and nothing else (I don't think I ever encountered any other characters)
- don't accept a space in a emoji name (*seriously?!*)
- don't *ever* look for emojis in code snippets and inline code (I doubt there's any use case for that, if really needed, people can always insert the matching unicode character directly)
- have emojis **opt-in** via a config option, because adding `\:` everywhere just because there's a new feature with dubious use is absolutely *not* acceptable

albert-github:
Some thoughts about the points mentioned:
- match emoji names as [a-zA-Z0-9_-] and nothing else (I don't think I ever encountered any other characters)
  When looking at the names of the as defined in https://unicode.org/emoji/charts/full-emoji-list.html we see here quite a few "CLDR short names" are with special characters e.g.:
  - `keycap: *`
  - `+1`
  - ` b button (blood type)`
  - `flag Antigua & Barbuda`
  - `mrs. claus`
[...]

PR #6699, opened 2018-12-20
issue #6691 Multiple issues with emoji matching

albert-github:
This fix is regarding the false positive emoji detection of scope rules like: `Platform::*Application::viewportEvent(const Vector2i&)`
In case the "emoji" is followed by and extra `:` it is not seen as an emoji anymore.

### DIFF
--- a/src/doctokenizer.l
+++ b/src/doctokenizer.l
@@ -689,7 +689,7 @@ REFWORD_NOCV   {FILEMASK}|{LABELID}|{REFWORD2_NOCV}|{REFWORD3}|{REFWORD4_NOCV}
                          g_token->name = yytext;
                          return TK_SYMBOL;
                        }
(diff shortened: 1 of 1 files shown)


mosra:
I'm absolutely serious here, if there are no regression test cases for this (actually, looking closer, there's no test for any of the emoji functionality at all), the project stability will only go further downhill.

lrineau:
In my opinion, that change of the parser, without a configuration option, was a bad idea. That changes the Doxygen language syntax, and thus the compatibility, just for a new marginal feature.

I really appreciate your work, @albert-github, but for that decision, it was an error to enable it by default. In my opinion, the option should be created, and set OFF by default, before the release of Doxygen 1.8.15 (Cc: @doxygen).

Issue #6690, opened 2018-12-19
Regression in handling of shorthand signed/unsigned types in function parameters (with bisect and test case)


mosra:
The following reduced code snippet produces a warning:

```
a.cpp:1: warning: unable to resolve reference to `foo(float)' for \ref command
```

```cpp
/**
@brief Foo

The following produces a warning: @ref foo(float)
*/
struct Foo {
    /** @brief Float */
    void foo(float value);

    /** @brief Unsigned */
    void foo(unsigned value);
};

// to fix the warning, either:
// - comment out this definition,
// - or change both to say `unsigned int value`
void Foo::foo(unsigned value) {}
```

This is a regression since 1.8.14, I bisected it to https://github.com/doxygen/doxygen/issues/2859 (commit 657a6b5348d453fd1351b8c3238426e25acdbbb9). In particular I think the check should explicitly check for `unsigned int` being unnamed parameter, while anything else (such as `unsigned value`) being named (and the same for `signed`). For completness, here's the (wrong) generated XML output:

```xml
      <memberdef kind="function" ...>
        <type>void</type>
        <definition>void Foo::foo</definition>
        <argsstring>(unsigned value)</argsstring>
        <name>foo</name>
        <param>
          <type>unsigned value</type>
        </param>
```

while it should be

```xml
      <memberdef kind="function" ...>
        <type>void</type>
        <definition>void Foo::foo</definition>
        <argsstring>(unsigned value)</argsstring>
        <name>foo</name>
        <param>
          <type>unsigned</type>
          <declname>value</declname>
        </param>
```

albert-github:
I've just pushed a proposed patch, pull request #6700

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.15.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6700, opened 2018-12-20
issue #6690 Regression in handling of shorthand signed/unsigned types in function parameters (with bisect and test case)

albert-github:
For some keywords not the value of the 'type' should decide whether the 'name' in the name field is part of the 'type' or is the 'name' in the argument.


mosra:
Yes, but can you please add the test case from #6690? Fixing bugs this way won't improve the test coverage at all and nothing prevents the bug from being reintroduced again.

doxygen:
I remove the check on type, and completed the list of default types and moved them into a dictionary. Also added a test case as suggested.

doxygen:
good point. I've added them along with `char8_t` (new in C++20) and `short`.

PR #6701, opened 2018-12-20
+DOT_SHAPE

makelinux:
To allow cleaner dot graph with text and edges only.


doxygen:
Code wise the change looks ok, but it is very specific option changing only one small detail. I'm not sure it is worth a separate option.

Issue #6277, opened 2018-07-02
Increasing access of inherited C++ members with 'using...' is not recognized by Doxygen (Origin: bugzilla #789168)


doxygen:
status REOPENED severity *normal* in component *general* for *---*
Reported in version *1.8.13* on platform *Other*
Assigned to: Dimitri van Heesch

On 2017-10-18 17:45:33 +0000, the.crazed.cat.herder@gmail.com wrote:

> Consider the following C++ code:
> 
> class BaseClass
> {
> public:
> 
> 	BaseClass() { };
> 	virtual ~BaseClass() {};
> 
> protected:
> 
> 	void SomeFunc() {};
> 
> };
> 
> 
> class DerivedClass1 : public BaseClass
> {
> public:
> 
> 	DerivedClass1() {};
> 	virtual ~DerivedClass1() {};
> 
> 	using BaseClass::SomeFunc;
> 
> };
> 
> 
> class DerivedClass2 : protected BaseClass
> {
> public:
> 
> 	DerivedClass2() {};
> 	virtual ~DerivedClass2() {};
> 
> 	using BaseClass::SomeFunc;
> 
> };
> 
> In both these sample derived classes, BaseClass::SomeFunc has had it's access changed to 'public' by employing the keyword 'using'. This is true regardless of the base class access level (public in DerivedClass1, protected in DerivedClass2), and regardless of the member access level originally defined in BaseClass (protected, in this example.)
> 
> Unfortunately, Doxygen insists that SomeFunc is still protected in both example derived classes. This is incorrect - employing 'using' in this manner to increase the access of the inherited BaseClass::SomeFunc is perfectly valid C++ (as of C++11, I believe), and VC++ recognizes this; it is, in fact, publicly accessible in both derived classes.
> 
> See http://en.cppreference.com/w/cpp/language/using_declaration for more information.

On 2017-10-29 10:48:09 +0000, Dimitri van Heesch wrote:

> Confirmed. Should be fixed in the next GIT update.

On 2017-10-31 01:29:06 +0000, the.crazed.cat.herder@gmail.com wrote:

> Thanks very much!
> 
> Soo... errr.... I don't suppose I could convince you to release an updated set of Windows binaries....?
> 
> :)

On 2017-11-19 12:09:39 +0000, VladimÃ­r VondruÅ¡ wrote:

> Hi,
> 
> was not sure what's the proper place where to report this, so I'm commenting here. 
> 
[...]

PR #6714, opened 2018-12-29
Don't crash or generate invalid XML for empty TOC

mosra:
The new XML TOC implementation from https://github.com/doxygen/doxygen/pull/736 is not handling TOC on pages without `@section`s correctly, accessing a null pointer. This adds a corresponding test and a fix. Interestingly enough, the docbook and XHTML generators are already handling this case properly.

I also added a test for TOC levels that could lead to bugs like those fixed in #6696 and #6697. The XML output is not affected by these but having an explicit test for it certainly won't hurt.

What concerns me and what I *would like to have fixed in additon to merging this PR* is that on a larger project ([Magnum](https://magnum.graphics)), `doxygen` didn't crash on this but rather exited with a success return code and generated empty XML for given page and truncated `index.xml`. That's similar to the issue with a fix for  https://github.com/doxygen/doxygen/issues/6277 that I reported a year ago (also accessing a null pointer, fixed in https://github.com/doxygen/doxygen/commit/1a1fdbed64de6ce01959b2e4d0988be823fb6bad). My question/request is:

- Why `doxygen` executed by `runtests.py` crashes in this case and why `doxygen` executed on a larger project fails silently, making the error much more confusing and harder to spot? Is it because threads are involved?
- Is it possible to propagate the aborted operation (from the thread?) all the way to the return code of the `doxygen` executable, so the crash is never silent?

### DIFF
--- a/src/xmlgen.cpp
+++ b/src/xmlgen.cpp
@@ -1818,10 +1818,10 @@ static void generateXMLForPage(PageDef *pd,FTextStream &ti,bool isExample)
     }
   }
   writeInnerPages(pd->getSubPages(),t);
-  if (pd->localToc().isXmlEnabled())
+  SectionDict *sectionDict = pd->getSectionDict();
+  if (pd->localToc().isXmlEnabled() && sectionDict)
   {
     t << "    <tableofcontents>" << endl;
-    SectionDict *sectionDict = pd->getSectionDict();
     SDict<SectionInfo>::Iterator li(*sectionDict);
     SectionInfo *si;
     int level=1,l;
--- a/testing/079/empty.xml
+++ b/testing/079/empty.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="">
+  <compounddef id="empty" kind="page">
+    <compoundname>empty</compoundname>
+    <title>An empty page</title>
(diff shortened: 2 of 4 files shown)


mosra:
> The underlying problem is that the writing of the XML output follows a quite different way than (X)HTML, LaTeX, RTF, Docbook and ... do 

Looking at how much of the logic is repeated for every output format, I'm thinking that the best way would be to have XML as an intermediate format with all others created from it to avoid repeating the same code. That could be for example via XSLT for the XML-based outputs (HTML and Docbook) and via some templating logic (similar as Python's Jinja) for the others. But that's for a separate discussion...

> Doxygen hardly uses threads of its own (only for dot as far as I know) [...] I think that the not crashing is "luck" in the sense that a not fatal part of the memory is hit and used correctly

It results in a file being written partially, so there has to be some sort of crash. It's accessing memory at address 0, no amount of "luck" should make it survive. Threads is one explanation, another could be that `SIGABRT` / `SIGSEGV` signals are intercepted (for whatever reason) and instead of crashing they get ignored. I know very little about the internals so I don't even know where to look -- and it happens on a project that takes several seconds to process. It's also maybe happening only on Linux, I have no idea how the behavior of this is on Windows. In case you'd want to try to reproduce:

```sh
git clone git://github.com/mosra/magnum
git clone git://github.com/mosra/magnum-extras # this one contains the empty page TOC
cd magnum
mkdir build
doxygen Doxyfile-mcss # ignore warnings, these are due to missing tagfiles from other projects

# if it doesn't crash, build/doc/index.xml should be truncated and build/doc/ui.xml empty
```

Issue #6708, opened 2018-12-28
Invalid UTF-8 characters in hover title


HayoBaan:
While the warning from https://github.com/doxygen/doxygen/issues/6370 is indeed gone in 1.8.15, the text shown in the hover in the dependency graph still has the UTF-8 characters showing incorrectly e.g. `“true”` gets replaced with `âtrueâ` (c3 a2 c2 80 c2 9c 74 72  75 65 c3 a2 c2 80 c2 9d in hex).

For an example, see:
[doxytest_utf8.zip](https://github.com/doxygen/doxygen/files/2715084/doxytest_utf8.zip)

albert-github:
Problem looks like to be in the conversion to png when using svg it looks OK (as far as I can see). 
png situation has to be investigated further.

albert-github:
I had a closer look and with the dot 2.38.0 version (on Windows, dot -V) I get the problems as indicated, with the version 2.40.1 (on Cygwin) I saw the right tooltips, so I think the problem is the Graphviz dot version.

I found a version 2.40.1 for Cygwin, but I haven't been able to locate a native 2.40.1 (or newer) version for Windows.

HayoBaan:
Interesting, so you're saying, the tooltip is generated by graphviz not doxygen itself?

However, as I'm already on 2.40.1 the version alone doesn't explain the issue. I'm on a Mac though, not on windows.

Note: the correct encoding for `“` would be e2 80 9c which gets translated into c3 a2 c2 80 c2 9c which is the double encoded form of e2 80 9c… So somewhere strings are getting doubly encoded…

albert-github:
I re-evaluated and you are right (I have no Mac so I cannot test it there, but looks like the platform is not relevant).
In the 1.8.14 it works with dot 2.38.0 and 2.40.1.
Problem was that the default version for me on Cygwin is 1.8.14 and on Windows 1.8.15.
I now ran both (Windows and Cygwin so 2.38.0 and 2.40.1) with 1.8.14 and it is OK and with 1.8.15 there is a problem.
Interesting is that the basic information (dot and map file) looks OK but the `area` information in the html file isn't (1.8.16 is in my case still identical to 1.8.15):
```
>diff -r -b -w html_windows_1_8_14/file1_8c.html html_windows/file1_8c.html

32c32
< <!-- Generated by Doxygen 1.8.14 -->
---
> <!-- Generated by Doxygen 1.8.16 -->
78,79c78,80
[...]

PR #6719, opened 2018-12-31
issue #6708 Invalid UTF-8 characters in hover title

albert-github:
The replace function has as side effect that it affect UTF-8 characters as well.
Removing   `id="node[0-9]*"` no by searching its start and end.

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -377,7 +377,6 @@ static bool convertMapFile(FTextStream &t,const char *mapName,
                            const QCString &context=QCString())
 {
   QFile f(mapName);
-  static QRegExp re("id=\"node[0-9]*\"");
   if (!f.open(IO_ReadOnly)) 
   {
     err("problems opening map file %s for inclusion in the docs!\n"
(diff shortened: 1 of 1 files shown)


src/dot.cpp
```diff
@@ -396,7 +395,17 @@ static bool convertMapFile(FTextStream &t,const char *mapName,
 
       if (buf.left(5)=="<area")
       {
-        t << replaceRef(buf,relPath,urlOnly,context).replace(re,"");
+	QCString replBuf = replaceRef(buf,relPath,urlOnly,context);
+        int indexS = replBuf.find("id=\""), indexE;
+        indexE=replBuf.find('"',indexS+4);
+        if (indexS>=0 && (indexE=buf.find('"',indexS))!=-1)
```
doxygen: This doesn’t look right. `indexE` is already assigned, you probably meant `indexE>indexS`?

albert-github: Looks indeed like some initial debug statement remained, I'll have a look (results were OK but this code isn't).

albert-github: Corrected code.

src/dot.cpp
```diff
@@ -397,8 +397,7 @@ static bool convertMapFile(FTextStream &t,const char *mapName,
       {
 	QCString replBuf = replaceRef(buf,relPath,urlOnly,context);
         int indexS = replBuf.find("id=\""), indexE;
-        indexE=replBuf.find('"',indexS+4);
-        if (indexS>=0 && (indexE=buf.find('"',indexS))!=-1)
+        if (indexS>=0 && (indexE=replBuf.find('"',indexS+4))!=-1)
```
doxygen: I guess `indexS` has to be bigger than 0 otherwise `left(indexS-1)` will refer to a negative size

albert-github: In principle yes, but in this case it is OK as the line `if (buf.left(5)=="<area")` already gives a length > 0. when the find is executed.

PR #6722, opened 2018-12-31
Make var & function declaration/definition location parsing not order-dependent

mosra:
The [m.css Doxygen theme](https://mcss.mosra.cz/doxygen/) provides include information also for free functions, variables and typedefs (instead of just classes), as [can be seen for example here](https://doc.magnum.graphics/magnum/namespaceMagnum_1_1Primitives.html#a5687268cfce168ff39376d0be228eab7):

image

To do that, the script is extracting the `<location>` information from the XML files. Previously, if a definition was processed before declaration, both the declaration and definition location was set to the definition. It was working correctly only when declaration was found before definition, but due to the fact that C/C++ files are named like this:

    foo.cpp
    foo.h  

And thus when globbing the filesystem, the definition is always seen before the declaration. Note that this affects only functions inside namespaces, (in my testing) functions in global scope were reported
twice, once for each file (which can be considered okay, as it doesn't exhibit this bug).

For variables, I'm distinguishing definitions and declarations based on presence of the `extern` keyword -- declarations have it, definitions not.

In case it's not clear what this PR does -- there's a test in the first commit, run it with current Doxygen `master` to see the difference. Oh, also -- I'm not well aware of all the interactions in Doxygen internals, so hopefully changing the source location this way doesn't lead to any inconsistencies in the internal state.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -2597,6 +2597,15 @@ static MemberDef *addVariableToFile(
               "    variable already found: scope=%s\n",qPrint(md->getOuterScope()->name()));
           addMemberDocs(rootNav,md,def,0,FALSE);
           md->setRefItems(root->sli);
+
+          // If md is not external and current is, update declaration location
+          // to current. In case md is external and current isn't, declaration
+          // and definition location is already correct.
+          if(root->explicitExternal && !md->isExternal())
+          {
+            md->setDefFile(root->fileName,root->startLine,root->startColumn);
+          }
+
           return md;
         }
       }
@@ -3748,10 +3757,22 @@ static void buildFunctionList(EntryNav *rootNav)
                 }
 
                 // if md is a declaration and root is the corresponding
-                // definition, then turn md into a definition.
+                // definition, then turn md into a definition. Declaration
+                // location and definition location is already correct in this
+                // case.
                 if (md->isPrototype() && !root->proto)
                 {
                   md->setPrototype(FALSE);
+                }
+                // If md is a definition and we are merging it with a
+                // declaration, update declaration location. Definition
+                // location is already correct. This is a bit random when there
+                // are multiple declarations for a single definition, but such
+                // behavior is still better than conflating everything to a
+                // single location.
+                else if(!md->isPrototype() && root->proto)
+                {
+                  md->setDefFile(root->fileName,root->startLine,root->startColumn);
                 }
               }
             }
--- a/src/definition.cpp
+++ b/src/definition.cpp
@@ -56,6 +56,7 @@ class DefinitionImpl
   public:
     DefinitionImpl();
    ~DefinitionImpl();
+    void setDefFileName(const QCString &df);
     void init(const char *df, const char *n);
 
     SectionDict *sectionDict;  // dictionary of all sections, not accessible
@@ -112,7 +113,7 @@ DefinitionImpl::~DefinitionImpl()
   delete inbodyDocs;
 }
 
-void DefinitionImpl::init(const char *df, const char *n)
+void DefinitionImpl::setDefFileName(const QCString &df)
 {
   defFileName = df;
   int lastDot = defFileName.findRev('.');
@@ -120,6 +121,11 @@ void DefinitionImpl::init(const char *df, const char *n)
   {
     defFileExt = defFileName.mid(lastDot);
   }
+}
+
+void DefinitionImpl::init(const char *df, const char *n)
+{
+  setDefFileName(df);
   QCString name = n;
   if (name!="<globalScope>") 
   {
@@ -395,6 +401,13 @@ Definition::~Definition()
   }
   delete m_cookie;
   m_cookie=0;
+}
+
+void Definition::setDefFile(const QCString& df, int defLine, int defColumn)
+{
+  m_impl->setDefFileName(df);
+  m_defLine = defLine;
+  m_defColumn = defColumn;
 }
 
 void Definition::setName(const char *name)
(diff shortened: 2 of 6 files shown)


src/doxygen.cpp
```diff
@@ -2597,6 +2597,15 @@ static MemberDef *addVariableToFile(
               "    variable already found: scope=%s\n",qPrint(md->getOuterScope()->name()));
           addMemberDocs(rootNav,md,def,0,FALSE);
           md->setRefItems(root->sli);
+
+          // If md is not external and current is, update declaration location
+          // to current. In case md is external and current isn't, declaration
+          // and definition location is already correct.
+          if(root->explicitExternal && !md->isExternal())
+          {
+            md->setDefFile(root->fileName,root->startLine,root->startColumn);
```
doxygen: This looks wrong to me. If `root->explicitExternal` is true and `md->isExternal()` is false it means root points to a variable _declaration_ not the single _definition_, and `md` probably already points to the definition. As the name suggests the `setDefFile()` should be passed the definition, not a declaration. Doxygen currently extracts documentation found at the declarations to add to the definition, and only falls back to showing the declaration if no definition has been found. So shouldn't the expression have been `!root->explicitExternal && md->isExternal()` instead?

mosra: With the (newly added) `setDefFile()` I just followed the current naming, where there's `getDefFile()` for getting declaration location and `getDefBodyFile()` / `setDefBody()` for getting / setting definition location. So, um, yes, it might be confusing, but is (I think) consistent with current state. If you have a better idea how to name it, please tell me, I'll happily update.

Besides that, the correctness is verified with the test :) Without the above, the resulting XML file would show

```xml
<location file="080_declaration_definition_order.cpp" line="11" column="1" 
          bodyfile="080_declaration_definition_order.cpp" bodystart="11" bodyend="11"/>
```

instead of (correct)

```xml
<location file="080_declaration_definition_order.cpp" line="32" column="1" 
          bodyfile="080_declaration_definition_order.cpp" bodystart="11" bodyend="11"/>
```

doxygen:
@mosra My idea was to always point to the definition and only fall back to the declaration if no definition is available (e.g. in case one only feeds the header file to doxygen). If I change it this way it is also order independent, but I don't know if you need the location of the declaration as well?

mosra:
That's not how the parsing works at the moment, though (from my experience, at least). TL;DR taken from the test file -- if I have this (with line numbers):

```cpp
1: // a.cpp
2:
3: void foo();
4: 
5: void foo() {}
```

then Doxygen produces this in the XML file (which makes sense, and nope, no fallback to the declaration is happening):

```xml
<location file="a.cpp" line="3" column="1" 
          bodyfile="a.cpp" bodystart="5" bodyend="5"/>
```

However, in the opposite case, when a declaration is found *after* a definition:

```cpp
1: // b.cpp
2: 
3: void foo() {}
4: 
5: void foo();
```

then, with current `master`, I have both definition *and* declaration pointing to the same place:

```xml
<location file="b.cpp" line="3" column="1" 
          bodyfile="b.cpp" bodystart="3" bodyend="3"/>
```

Since declarations and definitions are usually in separate files, this depends on the order in which the file are discovered and that depends on many external factors (order in `INPUT`, extensions used etc.). Thus, in this patch, I changed the second case to produce this instead:

```patch
-<location file="b.cpp" line="3" column="1" 
+<location file="b.cpp" line="5" column="1" 
           bodyfile="b.cpp" bodystart="3" bodyend="3"/>
```

The reason why I want this is that now `file`/`line`/`column` always points to the declaration (if found), and the declaration is generally in the header the user is meant to `#include` -- which I then use to show correct `#include` information for free functions.

In case one feeds only header files to doxygen, no pairs of declarations + definitions are found, thus this conflict-resolving code get never executed. And so the behavior in such case does not change with my patch.

Hope this finally clears my intentions up a bit :)

doxygen:
I've push the following commit https://github.com/doxygen/doxygen/commit/7f40e488e27bcea4bb15045df05479dc5fbd9d6d.

It adds new `declfile`, `declline`, and `declcolumn` attributes to `location` in order to store information about the declaration (if found), e.g.:

```xml
    <location file="082_decl_def.cpp" line="11" column="1" 
      bodyfile="082_decl_def.cpp" bodystart="11" bodyend="-1" 
      declfile="decl_def.h" declline="5" declcolumn="1"/>
```

As mentioned before, the `file`, `line`, and `column` attributes are meant to store definition information if found and fall back to the declaration info if not. This should now be done independent of the order. I've added test cases 082 and 083 for testing both orders.

So there are now 3 situations:
- doxygen finds a declaration and a definition for a variable or function: in this case `file`,`line`,`column` always points to the definition, and `declfile`,`declline`,`declcolumn` points to the declaration
- doxygen only finds a declaration for a variable or function: in this case both `file`,`line`,`column` and `declfile`,`declline`,`declcolumn` point to the declaration.
- doxygen only finds a definition for a variable or function: in this case `file`,`line`,`column` points to the definition and the declaration attributes are omitted.

mosra:
Oh, that's wonderful, thank you! :+1: Just tested and it seems to be doing the right thing for all cases I had -- so first I check if `declfile` is there and then if not, I fall back to `file`.

What's the desired use of `bodyfile`, by the way? My original assumption was that this is the definition location (which caused all the confusion above), but that's apparently `file`?

I guess this PR can be closed then, right? :)

doxygen:
`bodyfile` belongs together with `bodystart` and `bodyend` to indicate where the implementation body of a function is. I agree that `bodyfile` itself is rather redundant (it is always the same as `file` now, but it is kept for backwards compatibility.

PR #6723, opened 2019-01-01
Improvement of performance when use plantuml

cheoljoo:
Improvement of performance when use plantuml (FAST PLANTUML)

I followed @mosra 's comment.

# Happy New Year
## follow up
- The following is comments from ( #6467 ) second trial. 
	- In the debug statements the __PRETTY_FUNCTION__ is a GCC extension. But this project can run in windows.
		> I got it.  I removed __PRETTY_FUNCTION__.
	- In the PLANTUML_RUN_FAST all the .pu files are given to the plantuml.jar with their full path (see -d extcmd when running). I think this will lead to problems when there are quite a few files ("command line" overflow). Seen this merging the files into one file like done for formulas. A small test with one file revealed that it didn't bring a lot / nothing time wise, but here the problem with the "command line" overflow shouldn't / can't happen.
		> Yes. The command line has overflow problem.  So I change into one file.
		> I apply FAST solution as a default. So we do not need to add PLANTUML_RUN_FAST in config file.
	- I think it still would make sense to implement the 'md5' possibility for, especially, the non PLANTUML_RUN_FAST version, i.e. if the file already exists don't try to recreate the image.
		>  **This pull request solved this item.**
			- PlantumlManager::instance()->insert() : this function will be changed to skip creating png or svg images when plantuml image was existed (cached) in OUTPUT directory.

> **I tested including epstopdf in GENERATE_LATEX. It's done.**

# History
- 1st : #6465 
- 2nd : #6467 
- 3rd : #6472 

# Test Result
The example has 5 plantumls. 

|        Test Type          |  1.8.14  |  current ver (5 plantumls)    | current ver (modify 1 Plantuml in 5 plantumls)     | 
|---------------------------|------------|-------------|-----------------|
| Total Elapsed Time              | 23.370 sec | 7.349 sec   |  4.652 sec      |
| external tools Time       | 23.020 sec | 6.819 sec   |  4.301 sec      |

# Test Environments (How to test)
- The example has 5 plantumls. 
## Original Configration
- I follow the original path to process plantuml.
- Configuration of Doxyfile
    - OUTPUT_DIRECTORY       =  OUTPUT
    - SOURCE_BROWSER         = YES
      only set the html
    - GENERATE_LATEX         = NO
- Run
    - $ doxygen -d time

## Modify One Plantuml
- Modify one plantuml in 5 plantuml of example  with the same environment of PLANTUML_RUN_TEST

# How to Reduce the performance
## PLANTUML_RUN_FAST
- Make Just one pu (plantuml) file to run java once
    - Just run once when we generate HTML.
### another test with setting YES to all GENERATE_*
- When we generate several types , we run java several times according the different java arguments.
    - set YES to all GENERATE_*
    - There are 20 plantuml in this example and configuration.  But, it runs java just 4 times.
```text
Executing external command `java -Djava.awt.headless=true -jar "/Users/cheoljoo/bin/plantuml.jar" -o "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/docbook" -charset UTF-8 -tpng "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/docbook/inline_umlgraph_pngdocbook.pu" `
10.019 sec: Running PlantUML on png PlantumlFiles in html
Executing external command `java -Djava.awt.headless=true -jar "/Users/cheoljoo/bin/plantuml.jar" -o "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/html" -charset UTF-8 -tpng "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/html/inline_umlgraph_pnghtml.pu" `
17.906 sec: Running PlantUML on png PlantumlFiles in rtf
Executing external command `java -Djava.awt.headless=true -jar "/Users/cheoljoo/bin/plantuml.jar" -o "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/rtf" -charset UTF-8 -tpng "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/rtf/inline_umlgraph_pngrtf.pu" `
25.842 sec: Running PlantUML on eps PlantumlFiles in latex
Executing external command `java -Djava.awt.headless=true -jar "/Users/cheoljoo/bin/plantuml.jar" -o "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/latex" -charset UTF-8 -teps "/Users/cheoljoo/Code/github/doxygen/build/bin/B/OUTPUT/latex/inline_umlgraph_epslatex.pu" `
```

## Modify One Plantuml
- Cached file name : OUTPUT_DIRECTORY/inline_umlgraph_cache_all.pu
    - inline_umlgraph_cache_all.pu was created when doxygen was finished.
    - inline_umlgraph_cache_all.pu had all plantuml contents. and it will be used to check what is changed. 
- If you want to keep pu files , DOT_CLEANUP = NO
    - inline_umlgraph_<type><dictory>.pu includes changed plantuml. So it can reduce the java processing time.

# Explanation of the code
## old
- generatePlantUMLOutput() makes each plantuml from each @start~enduml with the name as inline_umlgraph_#.pu.

## new (PLANTUML_RUN_FAST)
- It will make a plantuml png at the end of processes of doxygen. So we can reduce the count of java runs.
- class PlantumlManager saves all plantuml information for running java at the end of doxygen and for comparing the plantuml contents as a cache.
    - I use writePlantUMLSource() to get more information (pu name and contents).  I modified writePlantUMLSource() because I do not want to read the file to get the contents in generatePlantUMLOutput().
    I add format argument in  writePlantUMLSource().
    -  The writePlantUMLSource() compares each plantuml contents and filename with cached information (inline_umlgraph_cache_all.pu). Then it will decide whether will  recreate or reuse. 
- Like dot , we make plantuml pictures at the end of doxygen.


mosra:
A friendly suggestion :wink: : you don't have to close pull requests and open new ones with updates, it's enough to just `git push` (or `git push -f`) to the original one. That will keep all the comment history and context for easier reviewing and you won't need to repeat everything every time.

Besides that (excuse me if I missed something), but why have an option for `PLANT_RUN_FAST`, why not make that the default? I don't see an use case where the user would want to explicitly run it slow. Having two code paths also means there is more code to test and more code that can have bugs. If I remember correctly, this optimization is done for LaTeX formulas as well, they're all being processed at once at the end, and there's no option for `LATEX_RUN_FAST` either.

albert-github:
@cheoljoo  When I don't have any plantuml diagrams in my code I still get (in this case) empty `inline_umlgraph_cache_all.pu`. I don't think this file should be created in this case or at least removed at the end of the run.

mosra:
Yeah, I'm getting that too, was about to open an issue :) Thanks for pointing that out!

albert-github:
Code of #6762 has been integrated in master on github.

Issue #6710, opened 2018-12-29
xml summary tag not identical to brief with < marker?


modersohn:
I'm documenting a C# file and want to generate documentation for members so that they *only* show up as brief descriptions in doxygen docs as well as show up in Visual Studio code hints.

In doxygen I get the desired result when using the additional "< marker":

`        public const int ErrorNoError = 0;   ///< No error, i.e. function was successful
`

But this won't show up in Visual Studio code hints.

If I'm using the summary tag instead:

```
        /// <summary> 
        /// No error, i.e. function call was successful
        /// </summary>
        public const int ErrorNoError = 0;

```

the Visual Studio code hints will work, but then the doxygen docs will contain a brief *and* a detailed description.

So it seems the summary tag is not fully identical to brief or at least to brief descriptions with a "< marker".

Is this as designed? Is there any other way I can get the desired behaviour?

albert-github:
This sounds a bit like a `JAVADOC_AUTOBRIEF` setting but is hard to tell from the description.
Can you please attach a, small, self contained example (source+config file in a tar or zip) that allows us to reproduce the problem?
Please also specify the doxygen version used.

modersohn:
Here's a testcase:

[TestCase6710.zip](https://github.com/doxygen/doxygen/files/2716494/TestCase6710.zip)

The doxyfile is freshly generated with "-g" without any modification.

The HTML output for MyConstant has no detail section, the one for MyOtherConstant unfortunately has.

albert-github:
I've just pushed a proposed patch, pull request #6726

modersohn:
Any chance this could still be tackled? I've switched now to using the summary tag, but this seriously bloats the resulting documentation because of the large number of items that only have summary.

albert-github:
In the issue there are 2 problems mentioned.

The `///<` is by design, as it is documentation after the entity.

The brief / detailed description problem.
In the 1.9.1 version of doxygen I still see that in the `MyNameSpaceDoesnt.MyClassWhichDoesnt` Class there is still a brief and detailed section for the attibute `MyOtherConstant`. In the current master (1.9.2 (9271e09d9aacba067e6a95340c9fbedbff9a2085)) there is only the brief description.

The proposed pull request #6726 is not integrated, so I will withdraw it.

I think the problem is fixed but not released.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.2.
[...]

PR #6726, opened 2019-01-02
issue #6710 xml summary tag not identical to brief with < marker

albert-github:
Main problem is that the `</summary>` tag lands in the detailed description as the brief description could not be appended (commentscan.l).
Furthermore a detailed description with just a space was not seen as empty (correct for isEmpty function), string has been stripped for test.
Added test case.

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -1002,7 +1002,7 @@ void ClassDef::writeDetailedDocumentationBody(OutputList &ol)
     ol.generateDoc(briefFile(),briefLine(),this,0,briefDescription(),FALSE,FALSE);
   }
   if (!briefDescription().isEmpty() && repeatBrief &&
-      !documentation().isEmpty())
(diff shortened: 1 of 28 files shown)


doxygen:
I don't like`stripWhiteSpace()` calls being added everywhere. I'd prefer to fix it at the place where the data is stored into the variable, which should be much more limited and thus easier to maintain.

albert-github:
I think this would indeed be a better solution though this might result in that when adding something to the detailed description it has to be tested whether or not string already contains something and in that case a space should be added. Problem with the later solution might be that things that need to be together might get separated.

albert-github:
Withdrawn

In the 1.9.1 version of doxygen I still see that in the `MyNameSpaceDoesnt.MyClassWhichDoesnt` Class there is still a brief and detailed section for the attibute `MyOtherConstant`. In the current master (1.9.2 (9271e09d9aacba067e6a95340c9fbedbff9a2085)) there is only the brief description.

Issue #4504, opened 2018-07-02
Private members are documented in XML when EXTRACT_PRIVATE = NO (Origin: bugzilla #666100)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.7.6.1* on platform *Other*
Assigned to: Dimitri van Heesch

On 2011-12-13 16:47:36 +0000, Paul A. Bristow wrote:

    when EXTRACT_PRIVATE = NO
    
    then private members are correctly not documented in html
    
    BUT are still documented in XML
    
    For example
    class test
    {
    private:
        /*!
          This is a dummy private function.
          \return zero always.
          \warning This should not be documented if the Doxygen option is
    selected.
        */
        int private_function(int z)
        {
          return 0;
        }
    }
    
    produces XML containing this:
    
    -<sectiondef kind="private-func"> -<memberdef prot="private" kind="function"
    id="classtest_1a9d5eb4b1d44630a644a2a85b9ec83c78" virt="non-virtual"
    inline="yes" explicit="no" const="no" static="no"> <type>int</type>
    <definition>int test::private_function</definition> <argsstring>(int
    z)</argsstring> <name>private_function</name> -<param> <type>int</type>
    <declname>z</declname> </param> <briefdescription> </briefdescription>
    -<detaileddescription> -<para>This is a dummy private function. -<simplesect
    kind="return"><para>zero always. </para></simplesect> -<simplesect
    kind="warning"><para>This should not be documented if the Doxygen option is
    selected. </para></simplesect> </para> </detaileddescription>
    <inbodydescription> </inbodydescription> <location line="35" file="I:/boost-
    sandbox/quickbook/boost/quickbook/quickbook_demo.hpp" bodyend="37"
    bodystart="34" bodyfile="I:/boost-
    sandbox/quickbook/boost/quickbook/quickbook_demo.hpp"/> </memberdef>
    </sectiondef>
    
    Is this a feature?
    
    Is there some other option that I need to select to prevent private
    documentation appearing the XML output?
    
    
    I have tried to supress using \internal thus:
    
    //! \internal
    private:
        /*!
          This is a dummy private function.
          \return zero always.
          \warning This should not be documented if the Doxygen option is
    selected.
        */
        int private_function(int z)
        {
          return 0;
        }
    //! \endinternal
    
    but the XML still contains the private information.
    
    
    (This is unwanted behaviour because the XML is post-processed).

jwakely:
Duplicate of #2405 ?

Issue #2405, opened 2018-07-01
XML output and EXTRACT_PRIVATE = NO ouputs private functions (Origin: bugzilla #413236)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.5.1* on platform *Other*
Assigned to: Dimitri van Heesch

On 2007-02-28 22:04:47 +0000, Ion GaztaÃ±aga wrote:

    I've been using doxygen with the Boost C++ documentation framework to
    generate reference files for Boost libraries. Although I specify the
    EXTRACT_PRIVATE = NO option, private functions were displayed in the
    documentation. I've written an small example of a simple class, using
    Doxywizard, requesting both XML and HTML outputs. In the HTML outputs,
    private functions are not displayed, but the XML ouput displays them.
    
    I'm attaching the Doxyfile and a very simple header file. I'm using Doxygen
    1.5.1/Windows XP. The same happens with older Doxygen versions.
    
    Thanks,
    
    Ion
    
    8<---8<---8<---8<---8<---8<---8<---8<---8<---8<---8<---8<---8<---8<---
    
    # Doxyfile 1.5.1-p1
    
    #---------------------------------------------------------------------------
    # Project related configuration options
    #---------------------------------------------------------------------------
    PROJECT_NAME           = 
    PROJECT_NUMBER         = 
    OUTPUT_DIRECTORY       = C:\doxygen
    CREATE_SUBDIRS         = NO
    OUTPUT_LANGUAGE        = English
    USE_WINDOWS_ENCODING   = YES
    BRIEF_MEMBER_DESC      = YES
    REPEAT_BRIEF           = YES
    ABBREVIATE_BRIEF       = "The $name class" \
                             "The $name widget" \
                             "The $name file" \
                             is \
                             provides \
                             specifies \
                             contains \
                             represents \
                             a \
                             an \
                             the
    ALWAYS_DETAILED_SEC    = NO
    INLINE_INHERITED_MEMB  = NO
    FULL_PATH_NAMES        = YES
    STRIP_FROM_PATH        = "C:/Archivos de programa/doxygen/bin/"
    STRIP_FROM_INC_PATH    = 
    SHORT_NAMES            = NO
    JAVADOC_AUTOBRIEF      = NO
    MULTILINE_CPP_IS_BRIEF = NO
    DETAILS_AT_TOP         = NO
    INHERIT_DOCS           = YES
    SEPARATE_MEMBER_PAGES  = NO
[...]

Issue #2187, opened 2018-07-01
cross-reference private methods without warning about them being undocumented (Origin: bugzilla #345974)


doxygen:
status NEW severity *enhancement* in component *general* for *---*
Reported in version *1.4.x* on platform *Other*
Assigned to: Dimitri van Heesch

On 2006-06-26 17:22:49 +0000, ergosys@gmail.com wrote:

> When SOURCE_BROWSER is enabled and EXTRACT_PRIVATE is not enabled, there is nothing mentioned in "referenced by" sections about private methods referencing public ones, as expected more or less.  
> 
> It would be useful however, if these members were still cross-referenced even though documentation is not generated for them.  In other words, if a private method references a public, documented one, the private method should show up in the documentation of the public method.  
> 
> Enabling EXTRACT_PRIVATE is not desirable in  most cases because private members are usually not important to users of a code base, clutters the documentation, and doxygen'ing them requires additional work.  Leaving the members undocumented is not a good option either because of the volumes of warnings which can be generated, making it difficult to find undocumented public members. 
> 
> An option to disable warnings for undocumented private methods might do the trick.

Issue #1910, opened 2018-07-01
Command or option to include private virtuals (Origin: bugzilla #321069)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.4.x* on platform *Other*
Assigned to: Dimitri van Heesch

On 2005-11-09 15:33:44 +0000, Marc Mutz wrote:

    Since private virtual functions are part of a class' interface for it's 
    subclasses, I'd like to have my private virtuals included in the
    documentation.
    However, atm, without saying EXCLUDE_PRIVATE = NO, I get _all_ private
    members
    - data as well as methods - and that's definitely not what I want. What
    makes
    this a bug IMO, is that private members that are documented with a doxygen 
    comment are nevertheless not included.

On 2005-11-13 15:49:07 +0000, Dimitri van Heesch wrote:

    Do I understand you correct if I summarize the problem as the lack of having
    the option: 
    EXCLUDE_PRIVATE = YES_UNLESS_DOCUMENTED
    ? If not please provide an example to clarify.

On 2006-02-07 07:50:28 +0000, Marc Mutz wrote:

    No, what I want is
      EXCLUDE_PRIVATE = YES_UNLESS_VIRTUAL_METHOD
    That's b/c private virtuals are reimplementable by subclasses in C++, and
    thus form part of the interface presented to subclasses. In this, they are
    different from private non-virtuals or private data members.
    
    YES_UNLESS_DOCUMENTED I can already get with the following:
    
    EXTRACT_ALL            = NO
    EXTRACT_PRIVATE        = YES
    HIDE_UNDOC_MEMBERS     = YES
    HIDE_UNDOC_CLASSES     = YES

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.16.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6729, opened 2019-01-02
Implement a new EXTRACT_PRIVATE_VIRTUAL option

mosra:
The classic article about [Virtuality from Herb Sutter](http://www.gotw.ca/publications/mill18.htm) suggests that virtual functions are always made private and public class interface is never virtual. Until now, it was not really possible to document these functions in Doxygen:

 * Enabling `EXTRACT_PRIVATE` would show all internals, not just virtual functions, which is not wanted.
 * Enabling `HIDE_UNDOC_MEMBERS` and `HIDE_UNDOC_CLASSES` would effectively disable warnings about *all* undocumented members, which is not wanted.

The usual workaround was to put the members into protected scope just for Doxygen:

```cpp
    #ifdef DOXYGEN_GENERATING_OUTPUT
    protected:
    #else
    private:
    #endif
        /** @brief Documented private virtual function */
        virtual doStuff();
```

The new `EXTRACT_PRIVATE_VIRTUAL` option makes these visible (and able to be linked to), but shows them *only* if they are documented. Real-world use of this feature can be seen for example in the [Magnum docs](https://doc.magnum.graphics/magnum/classMagnum_1_1Trade_1_1AbstractImporter.html#pri-methods):

image

I'm mainly using the XML output, for which private functions were already included (even though unconditionally, see https://github.com/doxygen/doxygen/issues/2405 and https://github.com/doxygen/doxygen/issues/4504), but until now it was not possible to `@ref` them. That's now fixed and a new test is added to verify that. I numbered it 81, assuming you'd want to merge this after my previous PR, #6722 (which has a test no. 80). Besides the XML output I verified that the private virtuals are shown also in the HTML output (using the same test file), but since I'm not primarily using it, it may happen that I missed some cases.

This feature was requested back in 2005 in https://github.com/doxygen/doxygen/issues/1910, I assume merging this would make that feature request fullfilled :) It's maybe also related to #2187.

Thanks in advance for merging!

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -620,6 +620,10 @@ void ClassDef::internalInsertMember(MemberDef *md,
     {
       addMemberToList(MemberListType_relatedMembers,md,FALSE);
     }
+    else if (md->isFunction() && md->protection()==Private && md->virtualness()!=Normal && Config_getBool(EXTRACT_PRIVATE_VIRTUAL))
+    {
+      addMemberToList(MemberListType_functionMembers,md,FALSE);
+    }
     else
     {
       switch (md->memberType())
--- a/src/config.xml
+++ b/src/config.xml
@@ -838,6 +838,14 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+    <option type='bool' id='EXTRACT_PRIVATE_VIRTUAL' defval='0'>
+      <docs>
+<![CDATA[
+ If the \c EXTRACT_PRIVATE_VIRTUAL tag is set to \c YES, documented private
+ virtual methods of a class will be included in the documentation.
+]]>
+      </docs>
+    </option>
     <option type='bool' id='EXTRACT_PACKAGE' defval='0'>
       <docs>
 <![CDATA[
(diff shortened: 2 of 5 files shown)


albert-github:
In the `config.xml` you use `*documented*`, his is the first place (to the best of my knowledge) of the usage of `*...*` in the `config.xml` file. A consequence is that it is shown in the documentation in italic. In the `Doxyfile` and in the help of the `Doxywizard` it is shown as `*...*` so I think we should abstain from it or otherwise adjust the script (configgen.py) so it can also handle the `*`.

mosra:
@albert-github Ah, thanks for pointing that out, Didn't know about this restriction, sorry. 

I wanted to emphasise the word, yes, because I think it's important. If I use `\em` instead, will it behave correctly also in Doxywizard? Or should I rather just not use it at all?

mosra:
Okay, removed it, then :)

doxygen:
Note that I shortened the option name to `EXTRACT_PRIV_VIRTUAL` to make `doxygen -s -g` produce a nicely aligned list of options with at least one space before the `=`.

Issue #6754, opened 2019-01-07
False positives for "multiple @param sections"


FFroehlich:
The 1.8.15 update of doxygen (https://github.com/doxygen/doxygen/commit/d39cdf826f15470def2ca36bc636868a62c42bbf specifically) has introduced a check for the identification of multiple `@param` sections, but this seems to be producing false positives when a `@param` and a `@retval` instructions both use the same variable name:

```
double fun(double a, double b) {
 }
 /** @file "fun.cpp"
   * @brief  function.
   */
 /** @fn double fun(double a, double b)
   * @brief  brief.
   *
   *
   * @param a   bar
   * @param b   foo
   *
   * @retval a      foobar
   */
```
will produce the following error message:
`.../fun.cpp:1: warning: argument 'a' from the argument list of fun has muliple @param documentation sections`

albert-github:
I've just pushed a proposed patch, pull request #6754

@doxygen
Maybe the entire call to `checkArgumentName` has to be omitted for `@retval` or the `@retval` values should be placed in a separate dictionary to overcome the fact that a `@retval` is documented twice.

albert-github:
Code has been integrated in master on github.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.16.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

igordesto:
I'm facing the same problem in version 1.8.17

```
/**-----------------------------------------------------------------------------
* \ingroup      MeasureStateMachine
* \brief        Measure state machine error handler
* \param[in]    eException Exception to be handled
* \param[in]    psMsrSttstc Pointer to the statistic structure
* \return       teException Updated exception.
*/
teException eMsrSttMchnErrHndlr(teException eException, tsMsrSttstc *psMsrSttstc)
```
will produce the following error message

`warning: argument 'eException' from the argument list of eMsrSttMchnErrHndlr has multiple @param documentation sections`

I performed tests changing the name of the variables and the result persists.

I appreciate any help.

albert-github:
Based on the given example I'm not able to reproduce it.

Most likely cause is that you have documented `eMsrSttMchnErrHndlr` but that the documentation is different at the both places.

[...]

PR #6757, opened 2019-01-08
issue #6754 False positives for "multiple @param sections"

albert-github:
Don't add the argument to list of parameters in case of a non-parameter call (i.e. retval  call)


albert-github:
@doxygen
Maybe the entire call to checkArgumentName has to be omitted for @retval or the @retval values should be placed in a separate dictionary to overcome the fact that a @retval is documented twice.

Another issue might be that `void` functions (in Fortran subroutines) cannot have `\retval` (or `\return`).

Issue #6781, opened 2019-01-18
Unable to use math in markdown table headers


AndyHeather:
version: recent master at commit 6a1b3708

adding math to markdown table headers breaks the table for html output - this worked previously in 1.8.14

```
TEST   {#TEST}
====

Correctly rendered table in html

| col1 | col2 |
|------|------|
|    a |    b |
|    c |    d |
|    e |    f |
|    g |    h |

Incorrectly rendered table in html

| \f$col_1\f$ | \f$col_1\f$ |
|-------------|-------------|
|           a |           b |
|           c |           d |
|           e |           f |
|           g |           h |
```

AndyHeather:
Added a test case

[TEST.tar.gz](https://github.com/doxygen/doxygen/files/2774045/TEST.tar.gz)

albert-github:
I've just pushed a proposed patch, pull request #6782

albert-github:
@AndyHeather unfortunately it does not work properly (in your case is is OK) see remarks with #6782

albert-github:
Previous proposed patch (#6782) has been withdrawn and has been replaced by the proposed patch #6784

AndyHeather:
Thanks

AndyHeather:
A bit more testing has shown some difference in behaviour with patch #6784:
- block equations could be indented and render correctly using master at 6a1b370 (but with problem of messing up tables using math in headers)
- using #6784 an indented block equation is rendered as a code/verbatim block, and only rendered correctly if left-aligned to the page (tho' problem of using math in headers resolved)

I.e. for the following
```
\f[
    a + b = c
\f]
```
vs
```
    \f[
        a + b = c
    \f]
```
Small test case: [TEST.tar.gz](https://github.com/doxygen/doxygen/files/2813102/TEST.tar.gz)

albert-github:
Actually I think that the old version (where in the indented bock the `\f[..` is converted as a formula) is wrong. As soon as markdown / doxygen discovers an indented block it converts it to a verbatim block, From the documentation (Chapter Markdown, paragraph code blocks):

> Preformatted verbatim blocks can be created by indenting each line in a block of text by at least 4 extra spaces

See also what already in the past happened when you just placed some indented text above the `\f[` formula.

AndyHeather:
Thanks - agreed

AndyHeather:
Hi - do you know if PR #6784 is going to be merged?

albert-github:
[...]

PR #6782, opened 2019-01-18
issue #6781 Unable to use math in markdown table headers for html output

albert-github:
There is no reason to ban the `\f$` command in markdown


albert-github:
Looks like that the building of the documentation fails, investigating ...

albert-github:
Withdraw and replaced bu proposed patch #6784

PR #6784, opened 2019-01-19
issue #6781 Unable to use math in markdown table headers

albert-github:
Due to the change of the place where the markdown processing is done the end of the line must be calculated a little bit differently.

Note: translator.py gave an error due to a strange indentation (did surface now), so had to be corrected as well.


AndyHeather:
Will this be merged into the main line?

Issue #6783, opened 2019-01-19
Doxygen does not copy Markdown images with relative paths


jkriege2:
Hi!

I have a project with several markdown files that link to image within the same repo with relative paths, e.g.:
`![jkqtplotter_simpletest1](../../screenshots/jkqtplotter_simpletest1.png)` (see https://github.com/jkriege2/JKQtPlotter/blob/master/examples/simpletest/README.md). 

For a better overview, here is the file structure:
```
|
+- doc   <-- I run doxygen from this directory
|   |
|   +- Doxyfile   <-- my Doxyfile (see https://github.com/jkriege2/JKQtPlotter/blob/master/doc/Doxyfile)
|   +- html   <-- here my output is written to
|   
+- screenshots   <-- here my images reside
|  |
|  +- jkqtplotter_simpletest1.png  <-- one of my images
|   
+- examples
   |
   +- simpletest
      |
      +- README.md <-- here I link to an image with ![](../../screenshots/jkqtplotter_simpletest1.png)

```

My Doxyfile contains `../screenshots` in the IMAGE_PATH.

Now here is my problem:
- If I use \image html jkqtplotter_simpletest1.png anywhere, all works fine and the image is copied to my HTML output directory.
- If I use ![](jkqtplotter_simpletest1.png)in my README.md above, all works fine and the image is copied to my HTML output directory.
- BUT: If I write `![](../../screenshots/jkqtplotter_simpletest1.png)` in my README.md above, the image is NOT COPIED to the output directory, but the path above is transformed into this HTML-Tag: `<img src="../../../../screenshots/jkqtplotter_simpletest1.png" alt="jkqtplotter_simpletest1" class="inline">`

When I run this through doxygen in my local windows machine, this transformed path works, because my HTML-output is in `/doc/html/...`, but if I only copy the HTML output to a webserver it fails: http://jkriege2.github.io/JKQtPlotter/d9/d6c/a02903.html

So could it be that doxygen does not correctly resolve these relative pathes? From my understanding, even if I use a relative path to an image that resides within the documentation directory structure (even within a path in IMAGE_PATH that should produce correct HTML output, i.e. also copy the file, or AT LEAST throw an error message ... correct?

For reference my doxygen output can be found here (if it helps):
https://travis-ci.org/jkriege2/JKQtPlotter/jobs/481784330

Best and hope you can help,
JAN

albert-github:
Thanks for the report.
[...]

PR #6787, opened 2019-01-20
issue #6783 Doxygen does not copy Markdown images with relative paths

albert-github:
In case of an image description in markdown format with a relative path it is transferred to just the filename so e.g. github and doxygen can use the same file (requirement will be that the, doxygen related, path to the image file is in the `IMAGE_PATH`).


src/markdown.cpp
```diff
@@ -893,7 +894,22 @@ static int processLink(GrowBuf &out,const char *data,int,int size)
...
-      out.addStr(link);
+      if (portable_isAbsolutePath(link) || isURL(link))
+      {
+        out.addStr(link);
+      }
+      else
+      {
+        // lets see if the file exists based on the path of the input file and the image file name (is a relative path / local filename)
+        QFileInfo fi(g_fileName);
+        QCString imgFile = g_fileName.left(g_fileName.length()-fi.fileName().length()) + link;
```
doxygen: The commit message says "requirement will be that the, doxygen related, path to the image file is in the IMAGE_PATH" but here I don't see image path being searched...

albert-github: Doesn't look like the `IMAGE_PATH` is involved. I have to do some extra tests.

albert-github: In case the `IMAGE_PATH` is not set the image is not copied to the HTML directory and thus not seen (unless the user copies it by "hand"). By setting the `IMAGE_PATH` the file is copied and is seen in the output. The copying is implicitly done outside the markdown code as the markdown image reference is converted to a `<img ...` tag.

Salmony:
I have verified this fix on master (commit 80dadfd5b57f1936b014ddc7f48be570a63d1d0b). What would it take to get it merged?

Salmony (approved):

PR #6793, opened 2019-01-23
Bug 757574 - Warning regarding subsection with anchor in markdown

albert-github:
In case `##` sections are used with and label marker and  the previous level had no label marker the previous level was rendered as `<h...>`, now they are rendered as doxygen sections and a label is generated. In case of a tag file the generated label is excluded as otherwise the same generated  label could exist in different tag files and give warnings.

### DIFF
--- a/src/definition.cpp
+++ b/src/definition.cpp
@@ -547,12 +547,15 @@ void Definition::writeDocAnchorsToTagFile(FTextStream &tagFile)
       {
         //printf("write an entry!\n");
         if (definitionType()==TypeMember) tagFile << "  ";
-        tagFile << "    <docanchor file=\"" << si->fileName << "\"";
-        if (!si->title.isEmpty())
-        {
-          tagFile << " title=\"" << convertToXML(si->title) << "\"";
+	if (si->label.left(strlen(GENERATED_NAME)) != GENERATED_NAME)
+	{
+          tagFile << "    <docanchor file=\"" << si->fileName << "\"";
+          if (!si->title.isEmpty())
+          {
+            tagFile << " title=\"" << convertToXML(si->title) << "\"";
+          }
+          tagFile << ">" << si->label << "</docanchor>" << endl;
         }
-        tagFile << ">" << si->label << "</docanchor>" << endl;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


doxygen:
I've committed a different solution, see https://github.com/doxygen/doxygen/commit/f40468618d8bd750c73c31592662fcf174a717a1. Let me know if you see problems with this.

albert-github:
When using the example:
```
# Header

## Header

### Header {#section_header}
```
I still get the warning:
```
aa.md:6: warning: found subsection command outside of section context!
```
with the current master.

doxygen:
The problem is due to the ambiguity between

     # Heading

     ## Heading


and 


     # Heading 1

     # Heading 2

In order to produce one page and still use the first ATX header as the page title, both of these should be translated to

    @page p1 Heading 1

    @section autotoc_md1 Heading 2


I've fixed this with this commit https://github.com/doxygen/doxygen/commit/8625ec5da0a505caf2dcfd16a5a8c1e668633ca8

albert-github:
Tested the alternative implementation and this looks OK for my test cases.

albert-github:
Another small problem related to the alternative implementation.
When we have a default `Doxyfile` aand the following source:
```
Page header
======


Documentation
=============


Getting Started
===============
```
we expect to get:
expected

but we get:
error

When looking comparing a number of versions:
- before alternative implementation May 19, 2019 10.53 (5fa5b3f4522231c30d7e67056f07cb5709815dfb ) => expected result
- first alternative implementation May 20, 2019 22.15 (f40468618d8bd750c73c31592662fcf174a717a1 ) => erroneous result
- second alternative implementation May 27. 2019 21.42 (8625ec5da0a505caf2dcfd16a5a8c1e668633ca8 ) => erroneous result
- before original pull request January 22, 2019 19:05 (5b1b37bc6b84da6d9641356c6a53741355150c08 ) => expected result
- this pull request January 23, 2019 16.53 (030a5929d8b677bd6a6c5a2d0a5c648c4e55a513) => expected result

doxygen:
The underlined headers needed the same treatment as the ATX headers. I've made [this commit](https://github.com/doxygen/doxygen/commit/a8b46600830be24ac189f798a1915f0ad1b86beb) to correct this.

albert-github:
Tests look OK

Issue #6794, opened 2019-01-23
C# code does not properly link to docs


johanngerell:
When generating docs from comments in C# source files, classes, namespaces, methods etc are all properly included in the doxygen index.

However, if I use \code, \snippet, \include etc, then only the first 2 lines have working code links.

For example, with this snippet in \code ... \endcode, the method calls and types on line 1 and 2 are properly linked to the generated docs, but line 3 have no links:
```cs
Acme.IService s = Acme.ServiceFactory.CreateService();
s.Foo();
s.Bar();
```
Now, if I swap lines 2 and 3, the same happens - only line 1 and 2 are linked to the docs.

The same thing happens if I use \snippet or  \include on a file with the same content as above.

albert-github:
Can you please attach a, small, self contained example (source+config file in a tar or zip) that allows us to reproduce the problem?
Please also specify the doxygen version used.

johanngerell:
Latest version:
```
> doxygen -v
1.8.15
```

I'll fix a small example

johanngerell:
Here's a minimal Acme example. I included the generated HTML produced with
> doxygen.exe Acme.doxygen
[Acme.zip](https://github.com/doxygen/doxygen/files/2789475/Acme.zip)

johanngerell:
![image](https://user-images.githubusercontent.com/247305/51642419-7d8e0f80-1f69-11e9-9a45-692fdaca3a8a.png)

albert-github:
I've just pushed a proposed patch, pull request #6799 

**NOTE**: I'm not sure if this patch is 100% correct. It works OK for this situation but there might be other situations where this fix backfires.

PR #6799, opened 2019-01-27
issue #6794 C# code does not properly link to docs

albert-github:
Only set a new  call scope in case typeClass exists.


albert-github:
NOTE: I'm not sure if this patch is 100% correct. It works OK for this situation but there might be other situations where this fix backfires.

Issue #6828, opened 2019-02-07
Physical newlines (^^) in ALIASES configuration tags not working with sections, subsections, subsubsections and paragraphs


rafaeltmc:
This alias
requirement{2}=\subsection sREQ\1 REQ#\1 \2 ^^ \ref REQ\1 \"Trace\" \n 

generates the following page:
![image](https://user-images.githubusercontent.com/3409197/52420349-b03f1880-2ad0-11e9-8c45-a2a470dcbd48.png)

when it should be
![image](https://user-images.githubusercontent.com/3409197/52420374-bfbe6180-2ad0-11e9-92a3-6aa797bd02c8.png)

There is a dirty solution to replace `^^` by `\\_linebr` 

Here is the example illustrated by the images [example.zip](https://github.com/doxygen/doxygen/files/2841391/example.zip)

The doxygen version being used is 1.8.15

See also: https://github.com/doxygen/doxygen/pull/602

albert-github:
I've just pushed a proposed patch, pull request #6829

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.16.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6829, opened 2019-02-07
issue #6828 Physical newlines (^^) in ALIASES configuration tags not working with sections, subsections, subsubsections and paragraphs

albert-github:
Previous changes corrected use of _linebr, missed one so this was not handled properly.


albert-github:
Works for section etc, but not for "normal" replacements.
Have to reconsider.

albert-github:
Works for normal case as well.

PR #6836, opened 2019-02-13
Adjustment of xhtml1-transitional.dtd

albert-github:
Adding attributes:
- doxygen
to be able to validate doxygen generated xhtml documents.

Analogous to #6681 but now for the area tag (found in the doxygen tag example).

### DIFF
--- a/src/context.cpp
+++ b/src/context.cpp
@@ -3854,7 +3854,7 @@ class TextGeneratorHtml : public TextGeneratorIntf
       if (ref)
       {
         m_ts << "<a class=\"elRef\" ";
-        m_ts << externalLinkTarget() << externalRef(m_relPath,ref,FALSE);
+        m_ts << externalLinkTarget();
       }
       else
       {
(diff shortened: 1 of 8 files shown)


doxygen:
The 'doxygen' attribute was add to help the obsolete tool to extract tag file information from doxygen generated output. I think it is better to remove this non-standard attribute and keep the dtd unmodified.

albert-github:
I do agree, does this also apply for the `a` tag? as here also the `doxygen` handling has been added?

I think the relevant code is in `util.cpp` in the `functionexternalRef`

doxygen:
> I do agree, does this also apply for the `a` tag? as here also the `doxygen` handling has been added?
> 
> I think the relevant code is in `util.cpp` in the `functionexternalRef`

Indeed, `externalRef()` looks like the only place where the doxygen attribute is still added.

albert-github:
Should I create a new pull request or will you change the 2 files util.cpp and xhtml1-transitional.dtd (removing the `doxygen=` part fro `a`

albert-github:
I've just pushed an update of the proposed patch.

There were some more places where the `doxygen=` part had an influence (on the output).

PR #6851, opened 2019-02-23
Layout of command description in documentation

albert-github:
There were some layout issues in the HTML and LaTeX versions of the doxygen documentation.
- `<hr>` to close to paragraph (due to enabling `<hr>` possibility in LaTeX).
- missing `<hr>` in the documentation

### DIFF
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -229,7 +229,8 @@ documentation:
 The following subsections provide a list of all commands that are recognized by
 doxygen. Unrecognized commands are treated as normal text.
 
-
+<br>
+<hr>
 \htmlonly</p><center><p>\endhtmlonly
 <h2>
 \htmlonly --- \endhtmlonly
@@ -237,6 +238,8 @@ Structural indicators
 \htmlonly --- \endhtmlonly
 </h2>
 \htmlonly</p></center><p>\endhtmlonly
+<hr>
+<br>
 
 \section cmdaddtogroup \\addtogroup <name> [(title)]
   \addindex \\addtogroup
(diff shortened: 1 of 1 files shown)


doxygen:
Why is the `<br>` needed in front of every `<hr>`. I would expect a <hr> will automatically stop and (if needed) restart the paragraph. Is that something not done in LaTeX and is this a workaround for it?

albert-github:
The idea behind `<hr>`, in my opinion, is that it can be used directly under the text and that the spacing is left to the user. When implementing the `<hr>` for LaTeX I had the mentioned reason in mind. 
In the implementation of the of the `<hr>`  in HTML output a `<p>` is added.

See also the small attached example: [example_6851.zip](https://github.com/doxygen/doxygen/files/2897796/example_6851.zip)

doxygen:
I think the DoxyHorRuler should be defined as something like this:

```tex
\newcommand{\DoxyHorRuler}{%
  \vspace{1em}%
  \hrule%
}
```

to support visually similar spacing as done in the HTML output, demonstrated by the modified example:

```c
/** \file
  * begin
  * <hr>
  * some text
  * <hr>
  * more
  */
```

albert-github:
I did some tests with different implementations:

The current implementation in doxygen:
current

The implementation from this pull request:
my_suggestion

The implementation as suggested by @doxygen:
doxygen_suggestion

Looking at the current implementation we see:
- the `<hr>` above paragraph 24.23 is very close to the end of the previous paragraph
- the `<hr>` between paragraph 24.24 and paragraph 24.25 is very close to the end of paragraph 24.24 and the beginning of paragraph 24.25

Looking at the suggestion from @doxygen:
- the gap between the different paragraphs is quite different especially when we look at the gap between the paragraphs 24.23 and 24.24 and the gap between the paragraphs 24.24 and 24.25

Looking at the implementation of this pull request:
- gap between the different paragraphs is not completely equal to disturbing (in my opinion)

Issue #6856, opened 2019-02-24
Doxygen only expands macro defined in header file once when referred multiple times


Jerry-Ma:
With

    ENABLE_PREPROCESSING   = YES
    MACRO_EXPANSION        = YES

and the following code snippet:

    // macro.h
    #define meta_enum_class(Type, UnderlyingType, ...)\
        enum class Type : UnderlyingType { __VA_ARGS__}

    // use1.h
    #include "macro.h"
    /** @enum Use1 */
    meta_enum_class(Use1, int, a, b, c);

    // use2.h
    #include "macro.h"
    /** @enum Use2 */
    meta_enum_class(Use2, int, d, e, f);

Doxygen could not correctly expand both of the macro usage, which results in failure of showing documentation for `Use2`, or `Use1`, depending on which file is processed first.

Say, if use1.h is processed first, which doxygen correctly expand the macro there and produces documentation for enum `Use1`, then at the time it processes  use2.h, the macro.h header is skipped so the `meta_enum_class` macro is not expanded, which is evident upon examining the output with `-d preprocess`:

    Preprocessing /path/to/use2.h...
    #include macro.h: already included! skipping...

Any suggestions on how to do this?

albert-github:
I tried it with version 1.8.15 and didn't see a problem (maybe I looked at the wrong results).
I added to the files:
- a `\file` statement to be sure the files are included
- some documentation o the Use` and Use2, now they only have a placeholder.

Which version of doxygen are you using?
Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent. In this case especially the Doxyfile is of importance.

Jerry-Ma:
I am using Doxygen 1.8.14, installed via home brew on macOS 10.14. Here I am attaching the code the demonstrate the problem.

Please note the

    INPUT = @CMAKE_CURRENT_SOURCE_DIR@/src/use1.h \
            @CMAKE_CURRENT_SOURCE_DIR@/src \
            @CMAKE_CURRENT_SOURCE_DIR@/docs

 in Doxyfile.in, this causes `Use1` in `use1.h` correctly documented while `Use2` in `use2.h` not. Removing the first item will make both `Use1` and `Use2` un-documented, because the ordering of files is `main.cpp`, `use1.h`, then `use2.h`, where the macro expansion happens only during processing `main.cpp`.

[...]

PR #6859, opened 2019-02-25
issue #6856 Doxygen only expands macro defined in header file once when referred multiple times

albert-github:
In case the include file is already is already in the global stack it should not be added to the global stack, but still be handled for the current file


albert-github:
This patch has some negative effects (hanging processes), see #6897.

albert-github:
Patch had to be reverted by #6903 due to #6897.

PR #6875, opened 2019-03-07
disambiguate duplicate id

fchapoton:
as suggested by lgtm 

https://lgtm.com/projects/g/doxygen/doxygen/


albert-github:
I'm not sure about this fix.
As a regular `html` file it would be correct that the `id`s have to be different, but this is not a regular html file but a template and in the template we see quite clear 2 branches:
```
<!--BEGIN PROJECT_NAME-->
```
and
```
<!--BEGIN !PROJECT_NAME-->
```

and the result will be that maximal 1 of these 2 `id`s will be in the end result.

fchapoton:
indeed. My bad. Please close and sorry for the noise.

albert-github:
I didn't know about lgtm, code reviews are never bad and maybe it does signal some other errors.

Issue #6880, opened 2019-03-10
mscgen links placed in the wrong place when using SVG output


RedDwarf69:
This

```
/*! \mainpage
 * \msc
 *   arcgradient=16;
 *
 *   a [label="ElementA"],b [label="ElementB"], c [label="ElementC"];
 *   ...;
 *   --- [label="Lala"];
 *   |||;
 *   a->b [label="Message1", URL="@ref refa"];
 *   b=>c [label="Message2"];
 *   b<<c [label="Message3"];
 *   a<-b [label="Message4", URL="@ref refa"];
 *   ...;
 *   a<-b [label="Message5", URL="@ref refa"];
 *   a->b [label="Message6", URL="@ref refa"];
 *   a<-b [label="Message7", URL="@ref refa"];
 *   a->b [label="Message8", URL="@ref refa"];
 *   a->b [label="Message9", URL="@ref refa"];
 *   ...;
 *   a->b [label="Message10", URL="@ref refa"];
 * \endmsc
 *
 * # ReferenceA @anchor refa #
 */
```

Works fine with a default doxygen 1.8.15 generated Doxyfile (DOT_IMAGE_FORMAT=png). But when doing `( cat Doxyfile; echo DOT_IMAGE_FORMAT=svg ) | doxygen -` the hyperlinks on the svg image are not in the correct place.

Doxygen 1.8.15
graphviz 2.40.1
mscgen 0.20

albert-github:
I had a look into the problem and with 'svg' the 'active' areas are not on the place as expected.
I extended the problem a little bit to get a better overview (unique references for each message and I also added references on the places where no references were present yet).
When looking at the map file generated this file is identical for 'png' and 'svg' type images
Based on this I manipulated the resulting 'svg' slightly so it shows the areas that should be the active areas: color red and the real active areas: color green. 
![result](https://user-images.githubusercontent.com/5223533/54143152-878ca480-4429-11e9-85c3-1a42abe6e420.jpg)

when looking at the png image (no coloring possible for me):
![result_png](https://user-images.githubusercontent.com/5223533/54143832-eef72400-442a-11e9-8466-52d88221b4c0.jpg)

we see that the green areas correspond with the text areas in the png image.
When looking in the mscgen 0.20 code we see that for the determination of the map there is no possibility to distinguish between the image types (the output type is in both case as follows `outType  = ADRAW_FMT_PNG`).

Possible solution:
- find the relation between the PNG and SVG coordinates
[...]

PR #6885, opened 2019-03-15
issue #6880 mscgen links placed in the wrong place when using SVG output

albert-github:
The mscgen implementation can only generate a correct map for png images (see #6880) and this results in incorrect links.
As the development of mscgen is not active this implementation sees to it that the map file in case of a svg image and also the svg image itself) is directly generated by doxygen. The handling on the other formats (png and eps, the later for LaTeX is left to mscgen itself so no extra libraries have to be installed and loaded i.e. gd library).

 Making mscgen suitable as library for doxygen and create the possibility for a good map in svg mode

Changes:
- adding cmake support
- adding support support for creating the ismap in case of svg (the issue of this request) by means of a new type '-T svgismap'
- generating lexer part with flex (bison part didn't work as newer bison versions cannot handle the yyrparse_result leaving the mscgen_language.y as reference)
- lexer now in non-interative mode
- mscgen_main.c see to it that the mscgen_main can be called multiple times (i.e. reset global static data to their initial value on each call)
- renaming include file references

Adding original mscgen sources (version 0.20) plus the config.h as generated on Cygwin (also some changes to CMakefiles and msc.cpp / msc.h slipped in).

renaming files

Making mscgen suitable as library for doxygen and create the possibility for a good map in svg mode

Changes:
- adding cmake support
- adding support support for creating the ismap in case of svg (the issue of this request) by means of a new type '-T svgismap'
- generating lexer part with flex (bison part didn't work as newer bison versions cannot handle the yyrparse_result leaving the mscgen_language.y as reference)
- lexer now in non-interative mode
- mscgen_main.c see to it that the mscgen_main can be called multiple times (i.e. reset global static data to their initial value on each call)
- renaming include file references
- casting values to overcome warnings (Windows)


doxygen:
I don't really like the inclusion of the mscgen source code (especially if it is just for one output format). 

Maybe an idea to look at https://github.com/mscgenjs/mscgenjs-core which is a javascript version of mscgen and use that to render the graphs for the HTML output. Just an idea didn't look at the implications too much yet.

albert-github:
I agree that the inclusion of the mscgen code is not nice but, as written, the problem is that there is no development on mscgen side.

The option for the  https://github.com/mscgenjs/mscgenjs-core might look nice, but here, probably, only HTML is really supported and the interaction with doxygen is unclear (to me). Also SVG could be used in other output formats, I don't know how this would work out.

There are a number of solutions:
1. don't do anything, so users cannot really use the SVG format, not a nice choice 
2.  start a new project / fork for maintaining the mscgen code. Not something I like to do and apparently nobody else wants to do either as I don't seen any actively maintained clones on e.g. github. And here is also the problems of the extra libraries required and creating distributions.
3. completely integrate mscgen into doxygen, including extra libraries. Would complicate the dependencies of doxygen.
4. adding the mscgen code as an addon to doxygen, would require to be able to build the complete mscgen application with the different required libraries. Would complicate the dependencies of doxygen.
5. using a mixture, i.e. doing the SVG work by doxygen and leave the rest to the original mscgen. SVG doesn't need any extra libraries so the footprint is minimal.

Seen the possibilities my choice was for option 5) the "mixture" possibility as this had the least impact.

doxygen:
I have pushed an [alternative implementation](https://github.com/doxygen/doxygen/commit/6fb61b2c6561149c1a318df26fd1acac619b3519) that links in mscgen as a standalone library and makes it produce any output format.

PR #6886, opened 2019-03-16
at sign (`@`) not handled correctly in preprocessor

albert-github:
When having the problem (reduced to show problem):
```
#if ('@' == 0x40)
#endif
```
we get the warning:
```
warning: preprocessing issue while doing constant expression evaluation: syntax error
```

The `@` was eaten although only `@@` and `@E` should be handled.

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -1178,15 +1178,21 @@ QCString removeIdsAndMarkers(const char *s)
     {
       if (c=='@') // replace @@ with @ and remove @E
       {
-	if (*(p+1)=='@')
-	{
-	  result+=c; 
-	}
-	else if (*(p+1)=='E')
-	{
-	  // skip
-	}
-	p+=2;
+        if (*(p+1)=='@')
+        {
+          result+=c;
+          p+=2;
+        }
+        else if (*(p+1)=='E')
+        {
+          // skip
+          p+=2;
+        }
+        else
+        {
+          result+=c;
+          p++;
+        }
       }
       else if (isdigit(c)) // number
       {


albert-github:
@doxygen What is actually the meaning of `@@` and `@E` in the preprocessor?

doxygen:
> @doxygen What is actually the meaning of `@@` and `@E` in the preprocessor?

- During macro expansion the places where arguments appear are replaced by `@<num>`. 
- `@E` represents a place where an argument is expected but nothing was passed when invoking the macro. 
- When a `@` appears in the macro, it is first escaped as `@@` and then later after expansion `@@` is replaced by `@` again and other markers are removed.
- Since macro are not expanded recursively `@-` is placed before a macro name to prevent recursion.

So if an `@` is eaten probably means it was not replaced by`@@` before passing it to the macro expansion routine.

albert-github:
So the `@@` and `@E` are internal doxygen codes.
In your comment you write about 'marcos' but in this case we  have to do with `#if` type constructs, so the problem might be in the fact that in 'macros' the `@` are replaced but not in the `#if` constructs and that the back-substitution is done in both cases.

doxygen:
> So the `@@` and `@E` are internal doxygen codes.
> In your comment you write about 'marcos' but in this case we have to do with `#if` type constructs, so the problem might be in the fact that in 'macros' the `@` are replaced but not in the `#if` constructs and that the back-substitution is done in both cases.

Indeed, the substitution needs to be done for any macro expression (so also for the guard of `#if` statements), so I think this would be the proper way to fix it (but maybe more is needed for the general case).

```diff
diff --git a/src/pre.l b/src/pre.l
index 1e45ca35..9b31161c 100644
--- a/src/pre.l
+++ b/src/pre.l
@@ -2185,6 +2185,7 @@ CHARLIT   (("'"\\[0-7]{1,3}"'")|("'"\\."'")|("'"[^'\\\n]{1,4}"'"))
                                          BEGIN(DefinedExpr1);
                                        }
 <Guard>{ID}                            { g_guardExpr+=yytext; }
+<Guard>"@"                              { g_guardExpr+="@@"; }
 <Guard>.                               { g_guardExpr+=*yytext; }
 <Guard>\n                              {
                                          unput(*yytext);
```

albert-github:
It looks to me that the @doxygen fix is better than mine (had a look in the code and its consequences). 

Not directly related to this but probably related to the handling of the `#if` / `#define` is the following example:
```
/** \file */

#if ('A' == 0x41)
/** the brief of fie A */
void fie_A(void);
#else
/** the brief of not fie A */
void not_fie_A(void);
#endif

#if ('@' == 0x40)
/** the brief of fie at */
void fie_at(void);
#else
/** the brief of not fie at */
void not_fie_at(void);
#endif
```

Here one would expect to see both the functions from the `#if` part (as the condition is in both cases true), so `fie_A` and `fie_at`, but we see `not_fie_A` and `fie_at`. Probably due to the different handling of `@` and `A` (the later is translated (?) to '0L').
Not sure if I should leave this question here or create an issue for it?

doxygen:
I've just committed https://github.com/doxygen/doxygen/commit/9fdeef7b5a1018872c7f07b8d03e374a3da840a0 which addresses both the escaping issue and the replacement of certain character literals by 0L.

albert-github:
Found a case (which was in my small set of tests for this issue, I think that even in the `''` the `\` should be escaped) that gives on windows:

windows_assert


```
/** \file */

#if ('\\' == 0x5c)
/** the brief of fie */
void fie_ee(void);
#else
/** the brief of not fie */
void not_fie_ee(void);
#endif
```

doxygen:
> Found a case (which was in my small set of tests for this issue, I think that even in the `''` the `\` should be escaped) that gives on windows:

Do you have a stack trace for this? I don't see this when I run the example on MacOSX.

albert-github:
Unfortunately not. It only happens on native Windows (2013 compiler). It doesn't happen on Cygwin (gcc).
I'll see if I can find the place by means of print statements.

albert-github:
Did some debugging work. Problem is in the code of `pre.l` with original line numbers 1183-1196.
Placing some debug statements:
```
--- a/src/pre.l
+++ b/src/pre.l
@@ -1167,6 +1167,7 @@ static void expandExpression(QCString &expr,QCString *rest,int pos)
  */
 QCString removeIdsAndMarkers(const char *s)
 {
+  fprintf(stderr,"pre.l %d %s\n",__LINE__,s);
   //printf("removeIdsAndMarkers(%s)\n",s);
   const char *p=s;
   char c;
@@ -1176,8 +1177,10 @@ QCString removeIdsAndMarkers(const char *s)
   {
     while ((c=*p))
     {
+  fprintf(stderr,"pre.l %d '%c' %d\n",__LINE__,c,c);
       if (c=='@') // replace @@ with @ and remove @E
       {
+  fprintf(stderr,"pre.l %d\n",__LINE__);
        if (*(p+1)=='@')
        {
          result+=c;
@@ -1190,17 +1193,20 @@ QCString removeIdsAndMarkers(const char *s)
       }
       else if (isdigit(c)) // number
       {
+  fprintf(stderr,"pre.l %d\n",__LINE__);
        result+=c;
        p++;
         inNum=TRUE;
       }
       else if (c=='\'') // quoted character
       {
+  fprintf(stderr,"pre.l %d '%c' %d\n",__LINE__,c,c);
         result+=c;
         p++;
         char pc=c;
         while ((c=*p) && (c!='\'' || pc=='\\'))
         {
+  fprintf(stderr,"pre.l %d '%c' %d\n",__LINE__,c,c);
           result+=c;
           pc=c;
           p++;
@@ -1210,6 +1216,7 @@ QCString removeIdsAndMarkers(const char *s)
       }
       else if (c=='d' && !inNum) // identifier starting with a `d'
       {
+  fprintf(stderr,"pre.l %d\n",__LINE__);
        if (qstrncmp(p,"defined ",8)==0 || qstrncmp(p,"defined(",8)==0)
                   // defined keyword
        {
@@ -1224,6 +1231,7 @@ QCString removeIdsAndMarkers(const char *s)
       }
       else if ((isalpha(c) || c=='_') && !inNum) // replace identifier with 0L
       {
+  fprintf(stderr,"pre.l %d\n",__LINE__);
        result+="0L";
        p++;
        while ((c=*p) && isId(c)) p++;
@@ -1258,6 +1266,7 @@ QCString removeIdsAndMarkers(const char *s)
       }
       else if (c=='/') // skip C comments
       {
+  fprintf(stderr,"pre.l %d\n",__LINE__);
        char pc=c;
        c=*++p;
        if (c=='*') // start of C comment
@@ -1278,6 +1287,7 @@ QCString removeIdsAndMarkers(const char *s)
       else
       {
 nextChar:
+  fprintf(stderr,"pre.l %d\n",__LINE__);
        result+=c;
        char lc=tolower(c);
        if (!isId(lc) && lc!='.' /*&& lc!='-' && lc!='+'*/) inNum=FALSE;
```
results on windows in:
```
Preprocessing D:/doxygen_bugs/_patches/issues/issues_6/issue_6886_preprocessor/ee.h...
pre.l 1170  ('\\' == 0x5c)
pre.l 1180 ' ' 32
pre.l 1290
pre.l 1180 '(' 40
pre.l 1290
pre.l 1180 ''' 39
pre.l 1203 ''' 39
pre.l 1209 '\' 92
pre.l 1209 '\' 92
pre.l 1209 ''' 39
pre.l 1209 ' ' 32
pre.l 1209 '=' 61
pre.l 1209 '=' 61
pre.l 1209 ' ' 32
pre.l 1209 '0' 48
pre.l 1209 'x' 120
pre.l 1209 '5' 53
pre.l 1209 'c' 99
pre.l 1209 ')' 41
pre.l 1180 '²' -3
```
and on Cygwin in:
```
Preprocessing /cygdrive/d/doxygen_bugs/_patches/issues/issues_6/issue_6886_preprocessor/ee.h...
pre.l 1170  ('\\' == 0x5c)
pre.l 1180 ' ' 32
pre.l 1290
pre.l 1180 '(' 40
pre.l 1290
pre.l 1180 ''' 39
pre.l 1203 ''' 39
pre.l 1209 '\' 92
pre.l 1209 '\' 92
pre.l 1209 ''' 39
pre.l 1209 ' ' 32
pre.l 1209 '=' 61
pre.l 1209 '=' 61
pre.l 1209 ' ' 32
pre.l 1209 '0' 48
pre.l 1209 'x' 120
pre.l 1209 '5' 53
pre.l 1209 'c' 99
pre.l 1209 ')' 41
pre.l 1180 'H' 72
pre.l 1234
pre.l 1180 '&' 38
pre.l 1290
pre.l 1180 'a' 97
pre.l 1234
pre.l 1180 '' 25
pre.l 1290
Parsing file /cygdrive/d/doxygen_bugs/_patches/issues/issues_6/issue_6886_preprocessor/ee.h...
```
So also on Cygwin the results are (in the debug output) not correct but has no consequences as the "accidental" character (i.e. `H`) there still in a valid range.
Line 1209 is the line in the while of the `else if (c=='\'') // quoted character`

Looks to me that a special handling of a `\` followed by an next `\` is required.

doxygen:
Ok, I see what you mean. Curiously this was also wrong in the `removeMarkers()` function. I've made a new fix: https://github.com/doxygen/doxygen/commit/01497f22d4bbce81d787cfa5a2d7ce19357b2a7c

Please check if there are still unhandled corner cases.

albert-github:
It is hard to find all unhandled corners as we don't know how many corners there are ( :wink: ).

I did try something with:
```
#define X "\\\""
#define Y "\\\'"
```
as multi-byte parts are not allowed (according to the compiler) in `#if``.

Looks all good.

In:
```
if (term==0 || (term!='\'' && term!='"')) return inputStr; // not a valid literal
```
I think `term==0 || ` is not necessary as it will be covered by `term!='\'' && term!='"'`

albert-github:
Alternative approach was necessary.

Issue #6892, opened 2019-03-18
xml not well-formed (invalid token) for c++


junghans:
Part of https://github.com/ECP-copa/Cabana/pull/94, generating the xml output for [Cabana_AoSoA.hpp](https://github.com/ECP-copa/Cabana/blob/master/core/src/Cabana_AoSoA.hpp), the output xml is invalid:
```
xml/classCabana_1_1AoSoA.xml:25:33: not well-formed (invalid token)
```
The lines in question are:
```xml
      <param>
        <type>typename</type>
        <declname>std::enable_if<(is_member_types< DataTypes >::value &&Impl::IsVectorLengthValid< VectorLength >::value), int >::type</declname>
        <defname>std::enable_if<(is_member_types< DataTypes >::value &&Impl::IsVectorLengthValid< VectorLength >::value), int >::type</defname>
        <defval>0</defval>
      </param>
```
which contain `<` instead of `&lt;`, the source lines are: https://github.com/ECP-copa/Cabana/blob/master/core/src/Cabana_AoSoA.hpp#L62-L64

This is Doxygen version 1.8.14

albert-github:
Please test with the current version (1.8.15) as well.
I tried to reproduce the problem, with 1.8.14, but was not able to.

Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

albert-github:
The problem did intrigue me so I downloaded the entire repository and did a build with XML enabled and reproduced the problem in 1.8.14 and newer versions.

I've just pushed a proposed patch, pull request #6893

In future  please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.16.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6893, opened 2019-03-19
issue #6892 xml not well-formed (invalid token) for c++

albert-github:
the `declname` and `defname` should  also be converted (compare as well the routine `generateXMLForMember`)


junghans:
Fixes #6892 for me!

Issue #6884, opened 2019-03-14
[PATCH ] Remove misleading More... links


AntoniMarcinek:
Hello,

Attached 
[patch_test.tar.gz](https://github.com/doxygen/doxygen/files/2964040/patch_test.tar.gz)
you can find a test case showing what I want to amend with the patch. Unpack it, run doxygen in the patch_test directory, then open html/classMyClass.html with a browser. 

You will see that all method names are links to detailed description which largely contains automatically generated content. This is fine.

In some cases, there will also be some description written by a developer, to which the user should pay more attention. I expect that "More..." links appear only in such cases. Indeed NoDescription() and BothBriefAndDetailedDescription()  follow this expectation. However OnlyBriefDescription() does not - it suggests that there is something more than automatically generated content and repetition of brief, while there is nothing.

So I propose the following patch to src/memberdef.cpp:2185 based on commit 1ea163ff9386266ef6ed6ad68773c29ae814dca2
```
diff --git a/src/memberdef.cpp b/src/memberdef.cpp
index 995901af..93c03628 100644
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -2182,7 +2182,7 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
     {
       ol.startMemberDescription(anchor(),inheritId);
       ol.writeDoc(rootNode,getOuterScope()?getOuterScope():d,this);
-      if (detailsVisible)
+      if (detailsVisible && !documentation().isEmpty())
```

Regards,
Antoni

PS If you look into patch_test/MyClass.h you will see that for OnlyDetailedDescription() I would also like to see the More... link (this time constituting the whole brief description). This is probably debatable (contrary to above which I perceive as a bug), so I do not include my proposed solution here. Should I open a separate issue after the above issue is concluded?

albert-github:
Some thoughts.

The problem here is the definition of what is a detailed description. Doxygen holds here that everything except what counts to the brief description is detailed description.
Like you wrote:

> You will see that all method names are links to detailed description which largely contains automatically generated content. This is fine.
> In some cases, there will also be some description written by a developer, to which the user should pay more attention. I expect that "More..." links appear only in such cases.

there are also generated descriptions, though the user did request to have these descriptions(!) by either a configuration parameter (e.g. `REFERENCED_BY_RELATION`) or a explicit command (e.g. `\showrefby`).
So we have:
- brief description
- detailed description
  - user written description
  - "automatically" generated information

[...]

PR #6894, opened 2019-03-19
Problem with referenced / references relations

albert-github:
Main problem is that when the documentation is in the `.h` file, including e.g. `\showrefby` this is overwritten by the global setting from the `.cpp` file.
Strategy is that in case something is explicitly set this should be used (resulting in a function with both fields, i.e. from `.h` and `.cpp`) when merging the documentation  (i.e. `addMemberDocs` in doxygen.cpp).
On other places where a `root` and member are merged the `||` operation has also been changed to  the 2 agruent version.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -3650,10 +3650,10 @@ static void buildFunctionList(Entry *root)
 
                 md->addSectionsToDefinition(root->anchors);
 
-                md->enableCallGraph(md->hasCallGraph() || root->callGraph);
-                md->enableCallerGraph(md->hasCallerGraph() || root->callerGraph);
-                md->enableReferencedByRelation(md->hasReferencedByRelation() || root->referencedByRelation);
-                md->enableReferencesRelation(md->hasReferencesRelation() || root->referencesRelation);
+                md->enableCallGraph(md->hasCallGraph(), root->callGraph);
+                md->enableCallerGraph(md->hasCallerGraph(), root->callerGraph);
+                md->enableReferencedByRelation(md->hasReferencedByRelation(), root->referencedByRelation);
+                md->enableReferencesRelation(md->hasReferencesRelation(), root->referencesRelation);
 
                 // merge ingroup specifiers
                 if (md->getGroupDef()==0 && root->groups->getFirst()!=0)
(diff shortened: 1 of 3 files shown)


albert-github:
Note the issue is a bit hidden in the discussion with issue #6884

src/doxygen.cpp
```diff
@@ -3650,10 +3650,10 @@ static void buildFunctionList(Entry *root)
 
                 md->addSectionsToDefinition(root->anchors);
 
-                md->enableCallGraph(md->hasCallGraph() || root->callGraph);
-                md->enableCallerGraph(md->hasCallerGraph() || root->callerGraph);
-                md->enableReferencedByRelation(md->hasReferencedByRelation() || root->referencedByRelation);
-                md->enableReferencesRelation(md->hasReferencesRelation() || root->referencesRelation);
+                md->enableCallGraph(md->hasCallGraph(), root->callGraph);
```
doxygen: I don't understand why the first argument has to be passed if it already a member of the same object. This could be called in the implementation, or am I missing something?

src/memberdef.cpp
```diff
@@ -4346,9 +4350,37 @@ void MemberDefImpl::enableCallGraph(bool e)
   if (e) Doxygen::parseSourcesNeeded = TRUE;
 }
 
+void MemberDefImpl::enableCallGraph(bool e, bool e2)
+{
+  static bool callGraph = Config_getBool(CALL_GRAPH);
+  if (e != e2)
```
doxygen: I don't understand the logic here, `e2` is not used other than comparing to `e`, and if both are not equal both are ignored...

src/doxygen.cpp
albert-github: Partly, the routine `enable...` consists with 1 and 2 arguments, so to distinguish we need this argument and in the original implementation there was a `||` which is now simply replaced by a `,` for simplicity and flexibility.

src/memberdef.cpp
albert-github: In case the `e` and `e2` are identical both are the same as the configuration parameter or both have been changed (in the same way by either `\show...`or `\hide...`).
I case of a difference one of the two has been changed and doesn't have the value of the configuration parameter anymore, but we don't know which one, to get the right value we need the opposite of the configuration parameter.

albert-github:
Tests reveal that it looks like that the problem has been solved in the mean time (see https://github.com/doxygen/doxygen/issues/6895#issuecomment-2838064268) so PR is obsolete

Issue #6897, opened 2019-03-22
Current Doxygen master with endless-loop while preprocessing a special cpp file


jschleus:
I just found that the current Doxygen master code seems to go in an endless-loop while preprocessing the file `AnchoredJointNode.cpp` of the software archive `wdune_0.99pl1456.orig.tar.gz` downloaded from https://wdune.ourproject.org/
(at least that suggest the last output line before "hanging").

I used the native Doxyfile generated by "doxygen -g" and only disabled LATEX and set
```
RECURSIVE = YES
```

Remark 1: The current release Doxygen 1.8.15 doesn't have that problem and probably also the master code from 2019-03-17 13:22 (and before) has worked without that problem.

Remark 2: The current Doxygen master code runs well for some other processed software packages.

Remark 3: The mention file is visible for e.g. via
https://fossies.org/linux/privat/wdune_0.99pl1456.orig.tar.gz/wdune-0.99pl1456.orig/src/AnchoredJointNode.cpp

albert-github:
Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

albert-github:
Quick analysis gives that it looks like an endless loop in the used include files through ´pre.l´ this looks like to be introduced by pull request #6859 for the issue: #6856 Doxygen only expands macro defined in header file once when referred multiple times.

Some debugging has to be don in respect to `g_allIncludes`, `absName` and `alreadyIncluded`.

jschleus:
Sorry, I'm afraid I know too little about doxygen to produce the desired example. But I will try to  ensure that the archive `wdune_0.99pl1456.orig.tar.gz` is downloadable at least for the next month under the URL https://fossies.org/linux/privat/wdune_0.99pl1456.orig.tar.gz .

By the way: Although it may not related I want to mention that I see since the above given date (2019-03-17 13:22) for some - but not for all - doxygen-processed packages a drastical increase of processing time by the factor 5-20. If I found more reliable facts I will issue a separate issue.

jschleus:
And still a small supplement, no concrete analysis, just an observation (but eventually nevertheless helpful):

[...]

PR #6903, opened 2019-03-28
Revert "issue #6856 Doxygen only expands macro defined in header file once when referred multiple times"

doxygen:
Reverts doxygen/doxygen#6859


albert-github:
Revert necessary due to problems reported in #6897

PR #6905, opened 2019-03-30
Bug 685714 - false positives reporting parameters or return value not being documented

albert-github:
The "validating" routines should always be called.
They were missed out in case of:
- only XML format there should be warnings as well (XML has its own write routines)
- no formats there should be warnings as well

### DIFF
--- a/src/outputlist.cpp
+++ b/src/outputlist.cpp
@@ -143,12 +143,16 @@ bool OutputList::generateDoc(const char *fileName,int startLine,
   {
     if (og->isEnabled()) count++;
   }
-  if (count==0) return TRUE; // no output formats enabled.
 
+  // we want to validate irrespective of the number of output formats
+  // specified as:
+  // - when only XML format there should be warnings as well (XML has its own write routines)
+  // - no formats there should be warnings as well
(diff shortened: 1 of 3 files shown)


src/outputlist.cpp
```diff
@@ -181,10 +185,15 @@ bool OutputList::parseText(const QCString &textStr)
...
   }
-  if (count==0) return TRUE; // no output formats enabled.
 
+  // we want to validate irrespective of the number of output formats
+  // specified as:
+  // - when only XML format there should be warnings as well (XML has its own write routines)
+  // - no formats there should be warnings as well
   DocText *root = validatingParseText(textStr);
 
+  if (count==0) return TRUE; // no output formats enabled.
```
doxygen: in case `count==0` the root object is no longer deleted, resulting in a memory leak. I think it is better to guard the `writeDoc` code with `count>0` and only have one return statement at the end.

albert-github: I agree.
I've corrected the pull request accordingly.
Also the (boll) return value is nowhere used, so replaced the generateDoc and parseText functions with a void function.

Issue #6908, opened 2019-04-02
Page title not using heading when comment starts page in markdown


ashley-human:
In a markdown page, if there is anything before the heading in the markdown source (including "comments") then the generated HTML output uses the file name rather than the heading in the source text as the page title.

In the example below `test_page_01.md` shows the issue, while `test_page_02.md` builds to give the expected output.

Note this does not effect the homepage (`index.md`) in my example below.

An opening comment is useful for contact details, copyright etc.

I'm not entirely clear if comments in markdown are something that is 'really' supported, but as they do not render I would argue the behaviour is still not expected. As I think comments in markdown are a bit of a grey area I have also tried this with comments of the form:
   `[//]: # (Some comment)` (from [here](https://stackoverflow.com/questions/4823468/comments-in-markdown/20885980#20885980))

Tested with Doxygen version 1.8.15 (on Windows 10, if that is at all important).

Minimum viable example:
----------------------------
* `index.md`:
```markdown
<!-- Some comment) -->

Page title
==========

Text body where is this text

```

* `test_page_01.md`:
```markdown
<!-- Some comment) -->

Intended page title (Page 01)
=============================

The source markdown page has a comment on line one.

```
* `test_page_02.md`:
```markdown
Intended page title (Page 02)
=============================

The source markdown page has no comments before the title.

```

To make the HTML output easier to see this, I set my `Doxyfile` settings to:
* `USE_MDFILE_AS_MAINPAGE = index.md`
* `GENERATE_TREEVIEW = YES`
With all other settings left to default (as done following command `doxygen -g`).

albert-github:
My first thought was that the problem is related to the proposed pull request #6543, but this is not the case.
I see some handling of `<!--` in markdown but it seems that this is to late as `doxygen -d markdown` gives:
```
======== Markdown =========
---- input -------
@page md_test_page_01 test_page_01
<!-- Some comment) -->

Intended page title (Page 01)
=============================

The source markdown page has a comment on line one.

---- output -----
@page md_test_page_01 test_page_01
<!-- Some comment) -->

@section autotoc_md1 Intended page title (Page 01)


The source markdown page has a comment on line one.

=========
```
the mainpage probably just goes OK due to a bit different handling of the mainpage.

For future issues: Please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

albert-github:
[...]

PR #6909, opened 2019-04-03
issue #6908 Page title not using heading when comment starts page in markdown

albert-github:
Support the following comments types also on the beginning of a markdown file (i.e. as white space):
HTML comment:
- `<!-- ... -->`
Markdown comment (see also https://stackoverflow.com/questions/4823468/comments-in-markdown/20885980#20885980):
- `[comment]: <> (This is a comment)`
- `[comment]: # (This is a comment)`
- `[//]: <> (This is a comment)`
- `[//]: # (This is a comment)`

File 055_markdown.md has to be updated as now the initial `@page` was mot present and thus the `# Foo`, `## Bar` etc. were out of place

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -2355,18 +2355,157 @@ static QCString processBlocks(const QCString &s,int indent)
   return out.get();
 }
 
-/** returns TRUE if input string docs starts with \@page or \@mainpage command */
-static bool isExplicitPage(const QCString &docs)
+static int calculateComment(const QCString &docs)
 {
   int i=0;
   const char *data = docs.data();
+  int size=docs.size();
   if (data)
   {
     int size=docs.size();
-    while (i<size && (data[i]==' ' || data[i]=='\n'))
+    while (i<size)
     {
-      i++;
+      if (data[i] == ' ' || data[i] == '\n')
+      {
+        i++;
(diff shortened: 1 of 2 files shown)


doxygen:
I don't understand why adding support for comments requires changes to example 055? The example without the `@page` command should also be valid I would expect.

albert-github:
The fact that test example 055 has to be changes as well is due to the fact that till this change the comment stopped searching for a page title so a page title was "invented" being the name of the file. With this change the comment is ignored and the first header found is the `# Foo` and this is used as page name. To be consistent with the filenames an extra `@page ` with the original title is added we have a minimal change, alternativ would be some other changes (new files etc.) for test example 055.

albert-github:
See also question:  https://stackoverflow.com/questions/64052923/markdown-files-in-doxygen-two-puzzles

albert-github:
Regarding the `@~...` a similar problem exists (see https://github.com/doxygen/doxygen/issues/9437#issuecomment-1177142471) the `@~...` problem can be implemented in a similar way.

albert-github:
I think it might be good to have this PR integrated especially for things like copyright messages that should be excluded from the documentation by means of `<!-- ...-->` and are at the beginning of the file.

PR #6913, opened 2019-04-05
Markdown file as internet link

albert-github:
See a link in case of a md file only as local link if it doesn't start with a supported protocol.
Example found was:
```
[Ansibullbot](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)
```


Lecrapouille:
I tried your patch and it worked for me (I also have github links) ! Thanks !

PR #6915, opened 2019-04-07
Show git version information

albert-github:
Show the git version on designated places (currently version, help and Doxyfile difference).
The implementation is based on the git_watcher.cmake (https://github.com/andrew-hardin/cmake-git-version-tracking).
The information is useful to see which. master, version of doxygen is used for a build on systems where a lot of builds are make (Fossies.org) or very regular builds are made (CGAL)

Furthermore the tracking of the VERSION file was not done (when changed the version.cpp was not updated), this is corrected as well.


doxygen:
Some improvement suggestions:

Looks like there are now two version files that use different methods:

- Python to produce `version.cpp` from `version.py` and `VERSION`
- A Cmake rule to produce `gitversion.cpp` from `gitversion.cpp.in`

I don't like the `correctGitVersion()` that needs to be called. 

Why not create one `version.h` which offers both `QCString getVersion()` and `QCString getGitVersion()` functions and one `version.cpp` created from `version.cpp.in` by cmake. Then the "correction" can be hidden inside the .cpp file.

PR #6923, opened 2019-04-11
Regression #6836 URL contains $ sign on second run in map of images

albert-github:
When running doxygen twice on a repository the generation of e.g. 'included directory' images  (also seen for a.o. 'class' image) is not done again and the available map file is used, but the replacement of some terms is not done properly (the $ at the beginning is not removed).

This is a regression on #6836.

Attached is minimal example ( [regression_6836.zip](https://github.com/doxygen/doxygen/files/3068013/regression_6836.zip) ) to show the problem when running doxygen twice.

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -322,6 +322,11 @@ static QCString replaceRef(const QCString &buf,const QCString relPath,
       }
       else
       {
+        int marker = link.find('$');
+        if (marker == 0 ) // we have just $url, e.g. include graphs
+        {
+          link = link.mid(1);
+        }
         result = href+"=\"" + link + "\"";
       }
     }


doxygen:
If I look what has caused the regression, I think the [this fix](https://github.com/doxygen/doxygen/commit/c637dede9ec30c2e35f19636edc0b3fd424b45e5) is more appropriate, i.e. urlOnly should have been FALSE for generated map files. Do you agree?

albert-github:
I'll have to check and report back.

albert-github:
At first, when creating the pull request, I was thinking in the same direction, but I saw the comment:
``
// ref$url (external ref via tag file), or $url (local ref)
```
and the `ref$url (external ref via tag file)` gave me not a good feeling as I don't have a case so I can test it.
I now created a test case for it and I think this patch is better (my proposed patch left the reference, incorrect, to the external ref as `<tagfile>$<reference>` instead of `<reference>`.

albert-github:
As mentioned the problem has been solved in a better way.

PR #6924, opened 2019-04-11
Remove superfluous '\n' in warnings

albert-github:
A number of warning messages have a double '\n' due to the fact that the message itself contains a '\n' and the warning routine automatically adds a '\n'.
Only add the 'n' in the warnings routine in case it is not present.


albert-github:
With the integration of #10135 this PR is obsolete.

PR #6928, opened 2019-04-14
'QGDict::hashAsciiKey: Invalid null key' messages

albert-github:
In different open source packages the message:
`QGDict::hashAsciiKey: Invalid null key`
appeared.

- doxygen.cpp
  - in case no group name is given ignore the group (as if group was not found, previously gd stayed 0 as well).
- scanner.l
  -  usage of php protection keywords led to the fact that the next argument was not seen and this was the relevant redefinition
  - termination of a php script can also happen on other places (quoted strings are exempt from this as handled a little bit differently)
  - in case of object C spaces are possible before the : as well
- tclscanner.l
  - a 'namespace eval ...' was not the definition of a namespace but a usage and should not be recorded as namespace
- util.cpp
  - only replace a '.' to a '::' class / namespace separator in case it is not the complete word.

Attachment with used source files: [qgdict.zip](https://github.com/doxygen/doxygen/files/3077333/qgdict.zip)


albert-github:
With #7068 a number of fixes have been made (partly from this proposed pull request), tests with packages mentioned in #7068 were successful.

PR #6932, opened 2019-04-15
Reentrant scanners

groleo:
Convert constexp.l declinfo.l sqlcode.l to reentrant scanners.

### DIFF
--- a/src/constexp.h
+++ b/src/constexp.h
@@ -22,12 +22,21 @@
 #include "cppvalue.h"
 #include <qcstring.h>
 
-extern bool parseconstexp(const char *fileName,int line,const QCString &s);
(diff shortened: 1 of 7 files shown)


doxygen:
> Looks like the used version on Travis for osx of bison is a bit old, so either
> 
>     * code should conform this older version as well
> 
>     * bison version on osx should be upgraded in the .travis,yml script (@doxygen is this possible)

With [this commit](https://github.com/doxygen/doxygen/commit/ac76b9a7a70b3c828c76442f5977937fcb87811a) I added newer versions of bison and flex via homebrew. Seems to work.

albert-github:
As far as I can see the code has been integrated in the master branch.

Issue #6950, opened 2019-05-01
XML - <initializer> for array of structure only contains references in first array element


krausedj:
For C code, if you have a array of structures

```
struct S_TAG {
     func_ptr foo;
     uint8 data:
}:
typedef struct S_TAG S_TYPE;

S_TYPE data[3] = {
    { (func_ptr)foo, 0 },
    { (func_ptr)bar, 1 },
    { (func_ptr}dog, 2 }
};
```

There will only ever be references created for the first func_ptr, and foo, the other array elements will not output references.  In my test setup, the HTML contains links for foo, bar, and dog.  Regular arrays work fine.

Would the target debug entry point be the following?

```
    t << "        <initializer>";
    linkifyText(TextGeneratorXMLImpl(t),def,md->getBodyDef(),md,md->initializer());
    t << "</initializer>" << endl;
```

1.8.15

albert-github:
Although an example is given (and a pointer for possible debug starting).
To be sure we are talking about the same: please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

krausedj:
[test_initializer.zip](https://github.com/doxygen/doxygen/files/3134510/test_initializer.zip)

Yes, that is fair, sorry it was late last night.  I was able to replicate it in a very small test case, which is attached.

You can see d5/d04/group___local_group_1ga5126e75ea4e84ac62fb05b1d3ec641b5 which shows the issue.

albert-github:
I've just pushed a proposed patch, pull request #6951

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.16.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #6951, opened 2019-05-01
issue #6950 XML - <initializer> for array of structure only contains references in first array element

albert-github:
Don't stop at a hexadecimal number, but output the hexadecimal number (plus the part that has not been printed yet) and continue with the next input part.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -2031,10 +2031,16 @@ void linkifyText(const TextGeneratorIntf &out, const Definition *scope,
   int floatingIndex=0;
   if (strLen==0) return;
   // read a word from the text string
-  while ((newIndex=regExp.match(txtStr,index,&matchLen))!=-1 && 
-      (newIndex==0 || !(txtStr.at(newIndex-1)>='0' && txtStr.at(newIndex-1)<='9')) // avoid matching part of hex numbers
-      )
+  newIndex=regExp.match(txtStr,index,&matchLen);
+  while ((newIndex=regExp.match(txtStr,index,&matchLen))!=-1)
   {
+    if (!(newIndex == 0 || !(txtStr.at(newIndex-1)>='0' && txtStr.at(newIndex-1)<='9'))) // avoid matching part of hex numbers
+    {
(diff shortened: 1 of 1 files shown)


albert-github:
The given solution is a general solution applying also for other places.
For the specific problem the change in routine `static void generateXMLForMember(const MemberDef *md,FTextStream &ti,FTextStream...` of `xmlgen.cpp` from:
```
     t << "        <initializer>";
     linkifyText(TextGeneratorXMLImpl(t),def,md->getBodyDef(),md,md->initializer());
     t << "</initializer>" << endl;
```
into
```
   t << "        <initializer>";
   SrcLangExt lang = md->getLanguage();
   ParserInterface *pIntf = Doxygen::parserManager->getParser(md->getDefFileExtension());
   pIntf->resetCodeParserState();
   XMLCodeGenerator *xmlGen = new XMLCodeGenerator(t);
   pIntf->parseCode(*xmlGen,0,md->initializer(),lang,FALSE,0,const_cast<FileDef*>(md->getFileDef()),
                    -1,-1,TRUE,md,FALSE,md);
   xmlGen->finish();
   delete xmlGen;
   t << "</initializer>" << endl;
```
would probably more consistent with the other output generators (xml has its own output generator separated from, most, other generators), but would introduce extra tags in the `<initializer>` part like:
`<codeline>`, `<highlight class="normal">`, <sp/>i and thus being inconsistent with the current situation. Also output of the standard test would have to be slightly adjusted with these extra tags.

krausedj:
I pulled and built this, and it is working in my use case.  I think I understand the change as described.

I know I am only an outsider, but your secondary comment about using parseCode, does look more aligned with the other generators.

doxygen:
@albert-github  Removed the redundant (?) line before the while loop, and made some simplications. See https://github.com/doxygen/doxygen/commit/b8ed98e5b9e84f70bbe485094320201caf1f1ecb

albert-github:
@doxygen 
Statement looks indeed redundant.

PR #6953, opened 2019-05-02
Handles fencedCodeBlock with lang=plantuml as umldiagram

ArjanKn:
Markdown editor VSCODE allows inline definitions of PLANTUML diagrams using the plantuml extension:
````
# Markdown doc with plantuml diagram

Text with uml diagram 

```plantuml
@startuml
Bob->Alice : Hello
Bob<--Alice : Hi
@enduml
```

## topic

text

````

This patch makes doxygen behave the same: render the diagram, not the code.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -2172,7 +2172,15 @@ static QCString processQuotations(const QCString &s,int refIndent)
     {
       if (isFencedCodeBlock(data+pi,size-pi,refIndent,lang,blockStart,blockEnd,blockOffset))
       {
-        writeFencedCodeBlock(out,data+pi,lang,blockStart,blockEnd);
+        if ( lang == "plantuml" )
+        {
+          int cmdStart = pi+blockStart+1;
+          processSpecialCommand(out, data+cmdStart, cmdStart, size-cmdStart );
+        }
+        else
+        {
+          writeFencedCodeBlock(out,data+pi,lang,blockStart,blockEnd);
+        }
         i=pi+blockOffset;
         pi=-1;
         end=i+1;


albert-github:
You removed the essential example, why? Now it is not clear how this feature could work, besides this the triple back-quote is, in my opinion, special there to signal a code section, in case one wants to have an image one should use `\startuml` ...`\enduml`.

TheDallenger:
Is there anything I can do to help getting this pull request accepted?

albert-github:
No I don't think so.

johnmcfarlane:
@albert-github the purpose of triple back-quote is indeed to print the content as code. The following code should not generate a diagram:

    ```
    @startuml
    Bob->Alice : Hello
    Bob<--Alice : Hi
    @enduml
    ```

And if users wish to display the source code of a PlantUML diagram, they always have that way to do it.

However, triple backticks qualified with an [info string](https://spec.commonmark.org/0.29/#info-string) are intended to mean something specific. For example, \`\`\`c++ hints to the render that instead of displaying monospace text, it should display C++ code (which is also often displayed monospace but also, possibly, with syntax colouring). And we are talking here about *qualified* triple backticks, not just triple backticks.

From [the commonmark spec](https://spec.commonmark.org/0.29/#fenced-code-blocks):

> The content of a code fence is treated as literal text, not parsed as inlines. The first word of the info string is typically used to specify the language of the code sample, and rendered in the class attribute of the code tag. However, this spec does not mandate any particular treatment of the info string.

From [PHP Markdown extra](https://michelf.ca/projects/php-markdown/extra/#fenced-code-blocks):

> You can specify a class name that will apply to a code block. This is useful if you want to style differently code blocks depending on the language. Or you could also use it to tell a syntax highlighter what syntax to use.

It's quite possible to add a little licence to passages like these and go arbitrarily far beyond `<pre>` blocks. That much licence has been taken by other renderers ([example](https://docs.gitlab.com/ee/administration/integration/plantuml.html#creating-diagrams)), which means that Doxygen is an outlier and users are mostly just impeded by this decision without a particularly compelling argument towards compliance and a dubious argument towards conformance.

Alternatively, if you reject this contribution, can the PR not be closed? The project looks like it's in a state of disuse with all of the issues and PRs that are open. I've been [deterred](https://stackoverflow.com/questions/5163478/doxygen-report-potential-recursive-class-relation#comment100901182_7498072) from filing issues in the past under the assumption that they would be unlikely to be addressed, given the large number of unresolved issues/PRs.

ogrodnikm:
I came upon the same problem. It is easier to write documentation with vscode/intellij, since doxygen generation takes long time in a bigger projects, but doxygen is better at presenting ready documentation. I'd love to have both.

jharden-ctek:
I would also like to see this merged, mainly because I also work with Markdown documents with inline PlantUML and these documents are included as parts of Doxygen based documentation.

I have created a new commit in my own fork that applies to the latest revision on master, builds and works as expected (as far as I can tell): https://github.com/jharden-ctek/doxygen/commit/d3ae283af9d5d754e697b9d8eef3b576a0a81108

TheDallenger:
Maybe we should convince the vscode plantuml plug-in devs to support the way doxygen handles plantuml instead?
I don’t know what the “correct” is, but could very much like it to work :-)

jharden-ctek:
@TheDallenger: Maybe so, but does it really make sense to have a `@startuml ... @enduml` block without a fenced code block? These tags have no meaning in Markdown syntax, so my opinion is that having the plantuml tags within a fenced code block is actually the variant that makes the most sense.

johnmcfarlane:
Last I checked, Doxygen was the outlier for my team. GitLab's rendering was
also out of line with Doxygen.

zheylmun:
@albert-github, If this were to be updated with the documentation demonstrating the options of handling the explicitly tagged plantuml code fence as a diagram and the code fence without an info string rendering as code, would it be considered for merge?  As currently implemented, Doxygen support for plantuml in markdown is effectively incompatible with most markdown renderers, including GitHub.

AScholan:
@ArjanKn  and @albert-github, has this MR died a death? 

I have spent some time trying to work out what is wrong with the PlantUML inside a markdown file where the MD was rendering correctly in VsCode and failing to render in Doxygen.

It appears to me to be entirely reasonable to accept this MR as:

1. It does not impact on existing functionality. Code with @startuml/@enduml outside of plantuml quoted code blocks will still work.
2. Adding the functionality will make Doxygen behave like the markdown preview in VsCode.
3. As @johnmcfarlane points out, if there is a need to see the plantuml source code in the documentation, which I think is unlikely for most users, then a block quote without the plantuml language description can be used.

Thank you in advance for your response.
Cheers,
Andrew.

mwerner96:
I would love to see this merged, too! Treating plantuml in markdown this way would also be in line with how GitLab does it. It would greatly increase the flexibility of combined Doxygen+Markdown documentation!

Issue #6968, opened 2019-05-07
INLINE_INHERITED_MEMB doesn't respect EXTRACT_PRIVATE=NO


jwakely:
This looks related to #800 but that is apparently fixed, so I'm opening a new issue.

Given this source:
````
/// This is my foo class
struct Foo
{
  /// public typedef
  typedef int value_type;
  /// public function
  void m_impl() { }
};

/// This is my bar class
struct Bar
: private Foo
{
  /// public member function
  void baz() { }
};
````

And this config:
````
EXTRACT_PRIVATE = NO
INLINE_INHERITED_MEMB = YES
````

I get this output with Doxygen 1.18.4
````
<p>This is my bar class.
 <a href="structBar.html#details">More...</a></p>
<div class="dynheader">
Inheritance diagram for Bar:</div>
<div class="dyncontent">
 <div class="center">
  <img src="structBar.png" usemap="#Bar_map" alt=""/>
  <map id="Bar_map" name="Bar_map">
<area href="structFoo.html" title="This is my foo class. " alt="Foo" shape="rect" coords="0,0,34,24"/>
</map>
 </div></div>
<table class="memberdecls">
<tr class="heading"><td colspan="2"><h2 class="groupheader"><a name="pub-methods"></a>
Public Member Functions</h2></td></tr>
<tr class="memitem:aed0079f81c54fe079eca5b25478a8e87"><td class="memItemLeft" align="right" valign="top"><a id="aed0079f81c54fe079eca5b25478a8e87"></a>
void&#160;</td><td class="memItemRight" valign="bottom"><a class="el" href="structBar.html#aed0079f81c54fe079eca5b25478a8e87">baz</a> ()</td></tr>
<tr class="memdesc:aed0079f81c54fe079eca5b25478a8e87"><td class="mdescLeft">&#160;</td><td class="mdescRight">public member function <br /></td></tr>
<tr class="separator:aed0079f81c54fe079eca5b25478a8e87"><td class="memSeparator" colspan="2">&#160;</td></tr>
</table><table class="memberdecls">
<tr class="heading"><td colspan="2"><h2 class="groupheader"><a name="pri-types"></a>
Private Types</h2></td></tr>
<tr class="memitem:a0975ea1542b18a60760599de38453a50"><td class="memItemLeft" align="right" valign="top"><a id="a0975ea1542b18a60760599de38453a50"></a>
[...]

PR #6970, opened 2019-05-07
issue #6968 INLINE_INHERITED_MEMB doesn't respect EXTRACT_PRIVATE=NO

albert-github:
Check besides the protection of the method also the protection of the class (but don't forget to ignore it when EXTRACT_PRIVATE is set).

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -3646,7 +3646,8 @@ void ClassDefImpl::mergeMembers()
 
   m_impl->membersMerged=TRUE;
   //printf("  mergeMembers for %s\n",name().data());
-  bool inlineInheritedMembers = Config_getBool(INLINE_INHERITED_MEMB);
+  static bool inlineInheritedMembers = Config_getBool(INLINE_INHERITED_MEMB);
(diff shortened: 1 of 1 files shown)


src/classdef.cpp
```diff
@@ -3646,7 +3646,8 @@ void ClassDefImpl::mergeMembers()
 
   m_impl->membersMerged=TRUE;
   //printf("  mergeMembers for %s\n",name().data());
-  bool inlineInheritedMembers = Config_getBool(INLINE_INHERITED_MEMB);
+  static bool inlineInheritedMembers = Config_getBool(INLINE_INHERITED_MEMB);
+  static bool extractPrivate         = Config_getBool(EXTRACT_PRIVATE);
```
jwakely: Why not `static const bool`?

albert-github: All other places where a similar construct was used just used `static bool` as well, so I did it for consistency.
Feel free to create a pull request (maybe look at all `Config_get...` methods.

jwakely: I agree consistency is best here, so I'll do as you suggest and send a pull request to change them all.

Thanks for the speedy fix.

PR #6974, opened 2019-05-08
Fixes #6973 - Fixed segmentation fault in do_warn

### LINKED ISSUE #6973: Segmentation fault in do_warn caused by va_list reuse
This issue was introduced by commit [2d46a05](https://github.com/doxygen/doxygen/commit/2d46a052563c1d362bfe4fa3c8fac1a258ce4cc2); re-use of the `va_list` parameter `args` causes a segmentation fault during the second call to vsnprintf.

Edit: Proposed fix in #6974.

mrhatch97:
Copied va_list to avoid reuse causing crash

### DIFF
--- a/src/message.cpp
+++ b/src/message.cpp
@@ -167,6 +167,10 @@ static void format_warn(const char *file,int line,const char *text)
 static void do_warn(bool enabled, const char *file, int line, const char *prefix, const char *fmt, va_list args)
 {
   if (!enabled) return; // warning type disabled
+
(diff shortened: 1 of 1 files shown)


albert-github:
Thanks!! :+1: :+1: :+1:


Together with someone else I've been searching the last few for the same problem on openSUSE (on Windows and Cygwin the problem did not appear though).
The patch has been tested for the case that crashed and it now worked!!

PR #6977, opened 2019-05-09
Fixes #6765 - Fixed C++17-namespace assignment bug

### LINKED ISSUE #6765: Error Parsing C++17-style Nested Namespaces
I am using the Windows version of Doxygen 1.8.15.

Doxygen is having trouble parsing nested namespaces in the C++17 style, in that it does not close them correctly. For example, if I have the following code:

```
namespace First::Second::Third {
class Foo;
}

namespace One::Two::Three {
class Bar {
```

Then Doxygen will, in its class list, categorize the Bar class as being in the First::Second::One::Two::Three namespace. That is, it will erroneously parse the closing bracket on the C++17-style First::Second::Third namespace as closing only the outermost namespace ("Third"), so whatever code follows next will be in the First::Second namespace.

This is problematic in the case of forward declarations of classes that are located in a nested namespace, as in the example above.

If you need any clarifications, let me know!

mrhatch97:
Added scope save and restore before and after namespace parsing

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -3941,6 +3941,7 @@ OPERATOR  "operator"{B}*({ARITHOP}|{ASSIGNOP}|{LOGICOP}|{BITOP})
 					  else
 					  {
 					    current->endBodyLine = yyLineNr;
+                        Entry * original_root = current_root; // save root this namespace is in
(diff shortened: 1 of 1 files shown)


h-2:
Any news here? We can't switch to curretn doxygen because it's broken due this error.

Issue #6960, opened 2019-05-03
dot file gets deleted too early if DOT_CLEANUP = YES (at least on windows)


cmorty:
Version: 1.8.15 Windows
Output: SVG

I have a reproduceable environment where the `.dot` file is deleted before dot is run if `DOT_CLEANUP` is set to `yes`. This only happens on the second run if the target folder is not yet empty. Deleting all dot files manually works fine if  `DOT_CLEANUP`is set to `no`.

I traced this using procmon. As you can see at `15:41:07,7761734` the files is marked for deletion, before `dot` tries to open it and fails.

[doxy.log](https://github.com/doxygen/doxygen/files/3142131/doxy.log)

albert-github:
As described it looks like a race condition, but it is a bit strange to me that every time the file `C:\mypath\doc\html\de\d5c\a01708_a11d6d240c4d4eb4224e907766ceea3af_icgraph.dot` is used in the messages.

Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

cmorty:
Ok, I spent some time digging into this. Seems like `DotClassGraph::writeGraph` gets called multiple times (5 in my case) for the same file.  For the last call the regenerate check matches and it deletes the `.dot` (`if (!regenerate) removeDotGraph(absDotName);`). 
Unfortunately I'm not too familiar with debugging QT using VS and QString is giving me a hard time.
Feel free to give me pointers, as I actually have a running build system. ;)

albert-github:
Is the code propriety code or open source code (so we can see if we can share it), it is hard to tell where to look without an example. The idea I currently have is that there is a class / namespace that is multiple times defined.

cmorty:
Unfortunately is proprietary and from a customer. But I'll try to invest some time to track this down.

thomas-haller:
I produced a code example for this bug(s).
[doxygen_test.zip](https://github.com/doxygen/doxygen/files/3161922/doxygen_test.zip)

Execute "doxygen doxygen/Doxyfile.Demo" twice. At the second run, it will fail (twice) to run dot because the .dot file got deleted and it will run dot twice on another .dot file. So there are two bugs.
This was reproduced here with trunc version in Windows 10 and with version 1.8.13 in Ubuntu.

My understanding of the bugs so far (names are related to the example):

[...]

PR #6980, opened 2019-05-10
fixed bug of .dot files being deleted to early or processed multiple times

thomas-haller:
issue #6960

in some cases dot is still called although not needed. But this is only a performance issue not a real bug.

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -1315,6 +1315,23 @@ void DotManager::addRun(DotRunner *run)
   m_dotRuns.append(run);
 }
 
+// check if the list of runs contains a run for this file
+bool DotManager::containsRun(const QCString& file, const QCString& path) {
+
+  QListIterator<DotRunner> li(m_dotRuns);
+  DotRunner* dr;
+  // go through all queued DotRunner
+  for (li.toFirst(); (dr = li.current()); ++li)
+  {
+    // check if the DotRunner matches
+    if (strcmp(dr->getFile().data(), file.data()) != 0) continue;
+    if (strcmp(dr->getPath().data(), path.data()) != 0) continue;
+    return true;
+  }
+  // no match found
+  return false;
+}
+
 int DotManager::addMap(const QCString &file,const QCString &mapFile,
                 const QCString &relPath,bool urlOnly,const QCString &context,
                 const QCString &label)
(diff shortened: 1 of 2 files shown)


albert-github:
Some thoughts about the patch.

I think it might work, though it looks at the end if a file already exists
- I think this is a bit late
- The question is also is it guaranteed the the contents is the same (except for the node numbers)


In the past the nodes were numbered per file / graph, but due to some problems with uniqueness this has been changed, so here a case of not the same files, probably, went unnoticed.
So the basic questions are:
- why is the file generated twice, can we prevent this
- is it possible that the files are not identical (except for the node numbers), in this case doxygen would have a bit a bigger problem.

thomas-haller:
You are right. This patch is more a damage limitation, because it is still possible that a file is regenerated which does not need a regeneration.

The first bug is that the file is generated (twice or more often) although not needed. This currently happens due to the node numbers changing. Do you think it is possible to renumber just inside the .dot file relative to the first node? So the first node becomes '0' and all others change their value relative to it?

I did not observe that not identical files are generated, but we could add a test for that. Just check if the file is queued for regeneration if the md5 do not match. That should not happen.

The second bug is that the .dot file becomes deleted while the file is queued for regeneration. This also has the side effect that in the html code the text for not-regenerated files is inserted. For this bug I think my patch is a good solution.

albert-github:
No, I don't think it would be OK to renumber the nodes. The numbering has been chosen so when having multiple images (with map files etc.) in one output file it won't be possible that there is more than one time a certain label.

I did not observe the problem of one file name with different content (excluding the different node numbers of course), but I'm not 100% sure it cannot happen.

A strategy that I could think of is that the md5 sum is not calculated over the complete image but over the "stripped" image where 'Node[0-9]*' is replaced by 'Node' so the structure is checked.

thomas-haller:
I am currently writing more improvements in dot.cpp and I am not sure if it is a good idea to remove the node numbers for md5. I can think of some cases in which the node numbers can partly change and produce a different output (which should have a different md5).

Can you explain in more detail why it is not possible to renumber the nodes inside the .dot files? I only found this line which says the node numbers are ignored.
https://github.com/doxygen/doxygen/blob/2840a9aede8fd841d8e0602f8f093665681d4fc7/src/dot.cpp#L398

albert-github:
I have in the back of my head that there otherwise would be a problem with xhtml, but at the moment I don't have prove / a case for it.
In case the id="..." is always striped out of the map part but I'm not 100% sure what happens in other parts of the output.
The case "in which the node numbers can partly change and produce a different output (which should have a different md5)." would be "disastrous" when this is possible as we have at that moment exactly the case with different images but with one filename!

cmorty:
@albert-github . It's important the the `<map>` tag's ID is correct, as that gets referenced. However the actual node ids shouldn't matter. They get removed from the map file and for producing a png they don't matter at all.
And as you said yourself: In case they would matter there would already be lots of issues.

albert-github:
Closed as obsolete due to implementation of alternate solution (#6987).

PR #6984, opened 2019-05-12
Bug 781306 - *** Error in `doxygen': realloc(): invalid pointer: 0x0000000001d45ba0 @ exit

albert-github:
Crash occurs, as indicated in error report, on line 131 of htags.cpp (line.resize(len+1);)
Looks like there is a buffer problem (enlarge after making the buffer smaller beforehand).
The readLine should place a `\0` at the end but to be on the save side.

### DIFF
--- a/src/htags.cpp
+++ b/src/htags.cpp
@@ -128,7 +128,7 @@ bool Htags::loadFilemap(const QCString &htmlDir)
       int len;
       while ((len=f.readLine(line.rawData(),maxlen))>0)
       {
-        line.resize(len+1);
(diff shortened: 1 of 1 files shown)


doxygen:
Although I merged this, I'm not so sure this is an actual fix. It probably just masks the problem that the pointer is invalid to start with. There are many other places where `readLine()` is followed by a `resize()`. If you look at the implementation of `resize()` it does make the last character '\0'.

albert-github:
Indeed a bit weird.

PR #7002, opened 2019-05-21
issue #6960 third fix

### LINKED ISSUE #6999: Regression on #6987 alternate fix for .dot file handling
In the CGAL BGL package we get the error message:
```
error: Problems running dot: exit code=2, command='dot', arguments='"/home/cgal-testsuite/cgal_doc_build/CGAL-5.0-Ic-39/doc/scripts/doc_master/doc_output/BGL/./inline_dotgraph_1.dot.dot" -Tpng -o "/home/cgal-testsuite/cgal_doc_build/CGAL-5.0-Ic-39/doc/scripts/doc_master/doc_output/BGL/dot_inline_dotgraph_1.png"'
error: Problems running dot: exit code=2, command='dot', arguments='"dot_inline_dotgraph_1.dot" -Tcmapx -o "/home/cgal-testsuite/cgal_doc_build/CGAL-5.0-Ic-39/doc/scripts/doc_master/doc_output/BGL/dot_inline_dotgraph_1.map"'
```
The executable used was the one that was created just after the mege of pull request #6987 

The problem here is caused by the `\dot` command but also the `\dotfile` commands gives these errors.

Some problem occurs with the doxygen test  procedure number 22 when running with the `--xhtml` option (same happens with other output formats as well).

I have created from the CGAL problem a small example: [regression.zip](https://github.com/doxygen/doxygen/files/3194050/regression.zip)

thomas-haller:
* md5 file creation is delayed after output creation. In this implementation the filename handling is based on the .dot file name, not on the basename. This is more like HEAD and should work better. should fix #6999 
* check if .dot file is already in processing queue. That part worked afaik.
* added node renumbering inside .dot files. As I now expect a high possibility for md5 collisions when the node numbers are removed for hashing ( especially in include graphs) I implemented a renumbering inside dot files.

related: #6987 #6980 #6960

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -745,11 +745,10 @@ static void removeDotGraph(const QCString &dotName)
 
 
 /*! Checks if a file "baseName".md5 exists. If so the contents
- *  are compared with \a md5. If equal FALSE is returned. If the .md5
- *  file does not exist or its contents are not equal to \a md5, 
- *  a new .md5 is generated with the \a md5 string as contents.
+ *  are compared with \a md5. If equal FALSE is returned.
+ *  The .md5 is created or updated after successful creation of the output file.
  */
-static bool checkAndUpdateMd5Signature(const QCString &baseName,
+static bool checkMd5Signature(const QCString &baseName,
             const QCString &md5)
 {
   QFile f(baseName+".md5");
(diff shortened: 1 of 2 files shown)


thomas-haller:
dotRenumberList stores the maping of node numbers. When a new graph is created it is reset (clearDotRenumberList). The size depends on the number of nodes in the largest graph, which should be lower than 1024

thomas-haller:
The renumbering can produce non-unique node numbers in the .dot file.
->closing

PR #7005, opened 2019-05-22
enable debug for re-entrant lexers

### LINKED ISSUE #7003: Regression on re-entrant sqlcode
When we have a small example (using default Doxyfile:

> \```sql
> SELECT a.id, a.name, b.id, b.name FROM foos AS a JOIN foos AS b ON a.parent = b.id;
> \```

and create doxygen with:
```
cd build
make
touch ../src/sqlcode.l
make LEX_FLAGS=-d
```
the situation before the re-entrant code possibility was integrated was that we got the debug output for the sql code fragment after the integration the code does not appear anymore.
(Might it be that the `yy_flex_debug` setting is not stored properly in the `struct yyguts_t` ?).

Problem was discovered after the integration of the pull request #7001 wheer the missing printlex statements were added and we now got in a "normal" (so without `LEX_FLAGS=-d`) executable the line:
```
--finished lexical analyzer: sqlcode.l
```

The example: [example.zip](https://github.com/doxygen/doxygen/files/3206974/example.zip)

This problem might also apply to the other re-entrant lexers (though I didn't notice it there yet).

(@groleo please have a look too).

(I did a quick search, maybe https://stackoverflow.com/questions/48850242/thread-safe-reentrant-bison-flex gives some extra information at the end of the answer)

groleo:
Albert points out that re-entrant lexers has an inverted debug flag from the non-re-entrant ones.
This should fix #7003

### DIFF
--- a/src/constexp.l
+++ b/src/constexp.l
@@ -113,6 +113,10 @@ bool parseconstexp(const char *fileName,int lineNr,const QCString &s)
   constexpYYlex_init_extra(&constexpYY_extra, &yyscanner);
   struct yyguts_t *yyg = (struct yyguts_t*)yyscanner;
 
+#ifdef FLEX_DEBUG
+  yyset_debug(1,yyscanner);
+#endif
+
   yyextra->g_constExpFileName = fileName;
   yyextra->g_constExpLineNr = lineNr;
   yyextra->g_inputString = s;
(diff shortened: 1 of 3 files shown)


albert-github:
I still see the line:
```
--finished lexical analyzer: sqlcode.l
```
in case of no debug, but this is due to the fact that the line:
```
printlex(yy_flex_debug, FALSE, __FILE__, fd ? fd->fileName().data(): NULL);
```
should be before the line:
```
sqlcodeYYlex_destroy(yyscanner);
```
and not after it (actual a problem I introduced, looks like it is only in sqlcode.l :-( , but actually happy as now we found the debug problem).

Issue #7009, opened 2019-05-23
Python class diagram not created for the base class while works for the derived class


pprebeg:
Python class diagram not created for the base class while it is created for the derived class. The derived 

class diagram shows that the base class is not documented, although it is present in the html text documentation. 
If the UML style is used, the base class diagram is shown, but it is still not connected with the derived class diagram

[test.zip](https://github.com/doxygen/doxygen/files/3212992/test.zip)

albert-github:
- which version of doxygen do you use
- please add some more reference to the questing regarding where you are missing some information. The problem with the missing diagram I d see in e.g. `classpy_d3_v_model_1_1_d3_v_model.html` in e.g.`classpy_d3_v_model_simple_1_1_d3_v_model_ctrl_simple.html` I do see a relation between 2 classes.

pprebeg:
Version 1.8.15 win64 binary
Class diagram is not shown for any of the classes in pyD3VModel.py

albert-github:
The missing images was clear but what are you missing in the images shown (with the UML look) and in the images in e.g. the file `classpy_d3_v_model_simple_1_1_d3_v_model_ctrl_simple.html`

pprebeg:
Sorry, I am not at the computer at the moment.

The problem when UML is on is that the diagram at the derived class is still not connected with the base class documentation. The base class is shown as it is not documented.

IThe related issue is that in the class list hierarchy diagram  two versions of the base classes are shown

albert-github:
When back at the computer (doesn't have to be today) show the missing information. in descriptive text with images or drawn into the images together with the names of the image files / the (html) files where they are used.

pprebeg:
In the classpy_d3_v_model_simple_1_1_d3_v_model_ctrl_simple.html the base is class drawn with a box with a gray border that denotes an undocumented struct or class. It should be drawn with a box with a black border that denotes a documented struct or class. Also, the grayed box does not contain link to the base class. It looks like that the derived class does not recognize that the base class is documented.  
As written above, the related issue is that that in the class list hierarchy diagram (inherits.html) two versions of the base classes are shown. One that comes from the base class documentation, while the other comes from the inherited classes documentation.  The problem remains no mater whether Graphwiz (DOT) is enabled or not.

The Python program that uses those two files runs without any issues.

albert-github:
[...]

PR #7011, opened 2019-05-24
issue #7009 Python class diagram not created for the base class while works for the derived class

albert-github:
- handle imports with parenthesis
- improve handling of line continuation in `from ... import`

### DIFF
--- a/src/pyscanner.l
+++ b/src/pyscanner.l
@@ -745,6 +745,9 @@ STARTDOCSYMS      "##"
 		  initEntry();
                     if (!importBracket) BEGIN(Search);
 		}
+  \\\n          |
+  {B},{B}\\\n   |
+  {B}\\\n{B},   |
   {B},{B}\n     |
   {B}\n{B},     { // handling line continuation
                   incLineNr();
@@ -756,6 +759,10 @@ STARTDOCSYMS      "##"
   {B}           {
                 }
   ","           {
+                }
+  "("           {
+                  if (importBracket) REJECT;
+                  importBracket = TRUE;
                 }
   .             {
                   unput(*yytext);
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github  I've pushed a slightly more elegant [solution](https://github.com/doxygen/doxygen/commit/7f0ea786de3153f535ca20280d6b8660ffefd2e6). Please check if that also addresses the problem.

albert-github:
@doxygen, this does not address the issue as mentioned in #7009 (at the end the test case is attached)

I created 3 executables:
1. current master 7f0ea78
2. before the fix 04776ad
3. fix as proposed in this pul request applied on 04776ad (naming it `04776ad _7009`)

I looked at the image as shown in the file `classpy_d3_v_model_simple_1_1_d3_v_model_ctrl_simple.html` and here we see that the version of 1) and 3) are wrong, they show the grey box and 2) is correct.

**Image produced by 1)**

img_7f0ea78


**Image produced by 2)**

img_04776ad


**Image produced by 3)**

img_04776ad_7009


Other problems occur also with other pages, the case and the results: [test_case.tar.gz](https://github.com/doxygen/doxygen/files/3466987/test_case.tar.gz)

doxygen:
Indeed, looks like I accidentally added a newline before the body of the scanner rule causing it to be ignored. Should be fixed with [this commit](https://github.com/doxygen/doxygen/commit/f554f9e4e4d06002e65aa49afc4560671348a42b)

doxygen:
I do see a warning: 

`pyD3VModel.py...
/issue7009/tst/src/pyD3VModel.py:149: warning: Reached end of file while still inside a (nested) comment. Nesting level 1 (probable line reference: 141)`

on both python files... probably a regression related to the earlier 'nested comments' counting change.

albert-github:
Regarding the `Reached end of file while still inside a (nested) comment. Nesting level 1 (probable line reference: 141)` ad the similar message from the other file, is strange all looks OK. Has to be investigated.

albert-github:
As the original problem has been fixed, I'll close this pul request.
The problem regarding "Reached end" is a separate issue.

PR #7016, opened 2019-05-29
issue 7015 Parsing issue with @includedoc and @startuml

albert-github:
In case of `\includedoc` the `\n` was not handled on the first line (`@startuml`)

Extended example: [example.zip](https://github.com/doxygen/doxygen/files/3233413/example.zip)


kanje:
I have verified this fix on my project. It has resolved the issue.

PR #7025, opened 2019-06-03
Refactoring of dot code

thomas-haller:
I did a lot of changes in the dot code. If you do not like parts of this pull request, tell me. I will then prepare a new one.

Changes:
* .md5 files are created after the output has been created. In case dot crashes and does not produce the correct output, there will be no correct .md5 now.
* added check if a .dot file is already queued for processing. This should prevent duplicate processing. Now I think this may be not working in case bitmap _and_ eps output is needed. Fixed in later commit.
* moved more code to DotGraph from its derived classes.
* cleaned up config variable handling (bugfix in DotFilePatcher::run)
* split dot.h and dot.cpp in multiple files to make it easier readable
* DotRunner and Jobs are now only created if they do not exist yet. (This fixes the problem in case bitmap and eps output are needed)
* Cleaned up DotRunner from code that was a workaround for bugs most probably not correctly understood.

I saw that rebase did something wrong in some commits. I fix that when I have feedback from you about this pull request.


src/dotgraph.cpp
```diff
@@ -0,0 +1,354 @@
...
+  m_baseName = getBaseName();
+
+  computeTheGraph();
+
+  m_regenerate = prepareDotFile();
+
+  if (!m_doNotAddImageToIndex) Doxygen::indexList->addImageFile(imgName());
+
+  generateCode();
+  //todo
```
doxygen: I'm looking at integrating your pull request and I was wondering what needs to be done here?

thomas-haller: the //todo needs to be removed ;)

doxygen:
I've merged the code and refactored it some more. Since I squashed the refactoring related commits, the pull request doesn't get closed automatically it seems. 

One thing I still do not like is that the Dot* classes are no longer passed as constant objects. From the client point of view these objects are constant, only the current implementation requires them to be non-const because DotGraph caches some info during writeGraph() used by derived classes. I think this could be further improved.

cmorty:
@doxygen What are the major advantages of making them const in your opinion? 
We diskussed keeping them const, but the hoops one had to jump through - using mutable or maps to save the additional information - didn't justify the minimal advantage of a more robust interface. Especially considering that doxygen is not a library.

albert-github:
@doxygen, underneath the comment box I see the button "Close and comment", this is not automatically but it probably closes the pul request.

albert-github:
When looking at the appveyor output (first saw it locally on my machine) I see a number of messages (Environment: VSVERSION=15 2017; Configuration: Release; Platform: x64: https://ci.appveyor.com/project/doxygen/doxygen/builds/25266523/job/p3xsben67hr884yj line 2569) like:
```
classdef.cpp
C:\projects\doxygen\qtools\qlist.h(109): warning C4150: deletion of pointer to incomplete type 'DotFilePatcher'; no destructor called [C:\projects\doxygen\build\src\_doxygen.vcxproj]
  c:\projects\doxygen\src\dot.h(35): note: see declaration of 'DotFilePatcher'
  C:\projects\doxygen\qtools\qlist.h(109): note: while compiling class template member function 'void QList<T>::deleteValue(type *) const'
          with
          [
              T=DotFilePatcher,
              type=DotFilePatcher
          ]
  C:\projects\doxygen\qtools\qlist.h(115): note: see reference to function template instantiation 'void QList<T>::deleteValue(type *) const' being compiled
          with
          [
              T=DotFilePatcher,
              type=DotFilePatcher
          ]
  c:\projects\doxygen\src\sortdict.h(81): note: see reference to class template instantiation 'QList<T>' being compiled
          with
          [
              T=DotFilePatcher
          ]
  c:\projects\doxygen\src\sortdict.h(125): note: see reference to class template instantiation 'SList<T>' being compiled
          with
          [
              T=DotFilePatcher
          ]
```

albert-github:
Currently we are some problems with the rose-0.9.11.33  package as run for the Fossies website (for the package: https://github.com/rose-compiler/rose/archive/v0.9.11.33.tar.gz and attached the used Doxyfile ([Doxyfile.zip](https://github.com/doxygen/doxygen/files/3290542/Doxyfile.zip) ) , note some placeholders).
The Fossies system is a SuSe system.

We see on the console the message:

> *** Error in `doxygen': corrupted double-linked list: 0x00000000025155b0 ***

and it looks like that all files have been generated.

An earlier version of the the rose package (rose-0.9.11.30) ran with the doxygen version from June 10, 2019 (527fa741a67b9bc34ef315ff0ff6f96bae726497 ) did complete without problems.
A small test with rose-0.9.11.33 and the version of May 27 completed as well.

Currently I'm running on Cygwin with the current master (lets call it 190613) and I see a few artifacts already:

1. number of dot graphs is bigger than as seen by Fossies (15490 against 15473)
2. I get messages like: `Error: dot: can't open /cygdrive/d/Fossies/rose-0.9.11.33/html/parser_8hpp__dep__incl.dot`

albert-github:
Beside the already mentioned package there are already a number of other packages found that crash as well: conda-4.7.3 (see for the source code: https://github.com/conda/conda/archive/4.7.3.tar.gz) and gromacs-2019.3 (see for the source code: ftp://ftp.gromacs.org/gromacs/gromacs-2019.3.tar.gz).

The crashes found only occur on systems with gcc 4.8.5.

An interesting observation is as well that it looks like the crash depends on the number of threads used (thread safe problem ?):
- rose-0.9.11.33 with 1 and 2 threads runs
- rose-0.9.11.33 with 3 and 4 threads crashes
- conda-4.7.3wit 1 thread runs
- conda-4.7.3 with 2 and 4 threads crashes (3 threads nor run)

Regarding the mentioned problems on Cygwin, when `CASE_SENSE_NAMES=NO` (should be the default, see #7056) there are no problems with the file names, with `CASE_SENSE_NAMES=YES` there are some problems but I think that this setting is asking for problems (I saw the problems with rose but not with conda).

doxygen:
@albert-github it would help if you have a stack trace of the crash you see.

albert-github:
I agree but I cannot reproduce the problem, The stack trace I saw sofar didn't have names in it just some "addresses".

doxygen:
I ran a small example through valgrind's helgrind tool, which reveals a race in `readBoundingBox`:

```
==23869== Possible data race during read of size 4 at 0x5CE68F0 by thread #2
==23869== Locks held: none
==23869==    at 0x44A202: QShared::ref() (qshared.h:49)
==23869==    by 0x5812C5: QString::QString() (qstring.h:688)
==23869==    by 0x8F207C: QUtf8Decoder::toUnicode(char const*, int) (qutfcodec.cpp:130)
==23869==    by 0x8D9526: QTextCodec::toUnicode(char const*, int) const (qtextcodec.cpp:760)
==23869==    by 0x8D0BBC: QString::fromUtf8(char const*, int) (qstring.cpp:14580)
==23869==    by 0x8CAD98: QString::QString(char const*) (qstring.cpp:12402)
==23869==    by 0x842057: DotRunner::readBoundingBox(char const*, int*, int*, bool) (dotrunner.cpp:116)
==23869==    by 0x842AAA: DotRunner::run() (dotrunner.cpp:219)
==23869==    by 0x8431ED: DotWorkerThread::run() (dotrunner.cpp:298)
==23869==    by 0x8F6287: QThreadPrivate::start(void*) (qthread_unix.cpp:91)
==23869==    by 0x4C34DB6: ??? (in /usr/lib/valgrind/vgpreload_helgrind-amd64-linux.so)
==23869==    by 0x4E476B9: start_thread (pthread_create.c:333)
==23869== 
==23869== This conflicts with a previous write of size 4 by thread #5
==23869== Locks held: none
==23869==    at 0x44A20B: QShared::ref() (qshared.h:49)
==23869==    by 0x5812C5: QString::QString() (qstring.h:688)
==23869==    by 0x8F207C: QUtf8Decoder::toUnicode(char const*, int) (qutfcodec.cpp:130)
==23869==    by 0x8D9526: QTextCodec::toUnicode(char const*, int) const (qtextcodec.cpp:760)
==23869==    by 0x8D0BBC: QString::fromUtf8(char const*, int) (qstring.cpp:14580)
==23869==    by 0x8CAD98: QString::QString(char const*) (qstring.cpp:12402)
==23869==    by 0x842057: DotRunner::readBoundingBox(char const*, int*, int*, bool) (dotrunner.cpp:116)
==23869==    by 0x842AAA: DotRunner::run() (dotrunner.cpp:219)
==23869==  Address 0x5ce68f0 is 0 bytes inside a block of size 32 alloc'd
==23869==    at 0x4C2F50F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_helgrind-amd64-linux.so)
==23869==    by 0x8CAA1C: QString::makeSharedNull() (qstring.cpp:12243)
==23869==    by 0x5812A6: QString::QString() (qstring.h:686)
==23869==    by 0x8B4743: QDir::QDir() (qdir.cpp:137)
==23869==    by 0x483181: __static_initialization_and_destruction_0(int, int) (htags.cpp:30)
==23869==    by 0x4831D8: _GLOBAL__sub_I__ZN5Htags8useHtagsE (htags.cpp:182)
==23869==    by 0x94FFAC: __libc_csu_init (in /home/dimitri/doxygen/build/bin/doxygen)
==23869==    by 0x591E7BE: (below main) (libc-start.c:247)
==23869==  Block was alloc'd by thread #1
```

Probably due to the use of QFile, which internally uses QString, which is not thread safe.

albert-github:
Interesting would be in accordance with the fact that it sometimes goes OK and in other cases doesn't. Strange is though that as far as I can see the crash occurs after all dot runs and updates (output file patches) have been completed. But the crash can be a delayed result of the not thread safe
I hope we soon can get a stack trace.

doxygen:
I've committed [this](https://github.com/doxygen/doxygen/commit/f6ed279645f1dbcf157e1d5efe25811ecc317c76) fix for the race conditions. Helgrind does not report any issues with this fix.

albert-github:
Due to some misunderstanding not all the information in the issue is correct regarding the version numbers / places where crashes occurred. The current master  form June 15  around 16.00 seems to work OK. As long as no new problems occur we can consider the issue closed. In case of new problems a new issue will be created.

PR #7031, opened 2019-06-05
Ci cache

thomas-haller:
This is a feature add on.
If there is an environment variable DOTCACHEDIR, it uses this folder for caching the output files of dot. This is usefull in continues integration for improving build times.

It is based on #7025 , but could be made independant.

I would be very happy about feedback.


doxygen:
I have two questions:
1. Why use an environment variable and not a configuration setting for the cache?
2. How is the cache different from checking the md5 checksum and regenerating the image only when it is different/absent? and what does the cache bring in terms of performance gain?

thomas-haller:
In CI you build typically in a clean folder to be independant from previous runs. So the output folder is empty at start.
The build machine can provide a caching folder. As this is not depending on the software to be build , but depends on the machine it is build on, it is better to use an environment variable which is set by the build machine.
The performance gain is the same as the md5 mechanism. But the md5 mechanism cannot work in CI while the caching mechanism can if the environment variable is set.

cmorty:
This has another advantage. Doxygen currently does not clean up files it created in one of it's previous runs (e.g. when a source file was deleted). The simplest way to fix this is to delete the documentation folder. However in that case the cache is lost. Having a seperate cache allows to simply delete the documentation folder and recreate it without performance loss.
Thus I suggest the following changes:
* Provide a configuration option to set a cache folder
  Add a note that the user is currently responsible for cleaning up the cache (maybe hinting at a pull-request being wellcome)
* Allow overriding the configuration option using an environment variable
  This will also allow multiple checkouts a CI or whatever to share the cache
* Do not create .md5 files if the cache is activated
  They are not necessary any more, and an optimistic copy (and creating a .dot in case it fails) should have a lower performance impact than reading the md5.

@doxygen Would those changes motivate you to merge this?

doxygen:
@cmorty yes these sound as useful enhancements making the feature more useful and less hidden. Please also update doxygen's documentation accordingly.

PR #7032, opened 2019-06-05
Relative markdown file reference

albert-github:
In case we have a relative reference to a local markdown file this file is not found when the relative path starts e.g. with `..`
So when we have a file:
```
docs\tutorial\security.md
```
and this references the file:
```
../api/browser-window.md
```
through the markdown syntax:
```
[`BrowserWindow`](../api/browser-window.md)
```
then the link was not found and a warning was given.

Example: [example.zip](https://github.com/doxygen/doxygen/files/3257344/example.zip)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -51,6 +51,7 @@
 #include "config.h"
 #include "section.h"
 #include "message.h"
+#include "portable.h"
 
 //-----------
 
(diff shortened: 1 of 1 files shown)


avostrik:
Dear maintainers,
When this change will be integrated into master in nearest future? 

It really helps to keep markdown linked files structure without special handling to be shown on hosted git and in generated Doxygen docs.

It will be required to have minor fix to use `Portable::isAbsolutePath` instead of portable_isAbsolutePath

ssproessig:
checked the patch with the `Portable::isAbsolutePath` fix locally - local Markdown links are now resolved as expected. 

👍  for merging + thank you @albert-github

PR #7036, opened 2019-06-06
Bug 504025 - preprocessor-directive doxygen is not able to recognize

albert-github:
No reason to limit preprocessor detection to states 'Body' and 'Bases'.

### DIFF
--- a/src/code.l
+++ b/src/code.l
@@ -2061,7 +2061,7 @@ RAWEND    ")"[^ \t\(\)\\]{0,16}\"
 					  endFontClass();
 					  BEGIN( Body );
   					}
-<Body,Bases>^[ \t]*"#"			{ 
+<*>^[ \t]*"#"			{ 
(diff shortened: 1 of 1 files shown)


src/code.l
```diff
@@ -2061,7 +2061,7 @@ RAWEND    ")"[^ \t\(\)\\]{0,16}\"
 					  endFontClass();
 					  BEGIN( Body );
   					}
-<Body,Bases>^[ \t]*"#"			{ 
+<*>^[ \t]*"#"			{
```
doxygen: Maybe `<Body,Bases>` is too limited, but `<*>` is too liberal. We don't want to handle preprocessor directives in comments or string literals for instance.

albert-github: I agree that `<*>` is a bit too liberal. It is hard to determine which states should be used, so I added just the state from this issue. Feel free to add other states.

PR #7037, opened 2019-06-06
Bug 664652 - Incorrect processing of the \date command in C++

albert-github:
In case of a `\brief` command terminate the current paragraph / section.


albert-github:
Checked against:
- 1.8.16: problem present
- 1.8.19: problem present
- 1.8.20: problem gone

So I think the problem has been solved in the mean time.

Example used: [example.tar.gz](https://github.com/doxygen/doxygen/files/5135944/example.tar.gz)

PR #7047, opened 2019-06-10
Bug 566109 - unable to resolve links to java functions with array arguments

albert-github:
Also handle '[' in the removeRedundantWhiteSpace, analogous to a '*'

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -1847,6 +1847,7 @@ QCString removeRedundantWhiteSpace(const QCString &s)
           *dst++=' '; // add extra space after
         }
         break;
+      case '[':
       case '*':
         if (i>0 && pc!=' ' && pc!='\t' && pc!=':' &&
                    pc!='*' && pc!='&'  && pc!='(' && pc!='/' &&


src/util.cpp
```diff
@@ -1847,6 +1847,7 @@ QCString removeRedundantWhiteSpace(const QCString &s)
           *dst++=' '; // add extra space after
         }
         break;
+      case '[':
```
doxygen: This will process `A[]` as `A []`. I afraid this might have unwanted site-effects.

albert-github: I do understand, any suggestions how to tackle the problem as when we have:
```
	/** {@link #main(String[])}
	 *  {@link #foo(Object...)}
	 * @param args */
	public static void main(String[] args) {}
```

at one place in the original the `String` and `[]` are separated and in the other place they are not.

doxygen: I've committed an alternative fix: https://github.com/doxygen/doxygen/commit/3360831da443551791f4115cd4c5373b6f99a8d7

albert-github:
An alternative approach has been implemented in the master on github.

Issue #7050, opened 2019-06-12
Physical newlines (^^) not working in group names and without spaces in 1.8.15


Inujel:
[physical_line_break_issue.zip](https://github.com/doxygen/doxygen/files/3280621/physical_line_break_issue.zip)

In a fresh Doxyfile, I used the following ALIASES: 
```
ALIASES                = "stuff1{2}=\1^^\2"   \
                         "stuff2{2}=\1 ^^ \2" \
                         "mygroup{2}=\defgroup MY-\1 \2 ^^ This is the group brief" 
```

I invoke them with this source file:
```
/** @page fruits Fruits
 *  
 *  @stuff1{Apple,Pear}
 * 
 *  @stuff2{Apple,Pear}
 */

/** @mygroup{Orange,Banana} 
 * 
 *  This is the detailed section
 */
```

In 1.8.14, all '^^' work as intenteded, i.e., as if I had actually entered a new line in the alias text. My group is named "Banana" with a brief section containing "This is the group brief".

In 1.8.15, only stuff2 seems to work correctly. 
stuff1 produces "Apple@_linebrPear"
The group is named "Banana @_linebr This is the group brief"

Note: In 1.8.14, mygroup also works without spaces around ^^.

albert-github:
I checked it against the current master as well and the `^^` as well as the ` ^^ ` don't give warnings (so with and without spaces).
Problematic is still the part:
```
/** @mygroup{Orange,Banana}
```
where we get in the output:
```
Banana \_linebr This is the group brief
```  
This probably is related to #6998, where headers are not processed and thus doxygen commands and e.g. `<a` remain.

albert-github:
I've just pushed a proposed patch, pull request #7053

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

Inujel:
Great, thanks!

Le dim. 16 juin 2019 à 13:14, albert-github <notifications@github.com> a
écrit :

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.16.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7053, opened 2019-06-12
issue #7050 Physical newlines (^^) not working in group names and without spaces in 1.8.15

albert-github:
Handling of `\_linebr` with `\defgroup` and also with the `\page` command.
Adding a missing linecount.

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -1576,7 +1576,7 @@ RCSTAG    "$"{ID}":"[^\n$]+"$"
   					  yyLineNr++; 
 					  addOutput('\n');
                                         }
-<GroupDocArg2>[^\n\*]+			{ // title (stored in type)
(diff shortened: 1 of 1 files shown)


Inujel:
It works, but something else is broken (don't know if it's from this patch or from 1.8.16 though): 

When using an alias like 
`defstuff{1}=@defgroup STUFF-\1 \1`

```
\defstuff{APPLE}
My apple brief
```

Sometimes, the brief appears in the details section, sometimes in the brief section. I can't seem to find a reason why. 
If I explicitely use the "\brief" command, it always works. It would be nice not to have to though.

albert-github:
I see also some problems with the example (chapter grouping) in the current master (with and without this patch). The version of March 24, 2019 (currently the version on the website) looks OK.

So I think this problem should be handled as a separate issue.

albert-github:
For the remark regarding the defgroup command I created a separate proposed pull request #7054.

Issue #6945, opened 2019-04-25
\class after \brief doesn't count as a new line - resulting docs show merged descriptions


TheMaverickProgrammer:
See [this post on SO.](https://stackoverflow.com/questions/55844783/doxygen-detailed-description-bleeds-into-the-brief-description?noredirect=1#comment98352341_55844783)

Consider the following:

```
/*! \brief Spawns two mettaurs on the field
 *  \class TwoMettaurMob
 * 
 * May optionally spawn an empty tile for challenge
 */
```

The output description will be

`Spawns two mettaurs on the fieldMay optionally spawn an empty tile for challenge`

The solution is to add space between the `\brief` and `\class` command

albert-github:
Besides the `\class` command, I think, also the following commands could be set to terminate a brief description:
- `\fn`
- `\var`
- `\typedef`
- `\property`
- `\protocol`
- `\enum`
- `\category`
- `\union`
- `\struct`
- `\namespace`
- `\package`
- `\dir`
- `\def`
- `\defgroup`
- `\idlexcepte`
- `\interface`

and to be consistent with the `\section` command:
- `\page`
- `\mainpage`
- `\file`

List is based on this issue and friend command as well as on the descriptions starting with: "Indicates that a comment block contains documentation..."

@doxygen your opinion ?

jwakely:
The "enhancement" label suggests there's no bug, but I think the missing whitespace between the brief description and the rest of the text is a real bug:

    Spawns two mettaurs on the fieldMay optionally spawn an empty tile for challenge

There should _at least_ be a space between "field" and "May".

doxygen:
> Besides the `\class` command, I think, also the following commands could be set to terminate a brief description:
> 
>     * `\fn`
> 
>     * `\var`
> 
>     * `\typedef`
> 
>     * `\property`
> 
>     * `\protocol`
> 
>     * `\enum`
> 
>     * `\category`
> 
>     * `\union`
> 
>     * `\struct`
> 
>     * `\namespace`
> 
>     * `\package`
> 
>     * `\dir`
> 
>     * `\def`
> 
>     * `\defgroup`
> 
>     * `\ingroup`
> 
>     * `\weakgroup`
> 
>     * `\idlexcepte`
> 
>     * `\interface`
> 
> 
> and to be consistent with the `\section` command:
> 
>     * `\page`
> 
>     * `\mainpage`
> 
>     * `\file`
> 
> 
[...]

PR #7054, opened 2019-06-13
Missing brief descriptions with `\defgroup`

albert-github:
Due to the fact that issue #6945 ended a brief description, all following was put into the detailed description (unless explicit \brief was used). The `\defgroup` should end the current brief description, but first line of the `\defgroup` should be seen as a new brief description.
Problem can be seen with in the doxygen documentation in the example in the Grouping part and also with #7053 a small example is given in the comment.

Test 19 was incorrect, didn't show group description either (see make tests `TEST_FLAGS="--keep --xhtml --id=19"`


albert-github:
Small comment. The current proposed pull request handles the `\defgroup` command., might be that there are other commands as well.

PR #7059, opened 2019-06-16
(X)HTML warning when `[` or `]` in constructed ids

albert-github:
When we have a construct like:
```
template <typename Element, size_t N>
class StlContainerView<Element[N]> {
 public:
  typedef internal::NativeArray<RawElement> type;
};
```
this will lead to files with `[` and `]` in it, as such not nice, but it is used in ids inside the code as well like:
```
id="aclass_stl_container_view_3_01_element[_n]_4_html_a1bf60158ff15896f2b53af11c09524fb"
```
running
```
xmllint --path ..../testing/dtd --noout  --nonet --postvalid html_org/*html
```
this leads to the message:
```
Syntax of value for attribute id of div is not valid
Document html_org/test_8h_source.html does not validate
```

escaping the `[` and `]` as done with other special characters solves this problem.

Example case with original and new results: [example.zip](https://github.com/doxygen/doxygen/files/3294367/example.zip)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -5479,6 +5479,8 @@ QCString escapeCharsInString(const char *name,bool allowDots,bool allowUnderscor
       case '$': growBuf.addStr("_0B"); break;
       case '\\': growBuf.addStr("_0C"); break;
       case '@': growBuf.addStr("_0D"); break;
+      case ']': growBuf.addStr("_0E"); break;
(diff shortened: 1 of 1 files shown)


albert-github:
NOTE: pull request #7061 contains also 4 "conversions" `_0A` till  `_0D`, these have been converted to lowercase so when #7061 is integrated this pull  (#7059) has to be adjusted as well.

PR #7083, opened 2019-06-27
Non existing MSC file crashes doxygen

albert-github:
- In case a msc file does not exist doxygen crashes (on Windows) as the msc conversion tries to access a NULL Pointer.
- Analogous for dot and dia to prevent messages about conversion errors fro a `.dot` / `.dia` file to the required output format.

example: [msc_2.zip](https://github.com/doxygen/doxygen/files/3335868/msc_2.zip)

### DIFF
--- a/src/docbookvisitor.cpp
+++ b/src/docbookvisitor.cpp
@@ -1241,41 +1241,41 @@ void DocbookDocVisitor::visitPre(DocDotFile *df)
 {
 DB_VIS_C
   if (m_hide) return;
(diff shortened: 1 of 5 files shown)


doxygen:
I decided to fix this differently; instead of checking for an empty file at many places, I propose to not add DocNodes that are not 'valid' in the first place. See https://github.com/doxygen/doxygen/commit/c131d65638effead8aa92a93fe16e1e61ce7180c

albert-github:
Pull request has been fixed in a different, better, way.

Issue #7109, opened 2019-07-08
Doxyapp can't use "[source_file | source_dir]" parameter


hakua-doublemoon:
# Issue(as is)

* I used doxyapp as follows:

> ./doxyapp src/test.c

* Src directory had some c files. And, doxyapp made lists about all of src's files.
* This always happens.

# My Expects(to be)

* I want a list about only src/test.c.

# Environment

* Ubuntu (WIndows Subsystem for Linux)

# Resolution

* I think this issue is caused by the location of `checkConfiguration()` and `Config_getList(INPUT).append()`
    * Doxyapp don't has a Doxyfile. So after `checkConfiguration()`, all of Configs are initialized.
    * So I think I should rewrite it as below:

```
  checkConfiguration();
  Config_getList(INPUT).append(argv[1]); // insert this sentence!
  adjustConfiguration();
```

-----

I'm sorry my English is poor. Please let me know if you can't understand what I say or want more information.

albert-github:
- Please specify the doxyapp version used.
- Which OS are you running on?
- did you make any changes ?
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent. (and also the commands given in the doxyapp).

The doxyapp is an example program, for what are you using it?

hakua-doublemoon:
Hello,

I answer your question:

* Please specify the doxyapp version used.

> commit 34c9bac4624c5b94172af4a9d9168a44482fe552 (HEAD -> master, origin/master, origin/HEAD)
> Merge: 54f758e7 e40b8afa
> Author: Dimitri van Heesch <doxygen@gmail.com>
> Date:   Mon Jul 8 15:48:16 2019 +0200
> 
>     Merge pull request #7111 from albert-github/feature/issue_7110
> 
>     issue #7110 URL Encoding for Hyperlinks in PDF

* Which OS are you running on?

> :~/Windows/Documents/program/doxygen$ cat /etc/os-release
> NAME="Ubuntu"
> VERSION="18.04.2 LTS (Bionic Beaver)"
> ID=ubuntu
> ID_LIKE=debian
> PRETTY_NAME="Ubuntu 18.04.2 LTS"
> VERSION_ID="18.04"

* did you make any changes ?
    + no change

[...]

PR #7118, opened 2019-07-11
issue #7109 Doxyapp can't use "[source_file | source_dir]" parameter

albert-github:
- initialization has to be done in an earlier stage (and only once)
- not only the INPUT tag was involved, none of the tags were set.

### DIFF
--- a/addon/doxyapp/doxyapp.cpp
+++ b/addon/doxyapp/doxyapp.cpp
@@ -261,8 +261,8 @@ int main(int argc,char **argv)
 
   // setup the non-default configuration options
 
-  Config::postProcess(FALSE);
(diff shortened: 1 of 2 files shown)


hakua-doublemoon:
Thank you for your work.
I tried it. And it works well for a file, but not good for directories.

```
$ ./doxyapp
Usage: ./doxyapp [source_file | source_dir]
$ ./doxyapp ../src/
> Type a symbol name or
> .list for a list of symbols or
> .quit to exit
> .list
> Type a symbol name or
> .list for a list of symbols or
```

albert-github:
I just updated the pull request so the directory is also recognized.

PR #7126, opened 2019-07-18
Latex output format fixes

BCiesla:
In this pull requests two issues related to LaTeX output are addressed
1. Align latex output to man format where each section name is terminated with colon.
2. Paragraph header in pdf has same indent as other sections like param or retval

### DIFF
--- a/src/latexdocvisitor.cpp
+++ b/src/latexdocvisitor.cpp
@@ -797,7 +797,7 @@ void LatexDocVisitor::visitPre(DocSimpleSect *s)
   // special case 1: user defined title
   if (s->type()!=DocSimpleSect::User && s->type()!=DocSimpleSect::Rcs)
   {
-    m_t << "}\n";
+    m_t << ":}\n";
   }
   else
   {
@@ -1517,7 +1517,7 @@ void LatexDocVisitor::visitPre(DocParamSect *s)
     default:
       ASSERT(0);
   }
-  m_t << "}\n";
+  m_t << ":}\n";
 }
 
 void LatexDocVisitor::visitPost(DocParamSect *s)
--- a/templates/latex/doxygen.sty
+++ b/templates/latex/doxygen.sty
@@ -306,16 +306,9 @@
 
 % Used by @par and @paragraph
 \newenvironment{DoxyParagraph}[1]{%
-  \begin{list}{}{%
-    \settowidth{\labelwidth}{40pt}%
-    \setlength{\leftmargin}{\labelwidth+\labelsep}%
-    \setlength{\parsep}{0pt}%
-    \setlength{\itemsep}{-4pt}%
-    \renewcommand{\makelabel}{\entrylabel}%
-  }%
-  \item[#1]%
+  \begin{DoxyDesc}{#1}%
 }{%
-  \end{list}%
+  \end{DoxyDesc}%
 }
 
 % Used by parameter lists


BCiesla:
Here's the key part of the example which shows the differences - left side is the output from 1.8.15, right is with the proposed changes. On the left sections "Errors" and "Restrictions" are not aligned with "See also" section. On the right section "See also" has added colon. Paragraphs Descriptions, Errors and Restrictions were created using @par command. Red line is added for reference.
comparison

I add also header file [API.h.txt](https://github.com/doxygen/doxygen/files/3406933/API.h.txt) and doxygen config [config.txt](https://github.com/doxygen/doxygen/files/3406932/config.txt) and whole pdf documents generated [with proposed changes](https://github.com/doxygen/doxygen/files/3406902/refman-1.8.15.pdf) and [without](https://github.com/doxygen/doxygen/files/3406901/refman-1.8.16.pdf).

BCiesla:
**Regarding the `:`**
if the presence of : in the LaTeX / pdf is decided by the language  used in the document and it is intended (design decision) then I'm fine to drop this change. 

**Regarding doxygen.sty change**
I'm not able to build doxygen manual - tex says "! TeX capacity exceeded, sorry [main memory size=17000000].". I'm using texlive under Linux.
Are you using the same LaTeX distribution?

albert-github:
I used on native windows:
`MiKTeX-pdfTeX 2.9.6839 (1.40.19) (MiKTeX 2.9.6840 64-bit)`
and under cygwin:
`pdfTeX 3.14159265-2.6-1.40.19 (TeX Live 2018/Cygwin)`

albert-github:
**Regarding the `:`**
My intention here is to get the general rule of thumb what should be used, the problem with Docbook is definitely not good as here it is a mixture, a soon as the general definition is available I can update the relevant code,

@doxygen use everywhere `:` or nowhere or output format depending ?

doxygen:
> @doxygen use everywhere `:` or nowhere or output format depending ?

@albert-github It should be consistent. If it is visually already clear that it is a heading, then the `:` should be omitted.

albert-github:
I do agree that the colon should only be present when it is not clear that it is after a heading and that the usage should be consistent.

Using the example of this pull request, where user added for 'Description', 'Errors\' and 'Restrictions\` paragraphs with colon to the documentation and the 'Parameters', 'Returns\' and 'See also\' are generated by doxygen.

**HTML**

html

Here it is clear that the mentioned texts are headers and no colon necessary.

**LaTeX**

latex

Here it is clear that the mentioned texts are headers and no colon necessary.

**rtf**

rtf

Here it is clear that the mentioned texts are headers and no colon necessary and the colons should be removed.

**docbook**

docbook

Here it is clear that the mentioned texts are headers and no colon necessary and the colons that are present should be removed ('Returns\', 'See also').

**man**

man

Here it is clear that the mentioned texts are headers and no colon necessary and the colons that are present should be removed.

**Conclusions / Actions**

- HTML: Here it is clear that the mentioned texts are headers and no colon necessary, just quick visual check necessary to see whether it is done consistent.
- LaTeX: Here it is clear that the mentioned texts are headers and no colon necessary, just quick visual check necessary to see whether it is done consistent.
- rtf: Here it is clear that the mentioned texts are headers and no colon necessary and the colons should be removed.
- docbook: Here it is clear that the mentioned texts are headers and no colon necessary and the colons that are present should be removed.
- man: Here it is clear that the mentioned texts are headers and no colon necessary and the colons that are present should be removed.


@doxygen correct?

doxygen:
@albert-github yes that looks correct to me

albert-github:
OK I'll have a go at the coming week.

albert-github:
Based on the above discussion I've just pushed a proposed patch, pull request #7146

PR #7131, opened 2019-07-19
code.l: make CallContext independent of global variables

groleo:


groleo:
> Some remarks.
> 
>     * I don't understand why on line 349 the empty string is used as initializer and on line 389 the NULL pointer is used.
> 
>     * I don't understand why on the mentioned places the empty string / NULL pointer is used, is it guaranteed that g_type / g_name have these values at that moment?
In the current source code, it should indeed be both "".
I've cherry-picked the patch from an internal branch (more reentrant changes) where g_type and g_name
are part of a lexer state struct, and there it made sense to have "" or NULL to catch possible invalid usages.
But, as i said, it's not the case yet with the current source code.

Issue #6769, opened 2019-01-13
Tabu is broken


sharris40:
Recent updates to LaTeX make it so Doxygen can't even generate its own documentation without the user patching tabu. See https://github.com/latex3/latex2e/issues/111 for an explanation. Tabu is no longer maintained and may be forked. There are a couple of open issues related to tabu not working properly, notably https://github.com/doxygen/doxygen/issues/6034 and https://github.com/doxygen/doxygen/issues/6332.

albert-github:
You claim "Recent updates to LaTeX make it so Doxygen can't even generate its own documentation without the user patching tabu.",  are you referring to the doxygen manual or something else? With the doxygen manual I have no problems at all.

Please give some more details.

sharris40:
I am referring to the Doxygen manual, yes. The problem only happens if you update LaTeX with tlmgr or whatever package manager your TeX Live distribution uses; specifically, the tools package (which may be renamed to latex-tools soon). If you're using the TeX Live bundled with your operating system, it may not have tlmgr, so this problem will only come up when your OS updates that package.

This won't be a major issue for a few months, since not many people will be updating TeX Live packages separately. However, the changes will likely show up in the annual updates in April.

albert-github:
This is not nice. Problem will to get it running with the newer versions but also with older versions of LaTeX. I did a small test with the distributions I have but this gave an error. I posted a comment with the answers from Ulrike Fischer and Frank Mittelbach about how to run with older versions (actually I hope that the tabu package will be fixed).

albert-github:
Some action is on its way ...
https://tex.stackexchange.com/questions/470107/incompatibility-of-fix-in-latex3-latex2e-in-respect-to-unmaintained-tabu-packa

https://github.com/tabu-fixed/tabu

sharris40:
The recent updates to tabu don't entirely fix the problem. I can now make it compile without errors by patching out the `\color@endgroup` lines in `tabu.sty`, but that seems to disable colored tables entirely. Using the old version of tabu with the `\color@begingroup` patch, I can get the PDF looking correct, but `pdflatex` complains and throws out several errors, which causes `make` to bail out.

albert-github:
Did you have a look at https://github.com/tabu-fixed/tabu this is already at CTAN and should enter the TexLive / MikTeX version in the next few days.

sharris40:
I pulled the most recent version from CTAN, as far as I can tell.

albert-github:
[...]

PR #7137, opened 2019-07-22
issue #6769 Tabu is broken

albert-github:
Created a workaround for the problems in the unmaintained tabu package due to changes in LaTeX 2019
Changes are based on the comments of David Carlisle and Frank Mittelbach of The LaTeX3 Project Team

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -497,7 +497,7 @@ static void writeDefaultHeaderPart1(FTextStream &t)
     documentClass = "book";
   t << "\\documentclass[twoside]{" << documentClass << "}\n"
        "\n";
-  t << "%% moved from doxygen.sty due to workaround for LatEx 2019 version and unmaintained tabu package\n"
+  t << "%% moved from doxygen.sty due to workaround for LaTex 2019 version and unmaintained tabu package\n"
        "\\usepackage{ifthen}\n"
        "\\ifx\\requestedLaTeXdate\\undefined\n"
        "\\usepackage{array}\n"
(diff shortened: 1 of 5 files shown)


templates/latex/doxygen.sty
```diff
@@ -3,14 +3,14 @@
...
+%%\RequirePackage{array} %% moved to refman.tex due to workaround for LatEx 2019 version and unmaintained tabu package
 \RequirePackage{calc}
 \RequirePackage{float}
-\RequirePackage{ifthen}
+%%\RequirePackage{ifthen} %% moved to refman.tex due to workaround for LatEx 2019 version and unmaintained tabu package
 \RequirePackage{verbatim}
 \RequirePackage[table]{xcolor}
-\RequirePackage{longtable}
-\RequirePackage{tabu}
+%%\RequirePackage{longtable} %% moved to refman.tex due to workaround for LatEx 2019 version and unmaintained tabu package
```
doxygen: is it really necessary to move these to refman.tex? Can't we just do the \RequirePackage{longtable_doxygen} here instead?

albert-github: Test with doxygen_manual indicates that this is possible, created updated pull request.

src/latexgen.cpp
```diff
@@ -497,7 +497,7 @@ static void writeDefaultHeaderPart1(FTextStream &t)
     documentClass = "book";
   t << "\\documentclass[twoside]{" << documentClass << "}\n"
        "\n";
-  t << "%% moved from doxygen.sty due to workaround for LatEx 2019 version and unmaintained tabu package\n"
+  t << "%% moved from doxygen.sty due to workaround for LaTex 2019 version and unmaintained tabu package\n"
        "\\usepackage{ifthen}\n"
        "\\ifx\\requestedLaTeXdate\\undefined\n"
```
doxygen: I'm not 100% confident about this construct. I think it would be safer to also include the matching version of array (as array_doxygen) along with doxygen. Then this can also be placed back in doxygen.sty

albert-github: I don't think so, this was suggested by the LaTeX group members as it is provided as part of the array  package (and I also wanted it to load as one of the first so that no other package could load another version of the array package).

PR #7142, opened 2019-07-26
take doc group out of commentscan.l

groleo:
most of the flex files have a simple API, usually consisting of a few functions,
making it easy to have a per-call context.
This is not the case with commentscan.l.
The parts that deal with doc group has no logical dependency to the things going on in commentscan.l.

This also helps to refactor commentscan.l into a re-entrant scanner, as the scanner state
will only be about the commentscanner, not commentscanner _and_ doc groups,
making it easier to know when to destroy the scanner state struct.


albert-github:
Small description about the reason would be nice.

PR #7156, opened 2019-08-02
warning: More #endif's than #if's found.

albert-github:
In case in a macro substitution string the last character was a backslash this was, incorrect, not seen as a sign for a line continuation.

Example: [example.zip](https://github.com/doxygen/doxygen/files/3462223/example.zip)
(updated, removed `.o` file and inserted `.c` file)


doxygen:
@albert-github The example zip contains a `.o` file instead of a `.c` file. Can you please upload an example with source code?

albert-github:
oops ...
updated, removed `.o` file and inserted `.c` file in example.zip in original description.

PR #7192, opened 2019-08-13
Fix for Fortran scanner; missing '.' in regex

ellio167:
Bug introduced in e9ebf43585bffee80c31dd69538feae2a4525178

### DIFF
--- a/src/fortranscanner.l
+++ b/src/fortranscanner.l
@@ -334,7 +334,7 @@ SCOPENAME ({ID}{BS}"::"{BS})*
 
  /*-----------------------------------------------------------------------------------*/
 
-<Prepass>^{BS}[&]*{BS}!.*\n             { /* skip lines with just comment. Note code was in free format or has been converted to it */
+<Prepass>^{BS}[&].*{BS}!.*\n             { /* skip lines with just comment. Note code was in free format or has been converted to it */
                                           lineCountPrepass ++;
                                         }
 <Prepass>^{BS}\n                        { /* skip empty lines */


albert-github:
Can you please give an example of where this does occur ?
Seen the comment I think that in this case the asterisk (`*`) should not be present.

ellio167:
@albert-github Thanks for the comment.  I'll put together a minimal example for this issue and post it soon.

Indeed, removal of the `*` also seems to fix the problem.  However, I guess I don't fully understand the what this line is trying to match. It seems like this should match lines without a leading `&` too.?  Am I missing something here?

ellio167:
Sorry, further inspection shows that the behavior is not fixed by simply removing the `&`.

ellio167:
Example source file
[fortran-scanner-bug-example.f90.txt](https://github.com/doxygen/doxygen/files/3497785/fortran-scanner-bug-example.f90.txt)

Resulting markup with 1.8.16:
image

Resulting markup with the patch in this PR:
image

Difference notes:
PR results in correct links created for:
* log_handle_type
* push_default_verbosity
* interface push_default_verbosity
* etc.

albert-github:
I see that some of the routines are now linked, but e.g. log_handle_equal is not everywhere linked (blue) i.e. on line 38.
I'm still a bit in doubt whether this solves the problem or is just a side effect, I will do some tests.

As a side note why are all the routines made recursive, for these routines I don't see the need.

ellio167:
Agreed, there does seem to still be some problems.  Your help is greatly appreciated!

The choice to use `recursive` everywhere is for thread safety (to ensure that function local variables are actually local...)

albert-github:
Can you please also provide the used doxygen configuration file (Doxyfile) or at least the difference from the default Doxyfile (doxygen -x)?

ellio167:
[Doxyfile.txt](https://github.com/doxygen/doxygen/files/3498359/Doxyfile.txt)

Sorry for the omission.

albert-github:
A small note:

I did a small test where I replaced the part:
```
  public &
    ! Derived types
    log_handle_type, &

    ! Routines
    operator (.eq.), &
    log_create, &
    push_default_verbosity
```
with
```
public log_handle_type, operator (.eq.), log_create, push_default_verbosity
```
and even when applying the proposed patch, we see in this one liner still the not linked routines.

I think that the 'fix' just has a side effect that looks like it solves the issue, but I think the real issue is that the public / private attribute is not properly handled in case there are multiple 'declarations' on 1 line.

albert-github:
I've just pushed a proposed patch, pull request #7199 that addresses the problem as shown in this proposed pull request, but also handles:
```
public log_handle_type, operator (.eq.), log_create, push_default_verbosity
```

As commented before I think the pull request 7192 just hit a side effect, and I think the pull request #7192 can be closed.

ellio167:
Closing in favor of #7199

Issue #7190, opened 2019-08-13
1.8.16: Blank FILE_PATTERNS => no files processed


balducci:
hello

apologies if I am missing some trivial point here

looks like there is a small problem/inconsistency with recently
released doxygen-1.8.16

The problem relates to the FILE_PATTERNS tag
Both in the manual and in the comment inside a generated configuration
file it is said:

    ----8<----
    # If left blank the following patterns are tested:*.c, *.cc,
    # *.cxx, *.cpp, *.c++, *.java, *.ii, *.ixx, *.ipp, *.i++,
    # *.inl, *.idl, *.ddl, *.odl, *.h, *.hh, *.hxx, *.hpp,
    # *.h++, *.cs, *.d, *.php, *.php4, *.php5, *.phtml, *.inc,
    # *.m, *.markdown, *.md, *.mm, *.dox, *.py, *.pyw, *.f90,
    # *.f95, *.f03, *.f08, *.f, *.for, *.tcl, *.vhd, *.vhdl,
    # *.ucf, *.qsf and *.ice.
    ---->8----

and, actually, 1.8.15 behaves accordingly.

But 1.8.16 behaves differently: AFAICS, if I leave FILE_PATTERNS
blank, as in:

    ----8<----
    FILE_PATTERNS =
    ---->8----

then C sources in a directory specified with INPUT are not
processed. Instead, to have them recognized and processed I have to:

=> either *comment* FILE_PATTERNS, like in:

    ----8<----
    # FILE_PATTERNS =
    ---->8----

=> or explicitly define extensions, like in:

    ----8<----
    FILE_PATTERNS = *.c
    ---->8----

In summary, manual and comment say that a *blank* FILE_PATTERNS should
trigger default glob patterns: this is true for 1.8.15, but not for
1.8.16 (where a blank FILE_PATTERNS causes, apparently, no recognition
whatsoever of any file extension and thus no file processing).

If this is intentional, then manual and comments should be updated;
otherwise, there seems to be a problem with doxygen-1.8.16 code

I enclose a small tar which demonstrates the problem (run the doit
script)

thanks a lot for your valuable work
ciao
-gabriele

albert-github:
You wrote:
> I enclose a small tar which demonstrates the problem (run the doit script)

But I don't see the file.

balducci:
here is the test case:

[demo.tar.gz](https://github.com/doxygen/doxygen/files/3496749/demo.tar.gz)

[...]

PR #7193, opened 2019-08-13
issue #7190 1.8.16: Blank FILE_PATTERNS => no files processed

albert-github:
This is actually a regression on #7095. Due to the fact that the init() routine has moved up, the settings for `FILE_PATTERNS` have already been done and a call to init does not operate on the `FILE_PATTERNS` variable but on the structure from which the `FILE_PATTERNS` was derived. We have should here operate on the `FILE_PATTERNS` directly as well., bu getting the default values out of the underlying structure.


doxygen:
Hmm, I already had a suspicion that #7095 might lead to regression but I didn't spot it... Are we sure FILE_PATTERNS is the only case where there is an issue (at least it seems like to only correction using `->init()`)

PR #7199, opened 2019-08-14
Protection of interfaces and types in Fortran

albert-github:
In the proposed pull request #7192 an attempt was made to get the protection of the types and interfaces correct. This looks like to work in case the types end interfaces are on continuation lines, but as soon as all types and interfaces are in one line this failed (see discussion with #7192).

in this patch it is solved by updating the so called modifiers also for CLASS_SEC (types and interfaces are seen as "classes"), furthermore in case the modifiers are not found a test is done whether or not the module (scope) name is prepended to the entry name as with the modifiers this module name is not present.

### DIFF
--- a/src/fortranscanner.l
+++ b/src/fortranscanner.l
@@ -2245,12 +2245,25 @@ static bool endScope(Entry *scope, bool isGlobalRoot)
     Entry *ce;
     for (;(ce=eli.current());++eli) 
     {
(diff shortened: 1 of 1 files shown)


ellio167:
This appears to work as intended.  Thanks!

yafshar:
Using the provided example by @ellio167, this fix is missing the `operator (.eq.)` interface. (see discussion at #7192)

PR #7201, opened 2019-08-15
Protection of interfaces and types in Fortran.

yafshar:
Minor modification to correct the behavior of the public/private
statement in Fortran (based on the #7199 patch).

Using the provided example by @ellio167, the fix in patch #7199 by @albert-github is missing the operator (.eq.) interface. (see discussion at #7192)

Screen Shot 2019-08-14 at 9 15 39 PM

But the provided minor modification correct this behavior as below:

Screen Shot 2019-08-14 at 9 17 10 PM

### DIFF
--- a/src/fortranscanner.l
+++ b/src/fortranscanner.l
@@ -587,7 +587,13 @@ SCOPENAME ({ID}{BS}"::"{BS})*
 <ModuleBody>private/{BS}(\n|"!")         { defaultProtection = Private;
                                            current->protection = defaultProtection ;
                                          }
+<ModuleBody>private/{BS}.*(\n|"!")       { defaultProtection = Private;
+                                           current->protection = defaultProtection ;
+                                         }
 <ModuleBody>public/{BS}(\n|"!")          { defaultProtection = Public;
+                                           current->protection = defaultProtection ;
+                                         }
+<ModuleBody>public/{BS}.*(\n|"!")        { defaultProtection = Public;
                                            current->protection = defaultProtection ;
                                          }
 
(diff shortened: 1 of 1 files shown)


albert-github:
Indeed the `operator(.eq.)` is not yet handled.

In my opinion this pull request is incorrect, as it resets the `defaultProtection` and this is the global protection. In case we have a public or private statement with 'definition(s)' this protection is only for the 'definition(s)' on the mentioned 'line'.
It is quite easily illustrated by adding the code:
```
  !> This is a private variable
  integer :: priv_int
```
e.g. after the `type log_handle_type`. This variable is a private variable but shows up as a public variable.

A small complete program (explicitly writing `priv_var` as the compiler will give an error on it):
```
!> the module
module test
  implicit none
  ! global setting of protection
  private
  ! setting pub_var to public
  public pub_var
  ! priv_var is private due to the global setting
  integer priv_var
  ! pub_var is public due to the previous setting
  integer pub_var
end module
program tst
  use test
  implicit none
  write(*,*) pub_var
  write(*,*) priv_var
end program
```

Seen my remark regarding the scope of the setting of `public` and `private`, just the removing of 
`defaultProtection = Public;`
and thus setting
`current->protection = Public;`
is not good either as it will not handle the real problem.

yafshar:
@albert-github  Thanks for bringing this point about `public\private statement` I would look into that.

yafshar:
@albert-github checking the reported problem, I also noticed the current solution for global protection is wrong as well. It only works if any other variables do not follow the `private` or `public` statement. 
As a simple test, you can do the traditional `cout` and see the results in the below test case:
`
!> the module
module test
  implicit none
  ! global setting of protection
  private priv_var
  ! setting pub_var to public
  public pub_var
  ! priv_var is private due to the global setting
  integer priv_var
  ! pub_var is public due to the previous setting
  integer pub_var
end module
program tst
  use test
  implicit none
  write(*,*) pub_var
  write(*,*) priv_var
end program
`

albert-github:
@yafshar I'm sorry but I don't understand your comment.
- what is the traditional `cout`?
- by default the global protection is `public`, when using the `private` statement without variables behind it it is a global default setting for all variables (etc.) in the module (doxygen has a small problem here as it only sees the global setting for the variables defined after the `private` statement. In case doxygen should work correctly either a stack of variables with global settings should be used so it can be corrected afterwards or doxygen has to scan twice, first to determine whether the global setting `private` or `public` is used and use this as default setting). In the example `piv_var` is explicitely `private` and `pub_var` is explicitly `public`. I don't get the point you want to make.

yafshar:
> @yafshar I'm sorry but I don't understand your comment.
> 
>     * what is the traditional `cout`?
> 
>     * by default the global protection is `public`, when using the `private` statement without variables behind it it is a global default setting for all variables (etc.) in the module (doxygen has a small problem here as it only sees the global setting for the variables defined after the `private` statement. In case doxygen should work correctly either a stack of variables with global settings should be used so it can be corrected afterwards or doxygen has to scan twice, first to determine whether the global setting `private` or `public` is used and use this as default setting). In the example `piv_var` is explicitely `private` and `pub_var` is explicitly `public`. I don't get the point you want to make.

Sorry for the confusion, your explanation made it clears to me. By `cout` I mean `std::cout` or any other print statements.

Issue #7200, opened 2019-08-14
Fortran warning: type not declared or defined


ellio167:
[fortran-type-undefined-warn-example.f90.txt](https://github.com/doxygen/doxygen/files/3503037/fortran-type-undefined-warn-example.f90.txt)

With the attached file, doxygen gives the following warnings:

```
Computing class relations...
Add enum values to enums...
Searching for member function documentation...
/.../f-dox/fortran-type-undefined-warn-example.f90:19: warning: documented symbol 'log_module::log_handle_type::type' was not declared or defined.
/.../f-dox/fortran-type-undefined-warn-example.f90:30: warning: documented symbol 'log_module::type' was not declared or defined.
Creating members for template instances...
```

Otherwise the generated content appears to be correct.

albert-github:
When I use this file I only get the message about line 30, type is not a variable so it is clear that it should not be present, Strange is that line 24 doesn't throw an error.
Looks like the bind statements cause the problem. In the output (I removed the private / public statements as they are not relevant for this issue), I see:

> type(log_handle_type), dimension(c, name="log_handle_two"), save, protected  log_module::bind
>    og_handle_one  More...



So please always:
- attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent. (In this issue the source is present but not the Doxyfile)
- Please also specify the doxygen version used. (At the moment this is probably the 1.8.16, but are there any changes already added).

albert-github:
I've just pushed a proposed patch, pull request #7202

ellio167:
My apologies for forgetting the Doxyfile and version used.  For completeness, here are the files and details:

[fortran-type-undefined-warn-example.zip](https://github.com/doxygen/doxygen/files/3505755/fortran-type-undefined-warn-example.zip)

version is 1.8.16 (via homebrew on macOS 10.12.6).

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.17.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7202, opened 2019-08-15
issue #7200 Fortran warning: type not declared or defined

albert-github:
problem in the statement:
```
  type(log_handle_type), protected, save, &
     bind(c, name="LOG_HANDLE_one") &
     :: log_handle_one
```
is the fact that the "bind" attribute is not handled.

- added handling of the bind attribute
- extending the bind definition also with the single quote version.


ellio167:
I confirm that this PR seems to resolve the warning associated with the identified issue.

PR #7204, opened 2019-08-16
Problem with fenced code block followed by horizontal ruler

albert-github:
When we have a program containing:
````
# First

# Second
Third
```
Fourth

```
---
````
then the end of the fenced code block combined with the horizonal ruler is not handled properly.
Adding an extra newline solves the problem (workaround for a user is also to add an extra newline after the close of the fenced code block.


albert-github:
The same problem was fixed by means of #7395 (so I fixed it twice 😞 )

PR #7207, opened 2019-08-18
Ignore non doxygen C-type comment

albert-github:
Only take doxygen C-type comment in consideration.

(also mad some warnings more readable / consistent).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3513088/example.tar.gz)

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -199,7 +199,7 @@ static void endCondSection()
 {
   if (g_condStack.isEmpty())
   {
-    warn(g_fileName,g_lineNr,"Found \\endcond command without matching \\cond");
+    warn(g_fileName,g_lineNr,"Found '\\endcond' command without matching '\\cond'");
     g_skip=FALSE;
   }
   else
@@ -389,7 +389,7 @@ void replaceComment(int offset);
 				     g_readLineCtx=YY_START;
 				     BEGIN(ReadLine);
                                    }
-<Scan>"//"/.*\n	                   { /* one line C++ comment */ 
+<Scan>"//[!\/]"/.*\n	                   { /* one line C++ comment */ 
 				     g_inSpecialComment=yytext[2]=='/' || yytext[2]=='!';
   				     copyToOutput(yytext,(int)yyleng); 
 				     g_readLineCtx=YY_START;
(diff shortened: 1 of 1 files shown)


doxygen:
> Only take doxygen C-type comment in consideration.

I don't think this is a good idea.. 
It now keeps in Scan state when inside a normal C/C++ comment... See what happens if we encounter, e.g.: 
```c
/* Normal C comment with a lonely quote
"
*/


/// Text
/// @cond
/// Conditional text
/// @endcode

class Test { };

```

The whole file is now processed while looking for the matching end quote, failing to process the conditional section.

albert-github:
I partly agree with the comment
- I think it is still a good idea to skip non doxygen C-type comment unconditionally, the problem the the double quote `"` gives a problem indicates to me that the handling of non doxygen C-type comment has to be improved
- I think it is not good that `@cond` is recognized inside non doxygen C-type comment

When looking at the code (currently line 392 of commentcnv.l) we see:
```
<Scan>"//"/.*\n                    { /* one line C++ comment */
                                     g_inSpecialComment=yytext[2]=='/' || yytext[2]=='!';
```

here actually the element `yytext[2]` is accessed although this will never filled with a character other than the null character as the pattern for `yytext` will always be `//`.

doxygen:
@albert-github I agree that special commands like `@cond` should not be handled in normal C/C++ comments as they are done now, and also that the line you point out is wrong. So I think in the context of `<Scan>` we should process normal C/C++ comments explicitly and differently from doxygen style C/C++ comments.

albert-github:
When I understood the comment correctly I think this is in the following way:
- rename `CComment` to `DoxyComment` (`CComment` does not cover the contents anymore as also e.g. Fortran, Python, use this state)
- introduce state `CComment` for real C-type comment
- extend all `<DoxyComment>` with `CComment` except (current) rules like:
  - `<CComment,ReadLine>[\\@]"cond"/[^a-z_A-Z0-9]    { // conditional section`
  - `<CComment,ReadLine>[\\@]"endcond"/[^a-z_A-Z0-9] { // end of conditional section`
  - `<CComment,ReadLine>[\\@]"cond"[ \t\r]*/\n`
- in case a `BEGIN(DoxyComment)` present see by means of an if statement whether it is a doxygen comment type otherwise use `BEGIN(CComment)`.
 Pay special attention to the (current) rules:
  - `<Scan>"//"/.*\n                           { /* one line C++ comment */ `
  - `<Scan>"/*"[*!]?                           { /* start of a C comment */`
- Critically look at the (current) rules like:
  - `<CComment,ReadLine>[\\@]("dot"|"code"|"msc"|"startuml")/[^a-z_A-Z0-9] { /* start of a verbatim block */`
  - `<CComment,ReadLine>[\\@]("f$"|"f["|"f{") {`
  - `<CComment,ReadLine>[\\@]("verbatim"|"latexonly"|"htmlonly"|"xmlonly"|"docbookonly"|"rtfonly"|"manonly")/[^a-z_A-Z0-9] { /* start of a verbatim block */`
- ...
- testing

PR #7215, opened 2019-08-21
Add '<thead>', '<tbody>' and '<tfoot>' HTML tags as dummies

albert-github:
Add '<thead>', '<tbody>' and '<tfoot>' HTML input tags but ignore them for the time being.
In case we have a table like:
```
<table>
  <thead>
  <tr>
  <th>Col 1</th>
  <th>Col 2</th>
  </tr>
  </thead> <tbody>
  <tr>
  <td>Row 1 Col 1</td>
  <td>Row 1 Col 1</td>
  </tr>
  </tbody>
</table>
```
the output would be garbled, now we get a table.


hedgepigdaniel:
I would love to see this merged - it's pretty confusing that Doxygen fails to parse tables in this case, since `<thead />` and `<tbody />` are common and widespread.

Maintainers: if I fixed the conflicts, would you merge?

PR #7230, opened 2019-08-27
Create possibility to define LaTeX commands for formulas

albert-github:
To be able to have building bocks for formulas one can create a `\newcommand` (or when one wants to change a command `\renewcommand`).
Due to the different handling of LaTeX commands in pure LaTeX code (latex output and formulas converted to images) and MathJax it is necessary to transform LaTeX commands to the MathJax equivalent.
This is done in a transparent way by providing the new commands in a file and add this verbatim to the pure LaTeX code and to translate the `\newcommand` and `\renewcomamnd` to MathJax macros.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3546817/example.tar.gz)


jonathanschilling:
Thanks for merging this :)

Issue #7217, opened 2019-08-21
[Patch] Fortran parser shortcomings for "end" keyword combinations.


dennisyoder:
[bug_fortran_end.zip](https://github.com/doxygen/doxygen/files/3527291/bug_fortran_end.zip)

Doxygen Version:  1.8.15
OS:  Red Hat Enterprise Linux Workstation release 7.3 (Maipo)

Details:

Fortran assignment statements for variable names beginning with "end" 
are seen as the end of the (program|subroutine|function|...) block.  
For example:

    endval = 1
    endsubroutine = 1
    endprogram(1) = 1
    endstruct%val = 1

Within the html source browser, these variables have incorrect keyword
syntax highlighting and subroutine calls appearing after them are not
decorated properly.  Presumably because the scoping unit has closed.

This seems to be due to the expression in fortrancode.l line 937
that is intended to parse:
    "^end [function|subroutine|module|program|...] [name]"
and the way it handles optional keywords and whitespace without any
look-ahead.

Proper Fortran 2008 syntax for the items in line 937 is:

    end [block data [name]]
    end [function [name]]
    end [subroutine [name]]
    end [module [name]]
    end [program [name]]
    end enum
    end type [name]
    end interface [name]

with whitespace optional.

Other Fortran 2008 "end" keywords that are part of the FLOW list and
will fail in the rare use cases shown:

    end do [name]   ! vs:  enddo = 1 assignment
    end if [name]   ! vs:  endif = 1 assignment

Also, the "end" FLOW parser construct is too loose and will make
end keyword combinations from variable assignments, such as:

    endelse = 1     ! "endelse" shown as keyword

This is due to the expression in fortrancode.l line 763.

Other Fortran 2008 keywords that might also cause problems:

    end associate [name]
    end block [name]
    end critical [name]
    end forall [name]
    end select [name]
    end where [name]
    end [submodule [name]]
    endfile unit_num

These would also likely be detected as an "end" keyword and trigger
an end of scope.

Attached is an example highlighting the problems and a possible patch.

It uses a more restrictive list of ENDFLOW keywords, making sure to
avoid (scalar, array, and structure) variable assignments.  It also
modifies the "end" (function|subroutine|module|...) keyword-matching
to limit false positives.

Dennis

dennisyoder:
[bug_fortran_end_1.8.16.zip](https://github.com/doxygen/doxygen/files/3528009/bug_fortran_end_1.8.16.zip)

[...]

PR #7232, opened 2019-08-27
Feature/issue 7217

dennisyoder:
Fix for issue #7217 [Patch] Fortran parser shortcomings for "end" keyword combinations. 

Fixes issues with variable names in assignment statements being detected as Fortran keywords.
Modifies the Fortran lexer to:
- be more restrictive in detecting end subprog keywords,
- be more restrictive in detecting end flow keywords,
- capture variable names on left-hand-side of assignment statements, and
- preserve whitespace between function call and opening parenthesis.

### DIFF
--- a/src/fortrancode.l
+++ b/src/fortrancode.l
@@ -807,14 +807,19 @@ LANGUAGE_BIND_SPEC BIND{BS}"("{BS}C{BS}(,{BS}NAME{BS}"="{BS}"\""(.*)"\""{BS})?")
                                           generateLink(*g_code, yytext);
 					  g_insideBody=FALSE;
 
-					  /* append module name to use dict */
-                                          useEntry = new UseEntry();
-					  //useEntry->module = yytext;
-                                          //useMembers->append(yytext, useEntry);
-					  //addUse(yytext);
-					  useEntry->module = tmp;
-                                          useMembers->append(tmp, useEntry);
-					  addUse(tmp);
+					  /* get module entry from use dict */
+					  /* else append module name to use dict */
+					  useEntry = useMembers->find(tmp);
+					  if (! useEntry) 
+					  {
+                                            useEntry = new UseEntry();
+					    //useEntry->module = yytext;
+                                            //useMembers->append(yytext, useEntry);
+					    //addUse(yytext);
+					    useEntry->module = tmp;
+                                            useMembers->append(tmp, useEntry);
+					    addUse(tmp);
+					  }
                                         }           
 <Use,UseOnly,Import>{BS},{BS}           { codifyLines(yytext); }
 <UseOnly,Import>{BS}&{BS}"\n"           { codifyLines(yytext);


doxygen:
@albert-github Can you have a look at the proposed patch? My Fortran knowledge is pretty limited so I don't want to pull in this request without your approval.

albert-github:
It was on my list already, I'll have a look in the next days.
(There are a few Fortran pull requests to be looked at).

src/fortrancode.l
```diff
@@ -732,6 +733,54 @@ LANGUAGE_BIND_SPEC BIND{BS}"("{BS}C{BS}(,{BS}NAME{BS}"="{BS}"\""(.*)"\""{BS})?")
...
+                                            if ( strlen(tmp) != strlen(yytext) ) {
+                                              // everything else is thrown back into input stream
+                                              // to prevent {BS} from being gobbled
+                                              //printf("     reprocess |%s|\n",yytext+strlen(tmp));
+                                              yyless(strlen(tmp));
+                                            }
+                                            free(tmp);
+					    g_insideBody=FALSE;
+                                        }
+
```
albert-github: Why not put the `{BS}` part after the `/` so the push back is not necessary
Why in the strtok an check for `(`, cannot be this rule

src/fortrancode.l
```diff
@@ -732,6 +733,54 @@ LANGUAGE_BIND_SPEC BIND{BS}"("{BS}C{BS}(,{BS}NAME{BS}"="{BS}"\""(.*)"\""{BS})?")
...
+                                            strcpy(tmp,yytext);
+                                            strtok(tmp," \t(");
+                                            //printf("     generateLink |%s|\n",tmp);
+                                            generateLink(*g_code, tmp);
+                                            // everything else is thrown back into input stream
+                                            //printf("     reprocess |%s|\n",yytext+strlen(tmp));
+                                            yyless(strlen(tmp));
+                                            free(tmp);
+					    g_insideBody=FALSE;
+                                        }
```
albert-github: Why not end the "search at the `(` like with a scalar variable so that the indices can be handled (and linked when necessary) separately
How is currently a multiline array handled, i.e. when we have a continuation line in the indices of the array?

src/fortrancode.l
```diff
@@ -938,25 +987,37 @@ LANGUAGE_BIND_SPEC BIND{BS}"("{BS}C{BS}(,{BS}NAME{BS}"="{BS}"\""(.*)"\""{BS})?")
                                           yy_pop_state();
                                           YY_FTN_RESET
                                         }
-<Start>"end"{BS}("block"{BS}"data"|{SUBPROG}|"module"|"program"|"enum"|"type"|"interface")?{BS}     {  // Fortran subroutine or function ends
-                                          //cout << "===> end function " << yytext << endl;
+<Start>^{BS}"end"{BS}/(\n|!|;)          { // bare end keyword.
+                                          //cout << "===> bare end keyword " << yytext << endl;
+                                          //printf("===> line number %d bare end keyword |%s|\n",g_yyLineNr,yytext);
```
albert-github: Why is here a begin of line character (`^`)?

src/fortrancode.l
```diff
@@ -938,25 +987,37 @@ LANGUAGE_BIND_SPEC BIND{BS}"("{BS}C{BS}(,{BS}NAME{BS}"="{BS}"\""(.*)"\""{BS})?")
...
                                           endScope();
    					  startFontClass("keyword");
   					  codifyLines(yytext);
 					  endFontClass();
+                                          yy_push_state(YY_START);
+                                          BEGIN(Subprogend);
+                                        }
+<Subprogend>{ID}/{BS}(\n|!|;)             {
+					  generateLink(*g_code,yytext);
+                                          yy_pop_state();
```
albert-github: I don't see the advantage of the new construct in splitting everything in different parts compared to the old construct.

src/fortrancode.l
```diff
@@ -1101,7 +1162,19 @@ LANGUAGE_BIND_SPEC BIND{BS}"("{BS}C{BS}(,{BS}NAME{BS}"="{BS}"\""(.*)"\""{BS})?")
...
+                                            strtok(tmp," \t");
+                                            //printf("===> line number %d function call |%s|\n",g_yyLineNr,yytext);
+                                            generateLink(*g_code, tmp);
+                                            if ( strlen(tmp) != strlen(yytext) ) {
+                                              // everything else is thrown back into input stream
+                                              // to prevent {BS} from being gobbled
+                                              //printf("     reprocess |%s|\n",yytext+strlen(tmp));
+                                              yyless(strlen(tmp));
+                                            }
+                                            free(tmp);
```
albert-github: Why not move the `{BS}` after the `/` so no push back is necessary.

src/fortrancode.l
dennisyoder: On line 739, {BS} after / would be nice, but generates "dangerous trailing context" compiler warning because {ID} contains a variable length wildcard (*).  Please suggest a simpler alternative that works.

On line 745, strtok is intended to capture {BS} and should NOT contain "(".  It is a remnant from a previous iteration that tried to capture array assignment within a common pattern match.  I will update.

src/fortrancode.l
dennisyoder: On line 768, ending the search at ( would not discriminate from certain fortran keywords, like "real(int_expression)" or "case(value)".  Placing / immediately after {ID} and treating the rest as trailing context generates the "dangerous trailing context" compiler warning.  So, the proposed context captures everything to the "=" and then uses the "(" in strtok on line 775 to push the indices back onto the stack so that links can be generated.

This construct will not capture array assignment if line continuation occurs before the = operator.  We rely upon the simple variable reference lexer of line 1281 to catch those.

src/fortrancode.l
dennisyoder: The ^ character should be removed.  "end" is likely at the beginning of a logical line, but could be preceeded by a line label.

src/fortrancode.l
dennisyoder: {ID} should not be allowed to follow a bare "end" keyword.  In hindsight, I have updated the PR to fix that with minimal changes to the original construct and without additional splitting.

src/fortrancode.l
dennisyoder: On line 1165, {BS} after / would be nice, but generates "dangerous trailing context" compiler warning because {ID} contains a variable length wildcard (*).  Please suggest a simpler alternative.

albert-github: I was thinking about directly "codifying" the spaces and not pushing them back. To be able to give a good suggestion do you have a small example where this benefit shows?

albert-github:
Running the code before your change (html_pre) and after your change (html_fix) on the following example (removed everything that was not really necessary):
```
      SUBROUTINE sub_1
          reg: DO ib = 1,10
              I_ClassicIndex_rad(ner_rad(idx)) = ib
              CYCLE reg
          END DO reg
      END SUBROUTINE sub_1
```
with in the `Doxyfile` only `SOURCE_BROWSER=YES` set.
I got the following differences in the output:
```
diff -r html_fix/aa_8f_source.html html_pre/aa_8f_source.html
70,71c70,71
< <div class="line"><a name="l00003"></a><span class="lineno">    3</span>&#160;              i_classicindex_rad(ner_rad(idx))<span class="comment"> = ib</span></div>
< <div class="line"><a name="l00004"></a><span class="lineno">    4</span>&#160;<span class="comment"></span>              cycle reg</div>
---
> <div class="line"><a name="l00003"></a><span class="lineno">    3</span>&#160;              i_classicindex_rad(ner_rad(idx)) = ib</div>
> <div class="line"><a name="l00004"></a><span class="lineno">    4</span>&#160;              cycle reg</div>
```
We see that:
- in the fixed version (top) on the first line the text ` = ib` is marked as comment, this is not the case.
- in the fixed version there is a "dummy" `comment span`, this is not a big problem but delays the rendering.


The example: [example_1.tar.gz](https://github.com/doxygen/doxygen/files/3718154/example_1.tar.gz)

dennisyoder:
Updates through c73d49a.  Resolved the issues identified in example_1.tar.gz, namely must adjust the fortran column counter when using yyless.  Some additional improvements to variable and array detection.  Also enabled multiple "use only" lines to append rather than replace the module dictionary entry.

For my f77/f90 test code, please see the example: [example_2.tar.gz](https://github.com/doxygen/doxygen/files/3990433/example_2.tar.gz)

AppVeyor build check failed on bibtex, but don't see how that is related to my changes.  Is there a server issue?

src/fortrancode.l
dennisyoder: One example would be preserving the whitespace alignment of function names and args across multiple source lines, as in the main program below.

```
PROGRAM main
  n = fun1  (1)  ! call fun1
  n = fun100(1)  ! call fun100
CONTAINS
  FUNCTION fun1(n)
    fun1 =  abs(n)
  END FUNCTION fun1
  FUNCTION fun100(n)
    fun100 =  abs(n)*2
  END FUNCTION fun100
END PROGRAM main
```

PR #7235, opened 2019-08-28
Consistent way to show scanner state

albert-github:
Create a consistent way to display the state mnemonics of the different scanners (analogous to the fortranscanner.l)


doxygen:
I like the idea of being able to print the state, but not that one has to duplicate all state definitions. It's strange that flex doesn't offer this, but maybe a nicer solution would be to create a small python script to extract the `%x State` lines and produce a `_states.h` file from a `.l` file and define the state to string mapping...

albert-github:
Interesting thought, it is indeed strange that (f)lex doesn't have by default this possibility. It is indeed not nice to copy the state definitions. I'll have a look into the possibilities.
Note besides the `%x` there are also `%s` that have to be handled.

doxygen:
@albert-github Since all rules already have a [start condition](http://ftp.gnu.org/old-gnu/Manuals/flex-2.5.4/html_mono/flex.html#SEC11), `%s` has the same effect as `%x` so we could also use `%x` everywhere.

albert-github:
Superseded by #7246.

PR #7237, opened 2019-08-29
Problem with  `<hr>` in LaTeX multicolumn cell

albert-github:
In case we have a `<hr>` in a table cell then it is displayed as a horizontal line (in LaTeX) unless it is in a multicolumn cell. In the later case it throws an error:
```
! You can't use `\hrule' here except with leaders.
\DoxyHorRuler ... }{0ex plus 0ex minus 0ex}\hrule
```

We have to use use `\hrulefill` in a table but in combination with the table option of xcolor this gives some problems so a new command has to be defined for this.

See also:
- horizontal rule / line inside a multicolumn field (https://tex.stackexchange.com/questions/506211/horizontal-rule-line-inside-a-multicolumn-field)
- interaction between \usepackage[table]{xcolor} and \hrulefill (https://tex.stackexchange.com/questions/506221/interaction-between-usepackagetablexcolor-and-hrulefill)

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3556098/example.tar.gz)

### DIFF
--- a/doc/doxygen_manual.tex
+++ b/doc/doxygen_manual.tex
@@ -66,6 +66,11 @@
 \usepackage{doxygen}
 \usepackage{manual}
 %%
+%gave problems when in doxygen.sty
+\makeatletter
+\newcommand\hrulefilll{\leavevmode\leaders\hrule\hskip 0pt plus 1filll\kern\z@}
+\makeatother
+%%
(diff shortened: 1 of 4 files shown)


albert-github:
Looks like there is a problem with the hyperlinks, investigating ...

PR #7245, opened 2019-09-05
HTML start comment with 3 dashes

albert-github:
A normal HTML comment `<!--` has 2 `-` signs but it is not prohibited to have 3, but doxygen translates `<!---` into `<!-&ndash;` and thus the comment is not recognized.
By checking and consequently handling, doing the right skipping, also the 3 `-` sign version the problem can be solved.
An end comment cannot contain 3 `-` signs, so here no changes have to take place.
The version with 3 or more `-` are in a start HTML comment does not give a problem with xmllint either.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3578532/example.tar.gz)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -480,7 +480,8 @@ static int processNmdash(GrowBuf &out,const char *data,int off,int size)
   {
     count++;
   }
-  if (count==2 && off>=2 && qstrncmp(data-2,"<!",2)==0) return 0; // start HTML comment
+  if ((count==2 || count==3)&& off>=2 && qstrncmp(data-2,"<!",2)==0) return 2-count; // start HTML comment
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I've tried to generalize the code a bit, see [this commit](https://github.com/doxygen/doxygen/commit/0f90a399a886ea7f69bf482415e6c4cdcb7269b3). 

I used the following example to test
```c
/** Some text
 *  <!- normal text ->
 *  <!-- a comment in html with 2 dashes -->
 *  <!--- a comment in html with 3 dashes -->
 *  <!---- a comment in html with 4 dashes -->
 *  <!----- a comment in html with 5 dashes -->
 *  <!------ a comment in html with 6 dashes -->
 *  <!------- a comment in html with 7 dashes -->
 *  More text
 */
class T {};
```

please double check if I didn't miss anything.

albert-github:
Looks OK to me.

Issue #7248, opened 2019-09-08
Including external tag files with TOC produces a broken index.qhp


antonio-rojas:
In some circumstances, when including an external tag file with TOC information from a md file, doxygen 1.8.16 produces a broken index.qhp. Minimal test case attached, just run `doxygen test.config` and notice how the generated index.qhp has an invalid `</section>` in line 13.

[test-doxy.zip](https://github.com/doxygen/doxygen/files/3587687/test-doxy.zip)

albert-github:
Complex situation.
All paragraphs are automatically given a label when they are translated from e.g. `## Foo 2` to `section autotoc_md1 Foo 2` and these labels are the same in the original project (from which the tag file comes) and the project where the tag file is used.
Doxygen issues a warning in this respect:
```
test.md:5: warning: multiple use of section label 'autotoc_md1' while adding section, (first occurrence: a)
```
This is not nice as the user has no influence on this automatic tag label other that to define for each paragraph its own unique label (e.g. `## Foo 2 ## {#labelid2}`.

This is not a good situation and it would be better to exclude the automatic generated tags as doxygen did add them and they are only used for non user defined references in the current project (and at least I don't know a way how to reference them by means of a tag file in another project).


The problem will also occur when the user defines its own label with the same name in the different projects. Here the user has to create an unique name.

A bit a more extensive example. First run doxygen in p1 and afterwards in p2: [example.tar.gz](https://github.com/doxygen/doxygen/files/3590200/example.tar.gz)
.

albert-github:
I've just pushed a proposed patch, pull request #7250 in respect to the automatic generated tags.

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.17.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7250, opened 2019-09-09
issue #7248: Including external tag files with TOC produces a broken index.qhp

albert-github:
Ignore automatically generated anchor names when importing a tag file.

### DIFF
--- a/src/tagreader.cpp
+++ b/src/tagreader.cpp
@@ -494,6 +494,23 @@ class TagFileParser : public QXmlDefaultHandler
 
     void endDocAnchor()
     {
+      // Check whether or not the tag is automatically generate, in that case ignore the tag.
+      switch(m_state)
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github Isn't it better to (also) fix the problem at the source? i.e. don't write auto generated labels to the tag file in the first place.

albert-github:
Probably it would be good to fix the source of the problem as well. . i.e in `definition.cpp` in the function `void DefinitionImpl::writeDocAnchorsToTagFile` where we should tets whether or not `si->label` starts with `autotoc_md` so something like:
```
if (!si->generated && si->ref.isEmpty()i && !si->label.startsWith("autotoc_md"))
```

We would still need this fix for backward compatibility for external tag references.

Feel free to fix it as such.

Issue #6404, opened 2018-07-22
baseline test corpus?


abathur:
I'm curious about locating/soliciting/building a small baseline test corpus for Doxygen.

Context:
- I'm upgrading Doxygen's experimental sqlite3 generator. 
- Alongside it, I'm building a Python module for using the generated database.
- I'm working on tests for this Python module that validate against the XML output.
- For now at least, my tests focus on assessing parity with the XML output by exposing missing memberdef/compounddef types and inter-element relationships (inner/outer, xrefs, reimplementation, etc.)
- I'm currently using the full 'examples' directory as my input, but it looks like this only covers 6 memberdef types and 8 compounddef types.

I'm not certain what this _should_ look like because I can imagine a canonical test corpus being useful beyond my fairly limited needs (a test corpus that just enumerates the main code entities/relations). I guess this could be in a new directory, be an expansion of the examples directory, or perhaps just a config that selects a specific set of inputs from `examples/` and/or `testing/`?

mosra:
Good idea!

I have a fairly extensive set of test cases in the [m.css Doxygen theme](http://mcss.mosra.cz/doxygen/) where I need to verify that most of Doxygen features are correctly parsed from the XML output. A side effect of that is that when something in Doxygen breaks, my tests spot it. But this doesn't cover everything, especially complex things like whitespace removal (#6299, would be nice to have this fixed) and link parsing/generation definitely need a much more extensive set of tests to avoid regressions.

Besides that, the [Magnum engine](http://doc.magnum.graphics) has a very large Doxygen-powered documentation that exercises many code paths in Doxygen (well, except those where I found problems/bugs and had to work around them). I think a good idea could be to have this (and other projects) as a "test suite" -- regularly generating documentation from them and spotting regressions (e.g. the "pass condition" could be having no warning output from Doxygen, and errors/warning/crashes would be treated as regressions). I'm trying to follow Doxygen development as close as possible and report new regressions, but at the moment (again due to #6299) I have to work with a fork to have warning-free output.

If I can do anything to make Doxygen better tested for regressions, please let me know.

mosra:
I meant using existing codebases just for "soak testing" (or what's the correct term) -- basically ensuring that current huge projects heavily using Doxygen are not broken when a new Doxygen version gets released. These will cover what custom minimal hand-written unit tests can't -- because their code evolves, their Doxygen usage also evolves and might hit new corner cases that weren't thought about before.

[...]

PR #7256, opened 2019-09-11
Sqlite3 tests

groleo:
This PR adds minimal sqlite3 support for testing, (uses the same steps done for xml)
for each test, a reference json file (the 'check:' file)
is compared against a json generated from doxygen_sqlite3.db.


groleo:
because you want to test sqlite3 only when support for it is being built ?
I don't see what use case you have in mind.

albert-github:
I cannot test with sqlite3 (not, yet, installed) but as far as I can see now '--sqlite3' will always be used in case a test is started and sqlite3 is enabled even when only normal tests are to be executed (like --xml or --xhtml or --pdf ...).

albert-github:
I now made a small build with sqlite3 enabled and did some tests.
The output from `use_sqlite3` is always outputted as there is e.g. no `GENERATE_SQLITE3` output flag, a bit contrary to the other output formats of doxygen.

Some findings / thoughts:
- when no options are given the xml and sqlite3 are tested
  - I assume that the sqlite3 test results will be provided for the tests 2..86 as well
  - a update possibility for sqlite3 will also be added?
- when running: `make tests TEST_FLAGS="--keep --xhtml --sqlite3 --id=1"`
  only the xhtml test is done as the 'sqlite3' test is blocked by `if 'check' in self.config and self.args.xml`
  when I enable `--xml` the `sqlite3` test is done but this is independent of the setting in TEST_FLAGS.
  (as an example of user / tester responsibility: For the clang parsing we have also the possibility to set the `CLANG_ASSISTED_PARSING=YES` by means of the test flag `--clang` and it is the responsibility of the user to use this `--clang` only when the executable is compiled with clang possibilities).

groleo:
the cmake change has been dropped
the json dumper now dumps the whole db minus meta (contains time fields
that change between runs)
the per-test Doxyfile is now using the OUTPUT_DIRECTORY to make it possible
to have per-test .db.

abathur:
I guess I'm a bit late, but I spotted this after noticing the 1.8.19->1.8.20 bump and searching for more about some of the sqlgen config changes in it.

I wanted to get sqlite3gen under test the last time I touched it, but I was also concerned about the test suite eroding unless sqlite3gen is part of regular builds and its tests are in the regular CI/test path.

Since I wanted to start building something on top of the output databases, my compromise for the time was to add a test module to its test suite testing the sqlite3 db against the XML output (for most, but not quite all, of the examples directory). It probably doesn't add much here, but I figured I should mention it just in case (I'm OK with parting it out if it is): https://github.com/abathur/doxy_db/blob/master/doxy_db/tests/t_xml_equiv.py

groleo:
Hi,

The pull-request you mentioned is a no-go since it is adding yet another
round of files to maintain and keep in sync with the xmls.

What I'm currently trying is to have
xml generated from the db and compare that to the test expected xml file.

But this quickly brings me to the still
unsolved problem of the briefdescription/description field format.
Right now, "description" is storing XML too,
which is wrong. I think sqlite's user-defined functions can help store
hierarchical data
inside "description".

With that being said, enabling sql tests
should only be a change in runtest.py.

abathur:
> Hi, The pull-request you mentioned is a no-go since it is adding yet another round of files to maintain and keep in sync with the xmls. 

I'm a little confused. Are you calling this PR (#7256) a no-go because it adds files to maintain?

> What I'm currently trying is to have xml generated from the db and compare that to the test expected xml file. 

In any case, the test I linked does something similar to what you're describing; it uses a subset of Doxygen's examples directory (not another round of files to maintain), generates the XML, generates the sqlite3 database, and cross-validates what it can using xpath queries on one side and SQL on the other. So, it's mostly trying to compare values, where it sounds like you intend to just re-synthesize the XML and diff them?

> But this quickly brings me to the still unsolved problem of the briefdescription/description field format. Right now, "description" is storing XML too, which is wrong. I think sqlite's user-defined functions can help store hierarchical data inside "description".

Not quite sure why it brings us to this, but a few thoughts:
- There may be a better representation for the descriptions in the database, but I'm reluctant to agree that XML is _wrong_ because I'm not aware of any "right" relational representation that can contain the spectrum of document-oriented XML without discarding information. 
- XML isn't really _right_, here, but the descriptions are already in it, and the Doxygen ecosystem already has tools for working with this XML, so I don't see an obvious win to converting it to a new format unless that format is both native to relational databases and doesn't discard information.
- I think it would be nice to have some convenience options for "simplifying" the description fields (i.e., converting them into plaintext or markdown or html)--those would meet many people's needs--but I'll be skeptical of any default conversion that discards information (AFAIR most of the output formats except for XML do this to some degree...?)
- Since there aren't really any structural/hierarchical guarantees about what the field contains, I'm curious what you have in mind regarding user-defined functions here. Do you have an example?

groleo:
On Thu, Aug 27, 2020 at 8:09 AM Travis A. Everett <notifications@github.com>
wrote:

> Hi, The pull-request you mentioned is a no-go since it is adding yet
> another round of files to maintain and keep in sync with the xmls.
>
> I'm a little confused. Are you calling this PR (#7256
> <https://github.com/doxygen/doxygen/pull/7256>) a no-go because it adds
> files to maintain?
>
Exactly that; unnecessary files, to be more precise. Having both XML and
JSON files for the same purpose is not really ok in my view.


> What I'm currently trying is to have xml generated from the db and compare
> that to the test expected xml file.
>
> In any case, the test I linked does something similar to what you're
> describing; it uses a subset of Doxygen's examples directory (not another
> round of files to maintain), generates the XML, generates the sqlite3
> database, and cross-validates what it can using xpath queries on one side
> and SQL on the other. So, it's mostly trying to compare values, where it
> sounds like you intend to just re-synthesize the XML and diff them?
>
I wouldn't pick the solution you mentioned, the reasons being a) it's
really easy to miss things b) it's a maintenance burden


> But this quickly brings me to the still unsolved problem of the
> briefdescription/description field format. Right now, "description" is
> storing XML too, which is wrong. I think sqlite's user-defined functions
> can help store hierarchical data inside "description".
>
> Not quite sure why it brings us to this, but a few thoughts:
>
Here's why. If you generate xml for, say compounddef table, it would be
something like this:
xml['@id'] = get_refid ...
xml['title'] = record['title']
xml['detaileddescription'] = record['detaileddescription']
xmlschema.etree_tostring(xml)

At this point [xmlschema.etree_tostring(xml)] the xml stored in
detaileddescription will
automatically use entities, resulting in a different output than expected.


>
>    - There may be a better representation for the descriptions in the
>    database, but I'm reluctant to agree that XML is *wrong* because I'm
>    not aware of any "right" relational representation that can contain the
>    spectrum of document-oriented XML without discarding information.
>
> what i'm testing now (don't know if it will work) is to have sql
statements (or some sort of format specifiers) inside description,
something like this:
"<para>test</para> something <ref>else</ref>" would become "para(test)
something ref(else)", with para(),refid() being a UDF.

abathur:
This isn't 100% in response to any specific thing you said here, but I had forgotten about it until something in your comment sparked my memory. As I mentioned before, I was hesitant to tailor a corpus that covered exactly the right things over concerns about duplication and rot as Doxygen marches on. At the time I opened an issue (#6404) about compiling an official test corpus that at least _aspires_ to being comprehensive (i.e. converges on having at least one known instance of everything Doxygen handles). 

---

The Doxygen ecosystem has a lot of additional tools/projects, and I suspect having a common reference corpus--versioned alongside Doxygen--to test against would help improve quality/compatibility and reduce duplicate effort.

> Exactly that; unnecessary files, to be more precise. Having both XML and JSON files for the same purpose is not really ok in my view.

Ok! Glad that interpretation is correct. I mis-read it the first time😁 

> I wouldn't pick the solution you mentioned, the reasons being a) it's really easy to miss things b) it's a maintenance burden

I suspect we mostly-agree on this. Full XML re-synthesis is a good test, and having a full XML parity test as a blocking check in the regular Doxygen test suite would be a big win for ensuring sqlite3gen stays stable. I have two concerns: 
1. sqlite3gen covers most of the essentials, but there is still a bit of work left to reach the near-complete parity I assume we'd need to actually generate identical XML (assuming you just plan to cmp/diff them)? 

    This is part of why I found it productive to generate both sides and then slice/compare with queries. It made it straightforward to prepare to test for things that the sql schema supports (but aren't extracted yet) and conditionally skip those tests on init until at least one of the relevant entities shows up in the database. This means the tests can automatically activate (and start blocking on CI) as soon as I target a Doxygen version that supports the underlying entity (here's [a segment of xml-equivalence test-results ](https://travis-ci.org/github/abathur/doxy_db/builds/721566588#L3054-L3124) showing some of what currently runs/skips).

2. This may be moot if you are already thinking/planning a way for end-users to generate XML from the database, but if this is just for testing I wonder to what extent the sqlite3 suite will have to duplicate a significant fraction of xmlgen's logic and track many of the changes to it going forward?

> At this point [xmlschema.etree_tostring(xml)] the xml stored in detaileddescription will automatically use entities, resulting in a different output than expected.

Ah! Gotcha. I didn't anticipate a round-trip back to XML in the use-cases since the XML output already exists and has decent tooling.

> what i'm testing now (don't know if it will work) is to have sql
> statements (or some sort of format specifiers) inside description, something like this: "<para>test</para> something <ref>else</ref>" would become "para(test) something ref(else)", with para(),refid() being a UDF.

Shooting from the hip since I haven't specifically looked into it before, but I suspect there's some workable form of this? A few thoughts:
- I'm not aware of anything like `eval()` for sqlite3, so I suspect the whole value/column would need to be wrapped in a function call (i.e., `SELECT expand_functions(description)` or `SELECT convert_xml(description)` to interpret the inner syntax anyways?
- I guess conversion to this format would need to escape internal code/markup, so that it doesn't try to over-resolve on the off-chance someone is documenting an entity with the same name? It would be bad to end up with: `<para>If you need a durable permanent identifier to refer to this event later, just call <computeroutput>refid(event)</computeroutput></para>` -> `para(If you need a durable permanent identifier to refer to this event later, just call computeroutput(refid(event)))`, but it might also just be sufficient to protectively namespace the function names?
- I haven't used it yet, but SQLite added an interesting [Generated Columns](https://www.sqlite.org/gencol.html) feature this year and I noticed it has an option for generating the columns values at write time and persisting them. It's obviously a little space-wasteful and may not mean much if you only need this for testing, but it might provide a way to distribute a database that can be read without the UDFs available. (Another downside; databases with generated columns aren't backwards-compatible).

    It looks like it could be as simple as having an option/flag to add a column like `converted_description TEXT GENERATED ALWAYS AS (convert_fmt1_to_fmt2(description)) STORED` to the schema before write would trigger write-time conversion. I'm not certain if clients will require the function to be defined on load, but if not it seems like it should be possible to read the resulting database without needing the conversion functions defined. 

    (It'd probably be ideal to have an option to declare it as VIRTUAL instead, so that projects with really large corpuses can make their own call about whether it's worth provisioning conversion functions to save space?)

groleo:
> I suspect we mostly-agree on this. Full XML re-synthesis is a good test, and having a full XML parity test as a blocking check in the regular Doxygen test suite would be a big win for ensuring sqlite3gen stays stable.
sqlite tests are actually required to mark it non-experimental.
after that, yes, as a bonus, it's easy to see when stuff breaks.

>     1. sqlite3gen covers most of the essentials, but there is still a bit of work left to reach the near-complete parity I assume 
we'd need to actually generate identical XML (assuming you just plan to cmp/diff them)?
yes. there's enum/enumfields to add, performance profiling on big projects


>     2. This may be moot if you are already thinking/planning a way for end-users to generate XML from the database, but if this is just for testing I wonder to what extent the sqlite3 suite will have to duplicate a significant fraction of xmlgen's logic and track many of the changes to it going forward?

This is why I want the sql2xml to be as small as possible (in the current P-R, there's only one class, SqliteJsonDumper)
meaning, if it gets too big then it's a good sign something is not ok, either in the db schema or in the tool itself.

> Shooting from the hip since I haven't specifically looked into it before, but I suspect there's some workable form of this? A few thoughts:
> 
>     * I'm not aware of anything like `eval()` for sqlite3, so I suspect the whole value/column would need to be wrapped in a function call (i.e., `SELECT expand_functions(description)` or `SELECT convert_xml(description)` to interpret the inner syntax anyways?
Yup. Something like that.

>     * I guess conversion to this format would need to escape internal code/markup, so that it doesn't try to over-resolve on the off-chance someone is documenting an entity with the same name? It would be bad to end up with: `<para>If you need a durable permanent identifier to refer to this event later, just call <computeroutput>refid(event)</computeroutput></para>` -> `para(If you need a durable permanent identifier to refer to this event later, just call computeroutput(refid(event)))`, but it might also just be sufficient to protectively namespace the function names?
right.
 
>     * I haven't used it yet, but SQLite added an interesting [Generated Columns](https://www.sqlite.org/gencol.html) feature this year and I noticed it has an option for generating the columns values at write time and persisting them. It's obviously a little space-wasteful and may not mean much if you only need this for testing, but it might provide a way to distribute a database that can be read without the UDFs available. (Another downside; databases with generated columns aren't backwards-compatible).
>       It looks like it could be as simple as having an option/flag to add a column like `converted_description TEXT GENERATED ALWAYS AS (convert_fmt1_to_fmt2(description)) STORED` to the schema before write would trigger write-time conversion. I'm not certain if clients will require the function to be defined on load, but if not it seems like it should be possible to read the resulting database without needing the conversion functions defined.
Yes, they do look interesting but they have a bit of limitations (not being able to use aggregate functions, subqueries etc).
I _don't_ know (needs testing) if UDF are also not permitted in the generated column's expression, but
it's definitely worth looking at.

groleo:
Forgot to mention it here: there yet another path to explore: storing the description fields as json.
sqlite3 3.9 added a bunch of nice-looking json functions which may be fiddled with.

abathur:
> Yes, they do look interesting but they have a bit of limitations (not being able to use aggregate functions, subqueries etc).

This is why I'm also a little uncertain about whether it'd object to a UDF in the column definition. I get the impression that the aggregate/subquery limitation is just on the definition of the generated column, and I don't imagine this conversion case requiring either.

> Forgot to mention it here: there yet another path to explore: storing the description fields as json. sqlite3 3.9 added a bunch of nice-looking json functions which may be fiddled with.

I guess something like jsonml can support a lossless round-trip. I'm not sure if there are meaningful queries into something that array-oriented without foreknowledge of what it contains or baking in assumptions about which nodes will appear where?

PR #7264, opened 2019-09-15
Bug 164073 - There should be a dummy tag that Doxygen ignores

albert-github:
Introduced the `\noop` command to ignore text till end of the line.


Krinkle:
Thanks! This is great. I have not edited the code of Doxygen before, but I would also be interested in seeing a basic demo of it to confirm/test it. Could you give me a hint about how I could contribute that?

albert-github:
You should get the code from github and get the patch. Use cmake to build.

PR #7267, opened 2019-09-16
Fix typo in source

luzpaz:
On the heels of #7266 this commit potentially breaks the API and therefore separated.


albert-github:
I don't think this breaks the API as I didn't find the word "contrained" in the repository, I did find "constrained".  (I compared the places with "maybevoid" and the places look similar).

luzpaz:
That's good news then, right?

PR #7275, opened 2019-09-23
Bug 340202 - @code: static_cast, const_cast, etc C++ keywords

albert-github:
The `*_cast` had already been introduced, but there were some other places that needed the "REJECT" as well.
Also the cast was missing a whitesace between the cast name and the `<`.

Tested also against he doxygen internal documentation.


doxygen:
@albert-github I've reworked the code a bit, see [this commit](https://github.com/doxygen/doxygen/commit/fe66145d44ad4cf79e2bb4928b190c415aaf5b90)

Issue #4465, opened 2018-07-02
Closing comment (*/) inside <param> tag in C# xml comments will break created documentation (Origin: bugzilla #663699)


doxygen:
status NEEDINFO severity *major* in component *general* for *---*
Reported in version *1.7.5.1* on platform *Other*
Assigned to: Dimitri van Heesch

On 2011-11-09 15:41:09 +0000, JK wrote:

    In C# source code the following will prevent the documentation for the
    method to be created:
    
            /// <param name="action">() =&gt; { /* your code here */ }</param>
            public void EnqueueControllerTask(ThreadingManagerTaskAction action)
    
    The problem seems to be the closing comment marker "*/". If removed the same
    class is documented correctly by doxygen.

On 2013-12-26 10:51:09 +0000, albert wrote:

    I tried to reproduce the problem with version 1.7.5.1 but was not able to do
    so.
    Please check if in the latest version the problem still exists and if so
    please add a (small) self-contained example (source+config file in a zip or
    tar) to reproduce the problem.

PR #7280, opened 2019-09-27
Formulas extension to "as is" images.

albert-github:
The idea was born during fixing "Crop png formula environment problem for HTML output (Origin: bugzilla #674005)" (issue #4465, pull request #7251) as the mentioned Tikz picture was shown in the LaTeX  version but not in the HTML version with images (parts converted to white) and in the MathJax version the image does not appear as MathJax doesn't know e.g. the figure environment, see http://docs.mathjax.org/en/latest/input/tex/macros/index.html#environments).
Furthermore formulas like `\f$ \color{yellow} A = \color{blue}{B} \mathbin{\color{red}{-}} \color{green}{C} \f$` were not shown properly as some letters (B) were converted to white characters.
This patch will with the new commands `\fig..` (analogous to the `\f...` commands) create images, but will not convert them to b/w images. The choice for a new set of commands is made as options would interfere with the environment version`\f{` command.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3662696/example.tar.gz)

### DIFF
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -107,6 +107,11 @@ documentation:
 \refitem cmdfbrclose \\f]
 \refitem cmdfcurlyopen \\f{
 \refitem cmdfcurlyclose \\f}
+\refitem cmdfigdollar \\fig\$
+\refitem cmdfigbropen \\fig[
+\refitem cmdfigbrclose \\fig]
+\refitem cmdfigcurlyopen \\fig{
+\refitem cmdfigcurlyclose \\fig}
 \refitem cmdfile \\file
 \refitem cmdfn \\fn
 \refitem cmdheaderfile \\headerfile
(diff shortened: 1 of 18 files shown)


doxygen:
@albert-github I see a lot of code duplication in this pull request (which I don't really like), without seeing the need for two different ways of creating Latex formulas. Is there really a need to have both and how would a user need to choose? And what do you mean with 

> "The choice for a new set of commands is made as options would interfere with the environment version `\f{` command."

Is the old style only needed to "force" black and white pictures?

albert-github:
I don't like code duplication either, but I didn't see another way.

A solution would have been to use the doxygen construct for options (like with e.g. `\snippet` giving  `\snippet{doc}`) this would have been possible for the commands `\f$ ... \f$` and `f[ ... \f]` (so giving `\f${color} ... \f$` and `f[{color} ... \f]` (doesn't look nice though). For `\f{env}{ ... \f}` this would lead to `\f{color}{env}{ ... \f}` or `\f{{color}env}{ ... \f}` or ... all quite confusing in my opinion and not logical.

As far as I know the old form is indeed needed for getting nice crispy formulas (as far as possible) in b/w.

The choice for the user will not always be easy and is dependent on the content of the formula / image to be displayed but as soon as colors or 'difficult' image constructs are used the new commands should be chosen, for "simple" formulas the old commands should be used.

PR #7286, opened 2019-09-30
Bug 335614 - HTML link incorrect when using tagfile

albert-github:
- See to it that when an extension is already present this extension is used and not a second extension is added
- let the tag file know what the original extension was.

### DIFF
--- a/src/definition.cpp
+++ b/src/definition.cpp
@@ -545,7 +545,7 @@ void DefinitionImpl::writeDocAnchorsToTagFile(FTextStream &tagFile) const
       {
         //printf("write an entry!\n");
         if (definitionType()==TypeMember) tagFile << "  ";
-        tagFile << "    <docanchor file=\"" << si->fileName << "\"";
+        tagFile << "    <docanchor file=\"" << si->fileName << (hasExtension(si->fileName) ? "" : Doxygen::htmlFileExtension) << "\"";
         if (!si->title.isEmpty())
         {
           tagFile << " title=\"" << convertToXML(si->title) << "\"";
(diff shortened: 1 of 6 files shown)


src/definition.cpp
```diff
@@ -545,7 +545,7 @@ void DefinitionImpl::writeDocAnchorsToTagFile(FTextStream &tagFile) const
       {
         //printf("write an entry!\n");
         if (definitionType()==TypeMember) tagFile << "  ";
-        tagFile << "    <docanchor file=\"" << si->fileName << "\"";
+        tagFile << "    <docanchor file=\"" << si->fileName << (hasExtension(si->fileName) ? "" : Doxygen::htmlFileExtension) << "\"";
```
doxygen: Wouldn't it be cleaner to instead of adding `hasExtension()`, add a `addHtmlExtension(fileName)` function that returns a adjusted file name?

albert-github: The name `addHtmlExtension` is not completely covering the purpose, the `HTML_FILE_EXTENSION` will be added only in case no extension is present. This is why I chose for the given construct. I think a better routine name for `addHtmlExtension` would be beneficial.

Notwithstanding the above:
You wrote "add a addHtmlExtension(fileName) function that returns a adjusted file name"

Would that be a function like
```
QCString addHtmlExtension(QCString fName)
{
  return ((QFileInfo(fName).extension(FALSE).isEmpty()) ? Doxygen::htmlFileExtension : "");
}
```
and in the code 
```
t << f << addHtmlExtension(f)
```

or
```
QCString addHtmlExtension(QCString fName)
{
  return fName + ((QFileInfo(fName).extension(FALSE).isEmpty()) ? Doxygen::htmlFileExtension : "");
}
```
with in the code
```
t << addHtmlExtension(f)
```

Issue #7295, opened 2019-10-06
Doxygen documentation of C++17 nested namespace erroneous


craffael:
Consider the following [code snippet](https://github.com/doxygen/doxygen/files/3694354/Example.zip) with a C++17 nested namespace:
```
/** @brief second desc */
namespace First::Second {

}
```

The generated namespace page using doxygen 1.8.16, respectively current master:
![grafik](https://user-images.githubusercontent.com/15138672/66269318-623d0900-e847-11e9-99d6-fa70466de159.png) 
This is incorrect: The comment that was applied to namespace second is now shown for namespace first!

Note that this used to work in 1.8.14:
![grafik](https://user-images.githubusercontent.com/15138672/66269343-9dd7d300-e847-11e9-8486-25ad8f120830.png)

albert-github:
I've just pushed a proposed patch, pull request #7296

craffael:
That was fast, thanks!

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.17.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7296, opened 2019-10-06
issue #7295 Doxygen documentation of C++17 nested namespace erroneous

albert-github:
documentation has to be added to the "last" entry not to the "first" / "top"

I used a bit more complicated test case: [example.tar.gz](https://github.com/doxygen/doxygen/files/3694604/example.tar.gz)

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -3989,6 +3989,7 @@ OPERATOR  "operator"{B}*({ARITHOP}|{ASSIGNOP}|{LOGICOP}|{BITOP})
                                             if (current->section == Entry::NAMESPACE_SEC && current->type == "namespace")
                                             {
                                               int split_point;
+                                              Entry *first_current = current;
                                               while ((split_point = current->name.find("::")) != -1)
                                               {
                                                 Entry *new_current = new Entry(*current);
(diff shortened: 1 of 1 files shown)


nrkramer:
Can we merge this? Does this fix a case like such:
`
namespace A {
    class B;
}

/**
* This is my class
*/
class A::B {
};
`
?
In such a case, the class is incorrectly being labeled as A.

albert-github:
Where do you see that the class is labeled as `A` I only see it is `A::B` i.e. full qualified and nowhere separately as `A`?

But probably not as the patch is about place of the documetation text, though:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.
- Please also specify the doxygen version used.

Best to add it to the original issue.

doxygen:
@albert-github I got a compiler error after merging since `Entry*` has now become a `std::unique_ptr<Entry>` so I decided to simplify things a bit, see [this commit](https://github.com/doxygen/doxygen/commit/4884295c6a3bc9af6e01c1688a3173ac6c3bc315)

Issue #7302, opened 2019-10-07
Parsing of template args in single-quotes is incorrect.


skyrich62:
Consider the following code:
```c++
/// @file
/// Something

template <char C>
struct one { };

/// Less than
struct lt : one<'<'> { };

/// Equal
struct eq : one<'='> { };

/// Greater than
struct gt : one<'>'> { };

/// Left paren
struct lp : one<'('> { };

/// Decimal point.
struct pt : one<'.'> { };

/// Right paren
struct rp : one<')'> { };
```
The parser incorrectly considers '<', (in struct lt), to be the start of some construct, rather than a simple char.  Parsing resumes at struct gt, where it finds '>'.  Similarly, '(' and ')' cause problems as do '[' and ']'.  This causes missing documentation for everything between struct lt and struct gt, (i.e. there is no documentation for struct eq nor for struct gt). Addionally, the HAVE_DOT diagrams look horrible:

![structlt__inherit__graph](https://user-images.githubusercontent.com/23705081/66352493-74d14400-e914-11e9-8b19-72a791b71a6b.png)

This may be related to #6802. 

### Warning, Speculation follows
It seems to me that the parser/scanner is counting matching things, like <, >, (, ), [, ], etc, and failing to recognize context, (such as in a character literal).

### Workaround:  (But it's a bit ugly and kludgy)
```c++
#define LT '<'
#define GT '>'
#define LP '('
#define RP ')'

struct lt : one<LT> { }; 
   ... etc
```
*NOTE* using: `const auto LT = '<';` also works, and is slightly less kludgy.

### Reference
I ran into this while writing a grammar for [taocpp/PEGTL](https://github.com/taocpp/PEGTL).

albert-github:
In your text you mention that you used `HAVE_DOT`, when not using dot but the doxygen 'internal' image generator we get a message like:
```
Error: /undefined in '
Operand stack:
   (one { };struct pt :one<'.'> { };struct rp :one<')
Execution stack:
   %interp_exit   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--   false   1   %stopped_push   2015   1   3
 %oparray_pop   2014   1   3   %oparray_pop   --nostringval--   1998   1   3   %oparray_pop   1884   1   3   %oparray_pop   --nostringval--   %errorexec_pop   .runexec2   --nostringval--   --nostringv
[...]

PR #7307, opened 2019-10-09
Determination of anonymous is to restrictive

albert-github:
In case we have a construct like (taken from #7302):
```
/// @file
/// Something

template <char C>
struct one { };

/// The struct str_063
struct str_063 : one<'?'> { };
/// The struct str_064
struct str_064 : one<'@'> { };
```
the resulting inheritance graph for `'@'` is not correct, compare with the output for `'?'`.
The determination of the anonymous / searching for `@` has to exclude parts inside a string.

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -753,7 +753,23 @@ void ClassDefImpl::IMPL::init(const char *defFileName, const char *name,
     isLocal=FALSE;
   }
   isGeneric = (lang==SrcLangExt_CSharp || lang==SrcLangExt_Java) && QCString(name).find('<')!=-1;
-  isAnonymous = QCString(name).find('@')!=-1;
+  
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I've made a more generic fix for this issue by hiding all the "anonymous" checks that search for `'@'` behind a `isAnonymous()` method in Definition. See [this commit](https://github.com/doxygen/doxygen/commit/b32afdbd35624376000aaf12d09930e311692c53). Please verify if this addresses the issue.

albert-github:
Problem has been solved in a more general way by hiding all the "anonymous" checks that search for '@' behind an isAnonymous() method in Definition. See b32afdbd35624376000aaf12d09930e311692c53 commit.

I cannot verify all situation, but looks  OK.

PR #7318, opened 2019-10-12
Warning when using empty HTML tag

albert-github:
When we use the construct:
```
<p/>
```
we get the warning:
```
warning: HTML tags may not use the 'empty tag' XHTML syntax.
```
tough the construct is handled properly, so we can omit the warning.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3720924/example.tar.gz)


albert-github:
Added:
- the HR tag has in xhtml the form `<HR/>`. Supporting empty tag
- better warning (with tag name, for empty tag).
- `</hr>` does not exist, correcting warning (analogous to `</br>`)

PR #7319, opened 2019-10-12
Bug 790856 - Namespace member functions links are broken

albert-github:
Applying patch #720 as used for python also for Cpp.
Tested against example of issue and against CGAL documentation


doxygen:
@albert-github I've made a different fix. 

Some background: for namespace members, the members are also mentioned in the file documentation. To avoid duplicate anchors (mainly an issue in e.g. latex) the `file_` prefix is added to the anchor, but these anchors are actually not used. The links will always point to the namespace documentation. 

Now if the namespace is not documented (or explicitly hidden) the member links would be broken if nothing is done. The idea is that in this case the links point to the file documentation instead. Since there are is no namespace documentation, the `_file` prefix should not be used.

albert-github:
I'll check against CGAL as soon as I  have the ability to access my right environment again.

(as reference the mentioned patch is 926986081bf9b97d4a25b11846af55e51d9f9871 ).

PR #7326, opened 2019-10-16
Incorrect display of multiline #define value

albert-github:
In case we have a `#define` like:
```
#define define_2 \
  2              \
  3
```
we see that the line with the `2` is not indented and shown as `2              \`, the line with the `3` is shown as `  3`. especially with more complex constructs this is distracting.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3734221/example.tar.gz)
this example shows also the distraction better in `define_3`


doxygen:
@albert-github I've created a new commit, since it was difficult to merge the pull request.

albert-github:
Looks OK to me (commit: 1dda553a6c80aae974c0bb90c1870bdbc09004ab)

(I think it is good to mention also the commit id for easier reference).

PR #7339, opened 2019-10-19
Support 'markdown/ github user' type of links to users and links

albert-github:
It is in change logs and release notes quite common to have a construct in the form of:
```
[@user](html link)
```
doxygen shows this correct but issues a warning:
```
warning: Illegal command @user as part of a <a>..</a> block
```
same when using a `#link` (is shown correctly but issues a warning. Pure numerical links don't pose a problem as links in doxygen cannot start with a digit)

In case the case we have a one word `@user` of `#link` escape the first character, unless the `user` is a doxygen command


albert-github:
Maybe this behavior should be "hidden" behind the setting `MARKDOWN_ID_STYLE = GITHUB`

doxygen:
@albert-github tweaked the implemention a bit, see https://github.com/doxygen/doxygen/commit/f5fc652042bc950503706bbd6cb597ce98f4b4ab

albert-github:
@doxygen my tests didn't show a difference with my test so tweaks look OK

PR #7341, opened 2019-10-19
Let $year support SOURCE_DATE_EPOCH

ellert:


albert-github:
What is the advantage of your change?
Can you add a small example with the output before and after your change?

ellert:
> What is the advantage of your change?
> Can you add a small example with the output before and after your change?

If I use an html footer file that contains:

Now is $datetime. This is $year.

and run doxygen the generated index.html contains:

Now is Sat Oct 19 2019 20:48:34. This is 2019.

as expected. However, if I set the SOURCE_DATE_EPOCH when running doxygen

SOURCE_DATE_EPOCH=$(date +%s -d '2018-07-01 1200Z') doxygen

The generated index.html contains:

Now is Sun Jul 1 2018 12:00:00. This is 2019.

I.e the output of $datetime shows the SOURCE_DATE_EPOCH, but the $year shows the current year. This (a) is inconsistent and (b) makes the output non-reproducible.

With the change $year supports SOURCE_DATE_EPOCH the same way $datetime does and the output is consistent:

Now is Sun Jul 1 2018 12:00:00. This is 2018.

simevo:
I agree with this patch and actually I have included a similar one in the currently experimental version of doxygen 1.8.16 in debian:

https://salsa.debian.org/debian/doxygen/blob/master/debian/patches/reproducible_year.diff

albert-github:
@simevo 
- Do I understand it correctly that the 1.8.16 version of debian would be different from the standard doxygen version? 
- Did you report your patch upstream?

simevo:
Debian version of doxygen has always differed slightly from upstream, and all differences are listed as patches here:
https://salsa.debian.org/debian/doxygen/blob/master/debian/patches/
Currently there are 15 and only 3 inluding this one have been forwarded upstream:
```
debian/patches/llvm_libs.diff:Forwarded: https://github.com/doxygen/doxygen/issues/6379
debian/patches/no-timestamps.diff:Forwarded: https://github.com/doxygen/doxygen/issues/5751
debian/patches/reproducible_year.diff:Forwarded: https://github.com/doxygen/doxygen/pull/7341
```
Some of these could be debian-specific. The [debian policy](https://www.debian.org/doc/debian-policy/ch-source.html#changes-to-the-upstream-sources) is that "_changes ... that are not specific to the needs of the Debian system ... should be sent to the upstream authors in whatever form they prefer so as to be included in the upstream version of the package_" so we definitely want to forward them to you.

I'm in the process of taking over maintenance of doxygen in debian (process not yet completed BTW) so TBH I have not gone through all patches created from the previous maintainers in order to check that they have been forwarded as required, but I trust them that they have.

Speaking of this one, as I said it's corrently in experimental so it was my intention to forward it but Mattias was faster (thanks !).

PR #7344, opened 2019-10-21
Small improvements of TCL output

albert-github:
- handling of the used filename, sometimes the filename was not set resulting in case of a warning with file `<unknown>`, also improving, slightly, the line reference.
- <strike>in case of `##` comment blocks the last line was not always show</strike>
- <strike>in case of multiple comment blocks, properly separate them</strike>
- correct handling tab character in source code output (*gen)


albert-github:
Missed the problem regarding the doxygen tests, investigating ...

PR #7348, opened 2019-10-25
Better warning in case a graph would have been to large

albert-github:
When a graph is to large we get a message like:
```
warning: Included by graph for 'tile.h' not generated, too many nodes. Consider increasing DOT_GRAPH_MAX_NODES.
```
but the user has no idea how big the graph really is so the value is added to the warning, so the user can decide to set a value larger than this value.


doxygen:
@albert-github I've decided to implement this change in a slightly different way. See the referenced commit.

albert-github:
No problem, the basic thing is that the warning is more clear, and this implementation is cleaner.

Issue #7020, opened 2019-06-02
Markdown bold, code  and link


albert-github:
Used doxygen version 1.8.15 and also master on June 2, 2019.

The test case: [example.zip](https://github.com/doxygen/doxygen/files/3244637/example.zip)

When having the simple file:
```

Based on README.md of https://github.com/lttng/lttng-tools

  - **[`l`]**

The
  - **n**  ' ' man
  - **0**

```
we get the mesages:
```
.../tst.md:11: warning: found </b> at different nesting level (5) than expected (2)
.../tst.md:12: warning: end of comment block while expecting command </b>
```

This looks to be from the back ticks part in combination wit the `[...]`  and the `**`.

When looking at the `-d markdown` output:
```
======== Markdown =========
---- input -------
@page md_tst tst

Based on README.md of https://github.com/lttng/lttng-tools

  - **[`l`]**

The
  - **n**  ' ' man
  - **0**

---- output -----
@page md_tst tst

Based on README.md of https://github.com/lttng/lttng-tools

  - <strong>[<tt>l</tt>]**

The
  - <strong>n</strong>  ' ' man
  - **0</strong>

=========
```
we see that most `**` are translated to the `<strong>...</strong>` pair but not all.

albert-github:
Running the example with doxygen 1.8.15 gave the results as indicated.
Running with the current doxygen version (1.15.0) gave the correct results.

I think the problem has been solved in the mean time

PR #7349, opened 2019-10-25
Warnings with quoted strings and underscores (emphasis) in markdown

albert-github:
When having an example like:
```
/** @file

<b>;_f</b>
<b>;_s</b>

<b>;_t</b>
<a name="i1_"></a><b>;_w</b>
*/
```
we get the warnings like:
```
.../example.cpp:6: warning: found </b> tag while expecting </em>
.../example.cpp:8: warning: end of comment block while expecting command </em>
.../example.cpp:8: warning: end of comment block while expecting command </b>
```
When having a non finished emphasis and a string is detected the string is not handled as of "higher order" (like brackets in mathematical calculations).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3772403/example.tar.gz)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -254,7 +254,25 @@ static int findEmphasisChar(const char *data, int size, char c, int c_size)
   {
     while (i<size && data[i]!=c    && data[i]!='`' && 
                      data[i]!='\\' && data[i]!='@' &&
-                     data[i]!='\n') i++;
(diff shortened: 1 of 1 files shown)


albert-github:
Looks like a similar issue  as in #7020 but at that place it is for backticks and `**`.
Maybe a more general approach is needed giving backticks, single quotes and double quotes a higher priority inside other tags.

PR #7387, opened 2019-11-08
feature: HTML_HEADING_ANCHOR

erlmann:
`HTML_HEADING_ANCHOR` lets you specify a text (or HTML code) that is inserted before all headings in order to produce clickable anchors for the headings.

Helpful if you want to share links to your docs: without this enhancement, you need to open the browser debugger, look up your heading in the HTML, copy the heading ID and create your URL manually.

With this, you get a clickable anchor right next to your heading that is configurable. Example for setting `HTML_HEADING_ANCHOR` to `&#128279;` (https://emojipedia.org/link-symbol/):

image

PR #7388, opened 2019-11-08
feature: TOC_INCLUDE_HEADINGS_PERMALINKS

erlmann:
Instead of having number-based anchors (`autotoc_md<NUMBER>`), setting `TOC_INCLUDE_HEADINGS_PERMALINKS` to `TRUE` will produce anchors built from page path and heading text which are immune against insertion of additional pages.


erlmann:
Same as with https://github.com/doxygen/doxygen/pull/7387: can I do anything for getting this merged?

src/markdown.cpp
```diff
@@ -1289,7 +1289,16 @@ static QCString extractTitleId(QCString &title, int level)
   {
     static int autoId = 0;
     QCString id;
-    id.sprintf("autotoc_md%d",autoId++);
+    if( Config_getBool(TOC_INCLUDE_HEADINGS_PERMALINKS))
+    {
+        id.sprintf( "autotoc_md_%s_%s",
+                    markdownFileNameToId(g_fileName).data(),
+                    title.lower().replace(QRegExp("[^a-z0-9]"), "_" ).data() );
+    }
```
albert-github, line 1297: The `autotoc` links are only intended for internal doxygen usage, when the user wants a deterministic / permanent anchor he has to provide one himself either by e.g. the doxygen `\anchor` command or by means of `{#anchor}` for examples like:
```
#The header {#head1}
```
or
```
The header {#head2}
=======
```

The pull request is a bit older so to get it quickly up and running for some tests I had to change the routine `[example.tar.gz](https://github.com/doxygen/doxygen/files/5917744/example.tar.gz)` a little bit, i.e. add an argument `fileName` and use this filename instead of `g_fileName` and furthermore with all calls to `[example.tar.gz](https://github.com/doxygen/doxygen/files/5917744/example.tar.gz)` add `m_fileName`).

When I've done this and used the attached example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5917744/example.tar.gz)

We can see that part of the implementation and the used path are shown in the anchors like: `autotoc_md_md_D__pull_7388_md1_README_another_sub` and `autotoc_md_md_D__pull_7388_README_another_sub` making the links not persistent / permanent as when the build path changes also the link changes.

erlmann: Understand. What I tried to solve the issue that, when you have `autotoc` active and shared a link to `The sub` in this document:

```markdown
# Page Title

# The Header
[TOC]
Some normal text


## The sub
Some normal text

## Another sub
Some normal text
```

Then the link points to `#autotoc_md2`.

Now you extend the documentation and add another heading **before** `The sub`:

```markdown
# Page Title

# The Header
[TOC]
Some normal text

## Inserted Heading
Inserted text

## The sub
Some normal text

## Another sub
Some normal text
```

Now, the link you shared (`#autotoc_md2`) no longer points to `The sub` but to `Inserted Heading`.

Yes, this does not happen when you explicitly set up your anchors, but especially when #7387 is active, you get shareable links on all headings without having to manually add anchors anywhere.

And this MR tries to resolve that issue by using the heading text instead of a simple counter for identifying a heading.

You're right: if you move your file around or change your heading, this MR won't help. But at least your links are still valid if you add new headings or move the sections around in the file.

albert-github: They will still be valid, with this pull request tough the user already added somewhere the reference by means of the, also, generated label which he has to figure out as well. By adding the anchor (e.g. by `{#head1}` he has full control over the name and this doesn't even change when he moves the file to a different directory (and is probably easier to remember as well). So I think the following is still valid:
> The autotoc links are only intended for internal doxygen usage, when the user wants a deterministic / permanent anchor he has to provide one himself either

erlmann: > The autotoc links are only intended for internal doxygen usage, when the user wants a deterministic / permanent anchor he has to provide one himself

Yes, if this is the case, than there's really nothing to do here. I wasn't aware of this fact.

Close?

albert-github: I gave my opinion on it, @doxygen what is your opinion?

erlmann: Rebased onto `Release_1_9_3`. With 1 year passed, I still think this is a useful addition.

Of course, a user **can** add a custom anchor to a heading. But this requires re-generating the docs.

Imagine the scenario where the docs are deployed to a server and you want to point a customer to a certain section in your docs where you didn't assign an anchor manually. Currently, you have to modify your sources, rebuild the docs, deploy them again and then can share the link with the updated anchor.

With the suggested change, you always have a permalink for each section that was created. No need to re-build and re-deploy.

erlmann:
Largely implemented with 45612e24023db52fe3c4e4dea43b8d7c8eb76e66, this PR is no longer required.

PR #561, opened 2017-04-06
Spelling fixes

jsoref:
Note: I've included certain files that you may choose to ignore (eg. the changelog). Personally, I'd rather read a corrected changelog w/ links to poor commit messages than a hard to comprehend changelog that happens to correspond one to one with a commit list.

I've excluded a number of files (basically anything that seemed to not be English). Those might need to be synchronized.

The majority of the changes are to comments, a couple are to output, and some are to data files.

It is possible to fold this into a single commit, but I'd rather the maintainers did that at acceptance. It's much easier to rebase an unfolded series (e.g. I rebased the changelog bits off my ignore series which had rewritten the changelog to omit hashes).

PR #7391, opened 2019-11-12
Spelling fixes

jsoref:
@albert-github this is a followup to #561 as you were comfortable splitting the commits as appropriate, I'm continuing w/ the same model.

There are a handful of misspellings which date to the original series (I've since added tooling to repurpose spelling fixes -- at some point, I'm going to see if I can use my github corpus to provide generalized suggestions without a human...). I am trying to avoid suggesting changes which you've dropped, although I may make mistakes (I'm human).

While I am submitting spelling fixes to the changelog. I'm leaving misspellings in lines that say:
> changing *incorrect spelling* to *corrected spelling*


doc/changelog.doc
```diff
@@ -812,7 +812,7 @@ href="https://github.com/doxygen/doxygen/commit/a697caadf1912d0d74faa208f4cff887
 <li>Fixed problem where automatic line breaking caused missing vertical bars in the parameter table for Latex output. [<a href="https://github.com/doxygen/doxygen/commit/5fc82b2275e202438ac61b070ac5f4be0df792d6">view</a>]</li>
 <li>Fixes for cross platform build with new LLVM/CLANG version [<a href="https://github.com/doxygen/doxygen/commit/b0aae61c97966cb9d424b500d7ced5bdf500d8db">view</a>]</li>
 <li>Fixup man page NAME section when page has title [<a href="https://github.com/doxygen/doxygen/commit/9d0908359363dbb43236767669c214721700acf7">view</a>]</li>
-<li>Function declaration following a function definition incorrectly listed as calling dependencing [<a href="https://github.com/doxygen/doxygen/commit/436fc7ed1158d517dd6f6d25aa3e05568f8c3d94">view</a>]</li>
+<li>Function declaration following a function definition incorrectly listed as calling dependency [<a href="https://github.com/doxygen/doxygen/commit/436fc7ed1158d517dd6f6d25aa3e05568f8c3d94">view</a>]</li>
```
jsoref: I'm not sure

src/defgen.cpp
```diff
@@ -288,7 +288,7 @@ void generateDEFForMember(MemberDef *md,
     {
       if (rmd->getStartBodyLine()!=-1 && rmd->getBodyDef())
       {
-        t << memPrefix << "referenceby = {" << endl;
+        t << memPrefix << "referencedby = {" << endl;
```
jsoref: This is an output change, but I think this is debug code...

src/doctokenizer.l
```diff
@@ -344,7 +344,7 @@ PHPTYPE   [\\:a-z_A-Z0-9\x80-\xFF\-]+
 CITESCHAR [a-z_A-Z0-9\x80-\xFF\-\?]
 CITEECHAR [a-z_A-Z0-9\x80-\xFF\-\+:\/\?]
 CITEID    {CITESCHAR}{CITEECHAR}*("."{CITESCHAR}{CITEECHAR}*)*|"\""{CITESCHAR}{CITEECHAR}*("."{CITESCHAR}{CITEECHAR}*)*"\""
-MAILADR   ("mailto:")?[a-z_A-Z0-9.+-]+"@"[a-z_A-Z0-9-]+("."[a-z_A-Z0-9\-]+)+[a-z_A-Z0-9\-]+
+MAILTOADDR   ("mailto:")?[a-z_A-Z0-9.+-]+"@"[a-z_A-Z0-9-]+("."[a-z_A-Z0-9\-]+)+[a-z_A-Z0-9\-]+
```
jsoref: There's a `MAILADDR` in another file in this directory. As I'm not sure if the files are related, I chose to give this a different name, but I think it's reasonable to argue that `adr` is a poor abbreviation, esp. given that `addr` is used for a similar token.

src/fortranscanner.l
```diff
@@ -2839,7 +2839,7 @@ static void scanner_abort()
 #if !defined(YY_FLEX_SUBMINOR_VERSION) 
 //----------------------------------------------------------------------------
 extern "C" { // some bogus code to keep the compiler happy
-  void fortranscannernerYYdummy() { yy_flex_realloc(0,0); } 
+  void fortranscannerYYdummy() { yy_flex_realloc(0,0); }
```
jsoref: I know this is a dummy thing, but I don't see any particular reason for the `ner` suffix.

templates/xml/compound.xsd
```diff
@@ -764,7 +764,7 @@
       <xsd:enumeration value="PHP" />
       <xsd:enumeration value="Objective-C" />
       <xsd:enumeration value="C++" />
-      <xsd:enumeration value="Javascript" />
+      <xsd:enumeration value="JavaScript" />
```
jsoref: I hope nothing outside this project cares.

doc/changelog.doc
albert-github: Looks OK to me

src/defgen.cpp
albert-github: "That is an output change", this is the reason why I didn't change it .

As far as I can tell the output is used in case `GENERATE_AUTOGEN_DEF` is set, but I think too that it wouldn't be a problem to change it (but I'm not sure as I don't know autogen, project looks a bit dead to me though).

src/doctokenizer.l
albert-github: Changing a "variable" name, in this case it is a symbolic pattern is dangerous, but looks like OK as it has been corrected on other places as well (problem probably originates from the fact that the English address is in Dutch adres).

src/fortranscanner.l
albert-github: Neither do I.

templates/xml/compound.xsd
albert-github: Not sure about this. @morsa would this hit you?

albert-github:
The splitting in different pull requests has been made for the mentioned. The split in different directories has been made after an offline discussion with @doxygen, would be more comfortable for me to make new changes and for @doxygen easier to integrate / less chances for merge conflicts.

Seeing this list the same might apply, though for e.g. the "Javascript" -> "JavaScript" proposed change multiple directories are necessary, but maybe this change should be alone in a proposed pull request.

Regarding the files in the vhdlparser directory these are generated files by means of JavaCC and have been reported upstream (as you saw) as proposed pull request https://github.com/javacc/javacc/pull/118. So these should not be fixed here unless it would be code from a more or less dead project which is not the case with JavaCC so hopefully they will be fixed and released soon.

The rest of the proposed fixes I still have to assess.

albert-github:
As a separate question, which package / workflow are you using to get the shown mistakes. So far I used codespell and missed a "few".

jsoref:
Sorry, it's my own home grown tool. Sometimes I include a reference to it when I make my initial PR -- https://github.com/jsoref/spelling -- I'm not sure why I didn't when I made the first one here. 
Basically it just generates a list of suspicious `word tokens` and then I use tooling to check the context and tooling to produce replacements.

It's almost certainly much more labor intensive than things like codespell as it isn't using any predefined corrections.
I generally just let Google Chrome/Mozilla Firefox suggest spelling corrections by pasting tokens into the urlbar, copying the suggested replacement, and pasting it back into a command prompt w/ a command to replace the original w/ the replacement. 

My current boilerplate for PRs is:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

 In theory if you maintained `fchurn` output you could use it to compare one version of your repo w/ a future version and then only review "newly added" suspicious `tokens`.

albert-github:
Thanks for the answer.
I will assess it.
You made it implicitly clear why you use commits per word.

orbitcowboy:
Did you try codespell?

albert-github:
I did try codespell and, to the best of my knowledge, the spelling errors reported by it have been fixed, see e.g. #7380 - #7372 which are spelling errors reported by OP in a previous pull request plus codespell reported errors (there are a lot of false positives so it is always possible to miss some and the list from codespell is not exhaustive and does not distinguish between US-English and UK-English).

orbitcowboy:
Thanks for feedback. Just to let you know, there is another dictionary for
codespell available here:
https://github.com/orbitcowboy/codespell_dictionary  It might find other
interesting typos.

albert-github <notifications@github.com> schrieb am Di., 19. Nov. 2019,
20:20:

albert-github:
Interesting list, comments will be made with the project.
An interesting list is also scripts/spelling.txt which can be found in e.g. https://git.kernel.org/torvalds/t/linux-5.4-rc8.tar.gz

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of a release).

jsoref:
@albert-github: it looks like you have a script which added https://github.com/doxygen/doxygen/pull/7391#issuecomment-568722568, I think that only applies to issues whereas this was a PR. Or, perhaps the text should say:

> Code has been integrated in master on github (please don't close the _corresponding_ issue as this will be done at the moment of a release).

(This PR may be the rare case where there is no corresponding issue.)

albert-github:
Well spotted. It is not unusual to have a pull request and no issue (I submitted quite a few proposed pull requests without corresponding issue).
My text should not have been with the merge issue, I normally add it only to the issue that is reporting a problem and not with a pull request. It landed here by mistake.

PR #7417, opened 2019-11-23
Warning in case no define name in PREDEFINED

albert-github:
In case no define name is specified with the PREDEFINED configuration tag, this was silently ignored. Added explicit configuration warning.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/3882439/example.tar.gz)

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -77,4 +77,5 @@ namespace Config
   void deinit();
 }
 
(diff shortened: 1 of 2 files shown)


albert-github:
Obsolete, will be replaced by better implementation

Issue #4214, opened 2018-07-02
htmlonly content appears in generated XML output (Origin: bugzilla #646002)


doxygen:
status RESOLVED severity *normal* in component *general* for *---*
Reported in version *1.7.4* on platform *Other*
Assigned to: Dimitri van Heesch

On 2011-03-28 19:17:36 +0000, Ray Lefuel wrote:

> The Doxygen docs state that the `\htmlonly` tag starts a block of text that will be verbatim included in the generated HTML documentation only.  However, the content appears in the generated XML documentation, too.  It is encased in `<htmlonly>` and `</htmlonly>` tags.
> 
> Shouldn't the content be ignored altogether as it's being generated?
> 
> Also, in the Doxygen docs, the link to the `\endhtmlonly` topic under the `\htmlonly` topic doesn't work.

On 2015-08-12 09:10:09 +0000, Thomas Martitz wrote:

> I can confirm this. Any resolution on the horizon?

On 2015-08-12 10:56:06 +0000, albert wrote:

> According to the documentation the `@htmlonly` (and other `*only` marked documentation) should not appear. For the other output formats this works.
> 
> I've just pushed a proposed patch (pull request #381) to github.
> 
> In case the old behavior should be maintained, i.e. in the xml file also the other *only documentation should appear, the documentation of the different commands has to be adjusted and I think it would be better that the code inside e.g. `<htmlonly> .. </htmlonly>` (etc.) should be placed in `CDATA` constructs to overcome possible incorrect XML code resulting from the other `*only` constructs.

On 2015-08-31 16:37:07 +0000, albert wrote:

> Code has been integrated in github

On 2015-12-30 10:19:01 +0000, Dimitri van Heesch wrote:

> This bug was previously marked ASSIGNED, which means it should be fixed in
> doxygen version 1.8.11. Please verify if this is indeed the case. Reopen the
> bug if you think it is not fixed and please include any additional information 
> that you think can be relevant (preferably in the form of a self-contained example).

PR #7481, opened 2020-01-05
Showing information from all \*only commands in XML output

albert-github:
This fix reverts the changes from #381 (i.e issue #4214 "htmlonly content appears in generated XML output (Origin: bugzilla 646002)", as XML is seen as an output format that can be further processed and therefore should contain all possible information.

### DIFF
--- a/testing/020/indexpage.xml
+++ b/testing/020/indexpage.xml
@@ -6,9 +6,21 @@
     <briefdescription>
     </briefdescription>
     <detaileddescription>
-      <para>Text.       
+      <para>Text. <htmlonly> 
+HTML
+</htmlonly> <htmlonly block="yes">
+HTML with block
+</htmlonly> <rtfonly> 
+RTF
+</rtfonly> <manonly> 
+Man
+</manonly> <latexonly> 
+LaTeX
+</latexonly>  
 XML
-  More text. </para>
+ <docbookonly> 
+DocBook
+</docbookonly> More text. </para>
     </detaileddescription>
   </compounddef>
 </doxygen>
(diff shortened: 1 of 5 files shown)


testing/020/indexpage.xml
```diff
@@ -6,9 +6,21 @@
     <briefdescription>
     </briefdescription>
     <detaileddescription>
-      <para>Text.       
+      <para>Text. <htmlonly> 
+HTML
+</htmlonly> <htmlonly block="yes">
+HTML with block
+</htmlonly> <rtfonly>
```
mosra: Thanks for doing this! 👍 

One "stretch goal" -- the documentation is stating this:

> When you want to insert a HTML fragment that has block scope like a table or list which should appear outside `<p>..</p>`, this can lead to invalid HTML. You can use `\htmlonly[block]` to make doxygen end the current paragraph and restart it after `\endhtmlonly`.

Ideally that would hold also for the XML output, so either `<htmlonly>` being inside a `<para>`, or `<htmlonly>` *not* being wrapped in a `<para>` (instead of current `<htmlonly block="yes">` being inside a `<para>`). Basically so it doesn't need to be extracted out of the paragraph on the client side. 

How hard would it be to implement it like this?

mosra: In other words, the output looking like this instead:

```xml
<para>Text. <htmlonly> 
HTML
</htmlonly></para>
<htmlonly block="yes">
HTML with block
</htmlonly>
<para><rtfonly> 
RTF
</rtfonly> <manonly> 
Man
</manonly> <latexonly> 
LaTeX
</latexonly>
```

albert-github: No idea at the moment how difficult this would be and whether this would be conceptual OK (as it is not really a "dump" anymore).
When really necessary I think best is to submit an issue for this as soon as this proposed pull request has been integrated in master on github.

mosra: Okay, no problem. Asked only in case it would be easy to do to do on your side -- I can do the paragraph extraction on my side as well ;)

Issue #7411, opened 2019-11-19
"warning: return type of member ... is not documented" for static void and virtual void functions


dweindl:
Doxygen from master (2972f6ddc491f615f814b1277ec715b6f707c419) complains about undocumented return types for `static void` and `virtual void` functions. For regular void functions everything works as expected.

Example `test.cpp`:
```
/** @file test.cpp */

/**
 * @brief This works
 */
void works() {}

/**
 * @brief This fails
 */
static void fails() {}

/**
 * @brief test class
 */
class Test {
    /**
    * @brief This works
    */
    void works() {}

    /**
    * @brief This fails
    */
    virtual void virtualfails() {}

    /**
    * @brief This fails
    */
    static void staticfails() {}
}
``` 

With default Doxyfile and:
```
EXTRACT_PRIVATE        = YES
EXTRACT_PRIV_VIRTUAL   = YES
EXTRACT_PACKAGE        = YES
EXTRACT_STATIC         = YES
WARN_NO_PARAMDOC       = YES
```

Results in:
```
[...]
test.cpp:11: warning: return type of member fails is not documented
[...]
test.cpp:25: warning: return type of member Test::virtualfails is not documented
test.cpp:30: warning: return type of member Test::staticfails is not documented
[...]
```

albert-github:
The reported problem looks like to be a consequence of issue issue "1.8.16: Function returning void pointer generates warning" (#7212), the resulting pull request (#7213) and making it more strict.

The mentioned functions have return types:
- virtual void
- static void

but we also have to take into consideration the examples as mentioned in the pull request #7213.

SteffenSeckler:
This is also broken for:
```c++
constexpr void f()
```
And will be broken in c++20 for:
```c++
consteval void f();
constexpr virtual void f();
consteval virtual void f();
```

Crayon2000:
I'm getting the same error with:
```c
INLINE void f()
```

The INLINE macro is something similar to:
```c
#ifndef INLINE
# if __GNUC__ && !__GNUC_STDC_INLINE__
#  define INLINE extern inline
# else
#  define INLINE inline
# endif
#endif
```
From: http://www.greenend.org.uk/rjk/tech/inline.html

But this can be fixed by using PREDEFINED with
```text
INLINE=
```

SteffenSeckler:
@Crayon2000 is this also fixed using 
```
MACRO_EXPANSION        = YES
```
?

Crayon2000:
[...]

PR #7483, opened 2020-01-05
A proposed fix for #7411

lelegard:
A proposed fix for #7411. Maybe this is not the most appropriate one. But without it, the latest version of doxygen generates 1806 false errors on [my project](https://github.com/tsduck/tsduck), making it unusable because actual errors would then become unnoticed.

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -4178,7 +4178,8 @@ void MemberDefImpl::warnIfUndocumentedParams() const
           qPrint(qualifiedName()));
     }
     if (!m_impl->hasDocumentedReturnType &&
-        isFunction() && hasDocumentation() && !returnType.isEmpty())
+        isFunction() && hasDocumentation() && !returnType.isEmpty() &&
+        returnType!="virtual void" && returnType!="static void")
     {
       warn_doc_error(getDefFileName(),getDefLine(),
           "return type of member %s is not documented",


dweindl:
Confirming that this fixes my issue #7411. Thanks @lelegard.

albert-github:
As is the commit looks like to solve the problem for the mentioned situations.
The question remains are the 2 mentioned situations:
- virtual void
- static void

the only ones to be considered or are there other situations possible?

A relevant previous commit in this respect 361f3e10a86370a568733a7e4d862b180cf45fd9 on August 31, 2019.

lelegard:
> are there other situations possible?

Probably. This is why I called it _"a proposed fix"_  and added _"maybe this is not the most appropriate one"_. Someone with a better knowledge of the doxygen code base could probably do better.

Specifically, the code is now:
~~~
    if (!m_impl->hasDocumentedReturnType &&
        isFunction() && hasDocumentation() && !returnType.isEmpty() &&
        returnType!="virtual void" && returnType!="static void")
~~~
Note that `"void"` is not explicitly tested. This means that some code, somewhere, probably empties `returnType` when it is `void`. A proper fix should occur there. Because `virtual void` and `static void` are not "return types". They are the combination of an attribute which applies to the function (`virtual` or `static`) and an actual type (`void` alone).

So, in my opinion, a proper fix should do the following:
- At the point where the function is analyzed,
- dissociate attributes and actual return type,
- then the return type becomes simply `void` again (and is probably emptied as it is now).
- Finally revert my proposed fix which will become useless.

But this must be done by some regular doxygen maintainer, someone who knows the code.

albert-github:
I was actually referring to other possibilities in cpp, whether or not it would be possible to have other situations than:
- void
- static void
- virtual void

that are actually not returning anything.

The reference to the mentioned commit, i.e. more or less the code a little bit earlier:
```
  else if ( // see if return type is documented in a function w/o return type
            hasReturnCommand &&
            (
              returnType=="void"   || // void return type
              returnType.find("subroutine")!=-1 || // fortran subroutine
              isConstructor()      || // a constructor
              isDestructor()          // or destructor
            )
          )
  {
    warn_doc_error(getDefFileName(),getDefLine(),"documented empty return type of %s",
                          qualifiedName().data());
  }
  else if ( // see if return needs to documented
            m_impl->hasDocumentedReturnType ||
           returnType=="void" || // void return type
           returnType.find("subroutine")!=-1 || // fortran subroutine
           isConstructor()    || // a constructor
           isDestructor()        // or destructor
          )
  {
    m_impl->hasDocumentedReturnType = TRUE;
  }
```
which is probably a more appropriate place.

lelegard:
> which is probably a more appropriate place.

It could even be earlier. The code you mention tests `returnType=="void"`. So, `returnType` has already been set earlier. And if its value is `static void` or `virtual void`, this is formally incorrect because in C++ "static void" is not the return type of a "function". In fact, "void" is the return type of a "static function" (the same applies with virtual).

So, I would say that assigning `"static void"` to `returnType` is an error. This is the one to fix.

In fact, with something like `static void f();`, `returnType` shall be `"void"` and some other field `returnAttribute` shall be `"static"`. Other possible attributes are `virtual` or non-portable ones such as `__declspec(..)` for instance. Looking at the C/C++ standards should give a list of possible attributes (plus some compiler specific docs).

If you properly separate return type and return attribute in the first place, then the problem will go away because `void` is the _only_ possible type which returns nothing.

SteffenSeckler:
I concur with @lelegard this should not be handled there, but rather should be handled properly when generating the returntype.
For the current workaround, the additional keywords `constexpr`, `consteval` (c++20) can also play a role (even in combination with virtual, since c++20), i.e.,
```c++
constexpr void f();
consteval void f();
constexpr virtual void f();
consteval virtual void f();
```

doxygen:
@lelegard Please check if [this commit](https://github.com/doxygen/doxygen/commit/6c10868a2b31076e8f5e42af0036b43e7f39d2b2) fixes the problem for you.

lelegard:
> @lelegard Please check if this commit fixes the problem for you.

It does, thank you. This fix is more appropriate than the quick & dirty hack I proposed. You may close this PR without merging.

dweindl:
> @lelegard Please check if [this commit](https://github.com/doxygen/doxygen/commit/6c10868a2b31076e8f5e42af0036b43e7f39d2b2) fixes the problem for you.

`result.stripPrefix("constexp ")` -> `result.stripPrefix("constexpr ")`

Thanks @doxygen

lelegard:
And, at the cost of being pedantic, checking `result.stripPrefix("volatile "))` is not formally correct because in something like `virtual volatile bool f()` the _attribute of the method_ is `virtual` only while the _returned type_ is `volatile bool`. In that case, `virtual` is an attribute of the method while `volatile` is an attribute of the returned type.

But, for the sake of efficiency in the context of doxygen, what would be incorrect in a compiler is probably sufficient here.

doxygen:
which means it should be fixed in doxygen version 1.8.18.

PR #7492, opened 2020-01-08
Updated Swedish translation to match the latest version

bjornpalmqvist:
Updated the translation for the Swedish language to match the latest version of doxygen.

### DIFF
--- a/src/translator_sv.h
+++ b/src/translator_sv.h
@@ -23,6 +23,7 @@ Xet Erixon           <xet@xeqt.com>
 Mikael Hallin        <mikaelhallin@yahoo.se>           2003-07-28
 Björn Palmqvist      <bjorn@aidium.se>                 2014-02-01
 Magnus Österlund     <magnus.osterlund@capgemini.com>  2016-09-12
+Björn Palmqvist      <bjorn@aidium.se>                 2020-01-08
 ==================================================================================
 Uppdateringar.
 1999/04/29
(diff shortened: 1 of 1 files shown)


src/translator_sv.h
```diff
@@ -70,16 +71,26 @@ Problem!
...
 2015/09/12
 * Fixat lite särksirvningar och inkonsekvenser
+2020/01/08
+* Uppdaterat den till senaste språkversionen 1.8.15
+
+Bytte ut Deprecated från Föråldrad till Obsolet
+
+VHDL översättningarna är kanske inte perfekta, då jag endast använt de en gång tidigare.
+     lämnade use clause orörd, då jag inte hittade en lämplig översättning för den.
+
```
albert-github: Would be nice to have an English translation of the update as well.

bjornpalmqvist: I can fix that, pusing an update soon.

src/translator_sv.h
```diff
@@ -70,16 +71,34 @@ Problem!
...
+Jag lämnade use clause orörd, då jag inte hittade en lämplig översättning för den.
+
+english:
+* Uppdaterat the language translation to 1.8.15
+
+Changed Deprecated from Föråldrad to Obsolet
+
+The VHDL translations may not perfect, when I only used it once before.
+I left use clause untouched when I didn't find a suitable translation for that.
+
```
albert-github: Intentions is quite clear, thanks, but I think:
- "english"
   should be
   "English"
- "Uppdaterat " 
   should be:
  "Updated"
- "The VHDL translations may not perfect, when I only used it once before."
  should be:
  "The VHDL translations may not perfect, as I only used it once before."
- "I left use clause untouched when I didn't find a suitable translation for that."
  should be:
  "I left use clause untouched as I didn't find a suitable translation for it."

bjornpalmqvist: Ok, will fix that. Thanks for the review!

albert-github: Thanks for the effort!!

PR #7502, opened 2020-01-09
Create .md version of markdown.html

excalamus:
I wanted to provide my users with a page which describes the Doxygen markdown syntax, however I could not find a markdown version of the [Markdown support page](http://www.doxygen.nl/manual/markdown.html)!  I could only find [the `.doc` version](https://github.com/doxygen/doxygen/blob/master/doc/markdown.doc) and wasn't sure how to make use of that.

This is a copy-and-paste of the Markdown support page with headers and examples formatted in markdown.  I also included the "complex tables in doxygen use html tables" example.  

Clearly having multiple formats of a document is not a good way to document things!  I'm making this pull request so that it will be available publicly (regardless of whether you accept it) and perhaps be of use to someone else.

Thanks for a lovely and well documented program!

### DIFF
--- a/examples/markdown_support.md
+++ b/examples/markdown_support.md
@@ -1,0 +1,747 @@
+Markdown support
+================
+
+Markdown support was introduced in doxygen version 1.8.0. It is a
+plain text formatting syntax written by John Gruber, with the
+following underlying design goal:
+
+> The design goal for Markdown's formatting syntax is to make it as
+> readable as possible. The idea is that a Markdown-formatted document
+> should be publishable as-is, as plain text, without looking like it's
+> been marked up with tags or formatting instructions. While Markdown's
+> syntax has been influenced by several existing text-to-HTML filters,
+> the single biggest source of inspiration for Markdown's syntax is the
+> format of plain text email.
+
+In the next section the standard Markdown features are briefly
+discussed. The reader is referred to the Markdown site for more
+details.
+
+Some enhancements were made, for instance PHP Markdown Extra, and
+GitHub flavored Markdown. The section Markdown Extensions discusses
+the extensions that doxygen supports.
+
+Finally section Doxygen specifics discusses some specifics for
+doxygen's implementation of the Markdown standard.
+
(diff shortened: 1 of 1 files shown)


albert-github:
As such having a markdown version of, parts of, the documentation would not be bad and the markdown file could be used in the doxygen documentation as well e.g. by means of a `\includedoc` statement.

Having just a "copy" of the file is a bad idea as this means two sources that should be maintained (like you already wrote).
Due to the cut / paste actions the lines between the original `.doc` file and the new `.md` file have line-breaks at different places  and are hard to compare. A better approach would have been to use the `.doc` file (which is already more or less in markdown format) and convert this to a `.md` file (also with the aid of the `\includedoc` possibility).

The current markdown file also is missing some parts of the original file as used by doxygen a.o.:
- table of contents (`[TOC]`)
- references to different sources like:
   - `[markdown]`
      to
      `[markdown]: https://daringfireball.net/projects/markdown/`
- ...

You wrote: 
> I wanted to provide my users with a page which describes the Doxygen markdown syntax

which users, which product?

excalamus:
> I wanted to provide my users with a page which describes the Doxygen markdown syntax

I mean users from a system administration point of view.  My users are various developers and engineers who create myriad little projects.   I have a server which hooks into our VCS and generates API documentation.  Doxygen is configured to render any additional .md documentation.  The .md version of markdown.html is included with each project for convenience, along with the .md source for reference.

image

I hope that clarifies!

> A better approach would have been to use the .doc file (which is already more or less in markdown format) and convert this to a .md file...

I whole heartedly agree!  However, I'm not familiar with `.doc` outside the MS sense and had no luck identifying the format through searches.  What is `.doc`?

albert-github:
The `.doc` is just plain ASCII, where `.doc` stand just for documentation. The original `.doc` files have been created in a time when doxygen probably was only able to parse C-type code and comments and theer was a need for separate documentation etc. These documentation files contain also C-type comment blocks.

PR #7524, opened 2020-01-19
Bug 744244 - @callgraph has no effect

albert-github:
To be able to distinguish between explicit setting of callgraph / hidecallgraph (etc.) compared to the configuration setting it is necessary to record what has been used and handle this (otherwise duringe a combine of a definition and declaration .
- entry.h, memberdef.h create possibility to record config / explicit setting
- entry.cpp set initial value
- memberdef.cpp, commentscan.l enable / disable functions have to "or" values to record the "history".
- memberdef.cpp when checking for use check individual values
- memberdef.cpp create functions to output "raw" value of call... / hide...
- doxygen.cpp properly initialize the members
- docparser.cpp warn in case use call... / hide... for the same member

Extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4082662/example.tar.gz)

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -37,6 +37,7 @@
 #include "searchindex.h"
 #include "parserintf.h"
 #include "objcache.h"
+#include "entry.h"
 
 #include "vhdldocgen.h"
 #include "arguments.h"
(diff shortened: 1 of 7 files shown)


src/memberdef.cpp
```diff
@@ -5523,24 +5536,56 @@ void MemberDefImpl::setAnonymousUsed() const
 
 bool MemberDefImpl::hasCallGraph() const
 {
-  return m_impl->hasCallGraph;
+  if (m_impl->hasCallGraph & Entry::CALL_SHOW) return true;
```
doxygen: I would think these these 4 lines can be done in one line like so
```c
return m_impl->hasCallGraph & (Entry::CALL_SHOW|Entry::CALL_CONFIG)
```

albert-github: No I don't think so as in case:
- CALL_SHOW = false
- CALL_HIDE = true  so explicit hide setting
- CALL_CONFIG = true so default is show
in the 4 line case the CALL_HIDE is honored, but in the reduced 1 line form it isn't .

PR #7534, opened 2020-01-23
Bug 791390 - Table formatting information ignored in Doxygen XML output

albert-github:
- added possibility (and checks) for xml
- added missing case for class in docbook

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4103470/example.tar.gz)

### DIFF
--- a/src/docbookvisitor.cpp
+++ b/src/docbookvisitor.cpp
@@ -1119,6 +1119,10 @@ DB_VIS_C
       {
         m_t << " align='center'";
       }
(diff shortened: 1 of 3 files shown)


doxygen:
@albert-github I've reworked the pull request such that it does not output unsupported attributes. For XML that is attributes not listed in the xsd schema file, and for docbook I looked at the spec to see which HTML attributes are supported. Please check if it is ok like this.

PR #7552, opened 2020-01-30
Improved Makefile for PdfLaTeX

wosrediinanatour:
Makefile for pdflatex: Parallel executation (make -j 2) is now possible. PdfLaTeX will not hang, if there are LaTeX errors.

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -296,6 +296,7 @@ static void writeLatexMakefile()
   QCString mkidx_command = Config_getString(MAKEINDEX_CMD_NAME);
   // end insertion by KONNO Akihisa <konno@researchers.jp> 2002-03-05
   FTextStream t(&file);
+  t << "CLEAN_FILES := *.ps *.dvi *.aux *.toc *.idx *.ind *.ilg *.log *.out *.brf *.blg *.bbl refman.pdf";
   if (!Config_getBool(USE_PDFLATEX)) // use plain old latex
   {
     t << "LATEX_CMD=" << latex_command << endl
@@ -346,11 +347,12 @@ static void writeLatexMakefile()
   }
   else // use pdflatex for higher quality output
   {
-    t << "LATEX_CMD=" << latex_command << endl
+    t << "LATEX_CMD:=" << latex_command << " -interaction=batchmode -halt-on-error" << endl
       << endl;
     t << "all: refman.pdf" << endl << endl
       << "pdf: refman.pdf" << endl << endl;
-    t << "refman.pdf: clean refman.tex" << endl;
+    t << "refman.pdf: refman.tex" << endl;
+    t << "\trm -f $(CLEAN_FILES)" << endl;
     t << "\t$(LATEX_CMD) refman" << endl;
     t << "\t" << mkidx_command << " refman.idx" << endl;
     if (generateBib)
@@ -373,7 +375,7 @@ static void writeLatexMakefile()
   t << endl
     << "clean:" << endl
     << "\trm -f " 
-    << "*.ps *.dvi *.aux *.toc *.idx *.ind *.ilg *.log *.out *.brf *.blg *.bbl refman.pdf" << endl;
+    << "rm -f $(CLEAN_FILES)" << endl;
 }
 
 static void writeMakeBat()
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -296,7 +296,8 @@ static void writeLatexMakefile()
   QCString mkidx_command = Config_getString(MAKEINDEX_CMD_NAME);
   // end insertion by KONNO Akihisa <konno@researchers.jp> 2002-03-05
   FTextStream t(&file);
-  t << "CLEAN_FILES := *.ps *.dvi *.aux *.toc *.idx *.ind *.ilg *.log *.out *.brf *.blg *.bbl refman.pdf";
+  t << "CLEAN_FILES := *.ps *.dvi *.aux *.toc *.idx *.ind *.ilg *.log *.out *.brf *.blg *.bbl refman.pdf" << endl
+   << endl;
   if (!Config_getBool(USE_PDFLATEX)) // use plain old latex
   {
     t << "LATEX_CMD=" << latex_command << endl
--- a/templates/latex/latexmakefile.tpl
+++ b/templates/latex/latexmakefile.tpl
@@ -1,22 +1,28 @@
+CLEAN_FILES := *.ps *.dvi *.aux *.toc *.idx *.ind *.ilg *.log *.out *.brf *.blg *.bbl refman.pdf
+
 {% if config.USE_PDFLATEX %}
+
+LATEX_CMD := pdflatex -interaction=batchmode -halt-on-error 
+
 all: refman.pdf
 
 pdf: refman.pdf
 
-refman.pdf: clean refman.tex
-	pdflatex refman
+refman.pdf: refman.tex
+	rm -f $(CLEAN_FILES)
+	$(LATEX_CMD) refman
 	{{ config.MAKEINDEX_CMD_NAME }} refman.idx
 {# TODO: generateBib #}
-	pdflatex refman
+	$(LATEX_CMD) refman
 	latex_count=8 ; \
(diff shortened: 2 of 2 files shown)


wosrediinanatour:
Test: Introduce an LaTeX-erlor in refman.tex (e.g. a \asfdsggsdfd). Then the make process should stop instead of hanging.
This is not another engine, only additional command lines. So I don't think it makes sense to change latexCommandName().
I updated the code - now changing latexgen.cpp. I cannot test make.bat (no Windows availabel), so I changed only the generation of the makefile.

albert-github:
> Test: Introduce an LaTeX-erlor in refman.tex (e.g. a \asfdsggsdfd). Then the make process should stop instead of hanging.

Wouldn't setting the configuartion `LATEX_BATCHMODE=YES` serve the same  purpose?

> This is not another engine, only additional command lines. So I don't think it makes sense to change latexCommandName().


The engine use I was referring to was in the `translator.h`:
```
    virtual QCString latexCommandName()
    {
      QCString latex_command = Config_getString(LATEX_CMD_NAME);
      if (latex_command.isEmpty()) latex_command = "latex";
      if (Config_getBool(USE_PDFLATEX))
      {
        if (latex_command == "latex") latex_command = "pdflatex";
      }
      return latex_command;
    }
```
and in the `translator_ke.h`:
```
    virtual QCString latexCommandName()
    {
      QCString latex_command = Config_getString(LATEX_CMD_NAME);
      if (latex_command.isEmpty()) latex_command = "latex";
      if (Config_getBool(USE_PDFLATEX))
      {
        if (latex_command == "latex") latex_command = "xelatex";
      }
      return latex_command;
    }

```

wosrediinanatour:
> 
> 
> > Test: Introduce an LaTeX-erlor in refman.tex (e.g. a \asfdsggsdfd). Then the make process should stop instead of hanging.
> 
> Wouldn't setting the configuartion `LATEX_BATCHMODE=YES` serve the same purpose?
> 
> > This is not another engine, only additional command lines. So I don't think it makes sense to change latexCommandName().
> 
> The engine use I was referring to was in the `translator.h`:
> 
> ```
>     virtual QCString latexCommandName()
>     {
>       QCString latex_command = Config_getString(LATEX_CMD_NAME);
>       if (latex_command.isEmpty()) latex_command = "latex";
>       if (Config_getBool(USE_PDFLATEX))
>       {
>         if (latex_command == "latex") latex_command = "pdflatex";
>       }
>       return latex_command;
>     }
> ```
> 
> and in the `translator_ke.h`:
> 
> ```
>     virtual QCString latexCommandName()
>     {
>       QCString latex_command = Config_getString(LATEX_CMD_NAME);
>       if (latex_command.isEmpty()) latex_command = "latex";
>       if (Config_getBool(USE_PDFLATEX))
>       {
>         if (latex_command == "latex") latex_command = "xelatex";
>       }
>       return latex_command;
>     }
> ```

You are right with both. I improved it at least for pdflatex now.

wosrediinanatour:
> 
> 
> > Test: Introduce an LaTeX-erlor in refman.tex (e.g. a \asfdsggsdfd). Then the make process should stop instead of hanging.
> 
> Wouldn't setting the configuartion `LATEX_BATCHMODE=YES` serve the same purpose?

Yes, it usuallly writes `\batchmode` into the LaTeX header. But, if `LATEX_HEADER` is defined/set, then a custom header section is used without batchmode. Anyway, using command line parameters I find better.

> 
> > This is not another engine, only additional command lines. So I don't think it makes sense to change latexCommandName().
> 
> The engine use I was referring to was in the `translator.h`:
> 
> ```
>     virtual QCString latexCommandName()
>     {
>       QCString latex_command = Config_getString(LATEX_CMD_NAME);
>       if (latex_command.isEmpty()) latex_command = "latex";
>       if (Config_getBool(USE_PDFLATEX))
>       {
>         if (latex_command == "latex") latex_command = "pdflatex";
>       }
>       return latex_command;
>     }
> ```
> 
> and in the `translator_ke.h`:
> 
> ```
>     virtual QCString latexCommandName()
>     {
>       QCString latex_command = Config_getString(LATEX_CMD_NAME);
>       if (latex_command.isEmpty()) latex_command = "latex";
>       if (Config_getBool(USE_PDFLATEX))
>       {
>         if (latex_command == "latex") latex_command = "xelatex";
>       }
>       return latex_command;
>     }
> ```

Yes, it is better there.

I have pushed an update.

src/latexgen.cpp
```diff
@@ -373,7 +377,7 @@ static void writeLatexMakefile()
   t << endl
     << "clean:" << endl
     << "\trm -f " 
-    << "*.ps *.dvi *.aux *.toc *.idx *.ind *.ilg *.log *.out *.brf *.blg *.bbl refman.pdf" << endl;
+    << "rm -f $(CLEAN_FILES)" << endl;
```
albert-github: Look a bit double here, shouldn't this just be `\t"rm -f $(CLEAN_FILES)" << endl;`

src/latexgen.cpp
```diff
@@ -346,11 +349,12 @@ static void writeLatexMakefile()
...
   else // use pdflatex for higher quality output
   {
-    t << "LATEX_CMD=" << latex_command << endl
+    t << "LATEX_CMD:=" << latex_command << endl
       << endl;
     t << "all: refman.pdf" << endl << endl
       << "pdf: refman.pdf" << endl << endl;
-    t << "refman.pdf: clean refman.tex" << endl;
+    t << "refman.pdf: refman.tex" << endl;
+    t << "\trm -f $(CLEAN_FILES)" << endl;
```
albert-github: Why removing here the "clean" dependency and thus duplicating the code?

src/latexgen.cpp
```diff
@@ -315,7 +317,8 @@ static void writeLatexMakefile()
       << endl;
     t << "refman.pdf: refman.ps" << endl;
     t << "\tps2pdf refman.ps refman.pdf" << endl << endl;
-    t << "refman.dvi: clean refman.tex doxygen.sty" << endl
+    t << "refman.dvi: refman.tex doxygen.sty" << endl
+      << "\trm -f $(CLEAN_FILES)" << endl
```
albert-github: Why removing here the "clean" dependency and thus duplicating the code?

src/translator.h
```diff
@@ -71,6 +71,10 @@ class Translator
       {
         if (latex_command == "latex") latex_command = "pdflatex";
       }
+      if (Config_getBool(LATEX_BATCHMODE))
+      {
+        latex_command +=  " -interaction=batchmode -halt-on-error";
+      }
```
albert-github: Why using `" -interaction=batchmode -halt-on-error";` and not `" -interaction=errorstopmode";`? (I don't know whether or not there is a difference between these 2 implementations).

Implementation should also be in:
- translator_kr.h
- translator_vi.h

src/latexgen.cpp
wosrediinanatour: yes... :-/

src/latexgen.cpp
wosrediinanatour: In general, there shouldn't be a prerequisite `clean` - since if executing make by e.g. `make -j 4` for parallel jobs, then target `clean` may be executed at last, and hence important files could be removed as well.

Here, it does not matter... but I think it is a better style. But of course, it is only my opinion.

src/latexgen.cpp
wosrediinanatour: See above.

src/translator.h
wosrediinanatour: > 
> 
> Why using `" -interaction=batchmode -halt-on-error";` and not `" -interaction=errorstopmode";`? (I don't know whether or not there is a difference between these 2 implementations).

Important notes from http://tug.org/texinfohtml/latex2e.html
 1.  In errorstop mode (the default), TeX stops at each error and asks for user intervention. In batch mode it **prints nothing on the terminal**, errors are scrolled as if the user hit <return> at every error, and missing files cause the job to abort. 
 2.  -halt-on-error **Stop processing at the first error.** 

Ad 1: Useful, because one sees anything in the refman.log. 
Ad 2: Then also "make" terminates with an error statement. Useful, if make is used in other Makefiles or Bash scripts.

 

> 
> Implementation should also be in:
> 
>     * translator_kr.h
> 
>     * translator_vi.h

Ok!

Issue #7360, opened 2019-10-30
use native tooltips


simevo:
avoid performance issues (#6934) and drop altogether the dependency on jquery-powertip using the `title` attribute on anchors as in:
```
<a title="Text of the tooltip 999" href="file1.html">other file 999</a><br/>
```

doxygen:
For pure text only tooltips this would work fine. If the tooltips also need to contain hyperlinks and markup however (as they do) I don't see how `title` is sufficient.

simevo:
Uhm if that's the requirement, I agree that a plugin is necessary. Also on second thought native tooltips are bad for screen readers and on mobile ...

Another option would be to implement tooltips with [Tippy.js](https://atomiks.github.io/tippyjs/) which relies on popper.js (also used by bootstrap4) but does not pull in jquery (for #7363).
I have set up a performance test [here](https://gitlab.com/simevo/tippyjs-perf-test), it is 15% slower than powertip which seems acceptable to me.

In that test I have put the HTML content for the tootip inside `data-tippy-content` attribute:
```html
<a
  class="tooltip"
  href="file1.html"
  data-tippy-content="<div class='ttc' id='afile1_html'><div class='ttname'>Text of the <b>tooltip 1</b></div></div>">
  other file 1
</a><br/>
```
this simplifies both markup and javascript:
```javascript
tippy('.tooltip');
```
but requires using single hyphens for quoting inside the double-hyphen-quoted string ...

doxygen:
@simevo I've dropped the dependency on jquery-powertip now.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in Doxygen version 1.17.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed, and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7562, opened 2020-02-02
WIP: fix #7361 and #7360

### LINKED ISSUE #7361: manage JavaScript dependencies the right way
ATM there are several 3rd-party JavaScript libraries copied in `jquery` dir and the update process is manual.

The right way to manage JavaScript dependencies is to delete all third-party library copies from this source, and have a `package.json` file in the root with something like:
```json
{
  "dependencies": {
    "jquery": "^3.4.1",
    "jquery-powertip": "^1.3.1",
    "jquery-ui-dist": "^1.12.1",
    "jquery-ui-touch-punch": "^0.2.3",
    "jquery.scrollto": "^2.1.2",
    "smartmenus": "^1.1.0"
  }
}
```
(the "caret range" `^x.y.z` means “_Compatible with version x.y.z and all updates that do not modify the left-most non-zero digit in the [x,y,z] tuple_”, see https://docs.npmjs.com/files/package.json#dependencies and https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004)

At build time, `npm install` or `yarnpkg install` will create a `node_modules` dir and pull all dependencies in there (the `node_modules` should be in `.gitignore`)

The current `jquery/Makefile` should be modified to pick the libraries from their new location in `node_modules` dir; if you need to patch them, patches can be included and applied by the `Makefile`.

When this machinery is in place, upgrading is as easy as `npm upgrade` or `yarnpkg upgrade` and then `make`.
If a new version of jquery is available, for example 3.5.0, that will be used instead of version 3.4.1 required in `package.json` (because the caret range allows it).

The actual version used of each library is stored in `package-lock.json` or `yarn.lock`; it is best to also put these files under source control, so that whoever downloads this source and builds it gets the same versions of all JavaScript libraries.

simevo:
this PR:
- avoids redistributing third-party JavaScript dependencies in the doxygen source-tree
- renames the JavaScript bundle to bundle.js to avoid confusion (the currently shipped "jquery.js" contains a lot more than just jquery)
- manages JavaScript dependencies the right way
- makes it easier to upgrade JavaScript dependencies in the future
- upgrades the JavaScript toolchain from obsolete compass yui-compressor

TODO:
- upgrade generated HTML to actually use tippy.js
- minify CSS
- test !
- use webpack to bundle JS ?


doxygen:
@simevo This pull request has become obsolete now that jquery has been removed altogether.

Issue #7570, opened 2020-02-04
Comments replaced by block comments inside Markdown code block


kuukunen:
(Windows 10, Doxygen version 1.8.17 (b5fa3cd1c6e6240e20d3b80a70e3f04040b32021*))

I had a weird issue where Doxygen seems to ignore almost everything inside a &grave;&grave;&grave; markdown code block in a .md documentation file... except it somehow replaces multiline /// comments with /** block comments */

Basically in my CodeConvention.md file this:

&grave;&grave;&grave;
/// Brief explanation of a function.
/// More explanation.
void Function1();

/// FF.
/// FF.
/&ast;&ast;\verbatim
Verbatim block
\endverbatim&ast;/
void Function2();

/// Brief explanation of a function.
void Function3();

/// Brief explanation of a function.
/// More explanation.
void Function4();
&grave;&grave;&grave;

becomes:

```
/// Brief explanation of a function.
/// More explanation.
void Function1();
 
/// FF.
/// FF.
/**\verbatim
Verbatim block
\endverbatim*/
void Function2();
 
/// Brief explanation of a function.
void Function3();
 
/** Brief explanation of a function.
 *  More explanation. */
void Function4();
```

Note the comments before Function4.

This seems to happen to multiline comments that appear after a /** block comment */ (The part with \verbatim, in this case, but doesn't need any commands inside)

kuukunen:
I don't know if this should be a completely new issue, but the block comment also somehow breaks aliases.

With:  `ALIASES                += ns{1}="\ref NS::\1 \"\1\""`

This:

---

\ref NS::Function
\ns{Function}

&grave;&grave;&grave;
/// something
&grave;&grave;&grave;

\ref NS::Function
\ns{Function}

&grave;&grave;&grave;
/** something */
&grave;&grave;&grave;

\ref NS::Function
\ns{Function}

---

Becomes:

[NS::Function](dummy_link_to_Function) [Function](dummy_link_to_Function)

```
/// something
```

[NS::Function](dummy_link_to_Function) [Function](dummy_link_to_Function)

```
/** something */
```

[NS::Function](dummy_link_to_Function) \ns{Function}

kuukunen:
On a bit further investigation, it seems this is a duplicate of #7117

albert-github:
I've just pushed a proposed patch, pull request #7572, for the 2 mentioned problems.

[...]

PR #7572, opened 2020-02-04
issue #7570 Comments replaced by block comments inside Markdown code block

albert-github:
Fenced codeblocks with backticks or tildes should be handled in a similar way as the `@code`  (etc.) command .
Due to the changed handling of fenced code blocks where is a small interaction with blocks starting with spaces, to overcome this those blocks have been rewritten (and are compatible with the old situation).
(Note; Proper of "spaced blocs" would require quite a bit of effort as spaces can be dependent on the context (not all "spaced blocks" are "spaced blocks" but can be a natural indentation for commands.)

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4154757/example.tar.gz)

### DIFF
--- a/doc/markdown.doc
+++ b/doc/markdown.doc
@@ -401,11 +401,13 @@ more tilde (~) characters on a line. The end of the block should have the
 same number of tildes. Here is an example:
 
 
-    This is a paragraph introducing:
+@verbatim
+This is a paragraph introducing:
 
-    ~~~~~~~~~~~~~~~~~~~~~
-    a one-line code block
-    ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


doxygen:
@albert-github I've push a [simpler fix](https://github.com/doxygen/doxygen/commit/137328681e7dcbd07a5ec0e4b068aea96ec41b4a) that does not have the side-effects you mentioned.
It basically prevents interpreting `*/` inside markdown. Let me know if you still see issues with this change that your fix would cover.

albert-github:
Fixed by means of separate commits: 1373286 and 3cbb8b8b56cb56f9cca84740baa065cb07997f90

PR #7574, opened 2020-02-07
Linking in PHP

albert-github:
In PHP the scope separator is the double backslash, but this was not considered here.
When having e.g.
```
@ref DigitalPublications\MidasWPX\Interfaces\PathsManagerInterface "PathsManagerInterface"
```
we got the message:
```
 warning: unable to resolve reference to 'DigitalPublications\MidasWPX\Interfaces\PathsManagerInterface' for \ref command
```
whilst the version with the double colon was OK.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4170711/example.tar.gz)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -4744,6 +4744,7 @@ bool resolveLink(/* in */ const char *scName,
   *resContext=0;
 
   QCString linkRef=lr;
+  QCString linkRefPHP=substitute(lr,"\\","::"); // PHP substitution
   QCString linkRefWithoutTemplates = stripTemplateSpecifiersFromScope(linkRef,FALSE);
   //printf("ResolveLink linkRef=%s\n",lr);
   const FileDef  *fd;
@@ -4796,6 +4797,12 @@ bool resolveLink(/* in */ const char *scName,
     return TRUE;
   }
   else if ((cd=getClass(linkRef))) // class link
+  {
+    *resContext=cd;
+    resAnchor=cd->anchor();
+    return TRUE;
+  }
+  else if ((cd=getClass(linkRefPHP)) && (SrcLangExt_PHP==cd->getLanguage())) // class link
   {
     *resContext=cd;
     resAnchor=cd->anchor();


src/util.cpp
```diff
@@ -4801,6 +4802,12 @@ bool resolveLink(/* in */ const char *scName,
     resAnchor=cd->anchor();
     return TRUE;
   }
+  else if ((cd=getClass(linkRefPHP)) && (SrcLangExt_PHP==cd->getLanguage())) // class link
```
andriokha, line 4805:
```suggestion
  else if ((cd=getClass(linkRefPHP)) && (SrcLangExt::PHP==cd->getLanguage())) // class link
```
This has changed in the meantime.

andriokha:
Thanks, I've tested locally and this appears to let me link to PHP classes and methods using the normal namespace separator in markdown and normal code comments, but **not** in `@see`. If I try to reference a class `B` from class `A`, both in namespace `\Foo\somit`, this is what I'm seeing (I'm sure not all of these are supposed to work, just trying to provide as much detail as I can!):
```php
<?php

declare(strict_types=1);

namespace Foo\somit;

use Foo\C;

/**
 * Test class A.
 */
class A {

  /**
   * Do something.
   *
   * - Inline @ref B reference.                                       <- no link
   * - Inline @ref Foo\somit\B reference.
   * - Inline @ref \Foo\somit\B reference.                            <- no link
   * - Inline @ref Foo.somit.B reference.
   * - Inline @ref Foo::somit::B reference.
   * - Inline @ref Foo\somit\B::doSomit() reference.                  <- no link
   * - Inline @ref \Foo\somit\B::doSomit() reference.                 <- no link
   * - Inline @ref Foo\somit\B::doSomit reference.                    <- no link
   * - Inline @ref \Foo\somit\B::doSomit reference.                   <- no link
   * - Inline @ref Foo.somit.B::doSomit() reference.
   * - Inline @ref Foo::somit::B::doSomit() reference.
   *
   * - Inline @link B link @endlink more text.                        <- link to current page
   * - Inline @ref Foo\somit\B reference.
   * - Inline @ref \Foo\somit\B reference.                            <- no link
   * - Inline @link Foo.somit.B link @endlink more text.
   * - Inline @link Foo::somit::B link @endlink more text.
   * - Inline @link Foo\somit\B::doSomit() link @endlink more text.   <- link to current page
   * - Inline @link \Foo\somit\B::doSomit() link @endlink more text.  <- link to current page
   * - Inline @link Foo\somit\B::doSomit link @endlink more text.     <- link to current page
   * - Inline @link \Foo\somit\B::doSomit link @endlink more text.    <- link to current page
   * - Inline @link Foo.somit.B::doSomit() link @endlink more text.
   * - Inline @link Foo::somit::B::doSomit() link @endlink more text.
   *
   * @see B                                                           <- no link
   * @see Foo\somit\B                                                 <- no link
   * @see \Foo\somit\B                                                <- no link
   * @see Foo.somit.B
   * @see Foo::somit::B
   * @see Foo\somit\B::doSomit()                                      <- no link
   * @see \Foo\somit\B::doSomit()                                     <- no link
   * @see Foo\somit\B::doSomit                                        <- no link
   * @see \Foo\somit\B::doSomit                                       <- no link
   * @see Foo.somit.B::doSomit()
   * @see Foo::somit::B::doSomit()
   */
  public function doSomit(B $b, C $c): void {
    $b->doSomit();
  }

}
```

andriokha:
Condensing from the above comment, it seems to me in an ideal world we'd be able to use a `\` separator to reference
- a method: at the moment that still seems to need the `.` or `::` separators
- a class in `@see` (a workaround is `@see @ref Foo\somit\B`)

Thanks!

albert-github:
I've updated (merged) this pull request so it us fully up to date (I hope) with the current master.

I ran your example but I get quite a few warnings and not many links present. Also I don't see anywhere a class / member / variable called `B` defined.

As this is a proposed pull request and you identified #8384 as a possible description of this PR best is to:
- attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem? Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

So we are talking about the same problem.

andriokha:
Thanks @albert-github, really appreciate your time on this, I'll try and post that today or tomorrow. I also think that auto linking wasn't working correctly, but I'll get it all in the repro.

Issue #7444, opened 2019-12-13
Support for deferred constants and package bodies  in general for VHDL


Prandr:
This one is more missing feature than a bug. 

In VHDL it is possible to have [_deferred constants_](https://www.doulos.com/knowhow/vhdl_designers_guide/tips/deferred_constants/): these are constants which are declared in the declarative part of the package, and are initialised in the package body (somewhat similar to C(++), where a variable is declared in a header file, but initialised in a source file). Doxygen ignores that, because it , as it looks like, doesn't look in the package bodies at all.

Generally it would be nice if doxygen looked in the package bodies and merge documentation for constants and functions, similarly how in  C(++) you can have brief description in the header file and detailed description in the source file.

albert-github:
- Please always attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

Prandr:
Alright here is the familiar piece of code. I added a detailed description to the function (although it isn't really necessary) to better illustrate the issue
[classinterface__pkg.zip](https://github.com/doxygen/doxygen/files/3961890/classinterface__pkg.zip)

albert-github:
A proposed patch has been pushed, pull request #7593

PR #7593, opened 2020-02-14
merge function/procedure/constants docs

maddox11:
this patch merges the VHDL  documentations for
 function/procedure/constants see #7444

### DIFF
--- a/src/vhdldocgen.cpp
+++ b/src/vhdldocgen.cpp
@@ -85,6 +85,11 @@ static void writeTable(QList<MemberDef>* port,FTextStream & t);
 static void endTable(FTextStream &t);
 static void writeClassToDot(FTextStream &t,ClassDef* cd);
 static void writeVhdlDotLink(FTextStream &t,const QCString &a,const QCString &b,const QCString &style);
+
+static void mergeDoc(const MemberDef* mdef);
+static MemberDef* findMemberFunction(const MemberDef* md,const QCString& package);
+  
+
 //static void writeVhdlPortToolTip(FTextStream& t,QList<MemberDef>* port,ClassDef *cd);
 static const MemberDef *flowMember=0;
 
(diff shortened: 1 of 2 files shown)


vhdlparser/vhdlparser.jj
```diff
@@ -1012,14 +1012,15 @@ QCString entity_class() :  {}
 | <TYPE_T> {return "type";}
 | <SUBTYPE_T> {return "subtype";}
 | <CONSTANT_T> {return "constant";}
-| <SIGNAL_T> {return "signal";}
```
albert-github: Why is `SIGNAL_T` removed here as I see it is still used on other places?used

src/vhdldocgen.cpp
```diff
@@ -200,7 +205,7 @@ void VhdlDocGen::writeOverview()
 
   if (!f.open(IO_WriteOnly))
   {
-    err("Warning: Cannot open file %s for writing\n",fileName.data());
+    fprintf(stderr,"Warning: Cannot open file %s for writing\n",fileName.data());
```
albert-github: Why is `err` replaced by `fprintf(stderr`, the `err` is the default way doxygen handles error

vhdlparser/vhdlparser.jj
maddox11: This is the correct grammer from VHDL 2008

entity_class:
    | entity
    | architecture
    | configuration
    | procedure
    | function
    | package
    | type
    | subtype
    | constant
    | variable
    | component
    | label
    | literal
    | units
    | group
    | file
    | property
    | sequence

src/vhdldocgen.cpp
maddox11: OOps that was my mistake. 
err is correct

albert-github:
@maddox11 any special reason for closing this proposed pull request?

PR #7611, opened 2020-02-24
Possibility to show non documented namespaces

albert-github:
With classes (though quite different from namespaces) we have to possibility to show non-documented classes or not (`HIDE_UNDOC_CLASSES`), with namespaces the non-documented names spaces are not shown (only when using `EXTRACT_ALL`).

- Creating possibility to hide / show all non-documented namespaces (default set so that compatibility with old situation is preserved).
- adjusting test cases so that they work with and without `HIDE_UNDOC_NAMESPACES`).

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -923,6 +923,17 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+    <option type='bool' id='HIDE_UNDOC_NAMESPACES' defval='1'>
+      <docs>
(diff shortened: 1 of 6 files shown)


doxygen:
> Should I create a new PR for the doxygen.cpp enhancement? (including fixing some typos there)

That's probably the easiest.

albert-github:
OK I'll have a look at it tomorrow.

albert-github:
Obsolete due to #7611 and #11143

PR #7614, opened 2020-02-24
default value of conditional path is 1

joenio:
this relates to analizo/analizo Cyclomatic Complexity metric

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -1414,7 +1414,7 @@ class MemberDefImpl::IMPL
     QCString declFileName;
     int declLine = 0;
     int declColumn = 0;
-    int numberOfFlowKW = 0;
+    int numberOfFlowKW = 1;
 };
 
 MemberDefImpl::IMPL::IMPL() :
@@ -1426,7 +1426,7 @@ MemberDefImpl::IMPL::IMPL() :
     categoryRelation(0),
     declLine(-1),
     declColumn(-1),
-    numberOfFlowKW(0)
+    numberOfFlowKW(1)
 {
 }
 


src/memberdef.cpp
```diff
@@ -1414,7 +1414,7 @@ class MemberDefImpl::IMPL
     QCString declFileName;
     int declLine = 0;
     int declColumn = 0;
-    int numberOfFlowKW = 0;
+    int numberOfFlowKW = 1;
 };
```
albert-github: - What is the first level of the numberOfFlow keywords?
- Do you have a reference why the default should be 1?

joenio: Hi, your question makes me discover that default value should be 0 in fact, then this merge request can be rejected and keeps number of conditionals path defaults to zero, I'm going to explain the confusion.

The original authors of the Cyclomatic Complexity (CC) metric on Analizo added the calculation on Doxyparse fork as starting by 1 as default, the CC can be calculated by the number of predicates (i.e., decision points) plus one, in this case the number of predicates is the same of conditional paths. I don't know why original authors prefers to count this "plus one" on the Doxyparse side, but now I see the right way is to count "plus one" on conditional paths on the Analizo side.

Follow some references about the subject:

1979, MARTIN R. WOODWARD, MICHAEL A. HENNELL, AND DAVID HEDLEY: A Measure of Control Flow Complexity in Program Text. Follow a excerpt on the specific point on this paper about this:

```
An interesting link exists between this last measure and the
intellectually very appealing cyclomatic complexity measure
V(G) of McCabe [10]. A program is first represented as a di-
rected graph and then:

V(G) = number of edges - number of nodes + 2.

McCabe has shown that this number equals the number of
predicates (i.e., decision points) in the program plus one.
```

The original author and paper of the Cyclomatic Complexity is:

1976, THOMAS J. MCCABE, A Complexity Measure

Thanks for your time @albert-github

albert-github:
Seen the comment with the code review, closing the pull request.

PR #7615, opened 2020-02-24
print protection information on doxyparse

joenio:
closes analizo/analizo#136

### DIFF
--- a/addon/doxyparse/doxyparse.cpp
+++ b/addon/doxyparse/doxyparse.cpp
@@ -236,6 +236,21 @@ static void referenceTo(MemberDef* md) {
   printReferenceTo(type, signature, defined_in);
 }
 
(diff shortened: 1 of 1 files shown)


joenio:
this add more information about protection of symbols, today doxyparse just prints this kind of information for "public" functions, variables and so on, this commit add to the output if simbols are "public", "protected", "private" or "package" as doxygen already have this kind of information, see a little output example, the two lines related to this commit are detached for the char **<**

```yaml
hello_world.cc:
  "HelloWorld":
    defines:
      - "destroy()":
          type: function
          line: 12
          protection: public              <
          lines_of_code: 3
          conditional_paths: 0
      - "private_method()":
          type: function
          line: 8
          protection: private             < added: private
          lines_of_code: 3
          conditional_paths: 0
```

Issue #7060, opened 2019-06-17
Function pointer typedefs render strangely in HTML documentation


kozross:
I have a header file with the following (notable) contents:

```c
// some not-important stuff
/**
 * @brief An allocator function, suitable for passing to the various allocation
 * routines in this header.
 *
 * @note This (deliberately) has the same type signature as @c malloc. Any
 * function used for this purpose should be morally similar to @c malloc in its
 * behaviour; in particular, it should return @c NULL when allocation fails (for
 * whatever reason).
 */
typedef void *(*ds9k_bv_allocator_function)(size_t);

/**
 * @brief A deallocator function, suitable for passing to the various
 * deallocation routines in this header.
 *
 * @note This (deliberately) has the same type signature as @c free. Any
 * function used for this purpose should be morally similar to @c free in its
 * behaviour.
 */
typedef void (*ds9k_bv_deallocator_function)(void *);

// some other not-important stuff
```

I render this into some documentation using the Doxyfile in the attached zip. However, when I open the documentation in my browser (see image in the attached zip), the results look strange. Am I misconfiguring something?

I have attached the details of this report: [bug-report-details.zip](https://github.com/doxygen/doxygen/files/3295426/bug-report-details.zip)

albert-github:
I assume that you mean that in the part `typedef void *(* ` and `typedef void(*`  the placing of the `*(*` and `(*`  looks a bit strange. 

I don't think you configured something wrong.

kozross:
@albert-github Yeah, that's exactly what I meant.

gregwillits:
I'm seeing problems with `typdef void` as well with 1.8.16, but not with 1.8.10.

My example has these two lines:

```c
typedef const char m_msgEvtName;
typedef void (*mMsgCntrSubscriberAction)(const void* theContext);
```

The dox comments are at the end of the file:

```c
/*!
 * @typedef typedef const char m_msgEvtName
 * m_msgEvtName is the data type for message names of broadcasted messages. The message name
 * itself is a string constant. The purpose of using m_msgEvtName is to give message name
 * parameters some semantic identification.
 */

/*!
 * @typedef typedef void (*mMsgCntrSubscriberAction)(const void* theContext)
 * Message subscriber callback prototype. The subscriber's callback function must follow this prototype.
 */
```

Two consequences:

[...]

PR #7622, opened 2020-02-29
issue #7060 Function pointer typedefs render strangely in HTML documentation

albert-github:
typedefs like:
```
typedef void *(*pMsgCntrSubscriberAction)(const void* theContext);
```
place the `(*` part with the type, though it should be in front of the name., especially in HTML this gives a strange layout..
Splitting of this part and show it just before the name instead of with the type. The closing `)` has been split off and added to the arguments before.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4270344/example.tar.gz)

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -2308,14 +2308,29 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
 
   // *** write type
   QCString ltype(m_impl->type);
+  QCString preTypedefString;
   if (isTypedef() && getLanguage() != SrcLangExt_Slice)
   {
     ltype.prepend("typedef ");
+    if (m_impl->args.at(0) == ')') // to be on the safe side and to exclude variable like "int (*bla)[10]"
+    {
+      int lastOpen = -1;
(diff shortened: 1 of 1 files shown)


doxygen:
The idea with rendering it with `(*` as part of the type is to make sure all member names are nicely aligned:

Screenshot 2020-07-28 at 11 24 35

If `(*` is made of part of the name this alignment is broken.

Given that
```.c++
typedef void *(*pMsgCntrSubscriberAction)(const void* theContext);
```

in equalent to
```.c++
using pMsgCntrSubscriberAction = void*(*)(const void* theContext);
```
one can argue that `(*` is really part of the type.

albert-github:
Something that looks quite inconsistent, in my opinion, is that there is an opening bracket in the type and the closing bracket is in the name, so neither type nor name are "complete".

doxygen:
I agree with the inconsistency, this is one of the ugly parts of C in my opinion, and probably also the reason why C++11 introduced the `using` construct. The name is really in the "middle" of the type, splitting it into two parts.

Issue #7626, opened 2020-03-03
"warning: documentation for unknown define ... found" without newline at end of file


tttapa:
When a macro is defined at the bottom of a header file, without a newline following its definition, doxygen issues a warning, and doesn't add the macro to the documentation.

[Doxygen-7625.zip](https://github.com/doxygen/doxygen/files/4284094/Doxygen-7625.zip)

Steps to reproduce: 
1. Download and extract ZIP (contains a header file and the default Doxyfile generated using `doxygen -g`).
2. Run `doxygen`.
3. Observe warning `~/Doxygen-7625/Test.hpp:7: warning: documentation for unknown define TEST found.`
4. Open the HTML documentation in a browser and observe that the documentation for the `TEST` macro doesn't exist.

When adding a newline at the end of `Test.hpp`, it does work.

Tested using doxygen 1.8.18 (fbebfae209139fccb15fde8fb7d1bdbb4b941875) and doxygen 1.8.16 on Ubuntu 18.04.

By no means a critical bug, but it took me longer than I'd like to admit before I figured out why my macro wasn't showing up in the documentation :)

albert-github:
Out of curiosity, which editor did you use to create this type of file?

albert-github:
I've just pushed a proposed patch, pull request #7628.

tttapa:
Thanks!

> Out of curiosity, which editor did you use to create this type of file?

I used Visual Studio Code. There's an option to automatically add a new line at the end of the file when saving, but it's not enabled by default.

albert-github:
Retested with doxygen 1.8.18 and got the warning as indicated.
Tested with the cu5rrent doxygen version (1.15.0) and the warning is gone (and the define is properly documented.

I think the problem has been solved in the mean time.

PR #7628, opened 2020-03-05
issue #7626 "warning: documentation for unknown define ... found" without newline at end of file

albert-github:
In case of non preprocessed code there was already a `\n` added when not present at the end. This has now been done in case of preprocessing as well.

The current code had a small flaw as it added the `\n` after the last character and this character could be a `\0` so we has to backup a little.

### DIFF
--- a/src/bufstr.h
+++ b/src/bufstr.h
@@ -98,6 +98,10 @@ class BufStr
     { 
       return m_writeOffset; 
     }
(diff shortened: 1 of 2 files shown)


albert-github:
Retesting with the current doxygen version (1.15.0) showed that the problem was solved in the mean time.

This proposed pull request is obsolete.

PR #7636, opened 2020-03-12
Fix C compiler warnings about C++ flags

lrineau:
This patch will fix the following warnings:
```
cc1: warning: command line option ‘-Wnon-virtual-dtor’ is valid for C++/ObjC++ but not for C
cc1: warning: command line option ‘-Woverloaded-virtual’ is valid for C++/ObjC++ but not for C
```

... using the generator expression `$<COMPILE_LANGUAGE:CXX>` to enable the warning flags only for C++ compilation units, and not for C.

### DIFF
--- a/BUILD.txt
+++ b/BUILD.txt
@@ -1,5 +1,5 @@
 Doxygen uses cmake (http://www.cmake.org/) to build executables for various platforms.
-It's required at least cmake version 3.1.3
+It's required at least cmake version 3.3.
 
 The first step is to create a build directory where the output should be stored.
 Doxygen can be fully build outside of the source tree.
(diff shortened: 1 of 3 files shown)


lrineau:
Note that I have only tested the patch with gcc. Hopefully, your CI will check with clang as well.

albert-github:
I did a quick check on one of the clang compilations and there were no new warnings.
It looks like the mentioned options are in case of clang accepted for the plain C compiler part as well as the Cpp part.

There are other compilation warnings that have to be checked as well, but that is outside the scope of this fix.

Regarding the error regarding `latex` and friends this also has nothing to do with this fix.

albert-github:
The problem with `latex` has been identified by travis as being on their side and they have made a fix (not yet tested by me, just a restart of the failing jobs won't help according to travis support).

lrineau:
I have added two commits, to state that CMake version 3.3 or later is required.

albert-github:
Good catch, also tested the travis update this works as well.

PR #7656, opened 2020-03-21
Use cstdint to define fixed width typeString

cmorty:
On some systems int64 wasn't 64 bit

### DIFF
--- a/qtools/qglobal.h
+++ b/qtools/qglobal.h
@@ -1,5 +1,5 @@
 /****************************************************************************
-** 
+**
 **
 ** Global type declarations and definitions
 **
@@ -38,6 +38,7 @@
 #ifndef QGLOBAL_H
 #define QGLOBAL_H
 
+#include <cstdint>
 
 #define QT_VERSION	223
 #define QT_VERSION_STR	"2.2.3"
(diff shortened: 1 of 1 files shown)


cmorty:
Ok, seems like this "Fix" actually broke something. I'll look into it.

albert-github:
@cmorty,

You wrote:
> On some systems int64 wasn't 64 bit

"some" is a bit vague:
- any special systems in mind?
- what happened on these systems?

cmorty:
Well, actually it was an extra-pedentic gcc that diverted me. My printf-checks (Coming in an upcoming PR) complained that `long unsinged int` (64 bit) isn't `long long unsigned int` (actually 64 bit, too).  
Yet the fact, that my changes broke something, hint at that there must actually be something wrong. I'll look into it.
I'll close this and report back, once I found out more....

cmorty:
Sweet. It seems like I triggered a bug in qcstring. :)

cmorty:
Nope, it's not qcstring, it's global initialisation: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-global-init

cmorty:
Ok, forget this one. Setting
```
const bool FALSE = true;
const bool TRUE = false;
```
is just plain stupid..... *very, very embarrassed*

PR #7657, opened 2020-03-21
Add message format-checking

cmorty:
I added printf-checks  `__attribute__((format(printf,[...]` for the formatted output. Also found some messages that probably never worked and fixed them. :)

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -209,7 +209,7 @@ class Statistics
     Statistics() { stats.setAutoDelete(TRUE); }
     void begin(const char *name)
     {
-      msg(name);
+      msg("%s", name);
       stat *entry= new stat(name,0);
       stats.append(entry);
       time.restart();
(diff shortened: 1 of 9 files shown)


src/doxygen.cpp
```diff
@@ -8379,7 +8379,7 @@ static void findDefineDocumentation(Entry *root)
       else
       {
         warn(root->fileName,root->startLine,
-             "found documented #define but ignoring it because "
+             "found documented #define %s but ignoring it because "
              "ENABLE_PREPROCESSING is NO.\n",
```
albert-github: Looks like the meaning changed here.

cmorty: I assumed that `root->name.data` contains the name of the define. To be honest, I didn't verify that, but assumed it wouldn't make things worse, as file and line are printed, but might help if I was correct. ;)

albert-github: After rechecking, this looks like to be better, warning changed but is better. Looks like due to the minimal view github gives I missed the `root->name.data()` and seen what we see above this is much better.

PR #7665, opened 2020-03-23
WIP / RFC: Cachedir

cmorty:
This is the current, working, state of the Cachedir feature. It currently works as the following:

Test if a file with the md5 hash can be found in the cache-dir. If so copy it and trigger postprocessing (basically a fast dot).
If not, build image and copy it to the cache-dir. Trigger postprocessing.

This creates reproducible results with and without cache-dir.

The CLEANUP-Patches will be removed before merging.

Todos / Open points / RFCs:
 * Create folder if it does not exists. What should be done if this fails? Print a warning and continue without cache?
 * What about postrprocessing? Should the files be copied after being post-processed? As links are patched and I don't have an overview over the whole process I'm not sure whether that is a good idea.
* Currently the images are generated into the target folder and then copied to the cache-dir. This kept the code simpler. Maybe it's a good idea to produce the images with their hash-name into the target-folder or the cache-folder, if is enabled, and than use that file as input for the post-procesing and delete the dot-generated file only in case it's not the cache-folder (This sentence is to long, but I hope it conveys the idea). That would save a move-operation to the temp-file in post-processing.
* In the context of caching: Is using md5 the right hash to use, or is sha1 (easily extracted from mbed-tls) a better option? It shouldn't be too expensive on modern CPUs.
* Documentation must be extended a bit. For example that there is no house-keeping. Anything else I should mention?
* What else should look into, to get this cleanly merged?


.gitignore
```diff
@@ -14,3 +14,6 @@
 /doxygen_docs
 /doxygen.tag
 /build*
+out
+out
```
albert-github: double? maybe also better to name it `doxygen_cache_out` (I assume this is the name you used for your cache directory).

src/configimpl.l
```diff
@@ -3,8 +3,8 @@
  * Copyright (C) 1997-2015 by Dimitri van Heesch.
  *
  * Permission to use, copy, modify, and distribute this software and its
- * documentation under the terms of the GNU General Public License is hereby 
- * granted. No representations are made about the suitability of this software 
+ * documentation under the terms of the GNU General Public License is hereby
+ * granted. No representations are made about the suitability of this software
```
albert-github: The review is at these 2 lines, but accounts for the entire file.
I do agree that the spacing should be consistent in the files, at the moment there is a mixture of tabs and spaces (especially in the lex files), but I don't think this should be "hidden" in other pull requests (and thus hiding real relevant changes for the pul request for easy review), I think it would be better to have a separate pull request for it.

src/configimpl.l
```diff
@@ -1713,6 +1713,30 @@ void Config::checkAndCorrect()
...
+    }
+    else
+    {
+        warn_uncond("The dot cache folder (DOT_CACHEDIR) %s could not be found (ignoring).\n",dotCachedir.data());
+        dotCachedir="";
+    }
+#if defined(_WIN32) // convert slashes
+    auto l=dotCachedir.length();
+    for (int i=0;i<l;i++) if (dotCachedir.at(i)=='/') dotCachedir.at(i)='\\';
+#endif
```
albert-github: This should be handled with `substitute` and `Portable::pathSeparator`

src/dotgraph.cpp
```diff
@@ -158,51 +164,82 @@ bool DotGraph::prepareDotFile()
...
+      }
+    }
   }
 
-  // need to rebuild the image
 
+
+  // need to rebuild the image
   // write .dot file because image was new or has changed
+  // It might be a good idae to write the file in the dot-Runner or even pass it via stdout
```
albert-github: idae <-> idea

.gitignore
cmorty: This is part of CLEANUP and will be removed before the final PR

src/configimpl.l
cmorty: Ok, I can easily move the Commit to its own PR. Will do so in the next days (have to spend some time on my day-job. :-) )

.gitignore
cmorty: Btw out ist the folder VS uses for Cmake projects

src/configimpl.l
cmorty: Ok, this is C&P code from a few lines above. Should those be changed, too (and be put in a separate PR?)

src/dotgraph.cpp
cmorty: Thanks - will fix

src/configimpl.l
albert-github: Amazing that this is in (in my opinion) and the amount of `_WIN32` / `WIN32` still present outside portable.cpp.
In doxygen.cpp I found also `static QCString fixSlashes(QCString &s)` I think a candidate to move to portable.cpp and to be used.

cmorty: @albert-github: Ok, so should I try to migrate the code or are you willing to do so (as you are probably faster. ;) )?

albert-github: Thanks for the confidence. I'll leave this patch to you.

cmorty: Not what I was hoping for. I'll see when I'll get around to it. Worst case I'd leave it like it is for now....

cmorty:
@albert-github  @doxygen ; Can one of you tell me something about post-processing? Especially whether it makes sense to Cache the un-postprocessed or rather postprocessed imgaes. Also where and how much should I document?

cmorty:
@albert-github @doxygen: I closed the open points (see OP). Testcases are open, but I'm not sure what the best approach is. I'm looking forward to your feedback to get this merged.

src/configimpl.l
cmorty: As this is C&P code and this PR is not about fixing that code, I'd keep it the way it is and maybe fix it later (but to be honest; I have other points on my agenda that I'm more interested in improving).

cmorty:
@albert-github @doxygen: Any chance to get this through in a timely manner. Would like to avoid having to rebase it too often. BTW: appveyor is green. Seems like some callback got lost.

PR #7676, opened 2020-03-30
Remove duplicate xsd fields for `docMarkupType`.

jtmcx:
The `docTitleCmdGroup` group defines choices `image` and `linebreak` (among others). The `docMarkupType` type inherits from `docTitleCmdGroup`, but includes redefinitions of `image` and `linebreak` in its definition. This is mostly harmless, but it violates the ["unique particle attribution rule"](https://www.w3.org/wiki/UniqueParticleAttribution).

Some xsd tools (in particular the `xsd` command from [code synthesis](https://www.codesynthesis.com/projects/xsd/)) complain about the duplicate fields, and error out.

This is my first PR to doxygen, so if there is anything additional that you need, please let me know! Thank you!

### DIFF
--- a/templates/xml/compound.xsd
+++ b/templates/xml/compound.xsd
@@ -447,7 +447,6 @@
   <xsd:group name="docCmdGroup">
     <xsd:choice>
       <xsd:group ref="docTitleCmdGroup"/>
-      <xsd:element name="linebreak" type="docEmptyType" />
       <xsd:element name="hruler" type="docEmptyType" />
       <xsd:element name="preformatted" type="docMarkupType" />
       <xsd:element name="programlisting" type="listingType" />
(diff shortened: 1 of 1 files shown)


doxygen:
Your first PR for doxygen has been integrated. Thanks for contributing!

albert-github:
As info:
 For xml testing there are in doxygen the test possibilities:
- make tests
   this is the default test strategy and uses xmllint to test
- make tests TEST_FLAGS=--xmlxsd
  here also xmllint is used

xmllint is apparently a bit less stringent.

- I had a quick look at the mentioned page of code synthesis but I get the impression that the XSD tool is a bit dead, last release in 2014 (version 4.0.0) or are there newer versions?
- Are there any other free (and easy) tools that can be used to test for these kind of problems.

jtmcx:
So sorry I didn't follow up sooner -- I'm not sure there are any good tools to test for these kinds of problems. I did some exploring, and it seems that only the codesynthesis tools complain. Xmlint, [Perl's LibXML](https://metacpan.org/pod/distribution/XML-LibXML/lib/XML/LibXML/Schema.pod), [HaXML](https://hackage.haskell.org/package/HaXml), and [go xsd](https://godoc.org/aqwari.net/xml/xsd) all seem okay with duplicate fields.

You're right that the codesyntesis xsd tool is a bit dead -- 4.0 is the latest version. Most other xsd tools seem to be a little stale, so it's hard to find one that's fully featured and regularly updated aside from xmllint.

PR #7679, opened 2020-03-31
Fix html file ext in external docs

avostrik:
This pull request solves following problem:

- project A generates tag list
- project B references external documentation from project B

In case of enabled tree view and 'Files List' entry in it, all links to project A are invalid, following pattern `somefile_8hpp.html.html`.

Root cause is in that `node2URL` always adds html extension, even for filenames that come from tag file.

Looking how htmlFileExtension is used, fixed several inconsistencies in writeTagFile methods and fixed missed extension for page definition filename output.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -10535,6 +10535,7 @@ static void writeTagFile()
                      << "</title>" << endl
                      << "    <filename>"
                      << convertToXML(Doxygen::mainPage->getOutputFileBase())
+                     << Doxygen::htmlFileExtension
                      << "</filename>" << endl;
 
     mainPage->writeDocAnchorsToTagFile();
--- a/src/filedef.cpp
+++ b/src/filedef.cpp
@@ -349,7 +349,7 @@ void FileDefImpl::writeTagFile(FTextStream &tagFile)
   tagFile << "  <compound kind=\"file\">" << endl;
   tagFile << "    <name>" << convertToXML(name()) << "</name>" << endl;
   tagFile << "    <path>" << convertToXML(getPath()) << "</path>" << endl;
-  tagFile << "    <filename>" << convertToXML(addHtmlExtensionIfMissing(getOutputFileBase())) << "</filename>" << endl;
+  tagFile << "    <filename>" << convertToXML(getOutputFileBase()) << Doxygen::htmlFileExtension << "</filename>" << endl;
   if (m_includeList && m_includeList->count()>0)
   {
     QListIterator<IncludeInfo> ili(*m_includeList);
--- a/src/ftvhelp.cpp
+++ b/src/ftvhelp.cpp
@@ -259,7 +259,7 @@ static QCString node2URL(const FTVNode *n,bool overruleFile=FALSE,bool srcLink=F
         url = fd->getOutputFileBase();
       }
     }
-    url+=Doxygen::htmlFileExtension;
+    url = addHtmlExtensionIfMissing(url);
     if (!n->anchor.isEmpty()) url+="#"+n->anchor;
   }
   return url;
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -4544,7 +4544,7 @@ void MemberDefImpl::writeTagFile(FTextStream &tagFile) const
     tagFile << "      <type>" << convertToXML(typeString()) << "</type>" << endl;
   }
   tagFile << "      <name>" << convertToXML(name()) << "</name>" << endl;
-  tagFile << "      <anchorfile>" << convertToXML(getOutputFileBase()+Doxygen::htmlFileExtension) << "</anchorfile>" << endl;
+  tagFile << "      <anchorfile>" << convertToXML(getOutputFileBase()) << Doxygen::htmlFileExtension << "</anchorfile>" << endl;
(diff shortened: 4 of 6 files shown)


albert-github:
Can you please attach a, small, self contained example (source+configuration file in a tar or zip)  showing the problem?

avostrik:
Here it is. Check in file list browser "class.hpp [external]", which is invalid.
[test.tar.gz](https://github.com/doxygen/doxygen/files/4415331/test.tar.gz)

albert-github:
This fix unfortunately gives a lot of warnings for the CGAL project that heavily uses tag file. The warnings are like:
```
index.html:-1: warning: multiple use of section label 'index', (first occurrence: index.html)
/home/cgal-testsuite/cgal_doc_build/CGAL-5.1-Ic-114/doc/AABB_tree/aabb_tree.txt:3: warning: multiple use of section label 'index' for main page, (first occurrence: index.html)
```

albert-github:
I just ran the current master on the test provided by @avostrik and this also gives similar errors:
```
test/projectB/README.md:1: warning: multiple use of section label 'index' for main page, (first occurrence: index.html)
```

avostrik:
It is caused by part in page defintion output that was emitting filename without extension:
```
diff --git a/src/pagedef.cpp b/src/pagedef.cpp
index 15d6ca8f..011b6470 100644
--- a/src/pagedef.cpp
+++ b/src/pagedef.cpp
@@ -162,7 +162,7 @@ void PageDefImpl::writeTagFile(FTextStream &tagFile)
     tagFile << "  <compound kind=\"page\">" << endl;
     tagFile << "    <name>" << name() << "</name>" << endl;
     tagFile << "    <title>" << convertToXML(title()) << "</title>" << endl;
-    tagFile << "    <filename>" << convertToXML(getOutputFileBase()) << "</filename>" << endl;
+    tagFile << "    <filename>" << convertToXML(getOutputFileBase())<< Doxygen::htmlFileExtension << "</filename>" << endl;
     writeDocAnchorsToTagFile(tagFile);
     tagFile << "  </compound>" << endl;
   }
```

albert-github:
I think that it was not intentionally and that it should have an extension and is correct.
I think the problem lies more in the `<name>`, when I change the `<name>` in the projectA.tag file from `index` into `indexA` I don't get the warning anymore (this is not advised to do).

As far as I know the names over the different projects (when using tag files) should be unique, but `index` is a "general name"  and normally won't be referenced by means of a tag.

avostrik:
It seems I found where it was 'specially' handled as well:
```
diff --git a/src/tagreader.cpp b/src/tagreader.cpp
index 275d66ab..e1f28c50 100644
--- a/src/tagreader.cpp
+++ b/src/tagreader.cpp
@@ -1486,7 +1486,7 @@ void TagFileParser::buildLists(const std::shared_ptr<Entry> &root)
   for (pgit.toFirst();(tpi=pgit.current());++pgit)
   {
     std::shared_ptr<Entry> pe = std::make_shared<Entry>();
-    pe->section  = tpi->filename=="index" ? Entry::MAINPAGEDOC_SEC : Entry::PAGEDOC_SEC;
+    pe->section  = tpi->filename=="index.html" ? Entry::MAINPAGEDOC_SEC : Entry::PAGEDOC_SEC;
     pe->name     = tpi->name;
     pe->args     = tpi->title;
     addDocAnchors(pe,tpi->docAnchors);
```
Will this patch do?

albert-github:
No this patch won't do as here the `.html` is hardcoded and by means of `HTML_FILE_EXTENSION` in the original projectA the `.html` could be e.g. `.xhtml` or ... 
I think this on the right track though, my first thought would be to use:
```
pe->section  = tpi->name=="index" ? Entry::MAINPAGEDOC_SEC : Entry::PAGEDOC_SEC;
```
instead of
```
pe->section  = tpi->filename=="index" ? Entry::MAINPAGEDOC_SEC : Entry::PAGEDOC_SEC;
```
but I'm not sure if the `name` is always `index`. So to stay consistent, we probably have to store the `HTML_FILE_EXTENSION` in the tag file as well and read it in e.g. `tag_extension` use this together with the `filename==index<tag_extension>` to do the check.

albert-github:
@avostrik are you working on a patch?

avostrik:
Not yet, I think we are still collecting ideas and agree on one to implement.

Another one is to not store anything but use some util function to extract basename from filename and compare as before to 'index'. I am new to Doxygen codebase, if you could point how to do this with available API, it would be great.

albert-github:
Your remark made me think (a bit more).

There is no direct function but the code (functions are available in `util.cpp`):
```
 pe->section  =  stripExtensionGeneral(tpi->filename,getFileNameExtension(tpi->filename)) == "index" ? Entry::MAINPAGEDOC_SEC : Entry::PAGEDOC_SEC;
```
instead of
```
pe->section  = tpi->filename=="index" ? Entry::MAINPAGEDOC_SEC : Entry::PAGEDOC_SEC;`
```
looks like to work.
I tested it:
- the provided test
  - projectA and projectB generated with current master
  - projectA generated with doxygen 1.8.17 and projectB with the current master plus the proposed fix as shown above
- against CGAL

avostrik:
Submitted fix https://github.com/doxygen/doxygen/pull/7684

PR #7681, opened 2020-04-01
Treat relative markdown files as subpages

avostrik:
This pull request is an improvement that helps to build Doxygen pages tree without any special Doxygen commands in markdown files.

Look at individual commit messages for more details.

Test project: 
[test_subpages.tar.gz](https://github.com/doxygen/doxygen/files/4417265/test_subpages.tar.gz)

Issues that exist after generation on master:
- All markdown pages in the tree at the same level
- Automatically generated link to relative markdown file is broken on the main page, but correct one is in the tree view. May be it is by design, but reference points to an empty page. Markdown is not a source code, so likely have to be excluded from file references.
- local folder image referenced from markdown is not copied to output, showing alt placeholder


albert-github:
Do you think that the proposed pull request #7032 has also a relation to this pull request as it also concerns relative markdown pages.

avostrik:
Yes, it actually based on it, there is a line `This change is an adaptation of e52b821` in first commit message. I also asked a question there, but it was left without an answer.

albert-github:
I was hoping that your comment would have been enough for @doxygen to integrate the proposed pull request #7032 into the master.

avostrik:
Implicit subpage creation for every link in text seems not the best idea. It does not allow to do things like:
- have multiple references to same page without adding more entries into treeview at same level
- circular references between files, which works with `@ref` syntax.

Looks like solution should be an explicit tag to translate relative markdown link as subpage or reference. At the same time there is a need to keep markdown relative link working to not clutter how it looks like on web hosting directly.

Would it be ok to hi-jack optional title in reference link declaration for Doxygen command as link text can be redefined? I doubt there is a need to start link title with `@subpage` sequence.

example:

page1.md
```
Some text with [New title for page 2][Page2] link.

Same [Page2][] referenced second time using ID as a title.

This [Page2] will be rendered as `@subpage`, which may
be minor issue than implicit subpages creation as it is now...

Reference link declaration is not visible at all, not on the web,
not in the Doxygen output.
[Page2]: page2.md "@subpage"

Inline [Page2](page2.md "@subpage") can be used when only single
reference in text needed with subpage relationship.

```

page2.md
```
References [Page1] without subpage relationship as before, but resolving relative
file name to link ID behind the scenes.

Or inline link [Page1](page1.md) will not create another subpage as it happens
in this pull-request

[Page1]: page1.md
```

Issue #7675, opened 2020-03-29
Wrong HTML output from markdown with Doxygen code sample


avostrik:
Reproducible on current master branch.
Attached test files.

Some class description with Doxygen comment examples:

```cpp
/**
    @brief Idea is to have markdown fenced code block that can be copy pasted or used as a template.

    Markdown shall be properly rendered by web service that hosts it (like github, bitbucket) and in Doxygen generated docs.
    So we can't escape Doxygen commands with `\@` in this comment block.

    @code
    Test t = TestFactory::create();
    t.doSomething();
    @endcode
*/
class Test
{
}
```

[test.zip](https://github.com/doxygen/doxygen/files/4399812/test.zip)

Tried to figure out options to fix:

- introduce nested scope in src/doctokenizer.l and ignore everything in language specific multiline comments that appears in `@code`
- fix only in src/markdown.cpp by additional check and generate escaped intermediate representation for special comment blocks

albert-github:
Some thoughts:

We have here indeed a nesting of `@endcode` / `@endverbatim` commands, where even an extra `@code` / `@endcode` pair is introduced by means of the triple backtick.
A similar problem would also surface when we have some code that is indented by (at least) 4 spaces and contains the `@verbatim` / `@endverbatim` pair.

Doxygen does no counting but ends a `@code` / `@verbatim` section as soon as it sees the first corresponding `end`.

The only solution I see is that a counting mechanism is introduced, but this probably has to be done not only in the 2 mentioned files but also in `markdown.cpp`, `commentcnv.l`, `commentscan.l`, `pre.l` and `scanner.l`. The counting means also that the number of start and end commands have to match. Furthermore attention has to be paid to that "escaped" commands e.g. `\@code` in a `@code` part, which should not be seen as the start command.

@doxygen your opinion?

doxygen:
@albert-github I think this is a case that doxygen should support. I'm not sure counting at `pre.l`, `commentcnv.l`, and `scanner.l` is really needed, as the markdown block isn't processed by those scanners. `markdown.cpp` is the first to introduce the nested block, so any pass after that should handle it properly. An idea could be to let the markdown processor, escape the nested `@code` and `@endcode` by something that is later placed back when writing the output.

albert-github:
[...]

PR #7683, opened 2020-04-02
issue #7675 Wrong HTML output from markdown with Doxygen code sample

albert-github:
A counting strategy has been implemented to support nested verbatim and code section.

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -387,6 +387,7 @@ struct commentscanYY_state
   int              condCount = 0;
 
   int              commentCount = 0;
+  int              blockCount = 0;
   QCString         spaceBeforeCmd;
   QCString         spaceBeforeIf;
   QCString         copyDocArg;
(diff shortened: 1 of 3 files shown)


avostrik:
Works for nested Doxygen tags, but does not for markdown samples:

```markdown
    ```markdown
        Code sample in sample
        ```cpp
           int a = test;
        ```
        Another sample within same block
        ```java
           class hello...
        ```
    ```
```

Doxygen breaks it into two code spans on first closing `` ``` ``:

```
<div class="contents">
<div class="textblock"><div class="fragment"><div class="line">```markdown</div>
<div class="line">    Code sample in sample</div>
<div class="line">    ```cpp</div>
<div class="line">       int a = test;</div>
</div><!-- fragment --> <pre class="fragment">    Another sample within same block
    ```java
       class hello...
    ```
```

When it shall be similar when we use mix of code span chars: first set with triple tilda
[markdown_sample.zip](https://github.com/doxygen/doxygen/files/4429862/markdown_sample.zip)


```
<div class="contents">
<div class="textblock"><div class="fragment"><div class="line">```markdown</div>
<div class="line">    Code sample in sample</div>
<div class="line">    ```cpp</div>
<div class="line">       int a = test;</div>
<div class="line">    ```</div>
<div class="line">    Another sample within same block</div>
<div class="line">    ```java</div>
<div class="line">       class hello...</div>
<div class="line">    ```</div>
<div class="line">```</div>
</div> </div></div>
```
</pre><p> ``` </p>
</div></div><!-- contents -->
```

albert-github:
The triple backticks and tildes have not been considered.
For these I don't think we need a counting strategy, but "just" a level strategy, i.e. when the indentation from the original level is < 4 a triple backtick is seen as the closing part, otherwise it isn't.

@avostrik, @doxygen what do you think?
When this is true I think we can have this triple backtick correction better in a separate issue and pull request.

avostrik:
My last comment about embedding tildas and backticks is markdown specific stuff. May be it is easier to handle them in markdown processing, if postprocessing of resulting `@code` tags can't deal with it.

albert-github:
I indeed think this has to be handled in the markdown processing, otherwise the markdown processing will replace the first occurrence of ` ``` ` with the `@code` and the second ` ``` ` with `@endcode` and the damage is already done here. For this reason I wanted to hear some opinions (especially about the "level strategy")  and I also mentioned that it would be better to have it in a separate issue / pull request.

(the input and output of `doxygen -d markdown` that is shown is very helpful here).

avostrik:
Is some action required from me as issue submitter to get this one integrated? It solves issue with embedded doxygen comments. I do not have any objections to follow up with markdown fenced blocks nesting as separate one.

albert-github:
@avostrik no action required (as far as I can see), it is up to @doxygen to decide whether or not to merge or to reject the pull request.

Issue #7678, opened 2020-03-31
Not all functions are created


erez-o:
Hi,

In the following package, most functions are not created. 

It's a source code with a single h file, looks like everything up until the function `ma_context_is_loopback_supported` is created successfully (see line 4318 in the h file), and everything that comes after it is ignored. Attaching also the doxygen html output.

[doxyfile_ini.txt](https://github.com/doxygen/doxygen/files/4407181/doxyfile_ini.txt)
[package_src.zip](https://github.com/doxygen/doxygen/files/4407184/package_src.zip)
[doxygen_html.zip](https://github.com/doxygen/doxygen/files/4407185/doxygen_html.zip)

Using doxygen from master a [month ago](https://github.com/doxygen/doxygen/tree/59fc1a94e84278dedc80569b72756f336a705693)

Thanks

albert-github:
Based on the given information I tried to reproduce the problem, but I was not able to reproduce the problem.

The issue speaks about the html output but the given `doxyfile_ini.txt` does not contain the generation of HTML output so I added:
- `GENERATE_HTML = YES`
- `GENERATE_TREEVIEW = YES`

I ran with the current master (March 31, 2020) 1d403d6449ef632d94ac0973c9128ddb584ffc6a, the mentioned master from January 28, 2020 (59fc1a94e84278dedc80569b72756f336a705693) and with the 1.8.17 version and in all 3 cases I do see in the HTML output (miniaudio_8h.html) all output.

erez-o:
Truly sorry for my mistake,

Sending a new package and ini file:

[doxy_cfg.txt](https://github.com/doxygen/doxygen/files/4410412/doxy_cfg.txt)
[new_package_src.zip](https://github.com/doxygen/doxygen/files/4410415/new_package_src.zip)

I double checked so hopefuly there isn't a problem now.

You can see that while the function `ma_context_is_loopback_supported` exists in the generated docs, the functions `ma_device_init` or `ma_device_init` that come after it in the header file do not.

Thanks, and sorry again.

albert-github:
Using your code I can reproduce the issue.

When looking at your comments in the code I see in `miniaudio.h` constructs like (Introduction section):
```
    ```c
    #define MINIAUDIO_IMPLEMENTATION
    #include "miniaudio.h
    ```
```
which renders in:
![code_err](https://user-images.githubusercontent.com/5223533/78119730-260aef80-7409-11ea-91d7-04b1a1af0005.jpg)
[...]

PR #7685, opened 2020-04-02
issue #7678 Not all functions are created

albert-github:
Fenced code blocks with tildes and backticks were not handled in the same way in the respect that when after the backticks a language was given the complete sequence was handled by another rule.
By adding the backtick to the rule  where also the tilde was excluded fixes the problem.

This was also the case in the original problem, but was hidden as the `WARNINGS=NO` was set and thus messages like:
```
warning: reached end of file while inside a '```' block!
The command that should end the block seems to be missing!
```
where not shown.

Minimal example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4421749/example.tar.gz)

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -6349,7 +6349,7 @@ OPERATOR  "operator"{B}*({ARITHOP}|{ASSIGNOP}|{LOGICOP}|{BITOP})
 					    REJECT;
 					  }
   					}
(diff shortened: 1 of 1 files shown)


albert-github:
Currently working on the, travis signalled, problems regarding the documentation.

albert-github:
Documentation problem fixed.

Issue #7692, opened 2020-04-08
\copydoc does not work with file paths including dots


mlipp:
Commit https://github.com/doxygen/doxygen/commit/67cd1a3bf4981fd755ada6eeb85622b46638a749
and pull request https://github.com/doxygen/doxygen/pull/7042/files made copydoc stop working with file names including dots. It also seems that copydoc is not tested for files.

albert-github:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.
- Please also specify the doxygen version used.

timokoch:
The following function introduced with #7042 
```
QCString replaceScopeSeparator(QCString str)
{
  // we don't know about the language so we have to go for the worse
  return substitute(substitute(str,"\\","::"),".","::");  // PHP and Java, CSharp, VHDL, Python
}
```
agressively replaces all dots in `findDocsForMemberOrCompound`'s `commandName` with `::`. According to the doxygen doc all strings with `.` in it are parsed as file names.

This is done in line 700 in the docparser https://github.com/albert-github/doxygen/blob/67cd1a3bf4981fd755ada6eeb85622b46638a749/src/docparser.cpp#L700
which is after the replacement done by `replaceScopeSeparator` in line 665. Hence, if this analysis is right, filenames cannot be passed to copydoc anymore.

albert-github:
@timokoch this analysis might be correct, though see my previous comment (about example).

timokoch:
@albert-github Sure. 
[sourcesfiles.zip](https://github.com/doxygen/doxygen/files/4451213/sourcesfiles.zip)
Attached are two small source files. 

`source.cpp`
```cpp
//!\file
//!\brief a brief
//! some details

//! doc me
int main(int argc, char* argv[])
{
    return 0;
}
```
and `main.txt`
```cpp
/*!
 * \file
 * \brief Documentation
 * \copydetails source.cpp
 */
```

When I run `doxygen .` I get the following warning 
```
main.txt:2: warning: @copydetails or @copydoc target 'source.cpp' not found
```

`doxygen --version` produces `1.8.17`

Let me know if you need more details.

albert-github:
This was a case that I didn't consider, but copying the file description is also possible I saw in the documentation.
[...]

PR #7693, opened 2020-04-08
issue #7692 \copydoc does not work with file paths including dots

albert-github:
Besides links to functions / classes etc. it is also possible to have a link to a file and a file can contain a dot, so we first have to check the "original" name  and when not OK we should also check the replacements.
Regression on #7042

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -662,7 +662,6 @@ static bool insideTable(DocNode *n)
 }
 
 //---------------------------------------------------------------------------
-
 /*! Looks for a documentation block with name commandName in the current
  *  context (g_context). The resulting documentation string is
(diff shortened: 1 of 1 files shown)


timokoch:
@albert-github Thanks so much for this quick fix! Is there no way to find out the code language at this point? Or at least in a way that is cheaper than parsing it twice?

This PR fixes #7692 for me.

doxygen:
@albert-github I think a better solution is possible in `findDocsForMemberOrCompound` by first checking if parameter `commandName` is a file (e.g. the `findFileDef` part) and only then do the substitutions and check for a symbol.  That way the whole function does not need to be called twice.

albert-github:
I didn't / don't see how I could accomplish this.
Feel free to use an alternative implementation.

PR #7701, opened 2020-04-14
Add support to translate KaTeX math formula from markdown to Doxygen

avostrik:
Requires USE_MATHJAX enabled

[test_katex.tar.gz](https://github.com/doxygen/doxygen/files/4476410/test_katex.tar.gz)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -517,6 +517,52 @@ static int processQuoted(GrowBuf &out,const char *data,int,int size)
   return 0;
 }
 
+/** Process KaTeX formula in between `$...$` or `$$...$$` and translate them to
+ *  Doxygen formula command
+ */
+static int processKaTeX(GrowBuf &out,const char *data,int offset,int size)
+{
+  // skip escaped \$
+  if (offset>0 && data[-1]=='\\')
+  {
+    return 0;
+  }
+
+  int i=1;
+  if (size>2 && data[1]!='$') // $...$
+  {
+    while (i<size && data[i]!='$')
+    {
+      i++;
+    }
+    if (i > 2 && i<size && data[i]=='$')
+    {
+      out.addStr("@f$",3);
+      out.addStr(&data[1],i-1);
+      out.addStr("@f$",3);
+      return i+1;
+    }
+  }
+  else if (size>3 && data[1]=='$' && data[2]!='$') // $$...$$
+  {
+    i = 2;
+    while (i<size && data[i]!='$')
+    {
+      i++;
+    }
+    if (i > 4 && i+1<size && data[i]=='$' && data[i+1]=='$')
+    {
+      out.addStr("@f[",3);
+      out.addStr(&data[2],i-2);
+      out.addStr("@f]",3);
+      return i+2;
+    }
+  }
+
+  // not a formula
+  return 0;
+}
+
 /** Process a HTML tag. Note that <pre>..</pre> are treated specially, in
  *  the sense that all code inside is written unprocessed
  */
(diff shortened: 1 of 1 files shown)


albert-github:
You write:
> Requires USE_MATHJAX enabled

shouldn't this be:
> Requires MARKDOWN_SUPPORT enabled

The crash on travis-ci looks quite strange but due to the use of `USE_MATHJAX=NO` and `SUPPORT_MARKDOWN=YES` settings for the documentation generation and the fact that in the documentation the `$` is used it results in this problem. Looks like that even generating the documentation with `USE_MATHJAX=YES`

- check generation of the documentation (`make docs`)
- useful might be when running a test the option `-d markdown` i.e. `doxygen -d markdown`

avostrik:
I found root cause, but hard to say what would be a proper fix.
1. config.xml has unescaped `$` chars in couple of places. Fixed.
2. config.xml has variable names `$name`, `$file`, `$line` and `$text` in following vars: ABBREVIATE_BRIEF and WARN_FORMAT which are emitted without escaping into config.doc.

Looking at config.doc structure I wonder why this file is processed by markdown backend in first place? It has a lot of html tags.

Looking how xml translated to doc by python I do not like idea to touch it and escape every code block just in case it contains `$` variable.

So now I can't decide on best option:
- disable this doc from markdown processing
- ignore any markdown stuff between html tags

What do you think is a best way, may be another idea?

albert-github:
> Looking at config.doc structure I wonder why this file is processed by markdown backend in first place? It has a lot of html tags.

The html tags in the `config.doc` are interpreted by doxygen as the `config.doc` is used for HTML (and CHM), LaTeX (PDF) output.

`config.doc` is a generated file based on `config.xml`. `config.xml ` is used on a number of places:
- generating code for processing the doxygen configuration settings
- generating the documentation to be used in the doxygen manual (all documentation files are processed by the Markdown backend as they can contain Markdown syntax and it is not possible to exclude just one file)
- generating the (default) doxygen configuration file (Doxyfile) and the comments here
- generating handling of the doxygen configuration settings in the doxywizard  and showing the help information there.

All in all the handling of the `config.xml` is used on a number of places and its generation is quite a sensible to changes.

I think that at the moment it will break quite a bit of documentation where in the comments the `$` is used. The `$` is used in a number of languages e.g. php.

What I don't yet understand:
- what is the special thing about the KaTeX syntax compared to (La)TeX syntax (direct (La)TeX syntax i.e. `$` and `$$` is not supported in doxygen, it has all go througt the `\f$` / `\f[` / `\f{` commands)
- why not use an `INPUT_FILTER` to change the `$$ ... $$` in `\f[ ... `f]` / `$ ... $` in `f$ ... `f$`.


I see the commit:  "Ignore KaTeX processing between \<code\> and \</code\> tags" , what about the commands like `\verbnatim`, `\htmlonly`, ...

avostrik:
Idea comes from having preview of markdown in editor, web browser and doxygen without special tags in the markdown. I.e. I could add `\f$` in markdown, then it breaks [markdown math](https://marketplace.visualstudio.com/items?itemName=goessner.mdmath) extension (uses KaTeX) that does not understand `\f` before last `$`.

If it is possible to preprocess markdown before doxygen will parse it. I think it could work. I will check this. Thanks for pointing it out.

If it is possible with INPUT_FILTER to transform expected `$...$ and $$...$$` to `\f$..\f$` and `\f[..\f]` then may be code changes are not needed. 

Besides couple of unescaped variables in comment of config.xml that potentially could trigger somewhere else.

PR #7713, opened 2020-04-20
Add doxygen version information to tag file

albert-github:
Add doxygen version information to doxygen tag file.
currently this information is not used inside doxygen , but is useful for checking with which version of doxygen is used when debugging problems.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4503545/example.tar.gz)


doxygen:
@albert-github I've decided to add the version info as attributes to `tagfile` rather then as a new tag (to avoid breaking 3rd party parsers).

albert-github:
No problems with this change.

PR #7651, opened 2020-03-20
Reduce Qt-dependencies in dot-code and completely remove QThread & Co

cmorty:
There is still work to do to get at least the dot-code modernized (for-each, smart pointers) and QT-free, but this is a first step. I managed to remove QThread, QMutex and  QWaitCondition from the code, reducing the code size by over 1.5KLOC.
Although I tested this with a large project, maybe someone should tested on a few others before merging.

PR #7728, opened 2020-04-26
Change container type of DirList from QList to STL container

dhebbeker:
Effort to reduce (outdated) Qt dependencies (similar to #7651) in order to simplify modern C++ coding.

Chosen `std::list` as container type as it brings the used `sort()` member function.

Note:
 - Implementation of `QList::inSort()` was copied to `DirList::inSort()` as there is no equivalent in the STL container
 - Signature of `GroupDefImpl::addDir()` and `GroupDefImpl::addDir()` needed a minor change.

- [ ] Testing the [`sortInDirList`](https://github.com/doxygen/doxygen/pull/7728/files#diff-c0b61dd85e8f538c4e4e369f0e52cb4cR1101-R1109)

### DIFF
--- a/src/dirdef.cpp
+++ b/src/dirdef.cpp
@@ -142,7 +142,7 @@ bool DirDefImpl::isLinkable() const
 
 void DirDefImpl::addSubDir(DirDef *subdir)
 {
-  m_subdirs.append(subdir);
+  m_subdirs.push_back(subdir);
   subdir->setOuterScope(this);
   subdir->setParent(this);
 }
--- a/src/dirdef.cpp
+++ b/src/dirdef.cpp
@@ -15,6 +15,7 @@
 #include "docparser.h"
 #include "definitionimpl.h"
 #include "filedef.h"
+#include <algorithm>
 
 //----------------------------------------------------------------------
 
@@ -159,7 +160,7 @@ void DirDefImpl::addFile(FileDef *fd)
 
 void DirDefImpl::sort()
 {
-  m_subdirs.sort();
+  std::sort(m_subdirs.begin(), m_subdirs.end(), &compareDirDefs);
   m_fileList->sort();
 }
 
@@ -1092,3 +1093,7 @@ void generateDirDocs(OutputList &ol)
   }
 }
 
+bool compareDirDefs(const DirDef *item1, const DirDef *item2)
+{
+  return qstricmp(item1->shortName(),item2->shortName()) < 0;
+}
--- a/src/dirdef.cpp
+++ b/src/dirdef.cpp
@@ -307,9 +307,7 @@ void DirDefImpl::writeDirectoryGraph(OutputList &ol)
 void DirDefImpl::writeSubDirList(OutputList &ol)
 {
   int numSubdirs = 0;
-  QListIterator<DirDef> it(m_subdirs);
-  DirDef *dd;
-  for (it.toFirst();(dd=it.current());++it)
+  for(const auto dd : m_subdirs)
   {
     if (dd->hasDocumentation() || dd->getFiles()->count()>0)
     {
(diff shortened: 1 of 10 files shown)


src/groupdef.cpp
```diff
@@ -327,13 +327,16 @@ bool GroupDefImpl::addNamespace(const NamespaceDef *def)
...
 
-void GroupDefImpl::addDir(const DirDef *def)
+void GroupDefImpl::addDir(DirDef *const def)
 {
   if (def->isHidden()) return;
   if (Config_getBool(SORT_BRIEF_DOCS))
-    m_dirList->inSort(def);
+  {
+    m_dirList->push_back(def);
+    m_dirList->sort();
```
dhebbeker, line 336: Does this really sort the objects, or the items?

dhebbeker: `QList::inSort()` [is using](https://github.com/doxygen/doxygen/blob/91370bf84ac299fcb773e1b9d81e8f5c56da0725/qtools/qlist.doc#L223) `compareItems()`.

doxygen: `inSort()` inserts one element at the right place, whereas `sort()` sorts the whole list. In most places I use append (push_back) while adding and then sort once after everything is in the list. so `sort()` is not a good replacement for `inSort()`

src/dirdef.h
```diff
@@ -34,7 +36,7 @@ class FTextStream;
 class DirDef;
 
 /** A list of directories. */
-class DirList : public QList<DirDef>
+class DirList : public std::deque<DirDef>
```
doxygen: I've learned that it is not a good idea to inherit from STL containers, since they do not have virtual destructors. See also: https://stackoverflow.com/questions/2034916/is-it-okay-to-inherit-implementation-from-stl-containers-rather-than-delegate

dhebbeker: Thank you for pointing that out. In the [article](https://punchlet.wordpress.com/2009/12/29/letter-the-fifth/) [referenced](https://stackoverflow.com/questions/2034916/is-it-okay-to-inherit-implementation-from-stl-containers-rather-than-delegate#comment1959577_2034916) by the resource you gave, it is well explained why one should not inherit from the STL containers. Thus I now applied an alternative discussed in that article: Free functions.

src/groupdef.cpp
dhebbeker: I copied the QList way of `inSort` into [a free function](https://github.com/doxygen/doxygen/pull/7728/files#diff-c0b61dd85e8f538c4e4e369f0e52cb4cR1101-R1109).

dhebbeker:
Some testing is still necessary.

src/dirdef.cpp
```diff
@@ -1096,3 +1093,17 @@ void generateDirDocs(OutputList &ol)
...
+
+void sortInDirList(DirList& list, DirDef *const newItem)
+{
+  auto potentialSuccessor = list.begin();
+  while (potentialSuccessor != list.cend() && compareDirDefs(*potentialSuccessor, newItem))
+  {
+    potentialSuccessor++;
+  }
+  list.insert(potentialSuccessor, newItem);
+}
```
dhebbeker, line 1109: I performed a [test of the correct integration of `compareDirDefs` and `sortInDirList`](https://github.com/doxygen/doxygen/pull/7728#issuecomment-622994606) by turning [this `<` operator](https://github.com/doxygen/doxygen/pull/7728/files#diff-c0b61dd85e8f538c4e4e369f0e52cb4cR1098) and checked that the output in the generated documentation was in reverse order.

- [used test configuration](https://github.com/doxygen/doxygen/files/4605861/test_sortInDirList_compareDirDefs.zip)

PR #7739, opened 2020-04-29
add a boolean configuration IGNORE_MISSING_PARAMDOC.

doj:
If this boolean is true, doxygen will not output messages
about missing function parameter documentation.

This is my first change for doxygen. It's possible that there are other places where it is useful to check this new configuration option. For my use case it did ignore the messages I was not interested in.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -1292,6 +1292,15 @@ FILE_VERSION_FILTER = "cleartool desc -fmt \%Vn"
 ]]>
       </docs>
     </option>
+    <option type='bool' id='IGNORE_MISSING_PARAMDOC' defval='0'>
+      <docs>
+<![CDATA[
+ This \c IGNORE_MISSING_PARAMDOC option can be enabled to ignore warnings
+ about missing parameter documentation in a documented function.
+ It is only effective if \c WARN_IF_DOC_ERROR tag is set to \c YES.
+]]>
+      </docs>
+    </option>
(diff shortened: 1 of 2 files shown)


albert-github:
I tried to reproduce the unwanted warnings but I didn't succeed (I did set `WARN_IF_DOC_ERROR=YES`. Note: this might be set as a dependency by means of `depends=...` in the config.xml).

Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that shows the problem?  Please don't add external links as they might not be persistent.

doj:
[pull-7739.zip](https://github.com/doxygen/doxygen/files/4561774/pull-7739.zip)
added a zip with 2 files.

doj:
yes, this is supposed to ignore those warnings about those incomplete documented parameters.
(my project just has thousands of them and I can't fix them all. I'd like to suppress those warnings to see and fix other (more important to me) issues.)

albert-github:
Thanks for the example, I see here the 2 messages as well.

I refer to the comment: https://github.com/doxygen/doxygen/pull/452#issuecomment-250974326

Sometimes through the sheer number of warnings it is had to get the right ones.
Possibility I see for you is just get the warnings list and do a bit of intelligent editing / filtering on it e.g. with awk or with grep.

Still the best thing to do is to fix also the missing parameters as they give information in the API documentation.

PR #7765, opened 2020-05-08
Running Inkscape 1.0

albert-github:
The command line interface (CLI) of Inkscape 1.0 has changed in comparison to previous versions. In order to invoke Inkscape, the used version is detected and based on the version the right syntax of the CLI is chosen.


doxygen:
@albert-github I've move the inkscape version detection code to a separate function, and made it robust for versions >1. Please verify if it still works as intended.

albert-github:
I don't expect a version 2 soon and when this will happen they might have again an incompatible command line but it doesn't hurt.

Looks like it is still working as intended.

Issue #7721, opened 2020-04-23
Error parsing VHDL file with if-generate


engycz:
VHDL file with a process in if-generate clause is not correctly parsed.
```
  Count_2 : if Counters_g = true generate

    --! @vhdlflow diagram
    Test2b : process(Clock_in, ResetRx_in) --received packet OK counter
    begin
      if ResetRx_in = '1' then
        UserDatab_out <= '0';
      elsif Clock_in'event and (Clock_in = '1') then
        UserDatab_out <= '1';
      end if;
    end process Test2b;

  end generate Count_2;
```
There is also an error in doxygen log
```
Generating docs for nested compound TestIfGen::TestIfGen_A...
operable program or batch file.
error: could not create dot file<unknown>:1: warning: syntax error 'end'
Generating namespace index...
Generating graph info page...
```
No flow diagram is generated for Test2b process.
Here is complete test file with doxygen config file [IfGen.zip](https://github.com/doxygen/doxygen/files/4521758/IfGen.zip)

albert-github:
- which version of doxygen did you use (`doxygen -v`)
- I saw that that you use a "strange" `OUTPUT_DIRECTORY       = !_DoxyDoc`, this should not be a problem but ...
- can you also run with -d extcmd and provide the entire log file

engycz:
Doxygen version 1.8.18 (a1b07ad0e92e4526c9ba1711d39f06b58c2a7459*)
I also tried git master version.
Same behavior with `OUTPUT_DIRECTORY = _DoxyDoc`

albert-github:
Please provide the:
can you also run with `-d extcmd` and provide the entire log file


@maddox11 any idea?

engycz:
```
doxygen.exe -d extcmd DoxyGenFPGA.Doxyfile
warning: Tag 'TCL_SUBST' at line 238 of file 'DoxyGenFPGA.Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'PERL_PATH' at line 2118 of file 'DoxyGenFPGA.Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'MSCGEN_PATH' at line 2140 of file 'DoxyGenFPGA.Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
[...]

PR #7771, opened 2020-05-11
fixed if-generate-bug 7721

maddox11:
fixed if-generate-bug #7721


engycz:
@maddox11 JavaCC generated files should also be included in PR.

albert-github:
@engycz the generated files will be added when the proposed pull request is integrated in master.

Do you have java installed on your computer, in that case the only thing needed is either:
- the javacc distribution or
or
- the jar file: javacc-7.0.5.jar
- small script called javacc in the path starting the jar

engycz:
It's OK if the files will be generated automatically.

PR #7782, opened 2020-05-17
issue #7781: allow '>' before Markdown emphasis

### LINKED ISSUE #7781: [markdown] <h1>_foo_</h1> ignored
**Describe the bug**

```md
<h1>_bad_</h1>
<h1> _good_</h1>
```

rendered into:

```html
<h1>_bad_</h1>
<h1> <em>good</em></h1>
```

image

That is, if an html-tag-starting and a markdown's `_em_` (or `**bold**`) are adjacent, `_em_` will be printed out _as is_, not be rendered into `<em>em</em>`.

**Expected behavior**

```html
<h1><em>bad<em></h1>
<h1> <em>good</em></h1>
```

image

**Additional context**
Please let me know if there are related issues or anything.

wataash:
Fixes #7781.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -70,8 +70,8 @@
 // is character at position i in data allowed before an emphasis section
 #define isOpenEmphChar(i) \
   (data[i]=='\n' || data[i]==' ' || data[i]=='\'' || data[i]=='<' || \
-   data[i]=='{'  || data[i]=='(' || data[i]=='['  || data[i]==',' || \
-   data[i]==':'  || data[i]==';')
+   data[i]=='>'  || data[i]=='{' || data[i]=='('  || data[i]=='[' || \
+   data[i]==','  || data[i]==':' || data[i]==';')
 
 // is character at position i in data an escape that prevents ending an emphasis section
 // so for example *bla (*.txt) is cool*
--- a/testing/055/md_055_markdown.xml
+++ b/testing/055/md_055_markdown.xml
@@ -28,7 +28,7 @@
         <sect2 id="md_055_markdown_1autotoc_md3">
           <title>Markdown in HTML</title>
           <para>
-            <heading level="3">**Header3** blah <emphasis>blah</emphasis> <computeroutput>blah</computeroutput></heading>
+            <heading level="3"><bold>Header3</bold> blah <emphasis>blah</emphasis> <computeroutput>blah</computeroutput></heading>
           </para>
         </sect2>
       </sect1>
--- a/testing/055/md_055_markdown.xml
+++ b/testing/055/md_055_markdown.xml
@@ -25,6 +25,12 @@
           </para>
           <para>Dash - NDash <ndash/> MDash <mdash/> EDash - ENDash -- EMDash --- E3Dash --- </para>
         </sect2>
+        <sect2 id="md_055_markdown_1autotoc_md3">
+          <title>Markdown in HTML</title>
+          <para>
+            <heading level="3">**Header3** blah <emphasis>blah</emphasis> <computeroutput>blah</computeroutput></heading>
+          </para>
+        </sect2>
       </sect1>
     </detaileddescription>
   </compounddef>
--- a/testing/055_markdown.md
+++ b/testing/055_markdown.md
@@ -22,3 +22,7 @@ More text
 [U]: http://example.com/last-line
 
 Dash - NDash -- MDash --- EDash \- ENDash \-- EMDash \--- E3Dash \-\-\-
+
+## Markdown in HTML
+
+<h3>**Header3** blah _blah_ `blah`</h3>


albert-github:
This fix results in a regression, for e.g. CGAL, for constructs like:
```
    <TD ALIGN=LEFT NOWRAP>
<TT>__GNUC__</TT>
```
Instead of the output:


res_OK

we get:
res_ERR

Note:
- the missing underscores (`__`) around `GNUC and the fact that in the error version it is bold.
- we get warnings:
  ```
  .../Chapter_portability.txt:47: warning: found </TT> tag while expecting </strong>
  .../Chapter_portability.txt:54: warning: found </strong> tag while expecting </tt>
  .../Chapter_portability.txt:102: warning: end of comment block while expecting command </strong>
  .../Chapter_portability.txt:102: warning: end of comment block while expecting command </tt>
  ```

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5021161/example.tar.gz)

wataash:
Thank you for reporting it, @albert-github. I'm sorry.

When I markdown is enabled, think `<TT>__GNUC__</TT>` should look like: <TT>__GNUC__</TT> (`<tt><b>GNUC</b></tt>`, just like GitHub's markdown parser does)

But apparently it's not what it should be: `__GNUC__` `<tt>__GNUC__</tt>`
To get it, we should write it as:

```
`__GNUC__`
```

Do you agree?

albert-github:
A workaround is indeed to use the back-ticks and this works. 
The thing is it is a regression and we have to see what the good construct is. A number of things are silently replaced and on other place there are the warnings (without the warnings I would not have seen the difference).
GitHub is a good guideline for this but not all on GitHub has to be followed.

The documentation of `<tt>` says:
```
 • <TT> Starts a piece of text displayed in a typewriter font.
 • </TT> Ends a <TT> section.
```

For back-tics it is written:
```
To indicate a span of code...
```
this points more in the direction of the current implementation.

@doxygen what will be the best?

wataash:
Let's say (a) and (b) as:

- (a): `<TT>__GNUC__</TT>` -> <TT>__GNUC__</TT>
- (b): `<TT>__GNUC__</TT>` -> `__GNUC__`

So the behavior of Doxygen before #7782 was (a).

[CommonMark](https://spec.commonmark.org/0.29/) is also a good (maybe the best and de facto) guideline, but it doesn't defines what backticks should become inside HTML tags, unfortunately.
It is even confusing!

```
<div><a href="bar">*foo*</a></div>
↓
<div><a href="bar">*foo*</a></div>

<del>*foo*</del>
↓
<p><del><em>foo</em></del></p>
```

[commonmark.js](https://github.com/commonmark/commonmark.js/) is a good guide too, which is organized by @commonmark and authored by jgm (author of CommonMark and [pandoc](https://github.com/jgm/pandoc)).
This implements (b).
https://spec.commonmark.org/dingus/?text=%3Cdiv%3E%3Ca%20href%3D%22bar%22%3E*foo*%3C%2Fa%3E%3C%2Fdiv%3E%0A

I vote (b), if this breaking change is allowed.

albert-github:
I'm a bit confused by your `(a)` and `(b)` isn't it the other way around:
- `(a)` after #7782 
- `(b)` before #7782

wataash:
Sorry, I corrected them

doxygen:
A small test shows that HTML block tags like `<div>` and `<table>` seem to end emphasis processing until the next paragraph, but other inline tags like `<tt>` can just be nested...

```md
__bold1__ normal1
<TT> __bold2+tt__ normal2+tt </TT>
__bold3__

__bold4__
<div> __normal3__ normal4 </div>
__normal5__

__bold5__
<div> <tt>__normal5+tt__</tt> normal6 </div>
__normal7__

__bold6__
<table><tr><td>__normal8__</td><td>__normal9__</td></tr></table>
__normal10__

<TT>__bold7+tt</TT> bold8 <TT>bold8+tt__</TT>
```

with output

__bold1__ normal1
<TT> __bold2+tt__ normal2+tt </TT>
__bold3__

__bold4__
<div> __normal3__ normal4 </div>
__normal5__

__bold5__
<div> <tt>__normal5+tt__</tt> normal6 </div>
__normal7__

__bold6__
<table><tr><td>__normal8__</td><td>__normal9__</td></tr></table>
__normal10__

<TT>__bold7+tt</TT> bold8 <TT>bold8+tt__</TT>

Is this indeed the behavior that we want?

albert-github:
It is a quite a difficult situation.
It all boils down to consistency versus compatibility.

I think it is for the users a bit confusing when "inline" and "paragraph" HTML tags give different results, it is inconsistent and now also incompatible with the old situation, so this is (in my opinion) not the situation we want to have. So the answer to the question in comment https://github.com/doxygen/doxygen/pull/7782#issuecomment-669232090 is for me NO.
The fact that e.g. `<tt`> behaves differently  when using inside a `<div>` / `<table>`  tag or not is also confusing.
(the `<pre>` tag is of course again a bit of a different story).

In the documentation (Emphasis and strikethrough limits) there is an interesting sentence:
> Furthermore, a `*` or `_` only starts an emphasis and a `~` only starts a strikethrough if
> • it is followed by an alphanumerical character, and
> • it is preceded by a space, newline, or one the following characters `<{([,:;`
> An emphasis or a strikethrough ends if
> • it is not followed by an alphanumerical character, and
> • it is not preceded by a space, newline, or one the following characters `({[<=+-\@`


1. the compatibility would result in the fact that the "inline" and "paragraph" HTML tags (but also e.g. the markdown tables as they are converted to HTML tables; looks like `|<tt>__bold+tt__</tt>|` is not bold but does strip the `__`) would not touch e.g. the `__`
2. for consistency with / without HTML tags it would be better to have the normal markdown behavior i.e. `__...__`  always being bold (there might be exceptions like between backticks and in `<pre>` parts).

Especially based on what is documented I would go for the compatibility version, in case of a change it has to be documented properly in the documentation and in the release notes.

albert-github:
In my previous comment I mentioned `<pre>` as possible exception. Thinking about it there might be some more (just some thoughts):
- inside strings (single and double quotes).
- backtick, single and triple
- fenced code block (3 or more tildes)
- `<pre>`
- `<code>` (`<c>` is a bit dubious as it says "Similar to using \<tt\>text\</tt\>.").
- `\code`
- `\verbatim`
- `\...only`
- `\f$`, `\f[`, `\f{`
- `\dot`, `\dia`, `\msc`

- `<b>` either just `__` or all specials like `_`, `~~` (analogous `<em>`/`<i>`, `<s>` /`<strike>`

Issue #7787, opened 2020-05-19
Doxygen 1.8.18: Markdown tables not working in ALIASES anymore?


effaremment:
I'm migrating a large set of docs from Doxygen 1.8.14 to Doxygen 1.8.18
I'm using ALIASES to generate tables in my docs.

I cannot succeed to generate tables with doxygen 1.8.18, whereas it was working fine with 1.8.14

[table-markdown-doxy1.8.18.tar.gz](https://github.com/doxygen/doxygen/files/4651399/table-markdown-doxy1.8.18.tar.gz)
Here's what I get :
![image](https://user-images.githubusercontent.com/2772432/82343837-9c80a280-99f3-11ea-8982-b029f6bb9dad.png)
The first table is defined in line in a .dox file, while the second one is defined as an ALIASES.

I'm attaching a small sample (.dox file and Doxyfile)

I'm running on
Welcome to Ubuntu 18.04.4 LTS (GNU/Linux 4.15.0-96-generic x86_64)

And the version of Doxygen is 1.8.18:
$ doxygen --version
1.8.18
$

albert-github:
I've just pushed a proposed patch, pull request #7788

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).
The total solution also contains some more changes, especially to make it more consistent and also to have `^^` work in case `MARKDOWN_SUPPORT=NO` is used.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.19.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7788, opened 2020-05-20
issue #7787 Doxygen 1.8.18: Markdown tables not working in ALIASES anymore?

albert-github:
The artificial newline characters in ALIASES (`^^`) or better said its doxygen replacements (`\\_linebr`) were not seen by the markdown parser as line terminator and as a consequence there was no table seen (it was just 1 long line, without header / data lines).

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -80,6 +80,15 @@
    data[i]=='\\' || \
    data[i]=='@')
 
+inline int isNewline(const char *data)
+{
+  // plain return
+  if (data[0] == '\n') return(1);
+  // doxygen return from ^^ in ALIASES
+  if (data[0] == '\\' && data[1] == '\\' && data[2] == '_' && data[3] == 'l' && data[4] == 'i' && 
+      data[5] == 'n'  && data[6] == 'e'  && data[7] == 'b' && data[8] == 'r') return(9);
+  return(0);
+}
 //----------
 
 struct LinkRef
@@ -1594,8 +1603,10 @@ int findTableColumns(const char *data,int size,int &start,int &end,int &columns)
   start = i;
 
   // find end character of the table line
-  while (i<size && data[i]!='\n') i++;
-  eol=i+1;
+  int j = 0;
+  while (i<size && !(j = isNewline(data + i))) i++;
+  eol=i+j;
+
   i--;
   while (i>0 && data[i]==' ') i--;
   if (i>0 && data[i-1]!='\\' && data[i]=='|') i--,n++; // trailing or escaped | does not count
@@ -2120,7 +2131,9 @@ static void findEndOfLine(GrowBuf &out,const char *data,int size,
   // find end of the line
   int nb=0;
   end=i+1;
-  while (end<=size && data[end-1]!='\n')
+  //while (end<=size && data[end-1]!='\n')
+  int j = 0;
+  while (end<=size && !(j = isNewline(data+end-1)))
   {
     // while looking for the end of the line we might encounter a block
     // that needs to be passed unprocessed.
@@ -2181,6 +2194,7 @@ static void findEndOfLine(GrowBuf &out,const char *data,int size,
       end++;
     }
   }
+  if (j) end += j-1;
   //printf("findEndOfLine pi=%d i=%d end=%d {%s}\n",pi,i,end,QCString(data+i).left(end-i).data());
 }
 


effaremment:
Thanks to the fix. I applied the patch to the 1.8.18 and it did solve the issue.

However, my real use case is more complex than the simple case I reported. 
I need to use conditional generation in my table ALIASES. And to make it pretty in the Doxyfile, I am indenting the ALIASES definitions. None of these 2 mechanisms work while they were working fine with 1.8.14
Below is the result I'm getting.
image
Here is the definitions I'm using:
`ALIASES                += MY_TABLE=" \
| Table header4 | ^^ \
|--------------| ^^ \
| ref 1  | ^^ \
| ref 2 | ^^ \
"

ALIASES                += MY_INDENTED_TABLE=" \
                                              | Table header4 | ^^ \
                                              |--------------| ^^ \
                                              | ref 1  | ^^ \
                                              | ref 2 | ^^ \
                                              "

ALIASES                += MY_COMPLEX_TABLE{2}="@section \2 My complex table ^^ \
| Table header4 | ^^ \
|--------------| ^^ \
@if MY_COND
| ref_\1 \"First\" | ^^ \
@endif
| ref 2 | ^^ \
"
`
[markdown-table-aliases.tar.gz](https://github.com/doxygen/doxygen/files/4681538/markdown-table-aliases.tar.gz)

Note: if you prefer I can open a new issue rather than expanding on this one.

albert-github:
I just tried the inline versions as well
- indented, the alias version results in a verbatim block (doxygen -d markdown):
  ```
  Indented
  @verbatim
                         | Table header4 | \\_linebr                         |--------------| \\_linebr                          | ref 1  | \\_linebr                            | ref 2 | \\_linebr
  @endverbatim
  ```
  A first test with an inline indented block like:
  ```
  The indented but now as inline version:
                          | Table header4 |
                          |--------------|
                          | ref 1  |
                          | ref 2 |
  ```
  results in a correct image (though not as an indented table but as a table at the beginning of the line).
  What is the difference between the alias and inline version? It is a newline!
  The original code is:
  ```
  Indented

  @MY_INDENTED_TABLE
  ```
  and now doxygen sees, correctly, the spaces at the beginning of the line as  an indication of a markdown code block, so removing the newline will solve this problem..
- complex
   - why are there no ` \` after the `@if` and `@endif` line?
   - `@if` in combination with tables are very nasty as the table detection depends on the use of `|`, by rewriting the inline version to:
      ```
      @section  bar_inline My complex table
      | Table header4 |
      |-----------|
      | @if MY_COND ref_foo "First" |
      | @endif  ref 2 |
      ```
      we get a nice table and the same is the case when we use as alias:
      ```
      "MY_COMPLEX_TABLE{2}=@section \2 My complex table ^^ | Table header4 | ^^ |--------------| ^^ | @if MY_COND ref_\1 \"First\" | ^^ | @endif ref 2 | ^^"
      ```

albert-github:
**The indented formulation**
The original code is:
```
Indented

@MY_INDENTED_TABLE
```
and by removing the extra empty line (better  formulation than the new line used before) theresult is correct, so:
```
Indented
@MY_INDENTED_TABLE
```

**The more complex conditional case**
The `||` is indeed a small problem as both the `|` signs are seen as table delimiters as well, so you have to do some logical rephrasing:
```
| Table conditional inline|
|--------------|
| ref 0        |
| @if MY_COND ref 1 |
| @endif ref 2 |
| @if ( MY_COND && MY_COND1 )  ref 3 |
| @endif ref 4 |
| @if (!(!MY_COND1 && !MY_COND2 ))  ref 5 |
| @endif ref 6 |
```

effaremment:
Thanks.
Regading the OR I was hoping for a more straight-forward solution.
All of these table ALIAS construction (condition, OR and indentation) were working fine in 1.8.14. 
I guess the parsing has evolved in between for some other reason.

albert-github:
Hallo Dimitri,

Ik zie een regressie met het voorbeeld dat ik gestuurd had (nogmaals
toegevoegd). Ik krijg hier (met en zonder Markdown):
../aa.h:3: warning: Found unknown command '\ilinebrtext'
../bb.md:4: warning: Found unknown command '\ilinebrtext'
lijkt er op dat een spatie wordt weggegeten in commentscan.l:
CommentScanner: .../bb.md:1
input=[
@page md_bb bb
Just another file:

text\ilinebr text\ilinebr
 ]
-----------
CommentScanner: .../bb.md:1
output=[
brief=[line=-1
]
docs=[line=2
Just another file:

text\ilinebrtext]
inbody=[line=-1
]
]

Groeten,

Albert

albert-github:
Problem has been solved with the aid of this pull request and some more further modifications (a.o. the `MATDOWN_SUPPORT=NO` gave problems with `^^` that weren't solved yet).

doxygen:
Not that the issue with using `||` inside remains, so the suggested workaround is still required.

PR #7795, opened 2020-05-23
Improve bison/flex debug possibility with gdb

wataash:
```
This changes "#line" in generated C++ source:
  #line 26 "doctokenizer.l"
to:
  #line 26 "/absolute/path/to/doctokenizer.l"
so that gdb can recognize the original .l .y source locations and step
into them.
```


albert-github:
Note: that compiling a lex file with `LEX_FLAGS=-d` also gives some nice information about how a lex file is parsed (though on input level, not on the low debug level).

Issue #7796, opened 2020-05-23
Backticks (`) in Doxygen-markup-in-C in Markdown collapses


wataash:
**Describe the bug**

This markdown file:

    good:

    ```c
    /**
     * @c good
     */
    int a;
    ```

    bad:

    ```c
    /**
     * `bad`
     */
    int b;
    ```

    end

will be rendered into a collapsed-look:

![image](https://user-images.githubusercontent.com/4846670/82729288-5730dd80-9d31-11ea-9ab3-dc0455cf57c9.png)

**Expected behavior**

`` `bad` `` becomes like `\c good`

**To Reproduce**

    mkdir test/

    cat > test/test.md << 'EOS'
    good:

    ```c
    /**
     * @c good
     */
    int a;
    ```

    bad:

    ```c
    /**
     * `bad`
     */
    int b;
    ```

    end
    EOS

    cat > test/Doxyfile << EOS
    GENERATE_HTML = YES
    GENERATE_LATEX = NO
    GENERATE_XML = YES
    INPUT=$PWD/test/test.md
    XML_OUTPUT=$PWD/test/out/
    HTML_OUTPUT=$PWD/test/html/
    HTML_FILE_EXTENSION=.xhtml
    STRIP_CODE_COMMENTS=NO
    EOS

    doxygen test/Doxyfile
    open test/html/md_test_test.xhtml

Note that this Doxyfile contains `STRIP_CODE_COMMENTS=NO` to avoid https://github.com/doxygen/doxygen/issues/7793.

**Version**

master (ff5ba4a40)

albert-github:
In doxygen there are 2 ways to create fenced code blocks in markdown syntax i.e. with tildes and with backticks. In case we need the other character in a fenced block we can easily switch to the other representation, though when both types are needed inside a fenced block we still have the same problem.

The problem also occurs only when the fenced character is the last non blank before the end of line.

I've just pushed a proposed patch, pull request #7797

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.19.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7797, opened 2020-05-23
issue #7796 Backticks (`) in Doxygen-markup-in-C in Markdown collapses

albert-github:
The problem occurs when the backtick is just before the end of line.
The fenced block will only terminate when the same number of fence characters are present, otherwise the search will continue.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -1500,7 +1500,7 @@ static bool isFencedCodeBlock(const char *data,int size,int refIndent,
       if (i==size || data[i]=='\n')
       {
         if (endTildes==startTildes)
-	{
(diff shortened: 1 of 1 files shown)


src/markdown.cpp
```diff
@@ -1499,8 +1499,11 @@ static bool isFencedCodeBlock(const char *data,int size,int refIndent,
       while (i<size && data[i]==' ') i++;
       if (i==size || data[i]=='\n')
       {
-        offset=i;
-        return endTildes==startTildes;
+        if (endTildes==startTildes)
+	{
```
wataash, line 1503: unintentionally TAB?

Issue #7727, opened 2020-04-25
warning: documented symbol `static bool (long-winded C++ type)' was not declared or defined.


svarshavchik:
No complaints from gcc 9 for the attached code, with -std=c++17, but doxygen emits the following warning:

/tmp/doxygen/test.cpp:19: warning: documented symbol `static bool sfinae< T, std::enable_if_t<(sizeof(T)+0 *sizeof(decltype(std::to_chars(nullptr, nullptr, std::declval< T && >()))) >1)> sfinae >::yes' was not declared or defined.

I started incrementally dropping some of the C++ goo, until the point where doxygen stopped complaining.

At this point the warning still exists:

`template<typename T> struct sfinae<T,std::enable_if_t<(sizeof(T) > 1)>> {
`

But at this point, the warning goes away:

`template<typename T> struct sfinae<T,std::enable_if_t<true>>
`
I think it's the `>` in the template parameter value expression, that's causing this confusion.
[doxybug.tar.gz](https://github.com/doxygen/doxygen/files/4534338/doxybug.tar.gz)

albert-github:
Fixed by means of aa89f8773a50887ad20fbf683ff3c9af38c8fc5e  (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.19.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7803, opened 2020-05-24
Added check for right angle bracket in case greater than used in specialization condition

JaredAll:


albert-github:
The handling of `<` and `>` is a very delicate in template specializations, also in respect to `<=`  /  `>=` but also in respect to using spaces (when I'm not wistaken the C++ standard writes something about this and also there have been changes between the different versions of the standard).

Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem you try to solve with this proposed patch?  Please don't add external links as they might not be persistent.

JaredAll:
Of course. This is attempting to solve the issue as described [here](https://github.com/doxygen/doxygen/issues/7727); this example is the same as the one in that issue.
[doxybug.tar.gz](https://github.com/doxygen/doxygen/files/4677118/doxybug.tar.gz)

doxygen:
@JaredAll I think I found a more generic fix (https://github.com/doxygen/doxygen/commit/aa89f8773a50887ad20fbf683ff3c9af38c8fc5e). Can you test if it solves your problem as well?

JaredAll:
@doxygen, looks like that fix takes care of it

PR #7813, opened 2020-05-29
Vhdl improvements (ALIAS, translation)

powARman:
Adds support for documenting VHDL alias constructs. Also adds a translation for classes: "Design Unit"
Example:
[test_alias.zip](https://github.com/doxygen/doxygen/files/4702283/test_alias.zip)
Without the patch the documentation of the alias is missing.

### DIFF
--- a/src/translator_en.h
+++ b/src/translator_en.h
@@ -396,7 +396,7 @@ class TranslatorEnglish : public Translator
       }
       else if (Config_getBool(OPTIMIZE_OUTPUT_VHDL))
       {
-          return "Design Unit Documentation";
+          return trDesignUnitDocumentation();
       }
       else
       {
@@ -2256,6 +2256,14 @@ class TranslatorEnglish : public Translator
     {
         return "Data Member Documentation";
     }
+
+//////////////////////////////////////////////////////////////////////////
+// new since 1.8.19
+//////////////////////////////////////////////////////////////////////////
+
+    /** VHDL design unit documentation */
+    virtual QCString trDesignUnitDocumentation()
+    { return "Design Unit Documentation"; }
 
 //////////////////////////////////////////////////////////////////////////
 
--- a/src/translator_en.h
+++ b/src/translator_en.h
@@ -394,6 +394,10 @@ class TranslatorEnglish : public Translator
       {
         return "Data Structure Documentation";
       }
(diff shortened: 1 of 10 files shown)


albert-github:
Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that shows the improvement?  Please don't add external links as they might not be persistent.

powARman:
@albert-github Added an example.

src/translator_en.h
```diff
@@ -394,6 +394,10 @@ class TranslatorEnglish : public Translator
       {
         return "Data Structure Documentation";
       }
+      else if (Config_getBool(OPTIMIZE_OUTPUT_VHDL))
+      {
+          return "Design Unit Documentation";
+      }
```
albert-github, line 400: As is this should be OK, though for update consistency with other output languages I have a bit of a doubt whether or not it wouldn't be better to call here a function `trDesignUnitDocumentation` do this can be in a category "new since 1.8.19" with possibilities to overload nicely (compare with e.g. `trDesignUnitHierarchy`).
@doxygen your idea ?

powARman: As for the German translation there already is a VHDL switch in trClassDocumentation I went that way also for English. The problem with the approach with adding an extra function trDesignUnitDocumentation is that trClassDocumentation is used in many places which all would have to be adapted when using a separate function.

albert-github: It is indeed consistent with the German translation and I see that in the Dutch (and other) translation the part `if (Config_getBool(OPTIMIZE_OUTPUT_FOR_C))` is missing, all in all not nice.

The intention I had was  to have the code as:
```
    virtual QCString trCompoundIndex()
    {
      if (Config_getBool(OPTIMIZE_OUTPUT_FOR_C))
      {
        return "Data Structure Index";
      }
      else if (Config_getBool(OPTIMIZE_OUTPUT_FOR_VHDL))
      {
        return trDesignUnitDocumentation();
      }
      else
      {
        return "Class Index";
      }
    }

```
and have this "else if" part in all translators so that as soon as a translator implements `trDesignUnitDocumentation` it is used up (still a number of changes to be done now though).

powARman: Ah, I see. I will do so and update my pull request

PR #7815, opened 2020-05-29
Option PDF_DST_FILE added

the78mole:
User can define a location, where latex makefile puts a copy of the PDF output.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -208,6 +208,7 @@ EXTRA_SEARCH_MAPPINGS  =
 GENERATE_LATEX         = NO
 LATEX_OUTPUT           =
 LATEX_CMD_NAME         =
+PDF_DST_FILE           = 
 MAKEINDEX_CMD_NAME     = makeindex
 LATEX_MAKEINDEX_CMD    = makeindex
 COMPACT_LATEX          = NO
--- a/src/config.xml
+++ b/src/config.xml
@@ -2603,6 +2603,15 @@ EXTRA_SEARCH_MAPPINGS = tagname1=loc1 tagname2=loc2 ...
 ]]>
       </docs>
     </option>
+    <option type='string' id='PDF_DST_FILE' format='file' defval='' depends='GENERATE_LATEX'>
+      <docs>
+<![CDATA[
+ The \c PDF_DST_FILE copies the generated refman.pdf to the defined destination. This makes the
+ PDF much easier to access than in the full blown latex build folder. It will take effekt at the
+ end of the makefile run. It is a path relative to the build folder of latex.
+]]>
+      </docs>
+    </option>
     <option type='string' id='LATEX_CMD_NAME' format='file' defval='' depends='GENERATE_LATEX'>
       <docs>
 <![CDATA[
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -293,6 +293,7 @@ static void writeLatexMakefile()
   }
   // inserted by KONNO Akihisa <konno@researchers.jp> 2002-03-05
   QCString latex_command = theTranslator->latexCommandName();
+  QCString pdf_dst_file = Config_getString(PDF_DST_FILE);
   QCString mkidx_command = Config_getString(MAKEINDEX_CMD_NAME);
   // end insertion by KONNO Akihisa <konno@researchers.jp> 2002-03-05
   FTextStream t(&file);
@@ -367,7 +368,12 @@ static void writeLatexMakefile()
       << "\t      latex_count=`expr $$latex_count - 1` ;\\" << endl
       << "\t    done" << endl
       << "\t" << mkidx_command << " refman.idx" << endl
-      << "\t$(LATEX_CMD) refman" << endl << endl;
+      << "\t$(LATEX_CMD) refman" << endl;
+    if(!pdf_dst_file.isEmpty())
+    {
+      t << "\tcp refman.pdf " << pdf_dst_file << endl;
+    }
+    t << endl;
   }
 
   t << endl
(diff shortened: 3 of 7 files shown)


the78mole:
I also fixed a bug with params (tabu) and groups and is slipped somehow into this pull request. Sorry.

albert-github:
Regarding remark you made heer: https://github.com/doxygen/doxygen/pull/7815#issuecomment-636253826 and the comment I placed with https://github.com/doxygen/doxygen/issues/7816#issuecomment-636299206:

the proposed "tabu" patch as part of another patch, please remove it from this patch (#7815) and create a separate patch out of it (In my opinion it is a bad idea to mix unrelated patches).

templates/latex/doxygen.sty
```diff
@@ -334,7 +334,7 @@
     \hline%
     \endhead%
 }{%
-    \end{longtabu*}%
+    \end{longtable}%
```
albert-github, line 337: What is the reason for changing here from longtabu to longtable, the longtabu tables look much better and are more flexible in respect to log / small parameter names (this said although it is known that longtabu is unmaintained, see https://github.com/tabu-issues-for-future-maintainer/tabu but we have found a reasonable workaround for this)

src/latexdocvisitor.cpp
```diff
@@ -1659,7 +1660,7 @@ void LatexDocVisitor::visitPost(DocParamList *pl)
                   parentType==DocParamSect::TemplateParam;
   if (useTable)
   {
-    m_t << "\\\\" << endl
+    m_t << "\\\\"
```
albert-github, line 1663: Any specific reason for removing the `endl`?

src/latexdocvisitor.cpp
```diff
@@ -1510,6 +1510,7 @@ void LatexDocVisitor::visitPre(DocParamSect *s)
       m_t << "\n\\begin{DoxyParams}";
       if      (hasInOutSpecs && hasTypeSpecs) m_t << "[2]"; // 2 extra cols
       else if (hasInOutSpecs || hasTypeSpecs) m_t << "[1]"; // 1 extra col
+      else m_t << "[0]"; // no extra cols
```
albert-github, line 1513: Probably a good fix, but awaiting example showing the problem wit issue #7816.

Doxyfile
```diff
@@ -208,6 +208,7 @@ EXTRA_SEARCH_MAPPINGS  =
 GENERATE_LATEX         = NO
 LATEX_OUTPUT           =
 LATEX_CMD_NAME         =
+PDF_DST_FILE           =
```
albert-github, line 211: Not really necessary as by default the value is empty.

src/latexgen.cpp
```diff
@@ -367,7 +368,12 @@ static void writeLatexMakefile()
       << "\t      latex_count=`expr $$latex_count - 1` ;\\" << endl
       << "\t    done" << endl
       << "\t" << mkidx_command << " refman.idx" << endl
-      << "\t$(LATEX_CMD) refman" << endl << endl;
+      << "\t$(LATEX_CMD) refman" << endl;
+    if(!pdf_dst_file.isEmpty())
+    {
+      t << "\tcp refman.pdf " << pdf_dst_file << endl;
+    }
+    t << endl;
```
albert-github, line 376: And the native Windows version (`writeMakeBat`) ?

Why copy the file and not rename / move?

src/config.xml
```diff
@@ -2600,6 +2600,15 @@ EXTRA_SEARCH_MAPPINGS = tagname1=loc1 tagname2=loc2 ...
...
  put in front of it.
+]]>
+      </docs>
+    </option>
+    <option type='string' id='PDF_DST_FILE' format='file' defval='' depends='GENERATE_LATEX'>
+      <docs>
+<![CDATA[
+ The \c PDF_DST_FILE copies the generated refman.pdf to the defined destination. This makes the
+ PDF much easier to access than in the full blown latex build folder. It will take effekt at the
+ end of the makefile run. It is a path relative to the build folder of latex.
```
albert-github, line 2611: Seen the implementation I think the directory can either be a full path and otherwise it is a path relative to the build folder of latex.

src/latexgen.cpp
the78mole: I was not sure, if the standard behaviour should be kept as much as possible. If any other software relies on finding a refman.pdf in the folder, it would break.

albert-github: The user requests the change so he is also responsible for "oher" software. Doxygen does not rely on it (as far as I know).

templates/latex/doxygen.sty
the78mole: Longtabu ist broken in group definitions. I know, you fixed it, but it seems, that it is still borken in this circumstances. I could not get it working for my external dependencies ([FreeRTOS](https://github.com/FreeRTOS/FreeRTOS-Kernel), [STM32CubeF4](https://github.com/STMicroelectronics/STM32CubeF4)) with longtabu.

albert-github: I cloned the FreeRTOS repository, created a default Doxyfile with just `RECURSIVE=YES` (as instructed in #7816), ran doxygen and ran make in the latex directory and got the same problem.

I'll have to look into this issue.

src/latexgen.cpp
the78mole: Simply missed it. I'm using the build from sources on MSYS2 and my make.bat is not generated.

templates/latex/doxygen.sty
the78mole: I know, but the workaround does not work for me. I tried to create a simple example, but the error does not show up.

albert-github: See the comment https://github.com/doxygen/doxygen/issues/7816#issuecomment-636332309 indicating incorrect usage of `<pre>` tags. Resolving the real problem properly resolves the issue with tabu.

src/latexdocvisitor.cpp
albert-github: Resolved by solving the real problem see #7816

src/latexdocvisitor.cpp
albert-github: Resolved by solving the real problem see #7816

PR #7826, opened 2020-06-04
remove boxes around nodes in call graphs

makelinux:
to make call graphs more clean and readable

```
before:
┌────────┐     ┌────────┐
│ caller │────►│ callee │
└────────┘     └────────┘
after:
    caller────►callee
```

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3335,6 +3335,14 @@ to be found in the default search path.
 ]]>
       </docs>
     </option>
+    <option type='string' id='DOT_SHAPE' format='string' defval='plain' depends='HAVE_DOT'>
+      <docs>
+<![CDATA[
+ Define 'record' to have a box around nodes.
+ More shapes are described here: https://www.graphviz.org/doc/info/shapes.html
+]]>
+      </docs>
+    </option>
     <option type='bool' id='CLASS_GRAPH' defval='1' depends='HAVE_DOT'>
       <docs>
 <![CDATA[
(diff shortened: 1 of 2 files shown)


albert-github:
- What do you mean by "ground nodes" ?
- Which graphs are affected?
- please give an example showing the effect of you change. i.e. by attaching a, small, self contained example (source+configuration file in a tar or zip) that shows the effect?  Please don't add external links as they might not be persistent.
- For consistency this should be an option and thus a project / user hs the choice to keep the old and new situation.

src/config.xml
```diff
@@ -3332,6 +3332,14 @@ to be found in the default search path.
  If you specify a
  different font using \ref cfg_dot_fontname "DOT_FONTNAME" you can set the path where \c dot
  can find it using this tag.
+]]>
+      </docs>
+    </option>
+    <option type='string' id='DOT_SHAPE' format='string' defval='plain' depends='HAVE_DOT'>
```
albert-github, line 3338: Note the shape "plain" is unknown in my version of dot (2.38), "plaintext" is known.

albert-github:
I just tried a little bit a more complex situation:

**Original**:
complain_8c_af4b1295281909f738210ebd619cb5707_cgraph

**With plaintext**:
plain

One thing that is explicitly missing is now the red boxes, indicating that the graph has further details but is not present here. Also strange is now the coloring of the head node.

albert-github:
This issue is also related to #6701

Issue #6741, opened 2019-01-05
Feature request: C++11 inline namespaces


mosra:
C++11 has the `inline` keyword for namespaces. Main uses of these are for versioning (e.g. there is the `std::__1` namespace in libc++) and for user-defined literals (see e.g. [here](https://en.cppreference.com/w/cpp/chrono/operator""h), `std::literals` and `std::literals::chrono_literals` are both inline so you can say `using namespace` on either of them in order to use the literals).

The following code snippet should explain their use:

```cpp
/** @brief A library */
namespace Library {
    /** @brief Version 1 */
    inline namespace V1 {
        void foo(); /**< @brief Version 1 of foo() */
    }
}

int main() {
    // Both do the same thing
    Library::foo();
    Library::V1::foo();
}
```

Right now, Doxygen doesn't recognize this keyword at all and so neither the HTML nor the XML output contain any information about the `inline` status. Note that I need mainly the XML output to contain this information, as it will be used by the [m.css Doxygen theme](https://mcss.mosra.cz/doxygen/).

While I'm usually trying to implement features myself and then submitting a PR, in this case I'm afraid I don't know the C++ parser internals well enough to even attempt that, which is why I'm submitting a feature request instead.

Thanks in advance!

Cc: @marzer

vinniefalco:
This would be extremely helpful for me as well, as I am using Doxygen's XML output to produce the reference in Beast.

jwakely:
See #5914 for a much older suggestion of how to handle inline namespaces in doxygen.

PR #7828, opened 2020-06-04
Add namespace inline flag in xml output

sizmailov:
(partially?) address #6741.


albert-github:
The question for me is what would be the possibilities to signal this as well with HTML / LaTeX / ... output ?

sizmailov:
I think all terminal output formats (i.e. not intermediate like XML) would require more changes to handle inline namespaces properly. Probably new settings are needed to allow fine rendering control. All this complexity is delegated to XML output consumer, so this PR is rather compact. 

IMO html/latex should be addressed separately.

PR #7839, opened 2020-06-09
remove boxes around nodes in call graphs 

makelinux:
Before:
image

After:
image


albert-github:
This issue is related to #7826 where already some comments were given.
- Especially the the remark about losing the "red" box information, is in my opinion a serious limitation
- The arrow as indicated in the image point into "thin air" as they don't dock to boxes, especially with large graphs this might be confusing.

This proposed pull request only addresses the "box" issue I think it would be better to have an option where a (XML?) file or string can be specified that contain the settings for the different items so also other issues regarding colors etc, can be handled.

testing/call_graph_test.c
```diff
@@ -0,0 +1,33 @@
+/*
```
albert-github, line 1: How is this file used?
It is placed in the testing directory but no test is executed for it.

makelinux: I've used it only manually. I am not  familiar enogth with your testing framework.
Can you please suggest how to incorporate it?
Is following good example?
 testing/085/085__tooltip_8cpp.xml
 testing/085_tooltip.cpp

albert-github: I'm not sure that the testing part is really relevant here as you want probably have here a visual comparison of the resulting graph. The tests are there to show differences in, some, basic functionalities.

makelinux: Better to have tests.
I added 
testing/087/087__call__graph_8c_ab50e9ccc8158d9b5cab04dba22244fbd_cgraph.dot testing/087_call_graph.c
But runtests.py doesn't find proper location of dot file. 
@albert-github can you please help to complete it?

testing/087_call_graph.c
```diff
@@ -0,0 +1,36 @@
+// objective: root and truncated nodes
+// check: 087__call__graph_8c_ab50e9ccc8158d9b5cab04dba22244fbd_cgraph.dot
+// config: DOT_GRAPH_MAX_NODES=8
+// config: GENERATE_HTML=YES
```
albert-github, line 4: Don't set this tests run in xml mode. When tests are run in html mode `doxygen tests TEST_FLAGS=--xhtml` this is added automatically.

testing/087_call_graph.c
```diff
@@ -0,0 +1,36 @@
...
+// config: HAVE_DOT=YES
+// config: CALL_GRAPH=YES
+// config: CALLER_GRAPH=YES
+// config: DOT_CLEANUP=NO
+// config: EXTRACT_ALL=YES
+// config: HTML_OUTPUT=out
```
albert-github, line 10: Will be automatically set when run in html mode.

makelinux: tests run automatically on command **make test** and it is right.
How to define my test html mode and run it automatically?

albert-github: The tests are not intended for automatic runs outside in xml mode the other modes are created for internal development testing (I found quite a few problems through it). The other modes don't test against the given files (here: "check: 087__call__graph_8c_ab50e9ccc8158d9b5cab04dba22244fbd_cgraph.dot") but do only some sanity checks on the format or convert to that format so it is known whether or not the format builds and the results can be inspected (e.g. rtf, pdf, docbook, chm). Interesting in that case is also the option `--keep`. 

I just ran your test case with (the old code) as `make tests TEST_FLAGS="--xml --xhtml --id=87 --keep"` but I got some messages:
```
Syntax of value for attribute id of map is not valid
...
Syntax of value for attribute id of map is not valid
Document test_output_087/html/087__callgraph_8c.xhtml does not validate
``` 
looks like the original already has some problem. The lines are like:
```
<map name="087__callgraph_8c_a1c4791d28548770b6e1bf28b18c95e96_cgraph" id="087__callgraph_8c_a1c4791d28548770b6e1bf28b18c95e96_cgraph">
```
I think I know the problem: an id cannot start with a number.  I also thought I fixed it (but apparently not or not everywhere, vaguely remember I had a discussion regarding xml / docbook about this, but I see in the xml also id starting with a number but these are internal names). This is something I have to look into.

Another problem might be that the `dot` program is not installed on travis and appveyor (I'm not sure).

makelinux: I've replaced testing/087_call_graph.c with examples/call_graph.c

makelinux: @albert-github

albert-github: What is the question?

makelinux:
@albert-github, I've updated the PR: replaced testing/087_call_graph.c with examples/call_graph.c
Would you like to review and suggest further actions?

albert-github:
I don't think the file in examples will do a lot as long as it is not bound into the documentation (and in the cmake procedure).

Furthermore in https://github.com/doxygen/doxygen/pull/7839#issuecomment-641252890 I made some suggestions for a more flexible approach.

makelinux:
@albert-github 
Latest changes:
- coma and quotations are optional in attributes are optional for graphviz
- using spaces and single words without quotations for better readability
- DOT_FONTNAME, DOT_FONTSIZE are replaced with DOT_GRAPH_ATTR,
DOT_COMMON_ATTR, DOT_EDGE_ATTR, DOT_NODE_ATTR
- removed duplicated code and used DotGraph::writeGraphHeaderwriteGraphHeader

doxygen:
@makelinux Would it be possible to make the default output to be the same as it currently is and still have the option to customize it? it is probably a matter of taste, but I find the version with boxes better to read and better looking.

makelinux:
> @makelinux Would it be possible to make the default output to be the same as it currently is and still have the option to customize it? it is probably a matter of taste, but I find the version with boxes better to read and better looking.



> @makelinux Would it be possible to make the default output to be the same as it currently is and still have the option to customize it? it is probably a matter of taste, but I find the version with boxes better to read and better looking.

Done

doxygen:
@makelinux I've tried out your pull request but I still see differences between the original dot images and the new ones; for instance the label text is too big, the arrow style is different.

The example I tried with default config file and `HAVE_DOT` enabled:

```cpp
/** A class */
class Base { };

/** Used */
class Used { };

/** A derived class */
class Derived : public Base
{
  public:
    Used *m_used;
};

```
file to compare `html/class_derived__coll__graph.png`

PR #7840, opened 2020-06-10
Incorrect label in map of dot files in xhtml

albert-github:
When a filename of a file starts with a digit the mapping of the resulting dot files results in message like:
```
Syntax of value for attribute id of map is not valid
```
an id cannot start with a digit, so an "a" is placed in front of it (unconditionally to overcome problems with a double label id i.e filename 087.cpp and a087.cpp).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4757882/example.tar.gz)
(To get the warnings: `xmllint --noout --path dtd --nonet --postvalid html/*xhtml`)

### DIFF
--- a/src/dotfilepatcher.cpp
+++ b/src/dotfilepatcher.cpp
@@ -460,7 +460,7 @@ bool DotFilePatcher::run() const
         convertMapFile(tt,map->mapFile,map->relPath,map->urlOnly,map->context);
         if (!result.isEmpty())
         {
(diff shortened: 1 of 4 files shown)


TrGameFn:
Unfortunately that change broke links in our class diagrams. Can we, please, have them back?

albert-github:
Code of pull request #8012 has been integrated in master on github

PR #7851, opened 2020-06-14
Updated formatting rules for paragraphs within table header cells.

### LINKED ISSUE #7849: Multiple paragraphs within table cells (td) are formatted wrong
When multiple paragraphs are present within table cells, the first paragraph is presented with normal font weight, all others are presented with strong font weight.
See attached example: 
[DoxygenTest.zip](https://github.com/doxygen/doxygen/files/4776377/DoxygenTest.zip)

This is due to an incorrect CSS Rule within doxygen.css:
https://github.com/doxygen/doxygen/blob/12515664fe30dd79f48fca4b1e3d6cdd3c1a5ce1/templates/html/doxygen.css#L69
Each selector should prepend "th".

infohoschie:
Only present paragraphs within table headers (th) with font weight strong, keep normal table cells (td) unaffected.
Resolves #7849

### DIFF
--- a/templates/html/doxygen.css
+++ b/templates/html/doxygen.css
@@ -66,7 +66,7 @@ p.startli, p.startdd {
 	margin-top: 2px;
 }
 
-th p.starttd, p.intertd, p.endtd {
+th p.starttd, th p.intertd, th p.endtd {
         font-size: 100%;
         font-weight: 700;
 }


albert-github:
Sorry just beaten you to it , I will remove  my proposed patch in favor of yours.

PR #7862, opened 2020-06-19
Problem with generating test output for Chinese

albert-github:
The log output generated for the Chinese language can contain some extended ASCII characters, this results in a message like:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf4 in position 36101: invalid continuation byte
```
and a crash of the test application during the read of the file.
In case of a crash it is attempted to open the file with another decoding and redo the read.

### DIFF
--- a/testing/runtests.py
+++ b/testing/runtests.py
@@ -391,25 +391,14 @@ class Tester:
 			elif (not outType and xopen(latex_output + "/temp",'r').read().find("Error")!= -1):
 				msg += ("PDF generation failed\n  For a description of the problem see 'refman.log' in the latex directory of this test",)
 				failed_latex=True
-			else:
-				res = False;
-				try:
-					res = xopen(latex_output + "/refman.log",'r').read().find("Error")!= -1
-				except:
-					res = xopen(latex_output + "/refman.log",'r', encoding = "ISO-8859-1").read().find("Error")!= -1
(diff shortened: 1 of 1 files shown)


testing/runtests.py
```diff
@@ -391,14 +391,25 @@ def perform_test(self,testmgr):
...
-				msg += ("PDF generation failed\n  For a description of the problem see 'refman.log' in the latex directory of this test",)
-				failed_latex=True
-			elif not self.args.keep:
-				shutil.rmtree(latex_output,ignore_errors=True)
+			else:
+				res = False;
+				try:
+					res = xopen(latex_output + "/refman.log",'r').read().find("Error")!= -1
+				except:
+					res = xopen(latex_output + "/refman.log",'r', encoding = "ISO-8859-1").read().find("Error")!= -1
```
doxygen, line 399: instead of doing this only as a fallback, I suggest to use this encoding (or "Latin-1", which is the same) from the start, which supposed to just read all characters as bytes and should always work. For just finding an ASCII string that should suffice I think.

albert-github: In this case it is indeed possible to use the `encoding='ISO-8859-1'` directly instead of the 
fallback solution.
I initially chose the fallback solution as not to endanger other tests.

PR #7864, opened 2020-06-21
improve using plantuml for  vhdl flowcharts

maddox11:
if  Plantuml  is installed we should use it for VHDL flowcharts without 
--!\vhdlflow [plant]

### DIFF
--- a/src/vhdljjparser.cpp
+++ b/src/vhdljjparser.cpp
@@ -397,14 +397,7 @@ void VHDLOutlineParser::handleCommentBlock(const char *doc1, bool brief)
     s->current->docLine = p->yyLineNr;
   }
 
-  int j = doc.find("[plant]");
-  if (j >= 0)
-  {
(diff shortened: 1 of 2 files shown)


src/vhdljjparser.cpp
```diff
@@ -397,14 +397,7 @@ void VHDLOutlineParser::handleCommentBlock(const char *doc1, bool brief)
     s->current->docLine = p->yyLineNr;
   }
 
-  int j = doc.find("[plant]");
-  if (j >= 0)
-  {
-    doc = doc.remove(j, 7);
-    s->current->stat = true;
-  }
-
```
albert-github, line 406: I think the new implementation is better, though 2 remarks:
- consistency, although `[plant]` is not documented at least I couldn't find it, might be useful to al least still strip the `[plant]`
- not for this pull request but for future new implementations: doxygen uses `{...}` for options on commands, see e.g. the `\snippet` command

Issue #7867, opened 2020-06-22
JAVADOC_AUTOBRIEF and @File regression


rodanrider:
**Describe the bug**
JAVADOC_AUTOBRIEF does not work as expected for @file documentation since v1.8.16.  Earlier releases work fine.

**Expected behavior**
I expect that the first line of text after a @file tag would be treated as the @brief description in a @file block.  Something changed in v1.8.16 and now even with the JAVADOC_AUTOBRIEF = YES, an explicit @brief tag is required.

**Screenshots**
v1.8.15 - Expected behaviour:
![image](https://user-images.githubusercontent.com/57236581/85332001-fad5f080-b48b-11ea-89dd-d0681a37d25c.png)

v1.8.16 - Unexpected behaviour:
![image](https://user-images.githubusercontent.com/57236581/85332045-104b1a80-b48c-11ea-9b14-624c96afd266.png)

**To Reproduce**
This is a default configuration file with the following 2 modifications:
JAVADOC_AUTOBRIEF = YES
OPTIMIZE_OUTPUT_FOR_C = YES

This is a pretty simple test case:
```
/**
 * @file test.h
 * This is a brief description.
 */
```
[test_case.zip](https://github.com/doxygen/doxygen/files/4815444/test_case.zip)

**Version**
I am using release versions of v1.8.15 - Expected behaviour, and v1.8.16 - Unexpected behaviour.
I have also confirmed that the released v1.8.18 exhibits the same buggy behaviour.

**Stack trace**
N/A

**Additional context**
N/A

albert-github:
I've just pushed a proposed patch, pull request #7867

rodanrider:
Thanks for the quick fix.  It fixes my reduced test case.  However, it causes other problems, such as if another tag is present after the file tag.  So I would suggest cancelling your pull request until those are also resolved.

I built it with the latest commit d5916c06fba9667911a19cf658e7c5ab54daf72c + your pull request.

```
/**
 * @file test_file_author_brief.h
 * @author John Doe
 *
 * This is a brief description.
 */
```
[v1_8_19_fix_bad.zip](https://github.com/doxygen/doxygen/files/4839388/v1_8_19_fix_bad.zip)

The author tag is used as the brief description!

![image](https://user-images.githubusercontent.com/57236581/85904351-591d0f00-b7bd-11ea-998e-d6012ee5019d.png)

albert-github:
[...]

PR #7868, opened 2020-06-23
issue #7867 JAVADOC_AUTOBRIEF and @File regression

albert-github:
The restriction of a command terminating a brief description is a bit to stringent, it should only be valid when the current mode is brief description and the brief descriptin is not empty.


albert-github:
A regression has been found and is shown in https://github.com/doxygen/doxygen/issues/7867#issuecomment-650422682

albert-github:
Withdrawn as most likely another strategy has to be used.

Issue #7872, opened 2020-06-25
A define containing a doxygen start of comment breaks the parser


RonaldHiemstra:
The following code fragment (in C++ code) 
`#define SELECT_ALL "/**"`
results in the following error:
warning: Reached end of file while still inside a (nested) comment. Nesting level 1 (probable line reference: 1)

Issue found in 1.8.17 and 1.8.18

albert-github:
- Always mention the version of doxygen you are using.

I could reproduce the problem with the current doxygen master: 1.8.19 (d5916c06fba9667911a19cf658e7c5ab54daf72c)

Some analysis:

Comment signs are always a difficult problem but according to the C standard it should be possible (in my interpretation):
> Except within a character constant, a string literal, or a comment, the characters /* introduce a
comment. The contents of such a comment are examined only to identify multibyte characters and
to find the characters */ that terminate it.

When looking at the doxygen preprocessor output we get:
```
00001 #define SELECT_ALL/**"
00002
```
so loos like the doxygen preprocessor doesn't recognize the comment properly.

When having e.g. `char *x = "/**";` doxygen doesn't give a warning.

albert-github:
I've just pushed a proposed patch, pull request #7874

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7874, opened 2020-06-26
issue #7872 A define containing a doxygen start of comment breaks the parser

albert-github:
Not only the preprocessor suffered from the "problem" but also the source code (e.g. with `SOURCE_BROWSER`) gave some strange / missing results.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4837278/example.tar.gz)
Note: the files look a bit strange (double and variables etc.) but the purpose it see the output in the "source browser" and in this case not the documentation.


doxygen:
@albert-github I noticed that the line numbers for the file cc.h in the example were still off by one in the source browser causing missing cross-references. I've fixed that with an additional commit.

Issue #7873, opened 2020-06-25
Combining `///`-style comments with macros containing @cond/@endcond causes a preprocessor error


chrisdembia:
Thanks @doxygen for such a great tool! I have used Doxygen for years now to document properties that are similar to `Q_PROPERTY`. The property is created by my macro, `DECLARE_PROPERTY`, and the macro generates doxygen documentation. This works fine unless the class using my macro uses `///`-style comments.

**Describe the bug**

Combining `///`-style comments with a macro that contains `@cond`/`@endcond` causes the preprocessor to fail.

I use the following input files:

doxyfile
```
# Doxyfile 1.8.18
QUIET                  = YES
INPUT                  = Bounds.h
ENABLE_PREPROCESSING   = YES
MACRO_EXPANSION        = YES
EXPAND_ONLY_PREDEF     = YES
EXPAND_AS_DEFINED      = DECLARE_PROPERTY
```

Bounds.h
```cpp
#ifndef BOUNDS_H
#define BOUNDS_H

#define DECLARE_PROPERTY(pname, T, comment)                                 \
    /** @cond */                                                            \
    /** Internal function for <b> pname </b> property.                   */ \
    void hidden_function_##pname();                                         \
    /** @endcond */                                                         \
    /** @name Properties (single-value)                                  */ \
    /** @{                                                               */ \
    /** comment                                                          */ \
    Q_PROPERTY(T, pname)                                                    \
    /** @}                                                               */ \
    /** @name Property-related functions                                 */ \
    /** @{                                                               */ \
    /** Get the value of the <b> pname </b> property.                    */ \
    const T& get_##pname() const;                                           \
    /** @}                                                               */

/** This works fine. */
class BoundsGood {
public:
    /** Constructor.
    Line two. */
    BoundsGood();
    DECLARE_PROPERTY(lower_bound, double, "Value of lower bound.");
    DECLARE_PROPERTY(upper_bound, double, "Value of upper bound.");
};

/** This causes a warning and the class ends up not documented. */
class BoundsBad {
public:
    /// Constructor.
    /// Line two.
    BoundsBad();
    DECLARE_PROPERTY(lower_bound, double, "Value of lower bound.");
    DECLARE_PROPERTY(upper_bound, double, "Value of upper bound.");
};

#endif
```

Running `doxygen doxyfile` with doxygen 1.8.18 generates the following output:
```
[...]

PR #7875, opened 2020-06-26
No default rule present in commentcnv.l

albert-github:
In the file `commentcnv.l` there is no default rule present, resulting in the, non understandable, output as indicated in https://github.com/doxygen/doxygen/issues/7873#issuecomment-650100405
(this is not the cause of the problem as reported with issue https://github.com/doxygen/doxygen/issues/7873, but quite annoying)


chrisdembia:
Thank you for quickly addressing this part of the issue :).

PR #7876, opened 2020-06-26
issue #7873 Combining `///`-style comments with macros containing @cond/@endcond causes a preprocessor error

albert-github:
At the end of a mult-line `///` section the state of readlineCtx not reset so the subsequent `@cond` block thought we were still in a `///` block`

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -753,6 +753,8 @@ MAILADR   ("mailto:")?[a-z_A-Z0-9.+-]+"@"[a-z_A-Z0-9-]+("."[a-z_A-Z0-9\-]+)+[a-z
 				     copyToOutput(yyscanner,yytext,(int)yyleng); 
 				     yyextra->inSpecialComment=FALSE;
 				     yyextra->inRoseComment=FALSE;
+                                     yyextra->readLineCtx = Scan; // reset, otherwise ther will be problems with:
+                                                                  //   static void handleCondSectionId
 				     BEGIN(Scan); 
                                    }
 <ReadLine>"/**"                    {


chrisdembia:
Yay! Should I test this locally? Or have you already tested it against the example I provided?

albert-github:
I tested it already against your example, but it never hurts to test it locally as well as your total project might have some other nice features as well.

chrisdembia:
@albert-github I tested this locally. While it fixes the warning, the output is now different, even for `BoundsGood` which does not use the `///`-style comments. 

Expected:

image

Actual:

image

albert-github:
I didn't test it yet, but I think it might have to do with:
- the `@cond` which defaults to `False` when nothing is set as the section label.
- not using `PREDEFINED             = Q_PROPERTY(x,y)=`

chrisdembia:
@albert-github I get the correct output with the same `doxyfile` by using the `master` branch.

I use `@cond` to intentionally hide a function `hidden_function...()`, so I do not expect to see documentation for this function.

Doxygen knows how to handle `Q_PROPERTY()`; if I use `PREDEFINED = Q_PROPERTY(x,y)=` then I believe Doxygen would replace my use of `Q_PROPERTY()` with nothing, rather than generating doxygen as if there were actually a Q_PROPERTY().

albert-github:
When I run the current master (1.8.19 (3e9d9386950cc67d155a920cf8483c4d5bdcec3c)) on the example as provided wit #7873 ([doxygen_macro_cond_bug.zip](https://github.com/doxygen/doxygen/files/4851626/doxygen_macro_cond_bug.zip)) I get:

master

and with the patched version (1.8.19 (85a96aee93346765126abe02234bfe4d83c0e6ae)) I get:

patched

So I don't see a difference.

chrisdembia:
Ah oops, I tested not with the `master` branch but with a previous release. So perhaps there's another bug since the 1.8.18 release.

albert-github:
Please always mention the version you used to test.

I now tested with 1.8.17 release version (1.8.17 (b5fa3cd1c6e6240e20d3b80a70e3f04040b32021*)) as well and here I do see:



So this looks more like your output.
(I also tested with the 1.8.18 release (1.8.18 (a1b07ad0e92e4526c9ba1711d39f06b58c2a7459)) and here I see the same output as with 1.8.17).

As the both problems are (in my opinion) independent, please submit a new issue for this issue.

chrisdembia:
Thanks @albert-github; my apologies regarding the version.

> As the both problems are (in my opinion) independent, please submit a new issue for this issue.

I agree.

chrisdembia:
@albert-github I created a separate issue: https://github.com/doxygen/doxygen/issues/7887

Issue #7879, opened 2020-06-28
Inline parameter documentation not working as expected


MartinMerkel:
Processing the following C code with Doxygen 1.8.18

```
cip_status_code_t cipNode_ctor(
    uint8_t numInstances,       ///< [in]Number of CIP node instances.
    cip_node_t* node              ///< [out]Pointer to CIP node structure.
)
{
 ...
}
```
results in the Parameters section in HTML out such as below:

> **Parameters**
> [in]         numInstances    [in]Number of CIP node instances.
> [out]      node                    [out]Pointer to CIP node structure.

instead of as probably intended:

> **Parameters**
> [in]         numInstances    Number of CIP node instances.
> [out]      node                    Pointer to CIP node structure.

Proposed fix for 1.8.18 code:

memberdef.cpp

Line 3753

Replace

`paramDocs+="@param"+direction+" "+a.name+" "+a.docs;`

with:

`paramDocs += "@param" + direction + " " + a.name + " " + docsWithoutDir;`

albert-github:
Thanks for the bug report and a\indicating the place of the, probable, bug.
Would have been nice when you had created a proposed pull request (patch) as well.

I've just pushed a proposed patch, pull request #7880

albert-github:
Has been fixed as part of #7881  (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.19.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #7880, opened 2020-06-28
issue #7879 Inline parameter documentation not working as expected

albert-github:
The direction was not stripped from the description when transforming to a `@param` command.

(fix as indicated by OP in the issue description)


doxygen:
Looks like this change is already included in master...

albert-github:
Indeed, has also been part of #7881 (explains also why I didn't include it in that pull request as there was a separate pull request for it).

PR #7881, opened 2020-06-28
More flexible and correct detection of direction of an argument

albert-github:
In the current version a line like:
```
uint8_t innInstances,       ///< [inn]Number of CIP node instances.
```
was seen as an input parameter.
Also not all combinations were checked (`[inout]` was OK but `[outin]` wasn't although `[in,out]` and `[out,in]` were working in the tokenizer.
Line up the detection of the direction between the tokenizer and the method `extractDirection`

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4842215/example.tar.gz)

Note: this problem was found when working on #7879.

### DIFF
--- a/src/doctokenizer.l
+++ b/src/doctokenizer.l
@@ -378,7 +378,7 @@ SPCMD2    {CMD}[\\@<>&$#%~".+=|-]
 SPCMD3    {CMD}_form#[0-9]+
 SPCMD4    {CMD}"::"
 SPCMD5    {CMD}":"
-INOUT	  "inout"|"in"|"out"|("in"{BLANK}*","{BLANK}*"out")|("out"{BLANK}*","{BLANK}*"in")
+INOUT	  "in"|"out"|("in"{BLANK}*","?{BLANK}*"out")|("out"{BLANK}*","?{BLANK}*"in")
 PARAMIO   {CMD}param{BLANK}*"["{BLANK}*{INOUT}{BLANK}*"]"
 VARARGS   "..."
 TEMPCHAR  [a-z_A-Z0-9.,: \t\*\&\(\)\[\]]
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -8138,10 +8138,25 @@ QCString extractDirection(QCString &docs)
   int l=0;
   if (re.match(docs,0,&l)==0)
   {
-    int  inPos  = docs.find("in", 1,FALSE);
-    int outPos  = docs.find("out",1,FALSE);
-    bool input  =  inPos!=-1 &&  inPos<l;
-    bool output = outPos!=-1 && outPos<l;
+    QRegExp re_in("\\[ *in *\\]"); // [in]
+    QRegExp re_out("\\[ *out *\\]"); // [out]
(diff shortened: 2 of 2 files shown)


doxygen:
@albert-github I've simplified `extractDirection()` a bit. I think it still covers the relevant cases, and in case of a user error, the direction part now (in most cases) becomes part of the documentation (so it is easier to spot mistakes). For some reason the parameter direction wasn't removed from the parameter documentation before. I've fixed that as well.

albert-github:
Strange that in my version the direction was not removed from the output.

In the new output I see for the explicitly mentioned example:
```
uint8_t innInstances,       ///< [inn]Number of CIP node instances.
```
the output:
```
[in]  innInstances Number of CIP node instances.
```
whilst in my version I see:
```
  innInstances  [inn]Number of CIP node instances.
```
the regular expressions were explicitly introduced to overcome these kind of problems.

doxygen:
@albert-github I've made some changes so that any extra characters other than "in" and "out" are also not accepted. Still no need for all the regexes.

albert-github:
Looks better though I found another regression on it (didn't test it against "my" version), but looks like a general problem
When having a line like:
```
uint8_t kommaInstances,    ///< [,]Number of CIP node instances.
```
the output is incomplete.

Example: [example2.tar.gz](https://github.com/doxygen/doxygen/files/5006173/example2.tar.gz)

doxygen:
@albert-github Your example revealed a bug in the quite generic `substitute()` function, triggered when the result of a substitution is an empty string.

albert-github:
Looks OK to me.

EduPonz:
I know it's been some time but today I stumbled upon an error on Ubuntu 20.04 because we had added a `@param[in out]` (mind the blank instead of `,`). [Here](https://doxygen.nl/manual/commands.html#cmdparam), that space (which this PR enables and was included in 1.8.19) is not documented (I guess we added it by mistake and had no problem in our local Ubuntu 22.04, but then CI for 20.04 failed). I wonder whether the possibility of adding a blank there should also be documented.

BTW, I really appreciate the traceability you have between release notes, issues, PRs, etc. It was rather easy to find this 4 years old PR, which is astonishing! 🚀

albert-github:
In the documentation I see:
> The `\param` command has an optional attribute, `<dir>`, specifying the direction of the parameter. Possible values
are "[in]", "[in,out]", and "[out]", note the `[square]` brackets in this description. Note that it is also possible to put
whitespace between the command and the `<dir>`. When a parameter is both input and output, `[in,out]` is used as
attribute.

So also `inout` and `outin` are not documented (`inout` is a typical Fortran term, `outin` is just a side effect for symmetry).
I think the situation mentioned was an unintentional side effect, but maybe it should be documented along with `inout` and `outin`.

EduPonz:
> So also `inout` and `outin` are not documented (`inout` is a typical Fortran term, `outin` is just a side effect for symmetry).
> I think the situation mentioned was an unintentional side effect, but maybe it should be documented along with `inout` and `outin`.

I've opened a PR to address this, would you mind reviewing it?
* #10927

PR #7888, opened 2020-06-30
Scalable search bar for high resolution displays

tttapa:
This pull request aims at replacing the pixelated PNG images used in the search bar by SVG images and CSS shadows, so it looks better on high resolution displays.

The following two images show  the differences between doxygen/master (d0f24b762e23b1b0c069209978aa403acc027cdf) and my feature branch:
old-select-drop-down-crop
new-select-drop-down-crop

- I've recreated the magnifying glass and the exit icons in Inkscape and included them in the templates folder, replacing all references to the original PNG files to the new SVG ones.
- I also updated the CSS styling of the search bar to replace the PNG background images that produced the shadow effect by actual CSS shadows.

I didn't touch any of the underlying search logic, and I didn't update the alternative stylesheets (`search_noidx.css` and `search_nomenu.css`), because I'd like some feedback first.

I tested both the JavaScript-based search and the server side search. For the server side search, I noticed a strange text field with the search query below the header, in both the doxygen/master version and my version:
old-server-results
new-server-results

I've attached the output of my feature branch for the default Doxyfile: [**Doxygen-scalable-search-bar.zip**](https://github.com/doxygen/doxygen/files/4854192/Doxygen-scalable-search-bar.zip)  
It also contains a `docker-compose.yml` file that launches a PHP container for testing the server side search. To use it, simply run `docker-compose up` in the same folder and go to `http://localhost:8888/index.html`.

I'm also attaching a ZIP with some screenshots before and after the changes: [**Doxygen-screenshots-scalable-search-bar.zip**](https://github.com/doxygen/doxygen/files/4854225/Doxygen-screenshots-scalable-search-bar.zip)

Finally: is there a vector source file for the doxygen logo available somewhere? It also looks pretty pixelated on modern displays:
doxygen-logo


doxygen:
@tttapa nice work on the search button! 

Strange thing you saw with the server side option. Can you file a bug report for it? 

As for the logo: no there is no vector graphics version. I once tried to generate one from the bitmap, but that was not successful, so it needs to be recreated by hand.

Issue #7884, opened 2020-06-30
Python to xml: Whitespace in method doc not preserved (MARKDOWN_SUPPORT=NO)


Stewori:
**Describe the bug**
Processing a Python code snippet like
```
class A(object):
    def __init__(self):
        """blah
        blah
                xyz
            blah
        """
        ...
```
results in an xml snippet
```
...
        <detaileddescription>
<para><verbatim>blah
blah
xyz
    blah
</verbatim> </para>
        </detaileddescription>
...
```
Note how the whitespace before "xyz" compared to "blah" is incorrect. This only occurs if there are equal or more spaces difference between them than one Python code indent. This issue seems to occur only in methods in classes, not in functions on module level.

**Expected behavior**
I expect the xml snippet to be
```
...
        <detaileddescription>
<para><verbatim>blah
blah
        xyz
    blah
</verbatim> </para>
        </detaileddescription>
...
```

**To Reproduce**
Find a Python file and doxygen config attached: [bugreport.zip](https://github.com/doxygen/doxygen/files/4850252/bugreport.zip)
The issue is demonstrated like mentioned above and with a more sophisticated example testing various whitespace amounts.
After running doxygen on the example, `xml/classexample_1_1A.xml` contains descriptions with messed up whitespace. Results in `xml/namespaceexample.xml` are fine (surprisingly), so it just happens inside classes.

**Version**
Tested with doxygen 1.8.11 from distribution and with self-compiled snapshot from 2020-06-29, 1.8.19 (d5916c06fba9667911a19cf658e7c5ab54daf72c).
Both produce the described problem.

`lsb_release -a`:
```
LSB Version:	core-9.20160110ubuntu0.2-amd64:core-9.20160110ubuntu0.2-noarch:security-9.20160110ubuntu0.2-amd64:security-9.20160110ubuntu0.2-noarch
Distributor ID:	LinuxMint
Description:	Linux Mint 18 Sarah
Release:	18
Codename:	sarah
```

**Additional context**
This issue hinders use of custom doc postprocessing on top of doxygen xml output. E.g. applying rst parsing is faulty if code samples with indentation are present in the doc.

I looked for duplicates but the bugtracker is difficult to search and has almost 2000 issues. Sorry if I overlooked a duplicate, pointers welcome! Maybe it is related to #7069 or #7072. Note that I tried to set various values for MARKDOWN_SUPPORT and TAB_SIZE and these do not seem to have any effect on the issue.

[...]

PR #7893, opened 2020-07-03
issue #7884 Python to xml: Whitespace in method doc not preserved (MARKDOWN_SUPPORT=NO)

albert-github:
Corrected handling of the comment blocks. Due to the possible nesting of classes and the handling of the first line special handling is required by means of the stripping  of spaces after a return in a doc block (otherwise it might be stripped away multiple times / left multiple times).

Extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4870473/example.tar.gz)

### DIFF
--- a/src/pyscanner.l
+++ b/src/pyscanner.l
@@ -97,6 +97,7 @@ struct pyscannerYY_state
   QGString *              copyString = 0;
   int                     indent = 0;
   int                     curIndent = 0;
+  int                     commentIndent = 0;
   bool                    importTuple = FALSE;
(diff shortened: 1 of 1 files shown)


doxygen:
Note that I renamed `stripSpaces` to `stripIndentation` and made it work directly on the input string so not copying is needed, see https://github.com/doxygen/doxygen/commit/d2b3bc01f81e842ecc5baaf74e4402bc24e841e5

Let me know if you see any issues with this change.

albert-github:
At first glance I didn't seen a problem.

PR #7895, opened 2020-07-06
Superfluous space in "TEST_FLAGS"

albert-github:
When having extra spaces in TEST_FLAGS like(note the double space before `--ip`):
```
make tests TEST_FLAGS="--keep  --id=1"
```
or
```
make tests TEST_FLAGS="--id=1  --id=2"
```
we get a message like;
```
[100%] Running doxygen tests...
usage: runtests.py [-h] [--updateref] [--doxygen [DOXYGEN]]
                   [--xmllint [XMLLINT]] [--id IDS [IDS ...]]
                   [--start_id START_ID] [--end_id END_ID] [--all]
                   [--inputdir [INPUTDIR]] [--outputdir [OUTPUTDIR]]
                   [--noredir] [--pool [POOL]] [--xml] [--rtf] [--docbook]
                   [--xhtml] [--xmlxsd] [--pdf] [--subdirs] [--clang] [--keep]
                   [--cfg CFGS [CFGS ...]]
runtests.py: error: unrecognized arguments:
NMAKE : fatal error U1077: 'D:\Programs\Python\Python37\python.exe' : return code '0x2'
Stop.
```

By stripping the spaces in the arguments this can be overcome

### DIFF
--- a/testing/runtests.py
+++ b/testing/runtests.py
@@ -487,6 +487,8 @@ def split_and_keep(s, sep):
 		vv = val.split(" ",1)
 		if ((len(vv) == 1) and not vv[0] == ''):
 			retVal += vv
+		if (len(vv) == 2):
+			vv[1] = vv[1].strip()
 		if ((len(vv) == 2) and not vv[1] == ''):
 			retVal += vv
 		if ((len(vv) == 2) and vv[1] == ''):


testing/runtests.py
```diff
@@ -487,6 +487,8 @@ def split_and_keep(s, sep):
 		vv = val.split(" ",1)
 		if ((len(vv) == 1) and not vv[0] == ''):
 			retVal += vv
+		if (len(vv) == 2):
```
doxygen, line 490: I was looking at what `split_and_keep(s,sep)` is doing, and would propose a more Pythonic way of doing it:

```python
def split_and_keep(s,sep):
    splitBySeparator = [sep+x.strip() for x in filter(None,s.split(sep))]
    splitByFirstSpace = [z.split(' ',1) for z in splitBySeparator]
    flattenedList = [i for s in splitByFirstSpace for i in s]
    return flattenedList
```

albert-github: I found a regression:
```
nmake tests TEST_FLAGS=" --keep --id 1 --cfg ALIASES=tst=\"test - - test\"
```

this regression was not present in the  first modification where the first line was:
```
    splitBySeparator = [sep+x.strip() for x in filter(None,s.split(sep))]
```
now the line reads:
```
    splitBySeparator = [x.strip() for x in re.split('(--[^\-]*)',s) if x.strip()]
```
in the original version and the first modification there was no problem with this construct.
- Was there a reason / corner case not caught with the original version of the pull request and the first modification suggestion?

My personal opinion regarding the "Pythonic way", Python has already a of cryptic things and the "Pythonic way" is even more cryptic. Goes for me even more against the "keep it simple and straightforward" principle, though I've seen worse constructs where everything was crammed in one big statement

doxygen: > Was there a reason / corner case not caught with the original version of the pull request and the first modification suggestion?

Yes, if the string started without a `--`, one was added anyway

> My personal opinion regarding the "Pythonic way", Python has already a of cryptic things and the "Pythonic way" is even more cryptic. Goes for me even more against the "keep it simple and straightforward" principle, though I've seen worse constructs where everything was crammed in one big statement

My goal was not to make it cryptic, but to use a set of simple filters to transform the data in small steps in order to end up with the right result, instead of complicated loop with a lot of branch conditions, which made the original code hard to understand.

Here is a new attempt:

```python
def split_and_keep(s,sep):
    addTokenSeparator = s.replace(sep,'\0'+sep)
    splitByNullDelimiter = addTokenSeparator.split('\0')
    stripAndRemoveEmptyElements = [x.strip() for x in filter(None,splitByNullDelimiter)]
    splitByFirstSpace = [z.split(' ',1) for z in stripAndRemoveEmptyElements]
    flattenedList = [i for s in splitByFirstSpace for i in s]
    return flattenedList
```

albert-github: I had a quick look at the new version, didn't see directly a problem.

Problem with the original code might have been that some extra comment would have been better and maybe an `if` / `elif` / `else` construct would have been more clear.

The construct now used with the `'\0'` is similar to the construct as previously used with the `p=chr(ord(max(s))+1)`.

Issue #6846, opened 2019-02-21
Doxygen parsing global variable as function in certain cases


yourb3stnightmare:
Doxygen 1.8.15 on Windows 7 (and all previous versions I have used) is not consistently parsing global variable declarations, sometimes confusing them with functions, and sometimes not even listing them in the output.   In the attached example project, I get two parse warnings:
```
TestFile.cpp:14: warning: Member toGUIReminder(int(0), BoxBase::NOTIFY_ON_SET) (function) of namespace Dictionary::OtherSettingsGroup is not documented.
TestFile.cpp:16: warning: Member toApplicationReminder(int(0), BoxBase::NOTIFY_ON_SET) (function) of namespace Dictionary::OtherSettingsGroup is not documented.
```
If I remove the header from TestFile.cpp (so the file starts at the #include), I no longer get warnings.

In addition to the warnings, the output doesn't even list toApplicationUpdateOtherSettings as part of the file.

[DoxygenTest.zip](https://github.com/doxygen/doxygen/files/2891044/DoxygenTest.zip)

albert-github:
See also: https://stackoverflow.com/questions/54812028/doxygen-sees-global-variable-declarations-as-functions-in-c?noredirect=1#comment96405458_54812028

yourb3stnightmare:
Being that this issue seems like it would be complicated to fix, may I suggest that a new optional command be to allow the user to specify the type?  Or maybe a '\variable' and a '\function' command that force the parser to parse the object as either a variable or a function.

albert-github:
Looking at the example:
 - `toApplicationUpdateOtherSettings` doesn't throw a warning
 - `toApplicationReminder` and `toGUIReminder`  throw a warning

the difference is that `toApplicationUpdateOtherSettings` is `bool` and the other 2 are `int`. Changing e.g. `toApplicationReminder`  from
`      Box<int> toApplicationReminder(int(0), BoxBase::NOTIFY_ON_SET);`
to
`      Box<int> toApplicationReminder(bool(false), BoxBase::NOTIFY_ON_SET);`
doesn't throw a warning anymore for `toApplicationReminder`
(Note this is not a solution, just analyzing the problem).

Question:
 - is the output for `toApplicationUpdateOtherSettings` as shown by doxygen correct?

yourb3stnightmare:
[...]

PR #7903, opened 2020-07-14
issue #6846 Doxygen parsing global variable as function in certain cases

albert-github:
A parameter can only be present for a function, so we should test on the member being a function it as well.
This fix is for the problem as reported by @tgpfeiffer


tgpfeiffer:
Thank you for fixing this!

PR #7920, opened 2020-07-25
added utf-8 support

maddox11:
Added utf-8 support for  VHDL.

### DIFF
--- a/vhdlparser/JavaCC.h
+++ b/vhdlparser/JavaCC.h
@@ -6,17 +6,21 @@
 #include <string>
 #include <memory>
 #include <cassert>
+#include <cwchar>
+#include "vhdlstring.h"
 #include <functional>
 
+
+
 #ifndef JAVACC_CHAR_TYPE
-#define JAVACC_CHAR_TYPE char
+#define JAVACC_CHAR_TYPE wchar_t
 #endif
 
 #ifndef JAVACC_STRING_TYPE
-#define JAVACC_STRING_TYPE std::basic_string<JAVACC_CHAR_TYPE>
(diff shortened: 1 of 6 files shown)


vhdlparser/JavaCC.h
```diff
@@ -35,12 +39,13 @@ class ReaderStream {
 const JAVACC_CHAR_TYPE EMPTY[] = { 0 };
 
 #ifndef MAX
-#define MAX(a,b) ((a)>=(b)?(a):(b))
+#define MAX(a,b) (a>=b?a:b)
 #endif
 #ifndef MIN
-#define MIN(a,b) ((a)<=(b)?(a):(b))
+#define MIN(a,b) (a<=b?a:b)
```
albert-github, line 45: I think it is better to have the round brackets as e.g:
```
MAX (3+5,7)
```
will otherwise give different results.

vhdlparser/vhdlparser.jj
```diff
@@ -7,27 +7,27 @@
  * granted. No representations are made about the suitability of this software
  * for any purpose. It is provided "as is" without express or implied warranty.
  * See the GNU General Public License for more details.
- *
+ *
```
albert-github, line 10: Better not to have the extra space.

doxygen:
@maddox11 I don't understand what you are trying to do with this pull request. Doxygen already converts input to UTF-8 before passing it to the VHDL parser. For UTF-8 you should *not* use wchar_t to store characters. See also http://utf8everywhere.org/. Do you have an example where UTF-8 input is now not handled correctly?

PR #7922, opened 2020-07-26
Fix compiling with clang parser support and with Clang compiler, too

Bagira80:
When compiling the code with the Clang compiler a new CMake option
`use_libc++` is available which is enabled by default. It determines if
the Clang compiler should use libc++ as its C++ standard library or uses
the systems default (e.g. libstdc++ on Linux).
Disabling that option might help when compiling with Clang and the
option `use_libclang` is enabled but the associated library was not
built with libc++ as its C++ standard library.

Additionally, the doxygen executable that is created when CMake option
`static_libclang` is enabled, no longer crashes.


albert-github:
To the best of my knowledge there is some misunderstanding. When compiling doxygen by means of the clang compiler this does not automatically mean that doxygen uses, internally, the clang parsing (as `use_libclang` is not set), so the travis tests won't reveal whether or not doxygen will build with clang support.

Bagira80:
> To the best of my knowledge there is some misunderstanding. When compiling doxygen by means of the clang compiler this does not automatically mean that doxygen uses, internally, the clang parsing (as `use_libclang` is not set), so the travis tests won't reveal whether or not doxygen will build with clang support.

There is no misunderstanding. I am aware of the fact that `use_libclang` has to be enabled explicitly. If Travis is not testing for it, I assume you do not have enough test-configurations configured.

You should probably have at least the following test-configurations for each of your configured operating systems:
* GCC with `use_libclang` disabled,
* GCC with `use_libclang` enabled and `static_libclang` disabled,
* GCC with `use_libclang` enabled and `static_libclang` enabled,
* Clang with `use_libc++` enabled, `use_libclang` disabled,
* Clang with `use_libc++` enabled, `use_libclang` enabled and `static_libclang` disabled,
* Clang with `use_libc++` enabled, `use_libclang` enabled and `static_libclang` enabled
* Clang with `use_libc++` disabled and `use_libclang` disabled,
* Clang with `use_libc++` disabled, `use_libclang` enabled and `static_libclang` disabled,
* Clang with `use_libc++` disabled, `use_libclang` enabled and `static_libclang` enabled

You can have even more configurations if the clang parser libraries (`libclang-cpp.so` and `libclangTooling.a`) are compiled differently. However, incompatible C++ standard-library combinations will then fail. (E.g. Clang with `use_libc++` and `libclang-cpp.so` built with `libstdc++`.)

---

Nevertheless, adjusting `use_libc++` should now guarantee that there is a combination in which Clang is able to compile Doxygen with clang parser support.

Bagira80:
I added one more commit to remove the wrong and obsolete forcing of `-std=c++11` on GCC. (C++14 is the minimal required mode anyway.)

Issue #7923, opened 2020-07-27
[BISECTED] source line numbers in warnings output by parser are off by 1


duncan-roe:
The first commit with wrong line numbers is 5bae9d9.
The commit message includes the line *change moment of calling markdown processing*, which is borne out by gdb run on 5bae9d9 and the previous version 824025b, in [gdb_logs.tar.gz](https://github.com/doxygen/doxygen/files/4979922/gdb_logs.tar.gz). These are best compared side-by-side (I find [tkdiff](https://sourceforge.net/projects/tkdiff/) particularly good for this). (Unimportant diffs have been x'd out).

Suggested fix is to update the `warn()` calls to use `lineNr + 1` (50 instances) (pull request coming shortly).

To reproduce: in a temporary directory, unpack [example.tar.gz](https://github.com/doxygen/doxygen/files/4979919/example.tar.gz) and run `doxygen 1>/dev/null`.
Observe some out-by-1 errors (all except fred and garply).

albert-github:
A small initial analysis:
- when using `MARKDOWN_SUPPORT=NO` it looks to me that the line numbers are better. 
- when changing the initial comment block in `bar.c` into:
  ```
  /**
   *
   * \defgroup qux
   * \endinternal
   * qux does not have a title line.
   * \n
   *
   *
   * @{
   */
  ```
  but the warnings are still:
  ```
  .../bar.c:1: warning: missing title after \defgroup qux
  .../bar.c:2: warning: found \endinternal without matching \internal
  ```
  instead of the now required values of 3 and 4 for the line numbers.
- looking at the code it looks like that the linenumber as given to the call of `CommentScanner::parseCommentBlock` does not take into account the number of "empty" lines in the beginning of a comment block.

Side note:
Instead of `doxygen 1>/dev/null` you could also set `QUET=YES` in your doxygen configuration file.

albert-github:
I've just pushed a proposed patch (in my opinion better patch), pull request #7925.

duncan-roe:
Agree #7925 fixes line numbers when there are empty comment lines while #7924 does not. Please commit #7925

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.19.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
[...]

PR #7924, opened 2020-07-27
issue #7923: source line numbers in warnings output by parser are off…

duncan-roe:
There should be no side-effects to this change, since it only alters warning message calls.


albert-github:
See my comments with the original issue #7923.

albert-github:
I've just pushed a proposed patch (in my opinion better patch), pull request #7925.

doxygen:
This has been superseded by pull request #7925.

PR #7932, opened 2020-07-29
crossref citations are shown unconditionally

albert-github:
The crossref items in a bib file are unconditionally added to the list of used citations although the the citation to which the crossref belongs to is not used.
This has been fixed. The problem was seen in the CGAL output.

Note: a crossref in LaTeX is a bit different from the normal understanding of cross-reference, from https://tex.stackexchange.com/questions/401138/what-is-the-bibtex-crossref-field-used-for: "crossref can be used if you have multiple entries referring to the same proceeding, book or similar."

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/4995973/example.tar.gz)

### DIFF
--- a/src/cite.cpp
+++ b/src/cite.cpp
@@ -139,6 +139,7 @@ void CitationManager::generatePage()
         input.at(fi.size())='\0';
         int pos=0;
         int s;
+        QCString label1;
         while ((s=input.find('\n',pos))!=-1)
         {
           QCString line = input.mid((uint)pos,(uint)(s-pos));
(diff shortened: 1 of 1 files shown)


src/cite.cpp
```diff
@@ -152,12 +153,48 @@ void CitationManager::generatePage()
...
+            // assumption entry like: "@book { name," or "@book { name" (spaces optional)
+            int j=line.find("{",0);
+            // when no {, go hunting for it
+            while (j==-1 && (s=input.find('\n',pos))!=-1)
+            {
+              line = input.mid((uint)pos,(uint)(s-pos));
+              j=line.find("{",0);
+              pos=s+1;
+            }
+            // search for the name
```
doxygen, line 176: Is it guaranteed that a "{" will always be found? otherwise j will be -1 and this will be used below in find...

tip: to find a character starting at position 0, one can do `line.find('{')` instead of `line.find("{",0)` which is shorter and more efficient.

A comment with a typical input example and what is ultimately assigned to `label1` would help to understand the code... if possible also use a more descriptive variable name.

albert-github: - tip about `find`, implemented
- replaced `label..` with more descriptive names
- found some regressions, especially when no `citeName` is present, this only occurs at the end of a bib fie (and actually the bib  is not correct). At the end of the bib file one has to add something like:
  - `@book ,`
  - `@book { ,`

src/cite.cpp
```diff
@@ -152,46 +152,49 @@ void CitationManager::generatePage()
...
+                if (k != -1)
+                {
+                  citeName = line.mid((uint)(j+1),(uint)(k-j-1));
+                }
+                else
+                {
+                  citeName = line.mid((uint)(j+1));
+                }
+                citeName = citeName.stripWhiteSpace();
+                j = -1;
```
doxygen, line 192: after this assignment the while loop at line 180 will do a find with position variable j==-1 on line 182. That doesn't look correct. May the `s!=-1` guard should be `j`? code is still a bit cryptic for my understanding.

albert-github: This index looks indeed a bit suspicious, but the guard definitely should not be based on `j` as we want to detect the end of the "file" / "buffer" in the while loop.
I have to check some "nasty" examples.

albert-github: Indeed there was a regression regarding the `,`. Added update and example (in separate comment, not allowed in review).

albert-github:
In review no attachment possible.
Indeed there was a regression regarding the `,`'. Added update and example: [example_3.tar.gz](https://github.com/doxygen/doxygen/files/5007040/example_3.tar.gz)

doxygen:
@albert-github I did a few more tweaks: 
- I move the cross reference searching code to a member function `insertCrossReferencesForBibFile()`
- I switched the order of the main if...else... to reflect that first a cite name has to be found and then a crossref inside it.
- I introduced a few small lambda functions to make the line processing a bit easier to read.

albert-github:
Looks OK, it remains )of course) difficult to process the file properly, but maybe in this way it is easier to adopt when necessary.

PR #7939, opened 2020-08-02
Travis has error in tests but doesn't stop

albert-github:
On case a travis run has e test error it does not stop.
Test:
- disable appveyor
- invoke error
- disable linux-ppc64le and linux-s390x

NOTE: THIS IS A TEST NOT MEANT FOR INCLUSION INTO MASTER


.travis.yml
```diff
@@ -125,4 +125,5 @@ script:
     - if [ ! "${TRAVIS_OS_NAME}" == "osx" ]; then
```
doxygen, line 125: it seems that each line under script is executed as a shell command. I think the `if..` construct causes the return value of the script not be given back to Travis. There is an option in bash to return an error for each command (`set -e`) but that doesn't work well with Travis (https://github.com/travis-ci/docs-travis-ci-com/issues/1672). Suggested workaround is to put all the commands under script in a build.sh file and do `set -e` inside the script.

albert-github: - `if ... osx` is not necessary as `osx` is not build (temporary), but it should work anyway
- we probably need a (temporary) `if` for the `clang` processing as well
- first test indicate that "workaround": `make tests && make docs;` works

@doxygen what is your opinion about this solution?

albert-github: - probably even better to use `make tests docs` instead of `make tests && make docs`

albert-github:
Superseded by #7940

.travis.yml
doxygen: I'm working on a change that allows run tests in parallel using `ctest -j`. I'll commit it soon. So then something like `ctest -j4 && make docs` is more appropriate.

albert-github: Seen the speed of the tests I'm not sure that it is necessary to have `-j4` here.

doxygen: I didn't know/forgot about the `--pool` option 😳. I guess using ctest still has the advantage of being more standardized for cmake based projects, but I'm still not too happy about having to spell out the whole list of tests in the `CMakeLists.txt` because of the issues with GLOB.

PR #7957, opened 2020-08-11
Support commented out argument names in declarations

tolnaisz:
The functionality is controlled by a new configuration EXTRACT_ANON_ARGUMENTS.

There are coding styles where function declarations, especially in headers, use commented out argument names or no argument names at all. In this case adding documentation of the function and its arguments to the header doesn't give the required/expected results. The arguments cannot be documented as named ones as the names are missing leading to warnings during output generation and also the resulting documentation will not be perfect, as the function prototype will show no names for the arguments, but there will be description for named parameters. The reader must then rely solely on order of the documented arguments.

This change adds a new config point (EXTRACT_ANON_ARGUMENTS) which by default is off, so functionality is 100% backward compatible. When it gets enabled, any parameter name from a function definition gets copied to the declaration of the function.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -916,6 +916,14 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+    <option type='bool' id='EXTRACT_ANON_ARGUMENTS' defval='0'>
+      <docs>
+<![CDATA[
+ If this flag is set to \c YES, the name of anonymous arguments in declarations will be
+ extracted from the definitions. By default anonymous arguments will stay nameless.
+]]>
+      </docs>
+    </option>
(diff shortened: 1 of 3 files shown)


tolnaisz:
When is it planned to get this reviewed and potentially pushed to master? @doxygen, do I need to create an issue for this pull request to ensure it gets some attention?

doxygen:
@tolnaisz no it is ok like this, be patient, I'll get to it.

doxygen:
@tolnaisz Note that I've enamed `EXTRACT_ANON_ARGUMENTS` to `RESOLVE_UNNAMED_PARAMS` and enabled it by default. See https://github.com/doxygen/doxygen/commit/09b04bcb61a0e127cf502bc0533cd9bce9293406. Let me know if you see any issues with these changes.

PR #7971, opened 2020-08-19
Fix generation of XML group ids

utzig:
This fixes an issue seen using Sphinx+Breathe to generate documentation, when references are added in the comment header to some other symbols, when those are loaded into Breathe using the `doxygengroup` directive.

The issue stems from a discrepancy in the id/refid generation. On the Doxygen XML output, all "refid" properties have the following format:

`<some_prefix_anchorid_1anchorid>`, where `anchorid` exists twice (it might be called compoundid in the code as well).

But the `id` property might have different formats depending whether it exists inside a <group_*.xml> file or any other XML file. In the non-group XML files the `id` has the following format:

`<some_prefix_anchorid_1anchorid>` which matches the `refid` and the symbol resolution works correctly.

The <group_*.xml> files though, have the following format:

`<some_prefix_1anchorid>`, which is missing the "duplicated" anchorid.

This commit adds the extra suffix before adding the anchor so all symbols get the same names and resolution works correctly no matter what directive is used in Breathe.


utzig:
This is related to https://github.com/michaeljones/breathe/pull/566

albert-github:
You wrote:
> The issue stems from a discrepancy in the id/refid generation. 

and
> The <group_*.xml> files though, ...

- One generation you found in xmlgen.cpp, where is the other generation?
- when creation of HTML output is the referencing OK in that case ?
- with the example from michaeljones/breathe#566 I don't see a `group....xml` file, are there special settings?
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

utzig:
This file should have both my test sources + the outputs [doxygen_pr7971.tar.gz](https://github.com/doxygen/doxygen/files/5096865/doxygen_pr7971.tar.gz)

If you want to build the html output yourself, you can just `sphinx-build -b html . html` in the directory where the tar.gz was uncompressed. I included it to avoid you having to install sphinx+breathe. But the broken links can be seen in the static `index.html` where `k_delayed_work_init()` and `k_delayed_work_submit()` where not resolved. This only happens if `.. doxygengroup::` is used in Breathe. If including with anything else, eg `.. doxygenfunction::` it works.

The origin of the "bug" can be seen looking at `doxygen/xml/refdoc_8h.xml`, one of the `refid`'s is `group__refdoc_gac42f9b985b8562386a00172c0417c0e3_1gac42f9b985b8562386a00172c0417c0e`, but if you open the group file `doxygen/xml/group__refdoc.xml` you'll see it has the `id="group__refdoc_1gac42f9b985b8562386a00172c0417c0e3` which does not match.

utzig:
> * when creation of HTML output is the referencing OK in that case ?

Do you mean the htmlgen.cpp in Doxygen? That is not used here, Breathe parses the XML output and Sphinx generates the HTML. We don't use the HTML generator in Doxygen, but I assume it works correctly.

albert-github:
Regarding: https://github.com/doxygen/doxygen/pull/7971#issuecomment-676405558
I was indeed pointing to the doxygen HTML output. The HTML and XML are very different, so when the problem is in the XML generator it is not said that the HTML generator has the same problem or not. It is always a good idea to check the doxygen HTML output as well, as when theer is also a problem it might be a more general problem.

utzig:
> Regarding: [#7971 (comment)](https://github.com/doxygen/doxygen/pull/7971#issuecomment-676405558)
> I was indeed pointing to the doxygen HTML output. The HTML and XML are very different, so when the problem is in the XML generator it is not said that the HTML generator has the same problem or not. It is always a good idea to check the doxygen HTML output as well, as when theer is also a problem it might be a more general problem.

Indeed there are no issues and now I understand why the variables are called `file` and `anchor`. The HTML generator does create a file with compound id name and also adds an anchor so a link looks like this: `href="group__refdoc_gac42f9b985b8562386a00172c0417c0e3.html#gac42f9b985b8562386a00172c0417c0e3"`, which works and makes sense. But for XML I don't think it does, because we link stuff using `refid` and `id` which IMO should match, independent from what the XML file is called.

albert-github:
Regarding: https://github.com/doxygen/doxygen/pull/7971#issuecomment-676403501, I do see the mismatch.

utzig:
> Regarding: [#7971 (comment)](https://github.com/doxygen/doxygen/pull/7971#issuecomment-676403501), I do see the mismatch.

I updated the group tests so it passes the CI, but I've this feeling that what should be done is to get rid of the duplicated `id`, so that the group `id` is actually correct, while all `refid` and other elements `id` that have this duplication are wrong.

albert-github:
When searching in the xml files we see (i.e. `grep` for `id=`) I get:

```
testing/test_output_019/out/019__defgroup_8c.xml:  <compounddef id="019__defgroup_8c" kind="file" language="C++">
testing/test_output_019/out/019__defgroup_8c.xml:      <memberdef kind="function" id="group__g1_1gae84463e3941387a9da58279761e18d7e" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_019/out/019__defgroup_8c.xml:      <memberdef kind="function" id="group__g2_1gafd310bbec27993e80a5dbaf6c54d5e0b" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_019/out/019__defgroup_8c.xml:      <memberdef kind="function" id="group__g3_1ga59de48fb40aec5ee2623cc453fe71643" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_019/out/019__defgroup_8c.xml:      <memberdef kind="function" id="group__g3_1gaf32d23522a5d9c3e7fed3dd5710001f7" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_019/out/group__g1.xml:  <compounddef id="group__g1" kind="group">
testing/test_output_019/out/group__g1.xml:      <memberdef kind="function" id="group__g1_1gae84463e3941387a9da58279761e18d7e" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_019/out/group__g2.xml:  <compounddef id="group__g2" kind="group">
testing/test_output_019/out/group__g2.xml:      <memberdef kind="function" id="group__g2_1gafd310bbec27993e80a5dbaf6c54d5e0b" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_019/out/group__g3.xml:  <compounddef id="group__g3" kind="group">
testing/test_output_019/out/group__g3.xml:      <memberdef kind="function" id="group__g3_1ga59de48fb40aec5ee2623cc453fe71643" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_019/out/group__g3.xml:      <memberdef kind="function" id="group__g3_1gaf32d23522a5d9c3e7fed3dd5710001f7" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">

```

So in different files there are the same `id`s.
This behavior can also be seen in other testing directories!

I'm not sure if this is a bug or that an `id` should be unique only in a file. I think that an `id` should be unique inside a project as when I search fir `group__g3` (as complete word) I get (and this makes sense to me):
```
testing/test_output_019/out/group__g2.xml:    <innergroup refid="group__g3">Third Group</innergroup>
testing/test_output_019/out/group__g3.xml:  <compounddef id="group__g3" kind="group">
testing/test_output_019/out/index.xml:  <compound refid="group__g3" kind="group"><name>g3</name>
```

@ doxygen
- what is correct:
  - `id` unique per file
  - `id` unique per project
- what to do with the other double `id`s? (I created a small but horrible script to search these doubles and the missing `id`s)

utzig:
I don't see why there would be a problem with having duplicated `id` in a project as long as a symbol can be uniquely identified.

What concerns me most here, is what to do if a `refid` does point to an `id` value for which the target `id` property's value does not exist, because both were created differently?

This is what happens in my previous example, where: `refid=group__refdoc_gac42f9b985b8562386a00172c0417c0e3_1gac42f9b985b8562386a00172c0417c0e` and `id="group__refdoc_1gac42f9b985b8562386a00172c0417c0e3`.

albert-github:
The:
> What concerns me most here, is what to do if a `refid` does point to an id value for which the target `id` property's value does not exist, because both were created differently?

This should not happen at all.

How would the correct reference be in the case of:
```
testing/test_output_078/out/078__xml__namespace__members__in__file__scope_8h.xml:      <memberdef kind="function" id="078__xml__namespace__members__in__file__scope_8h_1a0f1fe1a972c7c4196988a1bdde63ec77" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
testing/test_output_078/out/index.xml:    <member refid="078__xml__namespace__members__in__file__scope_8h_1a0f1fe1a972c7c4196988a1bdde63ec77" kind="function"><name>foo</name></member>
testing/test_output_078/out/index.xml:    <member refid="078__xml__namespace__members__in__file__scope_8h_1a0f1fe1a972c7c4196988a1bdde63ec77" kind="function"><name>foo</name></member>
testing/test_output_078/out/namespace_namespace.xml:      <memberdef kind="function" id="078__xml__namespace__members__in__file__scope_8h_1a0f1fe1a972c7c4196988a1bdde63ec77" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">

```

doxygen:
This issue seems to be related to setting `SEPARATE_MEMBER_PAGES` to `YES`. Was there a reason to enable this?
In any case we have to be very careful fixing this as it could impact what should be stable references in exiting output.

utzig:
> This issue seems to be related to setting `SEPARATE_MEMBER_PAGES` to `YES`. Was there a reason to enable this?
> In any case we have to be very careful fixing this as it could impact what should be stable references in existing output.

Sorry for the delay, but yes, you got it! I don't know why we use this option because it was done very very early in the project, before Sphinx documentation was even introduced and those pages are still maintained. Looking at one example: https://docs.zephyrproject.org/apidoc/latest/group__fifo__apis_gab1c5212040d12cbb92cede5cf54928ba.html#gab1c5212040d12cbb92cede5cf54928ba, I can kinda see why it might have been done; it looks pretty good to have each function in a separate page. 

For the HTML output this makes a lot of sense and even different file names are created. But for XML the files all have the same names, without the option set to NO or YES, only `refid` and `id` do change, so it begs the question, would not it be more correct for this option to not have any effect on the XML output? If you think this is reasonable I can work on a patch.

Issue #7914, opened 2020-07-19
Doxygen Treats C# Inline-Bodied Properties as Member Variables


justin-millman:
**Describe the bug**
C# allows you to implement a read-only property with the `=>` inline syntax; for example,
```csharp
namespace Example {
    public class Foo {
        public int Property => 10;
    }
}
```
This is equivalent to
```csharp
namespace Example {
    public class Foo {
        public int Property { get { return 10; } }
    }
}
```
but can be more concise in certain situations. Doxygen misparses properties of this form as member variables and renders them as such, rather than as properties with only a getter.

**Expected behavior**
Doxygen should recognize this syntax as a read-only property and generate documentation equivalent to that which is generates in the equivalence shown above.

**Version**
Doxygen version 1.8.18 (commit hash a1b07ad0e92e4526c9ba1711d39f06b58c2a7459) on a Windows 10 machine

**Additional context**
I wrote myself a short Python script to serve as a `.cs` filter for Doxygen. It finds these properties and replaces them with the getter-only equivalence, which allows me to get the output I desire. I don't know if the regular expression I generated is truly exhaustive or if there are any false-positives, but so far I haven't had any issues. I've attached the Python script here (as a `LOG` file to satisfy GitHub).

[ReadonlyPropertyFilter.log](https://github.com/doxygen/doxygen/files/4944081/ReadonlyPropertyFilter.log)

justin-millman:
Here is a minimal example using a default-generated Doxyfile. The relevant content can all be found at `Example/Example`.

[Example.tar.gz](https://github.com/doxygen/doxygen/files/4944117/Example.tar.gz)

justin-millman:
This is a duplicate of #7550, #6278, and #6064. Apologies for that . . . but I do think this should be fixed.

albert-github:
I've just pushed a proposed patch, pull request #7991

Issue #7550, opened 2020-01-29
C# Expression-Bodied Properties reveals underline implementation


peval27:
I've spotted that when writing something like:

```
/// <summary>
/// TCP port
/// </summary>
public int TcpPort => DefaultTcpPort;
```
the generated documentation is

> int Namespace.ClassName.TcpPort = >DefaultTcpPort

note the space between `=` and `>`.

This is not acceptable as it reveals an implementation detail to the public.

Used latest doxygen 1.8.17

albert-github:
I don't see the space between the `=` and `>` in the HTML output, but I don't have the complete example just the lines you gave, so:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

Maybe also have a look at the configuration setting `MAX_INITIALIZER_LINES`.

sgrigalunas:
I think this issue is with the Doxygen bug related to C# expression-bodied member properties. 

This has been reported in:
https://github.com/doxygen/doxygen/issues/6278
https://github.com/doxygen/doxygen/issues/6064

(not sure about the relevance of the space for this issue - but this bug exposes elements of the source code which may be the 'implementation detail' he is referring to)

albert-github:
I've just pushed a proposed patch, pull request #7991 for the generatl issue, the space between the `=` and `>` I never saw.

albert-github:
Code of #7991 has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

In case the current master (1.9.0 (9cf80a9)) does not solve the issue:
- please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.
- describe what is still going wrong.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

Issue #6278, opened 2018-07-02
c# expression-bodied property is not rendered correctly (Origin: bugzilla #789318)


doxygen:
status NEEDINFO severity *normal* in component *general* for *---*
Reported in version *1.8.13* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[DoxyGenBugRepro.zip](https://sourceforge.net/projects/doxygen/files/attachments/bug789318_362060_DoxyGenBugRepro.zip/download/download)_ (ID 362060)

On 2017-10-22 11:32:09 +0000, Harmen wrote:

    I have an expression-bodied property:
    
    public static MultiValueDictionary<TKey, TValue> Empty => new
    MultiValueDictionary<TKey, TValue>();
    
    that is renderd by doxygen as:
    
    MultiValueDictionary<TKey, TValue> MultiValueDictionary< TKey, TValue
    >.Empty => new MultiValueDictionary<TKey
    
    Expected output is:
    
    MultiValueDictionary<TKey, TValue> MultiValueDictionary< TKey, TValue
    >.Empty

On 2017-10-22 13:47:14 +0000, albert wrote:

    Can you please attach a (small) self-contained example (source+config file
    in a tar or zip) that allows us to reproduce the problem?

On 2017-10-22 20:35:59 +0000, Harmen wrote:

    Created attachment 362060
    Repro
    
    Source and doxyfile to reproduce the problem

On 2017-10-22 20:38:17 +0000, Harmen wrote:

    the zip file also contains the generated documentation (docs subfolder)

albert-github:
I've just pushed a proposed patch, pull request #7991

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

Issue #6064, opened 2018-07-02
Lambda for read-only property in C# 6 (Origin: bugzilla #771283)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *unspecified* on platform *Other*
Assigned to: Dimitri van Heesch

On 2016-09-12 10:53:15 +0000, Magnus Ãsterlund wrote:

    In C# 6 it is a new feature to use lambda to express a read-only property
    Eg.
    
    public string Name => _firstName + " " + _lastName
    
    Expected documentation
    **********************
    Properties
    string Name [get]
    
    
    Resulting documentation
    ***********************
    Public Attributes
    string Name => _firstName + " " + _lastName

On 2016-10-04 19:10:56 +0000, Andy Dent wrote:

    These are officially called "Expression-bodied properties" as per
    https://msdn.microsoft.com/en-gb/magazine/dn802602.aspx
    
    We are experiencing this using Doxygen to generate the Realm API docs
    
    Our internal bug https://github.com/realm/realm-dotnet/issues/854
    
    A sample of the output
    https://realm.io/docs/xamarin/0.78.0/api/class_realms_1_1_realm_object.html

magol:
I'm the reporter of this issue

albert-github:
@magol 
A bit of a late reaction
- with another issue I wrote 
  > Maybe also have a look at the configuration setting `MAX_INITIALIZER_LINES`

  when setting `MAX_INITIALIZER_LINES=0` the part `=> ...` is not present.
- You mentioned C# 6.0 is this the version described in? 
  ECMA-334 
  C# Language Specification
  5th Edition / December 2017

magol:
Of course, you never print the content of a function in the documentation. For the same reason, Doxygen should never print the content after `=>` to the documentation, as this is exactly the same thing as a function. `=>` is simply an easier way to write a single-line function.  So, `MAX_INITIALIZER_LINES=0` should not be a solution for this. 

I did try to find the definition for this in the C# Language Specification, but it was hard to read. I think you can look at chapter 12.16 for an specification of `=>`

albert-github:
I've just pushed a proposed patch, pull request #7991

albert-github:
Code of #7991 has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

[...]

PR #7991, opened 2020-09-01
bug 789318 c# expression-bodied property is not rendered correctly

albert-github:
Create possibility to see the CS construct `=>` not as an initializer (based on just the `=`) but as getter.
This problem solves besides this issue:
- #6278 c# expression-bodied property is not rendered correctly (Origin: bugzilla #789318)
also the issues:
- #6064 Lambda for read-only property in C# 6 (Origin: bugzilla #771283)
- #7550 C# Expression-Bodied Properties reveals underline implementation
- #7914 Doxygen Treats C# Inline-Bodied Properties as Member Variables


trovialdo:
You might have mislabeled this as C++; it should be C#?

Issue #7969, opened 2020-08-19
Relocatable binaries (--prefix=)


bmhm:
**Describe the bug**

The following command will not work:
`PREFIX=$HOME/.local/apps/doxygen make install --prefix=$HOME/.local/apps/doxygen`

This is useful if the user does not have root access.

**Expected behavior**

make install will install into the dir given with "prefix".

**Screenshots**

n/a

**To Reproduce**

`make install --prefix=$HOME/.local/apps/doxygen`
or
`PREFIX=$HOME/.local/apps/doxygen make install`

**Version**

git master as of 2020-08-19 and any previous version

**Stack trace**

n/a

**Additional context**

* https://www.gnu.org/software/gnulib/manual/html_node/Supporting-Relocation.html
* https://softwareengineering.stackexchange.com/questions/314327/what-is-a-relocatable-binary

bmhm:
Actually, `make DESTDIR=$PWD/installdir install` works just fine.

albert-github:
I myself never used the --prefix / PREFIX so far, but from a quick search in the cmake documentation I found `DESTDIR` (https://cmake.org/cmake/help/v3.18/variable/CMAKE_INSTALL_PREFIX.html), maybe this helps (it does not look 100% optimal to me though)

Bagira80:
While cmaking you need to set `CMAKE_INSTALL_PREFIX` to the location you later want to install to. Then a simple `make install` will install it there.

With the `DESTDIR` mechanism you can prepend another path to the configured install location. (However, it does not replace the configured install prefix!)

(See my comment here: https://github.com/doxygen/doxygen/pull/7992#issuecomment-684829354)

Bagira80:
Maybe this is of interested, too: https://stackoverflow.com/a/11307770/3115457

bmhm:
TY. I can create a PR for the docs if you like. I only do java and bash (mostly) and do not have experience with cmake.

But what I found out: using the strings Viber's there are references to the original install path. I don't think they should be there...

Just using another install dir does not make it relocatable.

Bagira80:
I am sorry, I cannot really follow? (Besides, I am just a normal contributor and not involved with Doxygen.)

What is it that you want to achieve with relocatable binaries?

[...]

PR #7992, opened 2020-09-01
Respect value of CMAKE_INSTALL_PREFIX when packaging

Bagira80:
Before this commit the value of `CMAKE_INSTALL_PREFIX` was ignored when creating a package with *CMake*/*CPack*. As a result the generated package would always install *Doxygen* to the default location (`/usr/bin` on Linux). `make install` on the other hand would install *Doxygen* to the location specified in variable `CMAKE_INSTALL_PREFIX`.

With this change the value of the variable `CMAKE_INSTALL_PREFIX` will be used when creating the package, too.

### DIFF
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -10,6 +10,7 @@
 # - http://archive.ubuntu.com/ubuntu/pool/main/d/doxygen (http://old-releases.ubuntu.com/ubuntu/pool/main/d/doxygen)
 # - http://rpmfind.net/linux/rpm2html/search.php?query=doxygen
 
+set(CPACK_PACKAGING_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}")
 set(CPACK_STRIP_FILES      ON)
 set(CPACK_PACKAGE_NAME     ${PROJECT_NAME} )
 set(CPACK_PACKAGE_VERSION  ${VERSION})


albert-github:
I assume this is the fix for :
> Relocatable binaries (--prefix=) #7969

Bagira80:
> I assume this is the fix for :
> 
> > Relocatable binaries (--prefix=) #7969

Not really.  
#7969 looks more like a usage error.

While configuring/cmaking one has to set `CMAKE_INSTALL_PREFIX` to the location where *Doxygen* shall be installed. It is the same as the `--prefix` option of *Autoconf/Automake*. `make install` without any extra option will then do the correct thing and install to that location.

If the user does not have write access to that location (s)he can "relocate" the installation to another directory using the `DESTDIR` mechanism. E.g.: `DESTDIR=/tmp/my_doxygen make install`  
This will then prepend `/tmp/my_doxygen` to the prefix configured before with `CMAKE_INSTALL_PREFIX`.

All this already works. (That's why I say that #7969 is a usage error.)

---

However, the current merge-request addresses the creation of an installation package for e.g. Ubuntu (`doxygen_1.8.20_amd64.deb`) which should install at the specified prefix (`CMAKE_INSTALL_PREFIX`).

Bagira80:
I am not sure if the Windows Appveyor builds have a problem with my change or some other problem (out of space?). Is there some way to see more details?

However, I found a possible different solution for using the value of `CMAKE_INSTALL_PREFIX` when packaging. At least CMake's help for [`CPACK_SET_DESTDIR`](https://cmake.org/cmake/help/latest/variable/CPACK_SET_DESTDIR.html) says:

>  Nevertheless, if CPACK_SET_DESTDIR is set then CPack will set DESTDIR before doing the local install. The most noticeable difference is that without CPACK_SET_DESTDIR, CPack uses CPACK_PACKAGING_INSTALL_PREFIX as a prefix whereas with CPACK_SET_DESTDIR set, CPack will use CMAKE_INSTALL_PREFIX as a prefix.

So, maybe one should just set `CPACK_SET_DESTDIR` instead.

PR #8005, opened 2020-09-03
Nice doxy code

deem0n:
Hello,

This code is for the process of converting Markdown to LaTeX.
I wanted to use [listings](https://ctan.org/pkg/listings?lang=en) package for code blocks. But that requires verbatim listing text, while `doxygen` parses text into LaTeX line by line:

```latex
\begin{DoxyCode}{0}
\DoxyCodeLine{\{}
\DoxyCodeLine{  "{}id"{}:98765,}
\DoxyCodeLine{  "{}parent\_id"{}:null}
\DoxyCodeLine{\}}
\end{DoxyCode}
```

`listings` package expects something like this:

```latex
\begin{DoxyCode}{0}
{
      "id":98765,
      "parent_id":null
}
\end{DoxyCode}
```

To keep backward compatibility a just added new config flag for Doxyfile named `VERBATIM_LISTINGS`. When it is set to YES doxygen generates verbatim listings enclosed into `\begin{DoxyVerbatimCode} ... \end{DoxyVerbatimCode}`.

User can attach `listings` environment to the DoxyVerbatimCode with:

```latex
\lstnewenvironment{DoxyVerbatimCode}{}{}
```

We can provide default settings for listings in the `doxygen.sty`. I also tried to get programming language name from the code blocks presented in the source markdown, but failed. It is always set to '.md' and comes from the source file name extension. It would be nice to parse lang names from code blocks:
```
===javascript

or

===sql
```

and send it as argument to the `listings`, so the code would be highlighted accordingly in the LaTeX document.
Example with listings (made with doxygen from markdown):

image

Another small fix is for narrow table cells. `pdflatex` is known to never wrap first word in the paragraph. So text in narrow table cells might flow to the neighbor cells:

image

With the fix it looks much better:

image

The idea comes from this post: http://sigstp.blogspot.com/2010/09/latex-hyphenate-single-long-words-in.html


src/config.xml
```diff
@@ -276,6 +276,17 @@ Go to the <a href="commands.html">next</a> section or return to the
...
+      </docs>
+    </option>
+    <option type='bool' id='VERBATIM_LISTINGS' defval='0'>
+      <docs>
+<![CDATA[
+ If the \c VERBATIM_LISTINGS tag is set to \c YES then doxygen will uses
+ \begin{DoxyVerbatimCode} \end{DoxyVerbatimCode} LaTeX commands for enclosing source code.
+ Source code is provided as is (verbatim).
+ When it is set to \c NO then doxygen uses \begin{DoxyCode}. The source code is processed and 
+ provided as LaTeX commands
```
albert-github, line 289: there are here some terms (like `\begin{DoxyVerbatimCode}`) that are not understood by a normal user but are internal LaTeX codes and should not be used in the documentation.

src/latexdocvisitor.cpp
```diff
@@ -1283,6 +1292,10 @@ void LatexDocVisitor::visitPre(DocHtmlCell *c)
   if (c->isHeading())
   {
     m_t << "\\cellcolor{\\tableheadbgcolor}\\textbf{ ";
+  } 
+  else 
+  {
+    if (c->isFirst()) m_t << "\\hspace{0pt}";
```
albert-github, line 1298: I see that it is for the hyphenation of the words (http://sigstp.blogspot.com/2010/09/latex-hyphenate-single-long-words-in.html).
I think this should not be hidden into a proposed pull request for a completely unrelated problem and should be a separate proposed pull request.

albert-github (commented):
What is the reason for wanting to use the LaTeX listings package. Won't you loose a lot of reference possibilities between code listings and code documenattion?

deem0n:
> What is the reason for wanting to use the LaTeX listings package. Won't you loose a lot of reference possibilities between code listings and code documenattion?

We have several .md documents with general documentation about different software modules. Source code listings include `curl` examples, XML & JSON configs, some bash snippets, JavaScript, SQL. And we need to display Cyrillic comments and parts in the listings as well. Default `doxygen` listing works quite well actually. But not always. Mainly I want 2 things: line numbering and smart word wrap. Both things probably could be done with LaTeX hacking, where I am not an expert. It is much easier to use `listings` package, but probably it is not right way with `doxygen`. Consider default doxygen output:

image

with listings I can define literate block like this:
```latex
literate=
         {\&}{}{1\discretionary{\&\return}{}{\&}}
         {;}{}{1\discretionary{;\return}{}{;}}
         {]}{}{1\discretionary{]\return}{}{]}}
         {/}{}{1\discretionary{/\return}{}{/}}
         {>}{}{1\discretionary{>\return}{}{>}}
         {=}{}{1\discretionary{=\return}{}{=}}
         {></}{}{1\discretionary{>\return}{</}{></}}
```
and get

image

Probably I can do the same with default `doxygen` output. Let me check. In any case `listings` package might be a viable option especially if users can switch it on/off. We can even make it work only for markdown documents to leave current working functionality for all other files (.c .cc etc) unaffected.

src/config.xml
deem0n: I will try to rephrase it, and I see that `\` char fails doc generation for `doxygen` anyway.

albert-github: As a side note: putting the `\end{DoxyVerbatimCode}` between backticks (`` ` ``) would most likely already solve the warning, but the "not understood" will still remain of course.

albert-github:
Can you make a small test project showing the output from the standard doxygen approach and your approach (will result in 2 runs), so we can have a look at it as well and see whether or not something is possible inside doxygen.
The self contained example (source+configuration file in a tar or zip) should also the include the latex directories for both approaches.

Is this problem limited to LaTeX or is it actually, in general, also valid for other output formats (like HTML, RTF, ...).

Note: I don't speak / read / write Russian so this might be quite challenging.

deem0n:
It was quite a dive into LaTeX, but finally i am quite happy with the results. Actually I made other small fixes for different things. Let's continue discussion on the verbatim source code here and I'll post different small PRs separately for each issue I found.

deem0n:
I made #8040 which is only for nice verbatim listings. This PR should be closed.

PR #8025, opened 2020-09-09
 without this patch tables can overflow when page contains floats

deem0n:
See https://www.latex-project.org/cgi-bin/ltxbugs2html?pr=tools/3512

### DIFF
--- a/templates/latex/longtable_doxygen.sty
+++ b/templates/latex/longtable_doxygen.sty
@@ -164,7 +164,7 @@
   \ifdim \dimen@>\z@\vfil\break\fi
       \global\@colroom\@colht
   \ifvoid\LT@foot\else
-    \advance\vsize-\ht\LT@foot
+    \global\advance\vsize-\ht\LT@foot
     \global\advance\@colroom-\ht\LT@foot
     \dimen@\pagegoal\advance\dimen@-\ht\LT@foot\pagegoal\dimen@
     \maxdepth\z@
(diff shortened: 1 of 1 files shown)


albert-github:
Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that shows the problem that should be fixed with the proposed patch? Please don't add external links as they might not be persistent.

albert-github (commented):
I don't think that the `longthable_doxygen.sty` is the way to go as here we have a frozen version of `longtable.sty` so that the tabu table environment continues to work as currently the tabu package doesn't work with the newer versions of LaTeX (see also https://github.com/tabu-issues-for-future-maintainer/tabu).

deem0n:
sad news though: https://github.com/latex3/latex2e/issues/158

albert-github:
Good to see that there is also a reference to the base of the problem.

My thoughts: Depending on the way how the  issue will be fixed in latex3 / latex2e (i.e. in the longtable part or in a kernel / other part) we can, when necessary, also update the longtable used by doxygen.

albert-github:
@deem0n can you have a look at the suggestion in https://github.com/latex3/latex2e/issues/158#issuecomment-690078001

deem0n:
I've checked quite a long doc (60 pages) with many tables. I see no problems when I use classic doxygen code formatter. I guess when I used my verbatim code env the bug was triggered, probably because of so named floats. Let me double check. Please keep this PR while I getting more info.

deem0n:
> @deem0n can you have a look at the suggestion in [latex3/latex2e#158 (comment)](https://github.com/latex3/latex2e/issues/158#issuecomment-690078001)

Thanks, I will try the `LaTeX` latest version on my long doc with classic code formatter and with my verbatim patch as well.

deem0n:
For now I can confirm that verbatim listings #8040 may results in overflowing tables. So this PR should be worked out if #8040 will be hopefully accepted.

PR #8026, opened 2020-09-12
Fix too aggressive hyphenation of abbr. words.

deem0n:


deem0n:
After the patch:
image

Before the patch:
image

I think we should not hyphenate all UPPERCASE words, it might look ugly. Upper-Case-Words style hyphenation is Ok though.

src/util.cpp
```diff
@@ -2263,7 +2263,7 @@ QCString argListToString(const ArgumentList &al,bool useCanonicalType,bool showD
   if (al.volatileSpecifier()) result+=" volatile";
   if (al.refQualifier()==RefQualifierLValue) result+=" &";
   else if (al.refQualifier()==RefQualifierRValue) result+=" &&";
-  if (!al.trailingReturnType().isEmpty()) result+=al.trailingReturnType();
+  if (!al.trailingReturnType().isEmpty()) result+=" -> "+al.trailingReturnType();
```
albert-github, line 2266: I don't think this fix is not a fix but an undo of #8013 and should not be present (It certainly has nothing tho do with hyphenation).

albert-github (changes requested):

src/util.cpp
deem0n: Hmm, I don't have good knowledge about `doxygen` internals, but latest master branch may produce this kind of LaTeX:

image

Should I try proposed change in `src/utils.cpp` at line 2266 to avoid this formatting?

albert-github: Line 2266 of `util.cpp` has nothing to do with hyphenation, so your real proposed patch (at line 6300) should stay. I think you should update your pull proposed request so that the line 2266 of `src/utils.cpp` is not changed.

deem0n: Fine, I had no intention to change that line. Fixed.

PR #8028, opened 2020-09-13
Force first word in table cell to hyphenate

deem0n:
Added invisible space before first word in every table cell to allow wraps.
http://sigstp.blogspot.com/2010/09/latex-hyphenate-single-long-words-in.html

Before the patch:
Screenshot 2020-09-13 at 22 55 58 PM

After the patch:
Screenshot 2020-09-13 at 23 02 33 PM

### DIFF
--- a/src/latexdocvisitor.cpp
+++ b/src/latexdocvisitor.cpp
@@ -1284,6 +1284,10 @@ void LatexDocVisitor::visitPre(DocHtmlCell *c)
   {
     m_t << "\\cellcolor{\\tableheadbgcolor}\\textbf{ ";
   }
+  else 
+  {
+    if (c->isFirst()) m_t << "\\hspace{0pt}";
+  }
   if (cs>1)
   {
     setCurrentColumn(currentColumn()+cs-1);
@@ -1307,7 +1311,11 @@ void LatexDocVisitor::visitPost(DocHtmlCell *c)
     setInColSpan(FALSE);
     m_t << "}";
   }
-  if (!c->isLast()) m_t << "&";
+  if (!c->isLast()) {
+    m_t << "&";
+    if (!c->isHeading()) m_t << "\\hspace{0pt}";
+  }
+
 }
 
 void LatexDocVisitor::visitPre(DocInternal *)
--- a/src/sqlcode.h
+++ b/src/sqlcode.h
@@ -1,12 +1,10 @@
 /******************************************************************************
  *
- * 
- *
- * Copyright (C) 1997-2014 by Dimitri van Heesch.
+ * Copyright (C) 1997-2020 by Dimitri van Heesch.
  *
  * Permission to use, copy, modify, and distribute this software and its
- * documentation under the terms of the GNU General Public License is hereby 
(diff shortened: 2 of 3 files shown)


src/latexdocvisitor.cpp
```diff
@@ -1284,6 +1284,10 @@ void LatexDocVisitor::visitPre(DocHtmlCell *c)
   {
     m_t << "\\cellcolor{\\tableheadbgcolor}\\textbf{ ";
   }
+  else 
+  {
+    if (c->isFirst()) m_t << "\\hspace{0pt}";
+  }
```
albert-github, line 1290: When not mistaken I saw this proposed patch also as part of #8025. Probably it is better to have as a separate proposed patch, but in that case it should be removed from #8025

src/latexdocvisitor.cpp
```diff
@@ -1307,7 +1311,11 @@ void LatexDocVisitor::visitPost(DocHtmlCell *c)
     setInColSpan(FALSE);
     m_t << "}";
   }
-  if (!c->isLast()) m_t << "&";
+  if (!c->isLast()) {
+    m_t << "&";
+    if (!c->isHeading()) m_t << "\\hspace{0pt}";
+  }
+
```
albert-github, line 1318: When not mistaken I saw this proposed patch also as part of #8025. Probably it is better to have as a separate proposed patch, but in that case it should be removed from #8025

src/sqlcode.h
```diff
@@ -1,12 +1,10 @@
 /******************************************************************************
  *
- * 
- *
- * Copyright (C) 1997-2014 by Dimitri van Heesch.
+ * Copyright (C) 1997-2020 by Dimitri van Heesch.
```
albert-github, line 3: What is the relevance of the changes in this file in respect to the hyphenation in tables?

src/sqlcode.l
```diff
@@ -1,10 +1,10 @@
 /******************************************************************************
  *
- * Copyright (C) 1997-2014 by Dimitri van Heesch.
+ * Copyright (C) 1997-2020 by Dimitri van Heesch.
```
albert-github, line 3: What is the relevance of the changes in this file in respect to the hyphenation in tables?

doxygen: @deem0n I think you need to pull in the externally made changes in the master branch, then switch to your feature branch and to a `git rebase master` before updating the pull request (probably you did a merge now?). Otherwise changes made by others will appear inside your pull request!

deem0n: Yes, it seems I messed up with git. Is it ok if I close this PR and make another one with just one file chaned (src/latexdocvisitor.cpp) ?

doxygen: @deem0n sure no problem.

deem0n: Please see #8030

src/latexdocvisitor.cpp
albert-github: You are correct I meant #8005 (got confused by nearly the same numbers).

Issue #7898, opened 2020-07-07
Internal search engine produce ".html" pages instead of using HTML_FILE_EXTENSION


luca-aep:
While generating the built-in client side search engine (SEARCHENGINE = YES ; SERVER_BASED_SEARCH = NO ; EXTERNAL_SEARCH = NO) doxygen generates files with extension fixed to ".html" instead of using the one specified by HTML_FILE_EXTENSION.

Looking at doxygen-1.8.18 source code, affected code should be:
- line 959 of src\searchindex.cpp
- line 1271 of src\searchindex.cpp
- line 350 and 356 of templates\html\search.js

luca-aep:
I propose this patch (done starting from tag Release_1_8_20).

[patch.zip](https://github.com/doxygen/doxygen/files/5209049/patch.zip)

albert-github:
Can you convert your patch into a pull request here on github?

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #8029, opened 2020-09-14
Proposed fix for issue #7898

luca-aep:
Proposed fix for issue #7898 - Internal search engine produce ".html"pages instead of using HTML_FILE_EXTENSION

### DIFF
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -1136,10 +1136,17 @@ void HtmlGenerator::startFile(const char *name,const char *,
   bool searchEngine = Config_getBool(SEARCHENGINE);
   if (searchEngine /*&& !generateTreeView*/)
   {
+    QCString htmlFileExtension=Config_getString(HTML_FILE_EXTENSION);
+    htmlFileExtension=htmlFileExtension.stripWhiteSpace();
+    if (htmlFileExtension.isEmpty())
+    {
+      htmlFileExtension = ".html";
+    }
+
     t << "<script type=\"text/javascript\">\n";
 		t << "/* @license magnet:?xt=urn:btih:cf05388f2679ee054f2beb29a391d25f4e673ac3&amp;dn=gpl-2.0.txt GPL-v2 */\n";
-		t << "var searchBox = new SearchBox(\"searchBox\", \""
-      << m_relPath<< "search\",false,'" << theTranslator->trSearch() << "');\n";
+		t << "var searchBox = new SearchBoxExt(\"searchBox\", \""
+      << m_relPath<< "search\",false,'" << theTranslator->trSearch() << "','" << htmlFileExtension << "');\n";
 		t << "/* @license-end */\n";
     t << "</script>\n";
   }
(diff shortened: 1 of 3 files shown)


src/htmlgen.cpp
```diff
@@ -2666,15 +2673,22 @@ void HtmlGenerator::writeSearchPage()
   QFile f(fileName);
   if (f.open(IO_WriteOnly))
   {
+    QCString htmlFileExtension=Config_getString(HTML_FILE_EXTENSION);
+    htmlFileExtension=htmlFileExtension.stripWhiteSpace();
+    if (htmlFileExtension.isEmpty())
+    {
+      htmlFileExtension = ".html";
+    }
+
```
albert-github, line 2682: I don't think this correction of the extension is necessary as in `configimpl.l` we see:
```
  // check & correct HTML_FILE_EXTENSION
  QCString htmlFileExtension=Config_getString(HTML_FILE_EXTENSION);
  htmlFileExtension=htmlFileExtension.stripWhiteSpace();
  if (htmlFileExtension.isEmpty())
  {
    htmlFileExtension = ".html";
  }
  Config_updateString(HTML_FILE_EXTENSION,htmlFileExtension);
```
so the correction is done and and the corrected value  is placed back in `HTML_FILE_EXTENSION` so that subsequent calls to `Config_getString(HTML_FILE_EXTENSION)` directly return the corrected value.

src/htmlgen.cpp
```diff
@@ -1136,10 +1136,17 @@ void HtmlGenerator::startFile(const char *name,const char *,
   bool searchEngine = Config_getBool(SEARCHENGINE);
   if (searchEngine /*&& !generateTreeView*/)
   {
+    QCString htmlFileExtension=Config_getString(HTML_FILE_EXTENSION);
+    htmlFileExtension=htmlFileExtension.stripWhiteSpace();
+    if (htmlFileExtension.isEmpty())
+    {
+      htmlFileExtension = ".html";
+    }
+
```
albert-github, line 1145: See review below.

src/searchindex.cpp
```diff
@@ -946,6 +946,13 @@ void writeJavaScriptSearchIndex()
   // write index files
   QCString searchDirName = Config_getString(HTML_OUTPUT)+"/search";
 
+  QCString htmlFileExtension=Config_getString(HTML_FILE_EXTENSION);
+  htmlFileExtension=htmlFileExtension.stripWhiteSpace();
+  if (htmlFileExtension.isEmpty())
+  {
+    htmlFileExtension = ".html";
+  }
+
```
albert-github, line 955: See review above.

src/htmlgen.cpp
```diff
@@ -1136,10 +1136,17 @@ void HtmlGenerator::startFile(const char *name,const char *,
...
     t << "<script type=\"text/javascript\">\n";
 		t << "/* @license magnet:?xt=urn:btih:cf05388f2679ee054f2beb29a391d25f4e673ac3&amp;dn=gpl-2.0.txt GPL-v2 */\n";
-		t << "var searchBox = new SearchBox(\"searchBox\", \""
-      << m_relPath<< "search\",false,'" << theTranslator->trSearch() << "');\n";
+		t << "var searchBox = new SearchBoxExt(\"searchBox\", \""
+      << m_relPath<< "search\",false,'" << theTranslator->trSearch() << "','" << htmlFileExtension << "');\n";
```
albert-github, line 1149: - Why create a new function and not just adjust the current function.
- Why only correct 2 out of 3 `SearchBox` (missing `void HtmlGenerator::writeExternalSearchPage()` ?

src/htmlgen.cpp
luca-aep: I found this correction in other parts of the code; not knowing the project I simply replicated it.
Doxygen::htmlFileExtension does the job and we get much cleaner code

src/htmlgen.cpp
luca-aep: Given my lack of knowledge of the project and the impossibility of testing external research, I wanted to be conservative.

Now I also applied the fix to void HtmlGenerator::writeExternalSearchPage() and restored the single function in search.js

PR #8036, opened 2020-09-16
Allow building doxygen as a subproject

### LINKED ISSUE #8035: Cannot build doxygen as a subproject
**Describe the bug**

For building doxygen, I use the cmake build scripts.
For configuring the compiler options, I add cmake as a subproject.

Currently, doxygen does not allow this because it expects to be built as the main project.
The cmake configure step fails.

Also, recent versions of xapian have a `xapian-config.cmake` script for including it.
Doxygen currently ships with a `FindXapian.cmake` script.

This problem was encountered while creating a conan package at https://github.com/conan-io/conan-center-index/pull/2912.
(conan is a C/C++ package manager)

Thanks!

**Expected behavior**
cmake configure succeeds and we can build.

**To Reproduce**

```sh
cd /tmp
mkdir superproject
cd superproject
cat >CMakeLists.txt <<EOF
cmake_minimum_required(VERSION 3.1)
project(superproject)
add_subdirectory(doxygen)
EOF
git clone https://github.com/doxygen/doxygen.git
mkdir build
cd build
cmake ..
```

**Version**
current master 0fc06d657d596adcc289a5f228973ea268efd66d

**Additional context**
I will open a pr with fixes.

madebr:
This patch was created by running:
`find . -type f -exec sed -i s/CMAKE_SOURCE_DIR/PROJECT_SOURCE_DIR/g {} \;`
and
`find . -type f -exec sed -i s/CMAKE_BINARY_DIR/PROJECT_BINARY_DIR/g {} \;`
in the root of the doxygen source tree.
(I undid a bad change change to `doc/changelog.doc`)

I also renamed `FindXapian.cmake` to `Findxapian.cmake`.

So the casing of the cmake package is the same as the one from upstream xapian-core.

Fixes #8035

This patch is used at https://github.com/conan-io/conan-center-index/pull/2912


addon/doxysearch/CMakeLists.txt
```diff
@@ -1,12 +1,12 @@
-find_package(Xapian REQUIRED)
+find_package(xapian REQUIRED)
```
albert-github, line 1: You changed `Xapian` to `xapian` for the package with the reason:
> So the casing of the cmake package is the same as the one from upstream xapian-core.

but in principle a package name doesn't have to have the same spelling as the directories used.
Internally in `FindXapian.cmake` we see that for directories etc `xapian` is used.

Furthermore on the Xapian site (https://xapian.org/) we see that in sentences also `Xapian` is used, like:
> Welcome to the Xapian project website.

Secondly this change has nothing to do with the "Allow building doxygen as a subproject" so mixing is is not really a good idea.

madebr: I'll remove the xapian related change and open a new issue about the xapian thing.

In the mean time, see https://oligarchy.co.uk/xapian/1.4.17/xapian-core-1.4.17.tar.xz.
It contains `cmake/xapian-config.cmake.in`.

albert-github: The mentioned file has indeed lowercase `xapian` in its name but inside I see e.g.
```
SET(XAPIAN_LIBRARIES "@libdir@/libxapian@LIBRARY_VERSION_SUFFIX@@SHLIBEXT@" CACHE FILEPATH "Libraries for Xapian")
SET(XAPIAN_INCLUDE_DIR "@incdir@" CACHE PATH "Include path for Xapian")
```

madebr: I've opened https://github.com/doxygen/doxygen/issues/8038

It's the casing of the file that matters.

PR #8040, opened 2020-09-16
Added Verbatim code listings usings LaTeX packages: listings & tcolorbox

deem0n:
Provided default settings with:

- line numbers
- border around listings
- nice UTF-8 compatible default font (important for Cyrillic and other languages)
- automatic line breaks, possibly in right places, keep readability!

Todo:

- keyword colouring or syntax highlighting (which is easy if we can provide source code language: C, javascript etc)
- copy & paste from PDF should not copy line breaks made by LaTeX.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -2411,7 +2411,7 @@ The \c DOCSET_PUBLISHER_NAME tag identifies the documentation publisher.
     <option type='string' id='FORMULA_MACROFILE' format='file' defval=''>
       <docs>
 <![CDATA[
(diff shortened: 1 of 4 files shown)


deem0n:
I'm not sure what was the symbols at src/config.xml line 2414 and templates/latex/doxygen.sty line 375. Hopefully I did not break some special formatting...

deem0n:
2 examples provided in the archive: 
[example.tar.gz](https://github.com/doxygen/doxygen/files/5234989/example.tar.gz)

albert-github:
- the "special characters" looks like to be just a space and I think they can safely be removed here.
- this request is a rewrite / follow up on #8005

PR #8056, opened 2020-09-24
Miscounting of lines in respect to page command

albert-github:
All examples together are a bit big so they are all in the attached file together with the warnings before and after this patch.

When looking at the output we see that a number of lines are off by 1 ore more due to the insertion of extra lines in the code or due to the fact that there is a reference to the start of the documentation of the page and not the line of the page command.

- commentscan.l
  in case of a rule contaiinge {DOCNL} this can be `=n` or `\ilinebr` in the later case this should not be replaced by `\n` as this results in an increase of the line count later on.
  in case of a page like command also register the line of the command as "topline".
- entry.h, entry.cpp
  storage space for the "topline" registering the line of the page like commands.
- doxygen.cpp
  setting and using the "topline"
- markdown.cpp, markdown.h
  don't add a `\n` as this increases the line count but use the `\ilinebr`
  to get correct warnings see to it that when having empty lines at the top of the page and a page is added that the empty lines appear after the page command.
- index.cpp
  using the "topline" instead of the "docLine" to get the correct warning
- pagedef.cpp, pagedef.h
  set and retrieve the "topline" for page like commands.
- util.cpp, util.h
   setting and using the "topline"
   in `addRelatedPage` use the known file name and  line for the warning regarding the section label

- cite.cpp, context.cpp, reflist.cpp
  changed to have good function calls.

Example and warnings: [example.tar.gz](https://github.com/doxygen/doxygen/files/5276354/example.tar.gz)


albert-github:
Investigating problems with tests 43 and 45

src/entry.h
```diff
@@ -264,6 +264,7 @@ class Entry
     QCString     includeName; //!< include name (3 arg of \\class)
     QCString     doc;         //!< documentation block (partly parsed)
     int          docLine;     //!< line number at which the documentation was found
+    int          topLine;     //!< line number at which the page / section definition was found
```
doxygen, line 267: In the example you provided `topLine` is always equal to `startLine` (and different from `docLine`) it seems. So is there a case where `topLine` is really needed or can we just use `startLine` instead?

albert-github: For commentscan.l this is true, but in e.g. fortranscanner.l this is not the case (no topline used).
So I was not certain that I could use startLine and decided to use topLine even though this would require some more memory.

doxygen:
@albert-github decided to go without `topLine`, see https://github.com/doxygen/doxygen/commit/c46e7ae8a25ec64b91ab39bf5246aa69e729d25b and let me know if you see any issues.

PR #8059, opened 2020-09-25
added support for mingw PRETTY_FUNC

ArmandZampieri:
Hello
I add an error building using mingw64 as __FUNCSIG__ was not defined and we enter the case to define it. gnu is using __PRETTY_FUNC__ and adding the Preprocessing case solved my problem (not sure it is the most elegant solution though) 
Don't hesitate to make any remark critic 
Thank you

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -59,9 +59,7 @@
 #ifdef ENABLE_TRACING
 #define IOSTREAM      stdout
 #define DATA_BUFSIZE  20
-#ifdef __MINGW32__
-#define PRETTY_FUNC __PRETTY_FUNCTION__
-#elif defined(_WIN32) && !defined(CYGWIN)
+#ifdef defined(_WIN32) && !defined(CYGWIN) && !defined(__MINGW32__)
 #define PRETTY_FUNC __FUNCSIG__
 #else
 #define PRETTY_FUNC __PRETTY_FUNCTION__
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -59,7 +59,7 @@
 #ifdef ENABLE_TRACING
 #define IOSTREAM      stdout
 #define DATA_BUFSIZE  20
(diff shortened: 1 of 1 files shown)


src/markdown.cpp
```diff
@@ -59,7 +59,9 @@
 #ifdef ENABLE_TRACING
 #define IOSTREAM      stdout
 #define DATA_BUFSIZE  20
-#if defined(_WIN32) && !defined(CYGWIN)
+#ifdef __MINGW32__
+#define PRETTY_FUNC __PRETTY_FUNCTION__
+#elif defined(_WIN32) && !defined(CYGWIN)
```
albert-github, line 64: Why not use:
```
#if defined(_WIN32) && !defined(CYGWIN) && !defined (__MINGW32__)
```
instead of the original:
```
#if defined(_WIN32) && !defined(CYGWIN)
```

ArmandZampieri: Hi, no particular reason in fact you're right this version is more compact, changed  in new commit

doxygen:
@albert seems the debug build for windows started failing on Appveyor after your changes. Can you look into this? The new expression is not completely equivalent (`_WIN32` check was not done in the original proposal; not sure if/why that would matter).

src/markdown.cpp
albert-github: Looks like when correcting the `#if` construct by accident you pasted `#ifdef` instead `#if`, now the fix reads:
```
#ifdef defined(_WIN32) && !defined(CYGWIN) && !defined(__MINGW32__)
```
instead of the intended:
```
#if defined(_WIN32) && !defined(CYGWIN) && !defined(__MINGW32__)
```

ArmandZampieri:
Hello, really sorry about that lack of attention mistake, corrected in last commit

albert-github:
I should have seen the problem with appveyor before as well, but due to the time appveyor takes (and some other problems regarding connecting to external source sto load packages) I don't always see it either.

I compiled, now, locally as well and fix looks OK to me.

Issue #7686, opened 2020-04-03
Can the generated UML function with arguments ?


625781186:
![image](https://user-images.githubusercontent.com/30521429/78316150-862c9d80-7591-11ea-93cb-f9f745cbf303.png)

albert-github:
I think the problem here will be that the diagrams will become very fast unreadable, due to the length of the argument list.
I made a very very quick test with a png and svg image (I cannot add the svg image here so it will be in an attachment):

![class_q_string__inherit__graph](https://user-images.githubusercontent.com/5223533/78355877-d1a17480-75ae-11ea-80b4-f237e4c97b10.png)

and the original image:

![class_q_string__inherit__graph](https://user-images.githubusercontent.com/5223533/78355901-d9f9af80-75ae-11ea-9ebd-bf86770d923b.png)

We see in the images that some "strange" ways the lines are folded (especially in the qdict images).
With the svg images we see even that there is an overflow of the boxes (also a bit depending on the zoom level). As it was a very quick test there might be some possibilities to improve the box sizes etc.

The change was just changing (in dotnode line 151 in function writeBoxMemberList) from:
```
          t << prot << " ";
          t << DotNode::convertLabel(mma->name());
          if (!mma->isObjCMethod() &&
            (mma->isFunction() || mma->isSlot() || mma->isSignal())) t << "()"; 
          t << "\\l";
```
into
```
          t << prot << " ";
          t << DotNode::convertLabel(mma->typeString());
          t << " ";
          t << DotNode::convertLabel(mma->name());
          if (!mma->isObjCMethod() &&
            (mma->isFunction() || mma->isSlot() || mma->isSignal())) t << DotNode::convertLabel(mma->argsString());
          t << "\\l";
```

When implementing the arguments some attention should be paid to:
- it should probably behind an `UML_WITH_ARGS` configuration setting.
- line breaking
- box overflow

@doxygen what do you think about it.

All images: [images.tar.gz](https://github.com/doxygen/doxygen/files/4426936/images.tar.gz)

625781186:
@albert-github
Hi , 
I can live with some typographical flaws, 
so how do I generate function parameters?

albert-github:
> I can live with some typographical flaws,

In the doxygen this would not be good.

> so how do I generate function parameters?

What I did was just some quick and dirty hack in the code as indicated. But I'm waiting for some more input about the pitfalls I might have overlooked.

625781186:
[...]

PR #8065, opened 2020-09-29
Can generated UML function with arguments

alexandreSalconiDenis:
Hi,
I've added two new option into the doxyfile to let user manage fold length on dot graph and uml details like in #7686.


camillebachmann:
Hello,
That is really something I was looking for, thanks!

I read some comments emitting doubts about such kind of feature. Yes, of course, the diagrams are heavier. But to me this adds a lot of value to them. Having the types, we can much better see the interactions between elements and we don't always need to read the text below to understand how things work. The diagrams give thus an efficient overview. That's what they are made for.

I just started with Doxygen few days ago, but I'm working with models in general for much longer. Diagrams can be a lot more than nice pictures in a doc. It's all about information they contain.

alexandreSalconiDenis:
All the feature added are done with doxygen option so for the people who doesnt want this uml the just dont need to use the new option, by default it stay like before. I think it can be a win/win solution to get that feature for the people who use it and dont change what was already there.

PR #8069, opened 2020-10-02
Flush warnings file

albert-github:
When having a large project the warnings file is buffered but when a long time there are no warnings emitted the last warnings are not yet written.
During testing this is most obvious when the `dot` process starts and one needs the `.dot` files but not the converted files at hat moment one wants to kill the doxygen process but the result is an incomplete warnings file of the warnings that should already be written.

Created a flush function for the warnings file and placed it at a few strategic points.

### DIFF
--- a/src/message.cpp
+++ b/src/message.cpp
@@ -279,6 +279,12 @@ void term(const char *fmt, ...)
   exit(1);
 }
 
+void warn_flush(void)
+{
+  fflush(warnFile);
+}
+
+
 void printlex(int dbg, bool enter, const char *lexName, const char *fileName)
 {
   const char *enter_txt = "entering";
--- a/src/message.cpp
+++ b/src/message.cpp
@@ -279,7 +279,7 @@ void term(const char *fmt, ...)
   exit(1);
 }
 
-void warn_flush(void)
+void warn_flush()
 {
   fflush(warnFile);
 }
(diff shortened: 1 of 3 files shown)


src/message.cpp
```diff
@@ -279,6 +279,12 @@ void term(const char *fmt, ...)
   exit(1);
 }
 
+void warn_flush(void)
```
doxygen, line 282: void here is also useless (it is the default in C++ unlike in C)

src/message.h
```diff
@@ -35,6 +35,7 @@ extern void err(const char *fmt, ...) PRINTFLIKE(1, 2);
 extern void err_full(const char *file,int line,const char *fmt, ...) PRINTFLIKE(3, 4);
 extern void term(const char *fmt, ...) PRINTFLIKE(1, 2);
 void initWarningFormat();
+void warn_flush(void);
```
doxygen, line 38: same here `(void)` is equal to `()`

src/message.cpp
albert-github: I'm more from the old sort, I like the clarity.
I can change it without problems.

src/message.h
albert-github: I'm more from the old sort, I like the clarity.
I'll change it for consistency reasons.

doxygen: It is also the recommended syntax in the C++ FAQ: http://www.dietmar-kuehl.de/mirror/c++-faq/newbie.html#faq-29.4

PR #8073, opened 2020-10-03
Give warnings even when EXTRACT_ALL is set

albert-github:
When `EXTRACT_ALL` is set a number of warnings is suppressed, sometimes  there is a request to see them as well.
To be able to see warnings in case `EXTRACT_ALL` is set the setting `WARN_EXTRACT_ALL` is introduced.

(See e.g. https://stackoverflow.com/questions/64156269/force-doxygen-to-pop-warnings-if-uncommented-c-methods)

### DIFF
--- a/src/message.cpp
+++ b/src/message.cpp
@@ -222,9 +222,12 @@ void warn_simple(const char *file,int line,const char *text)
 
 void warn_undoc(const char *file,int line,const char *fmt, ...)
 {
+  static bool warnIfUndocumented = Config_getBool(WARN_IF_UNDOCUMENTED);
+  static bool extractAll = Config_getBool(EXTRACT_ALL);
+  static bool warnExtractAll = Config_getBool(WARN_EXTRACT_ALL);
   va_list args;
   va_start(args, fmt);
-  do_warn(Config_getBool(WARN_IF_UNDOCUMENTED), file, line, warning_str, fmt, args);
+  do_warn(warnIfUndocumented || (extractAll && warnExtractAll), file, line, warning_str, fmt, args);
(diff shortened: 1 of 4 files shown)


src/message.cpp
```diff
@@ -222,9 +222,12 @@ void warn_simple(const char *file,int line,const char *text)
 
 void warn_undoc(const char *file,int line,const char *fmt, ...)
 {
+  static bool warnIfUndocumented = Config_getBool(WARN_IF_UNDOCUMENTED);
+  static bool extractAll = Config_getBool(EXTRACT_ALL);
+  static bool warnExtractAll = Config_getBool(WARN_EXTRACT_ALL);
   va_list args;
   va_start(args, fmt);
-  do_warn(Config_getBool(WARN_IF_UNDOCUMENTED), file, line, warning_str, fmt, args);
+  do_warn(warnIfUndocumented || (extractAll && warnExtractAll), file, line, warning_str, fmt, args);
```
doxygen, line 230: This logic seems wrong to me. If extractAll and warnExtractAll are both enabled, then the warnIfUndocumented setting is ignored.

albert-github: I think the logic is correct and is intentionally. When `EXTRACT_ALL` and `WARN_EXTRACT_ALL` are set we want to have all warnings, irrespective of the setting of `WARN_IF_UNDOCUMENTED`.

kammce:
Any chance of this change getting merged?

Snawe:
I would also vote for this to be merged. We are building of this branch now for quiet a long time now...

kammce:
Friendly ping

Issue #8137, opened 2020-10-28
Whitespace/Separator required to recognise custom command argument


asihera:
Hello,

Very seemingly simple problem.

In Doxyfile:

`  ALIASES += cmd{2}="\1\2"`

Then in source I do:

`  @cmd{x,y}`

and I get

`  x\2`

However,

`  ALIASES += cmd{2}="\1 \2"`

(note whitespace) produces:

`  x y`

with a whitespace in between.

Why is it that expansions of custom command arguments have to be something-separated? I want to concatenate the arguments of a custom command and I can't.

Incidentally, putting a command argument after a nested expansion doesn't work either. E.g.:

`  ALIASES += cmd{1}="@mycommand{x,y}\1"`

Will not expand \1. It just copies "\1" verbatim to the output.

Haven't tried the latest version but I have tested versions between 1.8.9 and 1.8.18 and they all behave the same.

Kind Regards,

albert-github:
Looks indeed like the `ALIASES` substitution doesn't find the second `\` properly to replace this as well.
(I tested it with the current master (1.9.0 (c5ce7512ad6482a3e928d99ddbb0cc68044b58dc))

Would have been nice when we had straight away a, small, self contained example (source+configuration file in a tar or zip) that allowed us to reproduce the problem (don't add it as external links as they might not be persistent. ) 

So here is the example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5454024/example.tar.gz)

asihera:
I had an example.

[Command-Args-Substition.tar.gz](https://github.com/doxygen/doxygen/files/5454065/Command-Args-Substition.tar.gz)

Not much different from yours but demonstrates the problem, I think.

albert-github:
I've just pushed a proposed patch, pull request #8138

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

asihera:
Thanks for taking a look at this.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #8141, opened 2020-10-29
Handling commands for backslash and at-sign in ALIASES

albert-github:
When we have an command like `\\`, `\@`, `@@` or `@\`in ALIASES this should not disturb the recognition of the  replacement strings like `\2`, so e.g. `\\2` should not be replaced by `\y` but remain `\\2` (assuming  the `\2` is linked to `y` in an ALIASES).

This issue was found after #8137 had been fixed and was already present in older versions as well.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5458311/example.tar.gz)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -5971,18 +5971,9 @@ static QCString replaceAliasArguments(StringUnorderedSet &aliasesProcessed,
   int markerEnd=0;
   for (i=0;i<l;i++)
   {
-    if (markerStart==0 && aliasValue.at(i)=='\\') // start of a \xx marker
+    if (markerStart>0 && !(aliasValue.at(i)>='0' && aliasValue.at(i)<='9'))
     {
-      markerStart=i+1;
-    }
-    else if (markerStart>0 && aliasValue.at(i)>='0' && aliasValue.at(i)<='9')
-    {
-      // read digit that make up the marker number
-      markerEnd=i+1;
-    }
-    else
-    {
-      if (markerStart>0 && markerEnd>markerStart) // end of marker
+      if (markerEnd>markerStart) // end of marker
       {
         int markerLen = markerEnd-markerStart;
         markerList.append(new Marker(markerStart-1, // include backslash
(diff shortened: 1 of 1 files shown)


src/util.cpp
```diff
@@ -5971,28 +5971,37 @@ static QCString replaceAliasArguments(StringUnorderedSet &aliasesProcessed,
   int markerEnd=0;
   for (i=0;i<l;i++)
   {
```
doxygen, line 5973: The logic in the loop has become too complex in my opinion. I'm pretty sure it can be simplified. Can you list the cases you now encountered and cover?

albert-github: I think the logic has decreased, it would be possible to remove the `continue` and the second `if` can in that case be replaced with `else if`.

The major test cases I encountered are shown in the example as attached to the pull request here on github.

doxygen:
@albert-github I think I managed to further simplify the logic (see https://github.com/doxygen/doxygen/commit/539b3bbef61dc7dec8cd5987dc2dbb7a2a956a83). Results on the example are the same, but please check if I missed any corner case.

albert-github:
I couldn't find a corner case (for the moment ;-) ), though I found another problem with `ALIASES` but this is independent from the problem solved here.
When I add:
```
ALIASES +=bs{1}="A\1\\"
ALIASES +=bs1{2}="A\1\\\2Z3"
ALIASES +=bs2{1}="A\1\\\\"
ALIASES +=bs3{2}="A\1\\\2Z3"
```
to the Doxyfile, I get:
```
warning: Missing end quote (") on line 15, file Doxyfile
warning: Missing end quote (") on line 17, file Doxyfile
error: Illegal ALIASES format '+=bs1{2}=A\1\\\2Z3'. Use "name=value" or "name{n}=value", where n is the number of arguments
error: Illegal ALIASES format '+=bs3{2}=A\1\\\2Z3'. Use "name=value" or "name{n}=value", where n is the number of arguments
```
when loading the Doxyfile. This problem is already present for a long time (at least since version 1.8.1), even 46 or 8 backslashes doesn't help. Also replacing the `\\` by `@\` doesn't help, like:
```
ALIASES +=as{1}="A\1@\"
ALIASES +=as1{2}="A\1\\\2Z3"
```

darsovit:
> I think this has broken a use case we had of building up an alias from argument and then calling that alias. Is there some mechanism to force to call an alias without putting the @ sign in front of it?
> 
> Example:
> ALIASES += SOMETHING_DESCRIPTION="Description 1"
> ALIASES += SOMETHINGELSE_DESCRIPTION="Description 2"
> ALIASES += table_entry{1}="<tr><td>\1</td><td>@\1_DESCRIPTION</td></tr>"
> 
> ...
> 
> Then usage like:
> <table>
> @table_entry{SOMETHING}
> @table_entry{SOMETHINGELSE}
> </table>
> 
> Expecting output:
> 
> SOMETHING	Description 1
> SOMETHINGELSE	Description 2

I found that I could get the same behavior by moving the constant part to the front instead of the end.

So, change to:
ALIASES += DESCRIPTION_SOMETHING="Description 1"
ALIASES += DESCRIPTION_SOMETHINGELSE="Description 2"
ALIASES += table_entry{1}="&lt;tr&gt;&lt;td&gt;\1&lt;/td&gt;&lt;@DESCRIPTION_\1&gt;&lt;/tr&gt;

albert-github:
@darsovit 
Interesting case.

Combination of command qualifiers (`\` and `@`) and replacement string (like `\1`) are very challenging.

Do I understand you correctly your second comment (https://github.com/doxygen/doxygen/pull/8141#issuecomment-764057413) gives a version for you that works for the problem you mention in your first comment (https://github.com/doxygen/doxygen/pull/8141#commitcomment-46174026)?

darsovit:
> @darsovit
> Interesting case.
> 
> Combination of command qualifiers (`\` and `@`) and replacement string (like `\1`) are very challenging.
> 
> Do I understand you correctly your second comment ([#8141 (comment)](https://github.com/doxygen/doxygen/pull/8141#issuecomment-764057413)) gives a version for you that works for the problem you mention in your first comment ([#8141 (comment)](https://github.com/doxygen/doxygen/pull/8141#commitcomment-46174026))?

Yes, it allows it to work if the new alias name built up doesn't start with the parameter (so effectively avoid having the @ and \1 together like originally we were doing in our usage (with 1.8.13). We wanted to upgrade for another issue and ran into this problem.

Issue #7556, opened 2020-01-31
ANSI-C anonymous (unnamed) struct/unions have duplicated names (__unnamed__)


DGA45:
In case of struct using nested anonymous (unnamed) inner struct/unions, the generated inner types will have the same name. In such case, it is not possible to reference them uniquely.
Example:
```
struct tm {
    int tm_sec;   /* seconds after the minute, 0 to 60
                     (0 - 60 allows for the occasional leap second) */
    int tm_min;   /* minutes after the hour, 0 to 59 */
    int tm_hour;  /* hours since midnight, 0 to 23 */
    int tm_mday;  /* day of the month, 1 to 31 */
    int tm_mon;   /* months since January, 0 to 11 */
    int tm_year;  /* years since 1900 */
    int tm_wday;  /* days since Sunday, 0 to 6 */
    int tm_yday;  /* days since January 1, 0 to 365 */
    int tm_isdst; /* Daylight Savings Time flag */
    union {       /* ABI-required extra fields, in a variety of types */
        struct {
            int __extra_1, __extra_2;
        };
        struct {
            long __extra_1_long, __extra_2_long;
        };
        struct {
            char *__extra_1_cptr, *__extra_2_cptr;
        };
        struct {
            void *__extra_1_vptr, *__extra_2_vptr;
        };
    };
};
```
The generated classes in PerlMod or XML for the 4 last struct will have the same name (here tm.__unnamed__.__unnamed__)

The proposal is to add the internal reference number (example @19) to the anonymous field name: __unnamed__19, instead of __unnamed__ alone.

DGA45:
Here is an example
[Doxy issue 7547 and 7556 examples.zip](https://github.com/doxygen/doxygen/files/4147433/Doxy.issue.7547.and.7556.examples.zip)

PR #8145, opened 2020-10-30
Fix #7547 and #7556 issues related to PerlMod and ANSI-C struct/union (struct/union were not discriminable + anonymous fields are duplicated)

### LINKED ISSUE #7547: Perlmod does not generate "kind" information to discriminate ANSI-C struct/union
When using PerlMod with ANSI-C code, the struct/union are generated as "classes".
However, they are missing the "kind" attribute which can be used to discriminate if this is a "struct" or a "union" (as it is correctly done in the XML output).

DGA45:
Fix issues:
- #7547 "kind" information added to PerlMod to discriminate ANSI-C struct/union
- #7556 ANSI-C anonymous (unnamed) struct/unions duplicated names

### DIFF
--- a/src/perlmodgen.cpp
+++ b/src/perlmodgen.cpp
@@ -1835,6 +1835,8 @@ void PerlModGenerator::generatePerlModForClass(const ClassDef *cd)
 
   m_output.openHash()
     .addFieldQuotedString("name", cd->name());
+  /* DGA: fix # #7547 Perlmod does not generate "kind" information to discriminate struct/union */
+  m_output.addFieldQuotedString("kind", cd->compoundTypeString());
 
   if (cd->baseClasses())
   {
--- a/doc/changelog.doc
+++ b/doc/changelog.doc
@@ -2,6 +2,35 @@
 \tableofcontents{html,latex}
 \section log_1_8 1.8 Series
 
+\subsection log_1_8_20 Release 1.8.20
+\htmlonly
+<b>(release date 24-08-2020)</b>
+</p>
+<h3>Bug fixes</h3>
+<ul>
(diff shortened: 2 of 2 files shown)


src/doxygen.cpp
```diff
@@ -1389,7 +1389,8 @@ static void processTagLessClasses(ClassDef *rootCd,
             if (type.find(icd->name())!=-1) // matching tag less struct/union
             {
               QCString name = md->name();
-              if (md->isAnonymous()) name = "__unnamed__";
+			  /* DGA fix #7556 ANSI-C anonymous (unnamed) struct/unions have duplicated names (__unnamed__) */
+              if (md->isAnonymous()) name = "__unnamed__" + name.right(name.length()-1); /* DGA: ensure unique name */
```
doxygen, line 1384: why the `name.right(name.length()-1)`, i.e. removing the first character?

src/perlmodgen.cpp
```diff
@@ -1835,6 +1835,8 @@ void PerlModGenerator::generatePerlModForClass(const ClassDef *cd)
 
   m_output.openHash()
     .addFieldQuotedString("name", cd->name());
+  /* DGA: fix # #7547 Perlmod does not generate "kind" information to discriminate struct/union */
```
doxygen, line 1838: please do not add comments like this in the code

src/doxygen.cpp
DGA45: Because the internal name for anonymous members is first set to something like "@19".
Then it is changed to "__unnamed__" all the time, leading to duplicated members name.
My proposal is to add the internal reference number, but not the "@" character which is not allowed in the identifiers naming.
Hence the removal of the first character to finally have a member name = "__unnnamed__19"

src/perlmodgen.cpp
DGA45: Ok, you mean to not add my shortname ("DGA") ?
But it is a good practice to still mention the fixed issue ?

Issue #8158, opened 2020-11-04
navtree js output varies from ASLR


bmwiedemann:
**Describe the bug**
doxygen produces nondeterministic navtreedata/index.js output from certain input (e.g. the `eigen3` 3.3.8 docs package in openSUSE), unless ASLR is disabled.

**Expected behavior**
output shoud be deterministic

**To Reproduce**
I made a minimal reproducer with just a few lines:
[eigen3-min.tar.gz](https://github.com/doxygen/doxygen/files/5485846/eigen3-min.tar.gz)

tar xf eigen3-min.tar.gz
cd doxygennondeterminism
for i in $(seq 1 10) ; do ./build.sh ; done|sort|uniq -c
      6 95846f882fbebd706b1cf5ee43e44cdf  -
      4 ded422b2f1ba4a1b1cbbba0a929c2554  -

**Version**
doxygen-1.8.20-3.1 of openSUSE-Tumbleweed-20201029

**Additional context**
Similar in nature to #7474

This PR was done while working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

albert-github:
I don't have an openSUSE system so I had to try it on windows.

I had to use a little bit an adjusted script (as I also have Cygwin I can use a lot of *nix commands as well):
```
rm -rf html
doxygen
md5sum html/navtree*.js|md5sum
md5sum html/navtree*.js
```
When running doxygen version 1.8.20 I get 2 different outputs
```
904d8112325aef5e415d70b8a5a49330 *-
2368aabf7d98fcf12ea9c2c346e96a24 *html/navtree.js
79ff46ed63b7500b6d56fd1f7e1df540 *html/navtreedata.js
e0180e4b4e0cf752a2f7647cc305454a *html/navtreeindex0.js
```
and
```
0469831f90d6ad2b114cda9628381d1c *-
2368aabf7d98fcf12ea9c2c346e96a24 *html/navtree.js
716c3828f321638e00d763ddcde804f1 *html/navtreedata.js
a2076aa20c74ac749015f5cc5301ee6f *html/navtreeindex0.js
```

Fortunately I saved the html directories ([output_1_8_20.tar.gz](https://github.com/doxygen/doxygen/files/5487075/output_1_8_20.tar.gz)) and could run a diff on them:
```
diff -r html_1/navtreedata.js html_2/navtreedata.js
40c40
< ""
---
[...]

PR #8164, opened 2020-11-06
Warnings in generated VhdlParser.cc

albert-github:
When compiling the generated VhdlParser.cc we get a number of warnings about the return type, these are no problems as they all follow an assert(false) but when "promoting" the missing to return type warning to an error (openSUSE) this is a problem.

The best way would be that the problem is solved in javaCC though this will be probably a long process, so we use the workaround to overcome the problem.
(The problem was detected as part of the reproducibility tests for openSUSE, see #8158)


doxygen:
@albert-github I think we could let JavaCC produce the `VhdlParser.cc` in the `generated_src` dir and only put the adjusted version in git, or do you see a problem with that?

albert-github:
I don't see a real problem here, but I wanted to leave the version generated by javaCC and the adjusted version both in the directory. In case the `VhdlParser.cc` is removed I think it has also to be added to the (local) `.gitingore`. I don't think we can just remove it.

doxygen:
The problem I now indeed see is that JavaCC generates all files in one go. So with the current rules it is hard to put `VhdlParser.cc` in a different directory without moving it there after it is generated. Alternatively, we could generate all files in the `generated_src` dir and then copying them back.

albert-github:
I think the best way would be to generate them in the generated_src directory and copy the relevant files back. We have to take into account that when no javaCC is present we have to copy the files from the src directory to the generated_src directory.

PR #8194, opened 2020-11-22
Doxygen warnings are not seen as errors during running tests

albert-github:
Doxygen warnings were on *nix suppressed but shown on Windows though the tests were nevertheless shown as OK
- correct redirection under windows
- catch doxygen warnings and make the corresponding test fail


albert-github:
Note the fails on e.g. Travis shows exactly the purpose of this patch.

doxygen:
> Note the fails on e.g. Travis shows exactly the purpose of this patch.

Indeed. Probably best to add the fix to this commit, so it passes again before merging.

albert-github:
OK I'll add the fix.

PR #8206, opened 2020-11-26
Incorrect XHTML results

albert-github:
Since commit d03e8d9411ab3e983fc3413c147fba1a5e5c9dad  stating
```
Refactoring: modernize TooltipManager class and source reference lists

- Tooltips are now collected per file instead of globally
```

We get for tests 021 and 085 XHTML warnings like:
```
test_output_021/html/index.xhtml:73: element div: validity error : ID aclass_test_xhtml already defined
<div class="ttc" id="aclass_test_xhtml"><div class="ttname"><a href="class_test.
```
when running
```
doxygen tests TEST_FLAGS=--xhtml
```
The tooltips should not be collected per input file type (or block of input file type), but should be collected for each output file and this requiring a global instance to collect the tooptip data  and writing the tooltip data at the end of the file.

### DIFF
--- a/src/clangparser.cpp
+++ b/src/clangparser.cpp
@@ -133,7 +133,6 @@ class ClangTUParser::Private
     CXToken *tokens = 0;
     uint numTokens = 0;
     StringVector filesInSameTU;
-    TooltipManager tooltipManager;
 
     // state while parsing sources
     MemberDef  *currentMemberDef=0;
@@ -573,7 +572,7 @@ void ClangTUParser::writeMultiLineCodeLink(CodeOutputInterface &ol,
                   const char *text)
 {
   static bool sourceTooltips = Config_getBool(SOURCE_TOOLTIPS);
-  p->tooltipManager.addTooltip(d);
+  TooltipManager::instance()->addTooltip(d);
   QCString ref  = d->getReference();
   QCString file = d->getOutputFileBase();
   QCString anchor = d->anchor();
--- a/src/code.l
+++ b/src/code.l
@@ -175,7 +175,6 @@ struct codeYY_state
 
   VariableContext theVarContext;
   CallContext     theCallContext;
-  TooltipManager  tooltipManager;
   SymbolResolver  symbolResolver;
 };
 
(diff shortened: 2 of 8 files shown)


doxygen:
@albert-github I made the collection per file to allow processing using multiple threads (still to be enabled, see `MULTITHREADED_SOURCE_GENERATOR` in `doxygen.cpp`), where each thread processes one source file independently and in parallel to another. 

Having a global tooltip manager would 1) requiring locking, and 2) makes the output non-deterministic. So I think another solution is needed, for instance by having a file prefix to make the anchor unique again or remove the duplicate anchors as a post-processing step.

albert-github:
I had the idea that the problem  was coming from problems related to multi-threading, but the current output is not XHTML conform.
I can imagine that the problem with multi threading might give problems as well, hence I added the mutex in the add function, but this might give indeed non deterministic results at the moment. Possibly a sort on the unique_ptr p in the write phase should do the deterministic trick.

I don't think that it is possible, but I'm not sure, that multiple files are written in parallel. Is this assumtion correct?

doxygen:
> I had the idea that the problem was coming from problems related to multi-threading, but the current output is not XHTML conform.

I agree that this should be fixed.


> I can imagine that the problem with multi threading might give problems as well, hence I added the mutex in the add function, but this might give indeed non deterministic results at the moment. Possibly a sort on the unique_ptr p in the write phase should do the deterministic trick.
> 
> I don't think that it is possible, but I'm not sure, that multiple files are written in parallel. Is this assumtion correct?

If `MULTITHREADED_SOURCE_GENERATOR` is set to `1` the "Generating code for" phase will be done with multiple threads (one per output file). The other parts like "Generating docs for..." are still done sequentially. There are still some race conditions that is why it is still disabled at the moment.

I have to find a better way to bind the tooltips to the output. I'm looking at passing the output generator also when adding tooltips to see if that would be a solution.

albert-github:
I think as a possible solution it would be possible to generate a "global array" (Doxygen::Tooltips ?) with as key the name of the output file or the file pointer to this file. Going into the *code.l the key could be used to retrieve relevant tooltipmanager (and when the tooltipmanager does not yet exist for the key, create one), in this way the keys are collected and can be written all at the end, just before the output file is closed. Direct after writing the tooltips of the file the tooltipmanager can be deleted.
I think that a sort before writing on the unique_ptr won't be a bad idea either and most likely the locking in the "addTooltip" should also stay.

What do you think of this?

albert-github:
As far as I can tell the commit a292601d5d06cf5993247f2f7a6bc7306997da23 fixes this issue.

Issue #6204, opened 2018-07-02
Doxygen cannot be configured to show 0 members in the UML class diagram (Origin: bugzilla #782366)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.8.13* on platform *Other*
Assigned to: Dimitri van Heesch

On 2017-05-09 05:06:28 +0000, Fabian wrote:

    I would like to use UML like diagrams in my doxygen documentation. However,
    I want to use them to show the relationships of the classes and nothing
    else. Hence I would like to set the number of fields and member variables to
    be shown to zero.
    
    UML_LOOK = YES
    UML_LIMIT_NUM_FIELDS = 0
    
    This does not work because 0 means "show all".
    
    Even if I set UML_LIMIT_NUM_FIELDS = 1 it looks annoying because I have
    grouped my member functions in groups and it shows up to
    UML_LIMIT_NUM_FIELDS *per group* and not in total.
    
    My current workaround is not to use UML_LOOK, but I am looking for a better
    way. Maybe doxygen can be modified to accept UML_LIMIT_NUM_FIELDS = -1 for
    "show no members".

timoffermann:
Today I was looking for exact this type of configuration for a two reasons:

- We prefer UML diagrams over non-standard looking diagrams
- We prefer at this place in the documentation the focus on the relationships and don't need the two optional compartments for attributes and operations as part of our diagrams

This would help us - the context of a larger project - to define one standard doxygen configuration (using UML) without the need of finetuning the number of fields.

Something like this:
![Screenshot from 2020-11-26 22-50-55](https://user-images.githubusercontent.com/3040731/100393421-dfbe6c80-3039-11eb-800b-6f0e69a1821f.png)

albert-github:
Seen the linked pull request #8207 I think that the setting `DOT_UML_DETAILS=NONE` (available as of version 1.9.0 meets the requirements.

I think the problem has been solved in the mean time.

Issue #8023, opened 2020-09-09
Add option to exit with non-zero exit code on warnings even if WARN_AS_ERROR is OFF


georg-emg:
Doxygen offers an option, WARN_AS_ERROR, that allows enforcing a no-warnings-allowed policy when generating documentation. However, when this option is set, processing stops at the first warning encountered. This forces the programmer into an endless loop of generating the docs, fixing the first warning, generarating the docs again, fixing the second warning etc. It would be helpful if doxygen would continue processing after warnings even with WARN_AS_ERROR set, and then just return a non-zero exit code at the end. This would allow the programmer to see and fix all warnings in one go even when WARN_AS_ERROR is set. This is analogous to the behaviour of modern compilers, which report as many errors as possible in a single compilation run.

albert-github:
As far as I know the idea behind the `WARN_AS_ERROR` is that doxygen will stop directly so no further computer resources are lost. As soon as this happens the problem lies again with the programmer, who made the mistake, to correct the issue. Best is to fix this error (and when obvious similar warnings as well) and just run doxygen locally but without the `WARN_AS_ERROR` so all warnings are shown (probably placing them in `WARN_LOGFILE` and can be solved, before starting the complete build process again.

georg-emg:
I understand the use case you describe, but it still makes it difficult to use doxygen in a modern continuous integration environment, where we currently have the choice of only logging a single (out of possibly dozens) of warnings, or of letting all warnings slip through unnoticed. This is because there is no way for the CI pipeline to determine whether there were warnings at all if `WARN_AS_ERROR` is `NO`, other than capturing the entire output and using some sort of heuristic on it. There is a reason why all compilers have moved to a strategy where compilation continues after an error, and why make, cmake, ninja etc. all have options to keep going after errors.

To preserve the existing behaviour, I suggest adding a new option instead, something like `FAIL_ON_WARNINGS`, that would return a non-zero exit code if any warnings occurred even if `WARN_AS_ERROR` is `NO`. [This thread](https://sourceforge.net/p/doxygen/mailman/message/34708445/) seems to suggest that this was already implemented at some point, but then (understandibly) rejected because it broke existing pipelines. But an option to reenable it would make implementing robust CI pipelines much easier, IMHO.

I will update the title of this issue to reflect this.

albert-github:
I've just pushed a proposed patch, pull request #8197

albert-github:
[...]

PR #8207, opened 2020-11-26
Enable UML diagrams without member fields

timoffermann:
The compartment for class attributes and the compartment for the
operations are optional.

By providing a new option "DOT_UML_SHOW_MEMBER" an user can
configure to use the UML_LOOK but without the additional
information regarding the content of the class.

#6204

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3444,22 +3444,22 @@ to be found in the default search path.
 ]]>
       </docs>
     </option>
-    <option type='bool' id='DOT_UML_DETAILS' defval='0' depends='UML_LOOK'>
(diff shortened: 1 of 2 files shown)


albert-github:
One is never happy when a new tag appears in the list of available tags (we already have so many).

Recently I was confronted with a similar problem (but at a completely different topic) and didn't want to add a new tag (see #8023) , so I created a proposed pull request where I changed an existing tag form a boolean tag to an enum tag.

In this case I think this could work for the tag `DOT_UML_DETAILS ` by means of the enum values:
```
YES
NO
NONE
```

what do you think about this?

timoffermann:
I appreciate your proposal - it fits my needs, is a logical extension of the existing functionality without and will not interfere with existing configuration.

PR #8215, opened 2020-11-29
fix: adding meta description and lang attribute, Closes #8214

### LINKED ISSUE #8214: missing lang attribute and meta description for SEO
**Describe the bug**
Missing lang attribute and meta type description

**Expected behavior**
Expecting certain attributes due to SEO penalisation without them

**Screenshots**

Before:
image

After:
image

**To Reproduce**
n/a

**Version**
n/a

**Stack trace**
n/a

**Additional context**
The SEO tool used was google lighthouse (developer tools, chromium)

allanbowe:
Three points:

1.  The `<!DOCTYPE html` attribute should really come first in the document.  Perhaps the autogenerated doxygen version line comment could come at the end?  or second line?
2. Perhaps lang attribute could be in the doxyfile, along with some other SEO attributes
3. The meta description tag needs to be in the `<head>` - so isn't wrapped in the PROJECT_BRIEF markers.  I believe that without PROJECT_BRIEF the attribute will be empty, which is harmless (no worse than without it)

See issue #8214 for before / after SEO results

### DIFF
--- a/templates/html/header.html
+++ b/templates/html/header.html
@@ -1,10 +1,11 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "https://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<html xmlns="http://www.w3.org/1999/xhtml">
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/xhtml;charset=UTF-8"/>
 <meta http-equiv="X-UA-Compatible" content="IE=9"/>
 <meta name="generator" content="Doxygen $doxygenversion"/>
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
+<meta name="Description" content="$projectbrief">
 <!--BEGIN PROJECT_NAME--><title>$projectname: $title</title><!--END PROJECT_NAME-->
 <!--BEGIN !PROJECT_NAME--><title>$title</title><!--END !PROJECT_NAME-->
 <link href="$relpath^tabs.css" rel="stylesheet" type="text/css"/>


templates/html/header.html
```diff
@@ -1,10 +1,11 @@
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "https://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<html xmlns="http://www.w3.org/1999/xhtml">
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
```
albert-github, line 2: I don't think this is correct. When choosing in the setting another language (see `OUTPUT_LANGUAGE`)  shouldn't this change as well?

allanbowe: indeed - this ought to be updated to correspond to the language of the output.   I'm not sure my skills are up to this, beyond some hacky inline JS if/then function.

albert-github: I don't think a JS solution is here a good solution as it probably will kick in after the page is loaded / during the load.

For XML and DocBook in #7825 the `lang` attribute has been added and when placing a placeholder `lang=$lang` in the default HTML header is is possible to replace this `$lang` attribute during the generation of the HTML pages by the correct language code (analogous to e.g. `$projectbrief`)

the question though is:
- Are all the language codes (for the doxygen supported languages) for XML the same as for HTML?
   The language codes are defined in the function `trISOLang()` in the different `translator_...h` files.

when this is the case the above solution will work, in the other case a new translation function has to be defined (or an optional boolean argument has to be added).

templates/html/header.html
```diff
@@ -1,10 +1,11 @@
...
 <head>
 <meta http-equiv="Content-Type" content="text/xhtml;charset=UTF-8"/>
 <meta http-equiv="X-UA-Compatible" content="IE=9"/>
 <meta name="generator" content="Doxygen $doxygenversion"/>
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
+<meta name="Description" content="$projectbrief">
```
allanbowe, line 8: Ideally this will be the @brief of the page, rather than the (singular) projectbrief.  This would enable far more meaningful previews when doc links are pasted into sites such as linkedin, facebook, twitter and slack.

albert-github:
Code of #9547 has been integrated in master on GitHub.

PR #8220, opened 2020-12-02
\name gives warnings regarding multiple defined labels

albert-github:
When having multiple statement like `\name` (without description) in one file or a statement like `\name Types` in multiple file an automatic label is created, but when xmllint checking in html we get messages like:
```
 element a: validity error : ID amgrp01747264fe7bf50731df0522c351974e already defined
```
or when building teh pdf from LaTeX:
```
LaTeX Warning: Label `_amgrp01747264fe7bf50731df0522c351974e' multiply defined.
```

A `\name` has no "anchor" so no need for a label either.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5631235/example.tar.gz)

(Problem found through CGAL code)

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -1122,7 +1122,7 @@ RCSTAG    "$"{ID}":"[^\n$]+"$"
                                           yyextra->current->type = yyextra->current->type.stripWhiteSpace();
                                         }
 <GroupDocArg2>{DOCNL}                   {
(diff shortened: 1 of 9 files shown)


albert-github:
Looks like that when using a `xrefitem` (or possibly also `\todo` and friends) give a small problem with this patch.

albert-github:
Superseded by #8238

PR #8232, opened 2020-12-07
Incorrect link from summary links for namespaces

albert-github:
When running a link checker on the CGAL BGL package we get the warning:
```
file:///.../doc_output/BGL/namespaceCGAL.html
 Lines: 164, 172, 1143, 1153
  Code: 200 (no message)
 To do: Some of the links to this resource point to broken URI fragments
        (such as index.html#fragment).
The following fragments need to be fixed:
        namespaces                      Line: 164
```
This is due to the fact that on the mentioned page no "section" namespaces is but in the summary links (top right) there is still "Namespaces".
Before writing the summary link it should be checked whether or not there are really namespaces shown (se also `NamespaceSDict::writeDeclaration`).

### DIFF
--- a/src/namespacedef.cpp
+++ b/src/namespacedef.cpp
@@ -893,10 +893,19 @@ void NamespaceDefImpl::writeSummaryLinks(OutputList &ol) const
     }
     else if (lde->kind()==LayoutDocEntry::NamespaceNestedNamespaces && namespaceSDict && namespaceSDict->declVisible())
     {
-      LayoutDocEntrySection *ls = (LayoutDocEntrySection*)lde;
-      QCString label = "namespaces";
-      ol.writeSummaryLink(0,label,ls->title(lang),first);
-      first=FALSE;
+      SDict<NamespaceDef>::Iterator ni(*namespaceSDict);
+      NamespaceDef *nd;
+      for (ni.toFirst();(nd=ni.current());++ni)
+      {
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I've implemented this a bit differently; `namespaces.declVisible()` should already tell if the summary link can be written or not. Please check if this solves the original problem as I don't know how to reproduce it.

albert-github:
It is quite a tedious process to find these kind of problems, it is done with the w3C link checker. I'll give tomorrow a go.

albert-github:
I reran the creation of the documentation of the CGAL package (I also ran the version without the corrections).
1. In the BGL package the mentioned broken links are gone (I only checked the BGL package, but seen the changes with the other packages I think for them theses broken links are gone as well).
2. In a number of packages the file `namespacestd.html` surfaces in a number of cases this file doesn't contain any information just an "identification / header" (like `Cone_spanners_2`) in other cases (like (`Geomview`) it contains a line for each STL class with the mentioning "STL class"
3. In the statements like `using namespace std;` the `std` is now hyperlinked to `namespacestd.html` (like e.g. Bounding_volumes/Min_sphere_of_spheres_d_2benchmark_8cpp-example.html).

In the project the `BUILTIN_STL_SUPPORT    = YES` so that the STL links are show should be OK.

PR #8233, opened 2020-12-07
Problem with reproducibility of MSC graph in HTML

albert-github:
The main problem was that the example / test didn't contain correct "code" i.e. it used `nref` instead of `\ref` (or `@ref`) (and also the definition of `Ack` was not correct).
- example made correct
- the reading and interpretation of the map file made more robust.

The problem showed with test 37.

### DIFF
--- a/src/msc.cpp
+++ b/src/msc.cpp
@@ -54,14 +54,20 @@ static bool convertMapFile(FTextStream &t,const char *mapName,const QCString rel
     //printf("ReadLine '%s'\n",buf);
     if (qstrncmp(buf,"rect",4)==0)
     {
-      // obtain the url and the coordinates in the order used by graphviz-1.5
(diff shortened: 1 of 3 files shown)


albert-github:
This fix (in the files of test 037) also fixes the problem regarding link checking (with and without sub directories):
```
Processing      file:///.../testing/test_output_037/html/dc/dae/037__msc_8cpp.xhtml


file:///.../testing/test_output_037/html/dc/dae/nref
 Lines: 81, 82, 83, 84, 85, 86
  Code: 404 File `.../testing/test_output_037/html/dc/dae/nref' does not exist
 To do: The link is broken. Double-check that you have not made any typo,
        or mistake in copy-pasting. If the link points to a resource that
        no longer exists, you may want to remove or fix the link.
```

PR #8234, opened 2020-12-07
Incorrect link from summary links for classes

albert-github:
When running a link checker on the CGAL Classification or Point_set_processing_3 package we get the warning:
```
file:///.../doc_output/Classification/classCGAL_1_1Classification_1_1Point__set__feature__generator.html
  Code: 200 (no message)
 To do: Some of the links to this resource point to broken URI fragments
        (such as index.html#fragment).
The following fragments need to be fixed:
        nested-classes
```
This is due to the fact that on the mentioned page no "section" classes is but in the summary links (top right) there is still "Classes".

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -2120,6 +2120,7 @@ void ClassDefImpl::writeAuthorSection(OutputList &ol) const
 
 void ClassDefImpl::writeSummaryLinks(OutputList &ol) const
 {
+  static bool extractPrivate = Config_getBool(EXTRACT_PRIVATE);
   ol.pushGeneratorState();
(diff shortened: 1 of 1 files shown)


albert-github:
This problem looks similar to the problem of #8232 but is apparently slightly different (#8232 was for namespaces and this is for classes) and is still present after the fix for #8232.

Note similar problems might exist for the other items in the summary link but did not, yet, surface.

albert-github:
@doxygen can you sine a light on this problem? Currently the "nested-classes" are present in the CGAL documentation?

albert-github:
Problem is still present for CGAL, might be good to have this PR integrated.

PR #8242, opened 2020-12-11
Incorrect specification of irc as URL protocol

albert-github:
- missing colon (`:`) at irc
- adding irc, sftp and ftps

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -6686,7 +6686,8 @@ bool isURL(const QCString &url)
 {
   QCString loc_url = url.stripWhiteSpace();
   return loc_url.left(5)=="http:" || loc_url.left(6)=="https:" ||
-         loc_url.left(4)=="ftp:"  || loc_url.left(5)=="file:";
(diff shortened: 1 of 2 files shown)


albert-github:
Interesting list: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml, though I don't think all protocols would be meaningful for doxygen to support.

src/util.cpp
```diff
@@ -6686,7 +6686,8 @@ bool isURL(const QCString &url)
 {
   QCString loc_url = url.stripWhiteSpace();
   return loc_url.left(5)=="http:" || loc_url.left(6)=="https:" ||
-         loc_url.left(4)=="ftp:"  || loc_url.left(5)=="file:";
+         loc_url.left(4)=="ftp:"  || loc_url.left(5)=="ftps:"  || loc_url.left(5)=="sftp:"  ||
+         loc_url.left(5)=="file:";
```
doxygen, line 6690: I was wondering if for consistency with the `URLPROTOCOL` pattern in `doctokenizer.l`, the prefixes `news:`, `irc:`, and `ircs`: should be added here as well?

albert-github: Probably both sets should have the same list.

Issue #8142, opened 2020-10-29
Doxygen using "override" keyword in function signature causing mis-matches


aotheoverlord:
This bug produces an error regardless of whether a function is documented or not.
```
class Foo : public BaseClass {
  void some_fn() const noexcept override;
};

Foo::some_fn() const noexcept
{
...
}
```
>> warning: no uniquely matching class member found for 
  Foo::some_fn()

It appears to be the override keyword; that cannot be placed in the function definition, only the declaration.

This is in 1.8.13 and 1.9.0, so seems to be an on-going problem.

albert-github:
I tried to reproduce your problem but I didn't succeed, probably due to the lack of a configuration file (Doxyfile) and / or the right input file(s) and their name(s), so:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

aotheoverlord:
This will be hard to do ... I do suspect it's a config issue; can you attach the config you used and I'll compare and make some adjustments.  If that doesn't resolve it, I'll work on getting a full minimal test case together.

albert-github:
It is a very small project.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5473812/example.tar.gz)

albert-github:
As no response and the comment in we have to conclude that either the issue is incomplete or usage, hence the issue will be automatically closed at the moment of an official release.

PR #8243, opened 2020-12-11
issue #8142 UTF-8 in URL in source generates truncated URL in HTML

albert-github:
Specifying the URL in a similar way as other IDs


albert-github:
Just found a small problem in case of LaTeX output.

Issue #8241, opened 2020-12-11
UTF-8 in URL in source generates truncated URL in HTML


paul-j-lucas:
**Describe the bug**
A Doxygen comment in C code contains:

    @sa https://en.wikipedia.org/wiki/Damerau–Levenshtein_distance

Note that the "–" there is a Unicode En Dash (U+2013) and is in the `.c` file encoded as UTF-8: E2 80 93.

The generated HTML is:

    <a href="https://en.wikipedia.org/wiki/Damerau">https://en.wikipedia.org/wiki/Damerau</a>–Levenshtein_distance

That is, the URL gets truncated at the en-dash character and the remaining part of the URL is put after the `</a>`.  Of course this means that the URL is wrong.

If you're going to suggest using percent-encoded characters, i.e.:

    https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance

that doesn't work either.  In that case, the generated HTML is:

    <a href="https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance">https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance</a>

i.e., the percent-encoded characters are preserved as-is rather than interpreted as being in UTF-8 and they're displayed as such in a browser which is wrong. (However, the link displayed in the URL tooltip at the bottom of a Safari window _is_ correct since Safari does the correct interpretation of the UTF-8. Hence, the correct link is clickable. It just _looks_ wrong on the screen.

**Expected behavior**
I expect the HTML generated _should_ be:

    <a href="https://en.wikipedia.org/wiki/Damerau–Levenshtein_distance">https://en.wikipedia.org/wiki/Damerau–Levenshtein_distance</a>

That is, I expect your URL parser _not_ to stop at the start of a UTF-8 encoded character. Your parser should stop only upon encountering one of a whitespace, `<`, or `>` character.

**To Reproduce**
```
/**
 * Implements the Damerau-Levenshtein algorithm to calculate an _edit distance_
 * between strings, i.e., how many letters need to be swapped within,
 * substituted within, deleted from, or added to \a source to get \a target.
 *
 * @param source The source string.
 * @param target The target string.
 * @return Returns said distance.
 *
 * @sa https://en.wikipedia.org/wiki/Damerau–Levenshtein_distance
 * @sa https://www.lemoda.net/text-fuzzy/damerau-levenshtein/
 */
[...]

PR #8244, opened 2020-12-11
Incorrect handling of UTF-8 characters in latex reference statement

albert-github:
In case in there is a literal utf-8 character  like the the Unicode En Dash (U+2013)  a pages is accessed with `\T1\endash` like:
```
https://en.wikipedia.org/wiki/Damerau\T1\textendash_Levenshtein_distance
```
and this page cannot be found, so the UTF-8 character has to be handled properly here.
This is found when solving issue #8241

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5680746/example.tar.gz)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -5309,7 +5309,63 @@ QCString latexFilterURL(const char *s)
       case '%':  t << "\\%"; break;
       case '\\':  t << "\\\\"; break;
       default:
-        t << c;
+        if (c<0)
+        {
+          char ids[5];
+          const unsigned char uc = (unsigned char)c;
+          bool doEscape = TRUE;
+          if (uc <= 0xf7)
+          {
+            const signed char* pt = (signed char *)p;
+            ids[ 0 ] = c;
+            int l = 0;
+            if ((uc&0xE0)==0xC0)
+            {
+              l=2; // 11xx.xxxx: >=2 byte character
+            }
+            if ((uc&0xF0)==0xE0)
+            {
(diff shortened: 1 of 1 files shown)


src/util.cpp
```diff
@@ -5309,7 +5309,63 @@ QCString latexFilterURL(const char *s)
...
+                unsigned char id = (unsigned char)ids[m];
+                t << "\\%" << map[id>>4] << map[id&0xF];
+              }
+              p += l - 1;
+            }
+          }
+          if (doEscape) // not a valid unicode char or escaping needed
+          {
+            static char map[] = "0123456789ABCDEF";
+            unsigned char id = (unsigned char)c;
```
doxygen, line 5361: `map` and `id` are not used in this case and nothing is escaped. Looks like something missing for this case...

albert-github: Copy paste error, we don't have an idea how to escape these characters here so so write the as is, the lines:
```
     static char map[] = "0123456789ABCDEF";
     unsigned char id = (unsigned char)c;
```
should not be present.

doxygen: ok, I think also the name of the `doEscape` variable is confusing. If it is `false' % escaping is done, so exactly the opposite of what the name suggests. Should this be corrected as well or am I missing something?

albert-github: Most likely yes, I copied the code from another part and (unfortunately) didn't pay attention to the names of all variables.

PR #8254, opened 2020-12-16
Improvements of reading the configuration file

albert-github:
- making the parsing of the configuration consistent with the reading of the configuration file a doxygen run
- adding a number of warnings in case of problems in the configuration file
- making the configuration warnings better visible (popping up the "Output log Window").

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5701862/example.tar.gz)

### DIFF
--- a/addon/doxywizard/CMakeLists.txt
+++ b/addon/doxywizard/CMakeLists.txt
@@ -98,6 +98,7 @@ wizard.h
 qt_add_resources(doxywizard_RESOURCES_RCC doxywizard.qrc)
 
 add_executable(doxywizard WIN32
+config_msg.cpp
 doxywizard.cpp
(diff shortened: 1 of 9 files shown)


albert-github:
@doxygen 
Currently the configuration warnings are only shown when the `--debug` flag is used when starting the doxywizard.
I think we should drop this restriction.
- what is your opinion?

doxygen:
> @doxygen
> Currently the configuration warnings are only shown when the `--debug` flag is used when starting the doxywizard.
> I think we should drop this restriction.
> 
>     * what is your opinion?

Sounds like a good idea indeed.

albert-github:
OK I'll create a PR for it (tomorrow).

PR #8258, opened 2020-12-17
Extend tests with some other default tests

albert-github:
During the change to GitHub Actions the extra  tests as added in #8217 were lost:

Extend tests with generating xhtml, docbook, rtf to see if the files can be generated and build.
Check also the generated xsd files in xml mode.

Note for the new checks no comparison is done with previous versions.


albert-github:
Investigating ...

albert-github:
Closed in favor of #8260 due to the numerous tests here

PR #8275, opened 2020-12-23
Concatenation of \param text and \details text

albert-github:
When having a simple file like:
```
/// \file

/// @brief Different xxxxxxxxxxxxxxxxxxxxx
/// @param[in]   var = zzzzzz
/// @details Different yyyyyyyy
void fie(int var);
```
we see in the version 1.8.18 that the text is displayed as:
```
Parameters
    [in] var = zzzzzz

Different yyyyyyyy
```
but in the newer versions this is like:
```
Parameters
    [in] var = zzzzzz Different yyyyyyyy
```
the inserted intermediate `\ilinebr` are not handled as newlines

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5736487/example.tar.gz)

### DIFF
--- a/src/doctokenizer.l
+++ b/src/doctokenizer.l
@@ -810,7 +810,7 @@ RCSID "$"("Author"|"Date"|"Header"|"Id"|"Locker"|"Log"|"Name"|"RCSfile"|"Revisio
                          }
                          lineCount(yytext,yyleng);
                        }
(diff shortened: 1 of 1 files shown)


albert-github:
I'm wondering a little bit whether the `\n` in the lines above this patch also should be replaced in the same way.
I do think so, but due to the lack of an example I didn't include it in the patch.

PR #8279, opened 2020-12-27
archive build artefacts in GitHub Actions

dhebbeker:
Adds an action which archives the build artefacts in `build/bin/` for each build configuration. This provides the binaries for all supported operating systems for every workflow run available.

This facilitates to have development builds available for all operating systems.

Preview
----------

Screenshot from 2020-12-27 11-23-55

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -60,6 +60,11 @@ jobs:
           }
     steps:
     - uses: actions/checkout@v1
+    
+    - name: print env
+      run: |
+        pwd
+        find
       
     - name: Download MikTex (Windows)
       run: |
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -283,6 +283,12 @@ jobs:
           message(FATAL_ERROR "Build failed")
         endif()
 
+    - name: Archive build artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: "${{ matrix.config.name }} build artifacts"
+        path: bin
+
     - name: Run tests (Linux / MacOS)
       shell: cmake -P {0}
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -283,6 +283,12 @@ jobs:
           message(FATAL_ERROR "Build failed")
         endif()
 
+    - name: Archive build artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: "${{ matrix.config.name }} build artifacts"
+        path: build/bin/
+
```
albert-github, line 291: - Wouldn't this explode i.e. give 8 set of executables for each build? or will only the last be preserved?
- How can these builds be accessed?

doxygen: @albert-github From what I read [here](https://github.com/actions/upload-artifact) the artifacts are generated per build but are stored for a maximum of 90 days only (can be made shorter). Build artifacts are accessed via the actions page, e.g. https://github.com/doxygen/doxygen/actions/runs/446837009 under Artifacts.

doxygen: @dhebbeker Nice feature!

albert-github: Interesting article! and also the link to the artifacts of the individual build.

I see there that it says that the debug version for windows is 511 MB, looks a bit of a calculation problem (downloading it gives about 99 MB and also the release version give a factor bigger) but that is not our problem, probably the way it is stored.

I'm still a bit worried about the amount of storage that is required (looks like now per build around 1Gb) and the number of pull requests / merges done with master.

doxygen: @albert-github I read [here](https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts):
 
> Storing artifacts uses storage space on GitHub. GitHub Actions usage is free for public repositories and self-hosted runners. 

and

> all files that were individually uploaded as part of the artifact get zipped together into a single file. This means that billing is calculated based on the size of the uploaded artifact and not the size of the zip file.

Let wait for Github to judge if we use too much data.

PR #8289, opened 2020-12-29
Add a CLANG_INCLUDE_INPUT_PATHS option

brycelelbach:
This PR adds a new option, `CLANG_INCLUDE_INPUT_PATHS`, which is on by default. If this option is on, then the directory of each input file is added to the include path when using Clang parsing. This behavior is the current status quo, but some projects need to be able to disable it.

Adding the directory of every input file to the include path is problematic for a number of projects. We ran into problems with it in [Thrust](https://github.com/NVIDIA/thrust).

For example, imagine a project that has this structure:

```
include/myproject/version.h
include/myproject/component0/coolstuff.h
include/myproject/component1/coolstuff.h
include/myproject/component1/widget.h
```

And those files are intended to be included by users as:

```
#include <myproject/version.h>
#include <myproject/component0/coolstuff.h>
#include <myproject/component1/coolstuff.h>
#include <myproject/component1/widget.h>
```

But within `include/myproject/component1/widget.h`, you might have:

```
#include "coolstuff.h" // Should resolve to `include/myproject/component1/coolstuff.h`.
```

If you have `INPUT = include`, then the directories of all headers found in `include` will be added as includes by Doxygen:

```
-Iinclude/myproject -Iinclude/myproject/component0 -Iinclude/myproject/component1
```

This is problematic, because now `include/myproject/component1/widget.h` will include `include/myproject/component0/coolstuff.h` instead of `include/myproject/component1/coolstuff.h`.

Alternatively, consider the case we ran into in Thrust, where our project structure looks like:

```
thrust/version.h
...
thrust/system/detail/errno.h
...
```

With `INPUT = thrust`, Doxygen will add the following as includes:

```
"-Ithrust ... -Ithrust/system/detail"
```

This is problematic because `errno.h` is also the name of a system header, and will now be found instead of the system header.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -982,14 +982,14 @@ Go to the <a href="commands.html">next</a> section or return to the
       <docs>
 <![CDATA[
  With the correct setting of option \c CASE_SENSE_NAMES doxygen will better be able to match the
- capabilities of the underlying filesystem. 
+ capabilities of the underlying filesystem.
 
- In case the filesystem is case sensitive (i.e. it supports files in the same directory 
+ In case the filesystem is case sensitive (i.e. it supports files in the same directory
  whose names only differ in casing), the option must be set to \c YES to properly deal with such files
- in case they appear in the input. 
- 
- For filesystems that are not case sensitive the option should be be set to \c NO to properly 
- deal with output files written for symbols that only differ in casing, such as for two classes, 
+ in case they appear in the input.
+
+ For filesystems that are not case sensitive the option should be be set to \c NO to properly
+ deal with output files written for symbols that only differ in casing, such as for two classes,
  one named \c CLASS and the other named \c Class, and to also support references to files without
  having to specify the exact matching casing.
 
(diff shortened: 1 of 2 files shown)


src/config.xml
```diff
@@ -1737,6 +1737,15 @@ to disable this feature.
...
   was generated with the `-Duse_libclang=ON` option for CMake.
+]]>
+      </docs>
+    </option>
+    <option type='bool' id='CLANG_INCLUDE_INPUT_PATHS' setting='USE_LIBCLANG' defval='1'>
+      <docs>
+<![CDATA[
+  If clang assisted parsing is enabled and the \c CLANG_INCLUDE_INPUT_PATHS tag
+  is set to \c YES then doxygen will use add the directory of each input to the
+  include path.
```
albert-github, line 1748: I think this sentence is not correctly formulated (especially the "use add").

brycelelbach: Fixed.

PR #8290, opened 2020-12-29
Change container type of UsedDir from QDict to STL container

dhebbeker:
Effort to reduce (outdated) Qt dependencies (similar to #7651) in order to simplify modern C++ coding. This allows to use range based for loops.


doxygen:
@dhebbeker I reworked the code a bit to make use of LinkedMap, see also https://github.com/doxygen/doxygen/commit/0c1bc5557d811d662b4144f14ef716d9e950d60d

dhebbeker:
Thank you for the prompt review and for improving my suggested changes @doxygen! Now I am aware of `LinkedMap`. :+1:

PR #8306, opened 2021-01-04
Optimize: usedDir can not be parent of dd if they have the same parent.

dhebbeker:
How can there be a case where the first two conditions are true, but not the third? It should be possible to omit the recursive check of `isParentOf()`.

### DIFF
--- a/src/dotdirdeps.cpp
+++ b/src/dotdirdeps.cpp
@@ -81,7 +81,7 @@ void writeDotDirDepGraph(FTextStream &t,const DirDef *dd,bool linkRelations)
       // draw all directories which have `dd->parent()` as parent and `dd` as dependent
       const auto newEnd = std::remove_if(usedDirsNotDrawn.begin(), usedDirsNotDrawn.end(), [&](const DirDef *const usedDir)
       {
-        if (dd!=usedDir && dd->parent()==usedDir->parent() && !usedDir->isParentOf(dd))
+        if (dd!=usedDir && dd->parent()==usedDir->parent())
         {
           drawDirectory(t, usedDir, usedDir->isCluster() && !Config_getBool(DOT_TRANSPARENT), dirsInGraph);
           return true;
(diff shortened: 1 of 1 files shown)


src/dotdirdeps.cpp
```diff
@@ -136,8 +136,7 @@ void writeDotDirDepGraph(FTextStream &t,const DirDef *dd,bool linkRelations)
         //    shortName().data(),
         //    !usedDir->isParentOf(this)
         //    );
-        if (dir!=usedDir && dir->parent()==usedDir->parent() &&
-            !usedDir->isParentOf(dd))
```
albert-github, line 140: might this have something to do with recursive directories (maybe based on symbolic links)?

dhebbeker: Do you mean, that a chain of `parent()` may be cyclic because of (symbolic) links in the file system?

albert-github: Yes this is what I meant, it is a wild guess though.

dhebbeker: I did some speculative tests in order to break it. But with no success: Apparently, the symbolic links are not used. Although I had them not excluded. This is my Doxyfile:

```
EXTRACT_ALL            = YES
RECURSIVE              = YES
HAVE_DOT               = YES
DOT_CLEANUP            = NO
EXCLUDE_SYMLINKS       = NO
DOT_IMAGE_FORMAT       = PNG
```

```
$ tree src/
src/
└── usedDir
    ├── dependee.h
    └── parent
        ├── dir
        │   └── dependent.h
        ├── link -> ../../usedDir/
        ├── otherDir
        │   └── dependee3_original.h
        ├── otherDir2
        │   └── dependee3.h -> otherDir/dependee3_original.h
        └── usedDir2
            ├── dependee2.h
            └── dependee.h -> usedDir/dependee.h
```

`dependent.h`:
```
#include "dependee.h"
#include "dependee2.h"
#include "dependee3.h"
```

The result is, that `dependee3.h`, `usedDir2/dependee.h`, `link` do not appear in the File List page neither in the directory dependency graphs.

[recursive_link_example.zip](https://github.com/doxygen/doxygen/files/5766626/recursive_link_example.zip)

doxygen: I recently wondered as well why this check was there. I also couldn't think of any reason.

PR #8313, opened 2021-01-07
Missing info in XML: location, width, valign, alt

### LINKED ISSUE #8312: Missing info in XML: location, width, valign, alt
**Describe the bug**
While building our custom rendering on top of DoxyGen generated XML, we found that some XML tags/attributes are missing for our usecase:
- `<location>` of pages
- width attribute in `<table>` and `<entry>`
- valign in `<entry>`
- alt in `<image>`

Since that are quite common attributes which work fine for html output, can you please add them in XML as well? We did this already locally and I will propose the change as pull request.

**Expected behavior**
The XML for the minimal sample attached below should include:

~~~
<para><table rows="1" cols="2" width="500"><row>
<entry thead="no" width="50%" valign="middle"><para></para>
</entry><entry thead="no"><para></para>
</entry></row>
</table>
</para>
<para><image type="html" name="image.png" alt="Image description" inline="yes"></image>
~~~

What we currently get:
~~~
<para><table rows="1" cols="2"><row>
<entry thead="no"><para></para>
</entry><entry thead="no"><para></para>
</entry></row>
</table>
</para>
<para><image type="html" name="image.png" inline="yes"></image>
~~~

**To Reproduce**
`doxygen Doxyfile` with [missing_xml_info.zip](https://github.com/doxygen/doxygen/files/5782158/missing_xml_info.zip)

**Version**
Tested with 1.9.0

Dschaehn:
Fixes #8312


src/xmldocvisitor.cpp
```diff
@@ -774,7 +778,17 @@ void XmlDocVisitor::visitPre(DocHtmlTable *t)
...
+  HtmlAttribListIterator li(t->attribs());
+  HtmlAttrib* opt;
+  for (li.toFirst(); (opt = li.current()); ++li)
+  {
+    if (opt->name=="width")
+    {
+      m_t << " " << opt->name << "=\"" << opt->value << "\"";
+    }
+  }
+  m_t << ">";
```
albert-github, line 791: In the HTML implementation we see that there is the usage of `htmlAttribsToString` and in this implementation the choice, now, has been made to program it for every part.
What is the reason for not using `htmlAttribsToString` or an equivalent version to write to the XML file?

albert-github:
To successfully run the tests you also need to update the tests, i.e.  `make tests TEST_FLAGS=--updateref`, see also `make tests TEST_FLAGS=--help`.

albert-github:
I saw the problem in the test and could locate the place of the problem (strangely I didn't see it during my tests).
I think I identified the problem , so I've just pushed a proposed patch, pull request #8317

src/xmldocvisitor.cpp
Dschaehn: I tried to be consistent to the existing code in xmldocvisitor.cpp and keep the change minimal. For example  `XmlDocVisitor::visitPre(DocHtmlCell)` also carefully selects just a few attributes. As far as I see, `htmlAttribsToString` is not used anywhere in xmldocvisitor.cpp so far.

From my perspective it would probably make sense to use `htmlAttribsToString`, that is to pass through all html attributes like the html exporter does. But this would apply to many other tags as well then. I see `htmlAttribsToString` being used 37 times in htmldocvisitor.cpp.

Dschaehn:
Thank you for the quick fix!

I first thought it was my fault (I updated the tests on windows and manually removed some different whitespaces from the xml output). But now I have setup ubuntu and the tests pass succesfully. I still get slightly different whitespace when updating the references, but it looks like your tests ignore those when comparing.

albert-github:
Where do you see differences  in the spacing?
For testing the results are transformed to a canonical form with xmllint like (from runtest.py):
```
                                        # convert output to canonical form
                                        data = xpopen('%s --format --noblanks --nowarning %s' % (self.args.xmllint,check_file))
```

doxygen:
Think this change also requires updating the XML schema file (e.g. `templates/xml/compound.xsd`)

Dschaehn:
@albert-github: I get a difference in the two xml files below in the text content of some XML tags. Maybe it's because of my setup (I used a WSL ubuntu installation under windows). If you think it's worth investigating, I could dig further and provide more details.

```diff
--- a/testing/021/indexpage.xml
+++ b/testing/021/indexpage.xml
@@ -6,7 +6,7 @@
     <briefdescription>
     </briefdescription>
     <detaileddescription>
-      <para> Our main function starts like this: <programlisting filename="example_test.cpp"><codeline><highlight class="keywordtype">void</highlight><highlight class="normal"><sp/>main()</highlight></codeline><codeline><highlight class="normal">{</highlight></codeline></programlisting> First we create a object <computeroutput>t</computeroutput> of the <ref refid="class_test" kindref="compound">Test</ref> class. <programlisting filename="example_test.cpp"><codeline><highlight class="normal"><sp/><sp/><ref refid="class_test" kindref="compound">Test</ref><sp/>t;</highlight></codeline></programlisting> Then we call the example member function <programlisting filename="example_test.cpp"><codeline><highlight class="normal"><sp/><sp/>t.<ref refid="class_test_1a47b775f65718978f1ffcd96376f8ecfa" kindref="member">example</ref>();</highlight></codeline></programlisting> After that our little test routine ends. <programlisting filename="example_test.cpp"><codeline><highlight class="normal">}</highlight></codeline></programlisting> </para>
+      <para> Our main function starts like this: <programlisting filename="example_test.cpp"><codeline><highlight class="keywordtype">void</highlight><highlight class="normal"><sp/>main()</highlight></codeline><codeline><highlight class="normal">{</highlight></codeline></programlisting>First we create a object <computeroutput>t</computeroutput> of the <ref refid="class_test" kindref="compound">Test</ref> class. <programlisting filename="example_test.cpp"><codeline><highlight class="normal"><sp/><sp/><ref refid="class_test" kindref="compound">Test</ref><sp/>t;</highlight></codeline></programlisting>Then we call the example member function <programlisting filename="example_test.cpp"><codeline><highlight class="normal"><sp/><sp/>t.<ref refid="class_test_1a47b775f65718978f1ffcd96376f8ecfa" kindref="member">example</ref>();</highlight></codeline></programlisting>After that our little test routine ends. <programlisting filename="example_test.cpp"><codeline><highlight class="normal">}</highlight></codeline></programlisting></para>
     </detaileddescription>
   </compounddef>
 </doxygen>
diff --git a/testing/055/md_055_markdown.xml b/testing/055/md_055_markdown.xml
index 423a28c7..e0a11082 100644
--- a/testing/055/md_055_markdown.xml
+++ b/testing/055/md_055_markdown.xml
@@ -23,7 +23,7 @@
           <para>
             <ulink url="http://example.com/last-line">Upper-cased reference link on last line</ulink>
           </para>
-          <para>Dash - NDash <ndash/> MDash <mdash/> EDash - ENDash -- EMDash --- E3Dash --- </para>
+          <para>Dash - NDash <ndash/> MDash <mdash/> EDash - ENDash -- EMDash --- E3Dash ---</para>
         </sect2>
         <sect2 id="md_055_markdown_1autotoc_md3">
           <title>Markdown in HTML</title>
```

albert-github:
Regarding the differences in https://github.com/doxygen/doxygen/pull/8313#issuecomment-756381140
The difference is at the beginning and end of a line / paragraph and should not have any influence in any result.
Unfortunately I'm not able to reproduce it.

PR #8327, opened 2021-01-12
Incorrect type for enum struct

albert-github:
In case we have the simple file:
```
/// \file

/// docu with nothing
        enum MyEnum_not { v1_0, v2_0 };
/// docu with class
        enum class MyEnum_cls { v1_2, v2_2 };
/// docu with struct
        enum struct MyEnum_str { v1_3, v2_3 };
```
with settings:
```
QUIET=YES
ALWAYS_DETAILED_SEC=YES
```
the `enum struct` is shown as `enum class` in the brief description and in the detailed description just as `enum` (the later also is the case for the `enum class`).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5802540/example.tar.gz)

### DIFF
--- a/src/entry.h
+++ b/src/entry.h
@@ -137,6 +137,7 @@ class Entry
     static const uint64 Local           = (1ULL<<16); // for Slice types
 
     // member specifiers (add new items to the beginning)
+    static const uint64 EnumStruct      = (1ULL<<18);
     static const uint64 ConstExpr       = (1ULL<<19); // C++11 constexpr
     static const uint64 PrivateGettable     = (1ULL<<20); // C# private getter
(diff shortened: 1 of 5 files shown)


doxygen:
Although it maybe good to output in the documentation what is in the source code, there is no difference between `enum class` and `enum struct` in C++ see also [this discussion](https://stackoverflow.com/questions/38536571/what-is-the-difference-between-enum-struct-and-enum-class).

albert-github:
This is correct, though the intention was so show exactly the words `struct` and `class` as it is normally also done for classes:

image

 versus
image

in the files html/struct_my_struct.html and html/class_my_class.html

Example: [example1.tar.gz](https://github.com/doxygen/doxygen/files/6404988/example1.tar.gz)

Issue #8336, opened 2021-01-15
Python class only with async methods support


yosshy:
**Describe the bug**
Doxygen doesn't generate a class reference of python properly if the class had multiple async methods and no sync method. It generates one with ONLY async method name.

**Expected behavior**
Doxygen generates a class reference with multiple async method names properly.

**To Reproduce**

1. Create a working directory and enter it with cd command.
2. Put a sample source code like below:
    ```
    ## Documentation for an async class.
    #  More details.
    class AsyncPyClass:
    
        ## Documentation for a method_a.
        #  @param self The object pointer.
        async def method_a(self):
            pass
    
        ## Documentation for a method_b.
        #  @param self The object pointer.
        async def method_b(self):
            pass
    
        ## Documentation for a method_c.
        #  @param self The object pointer.
        async def method_c(self):
            pass
    ```
3. Run "doxygen -g Doxyfile" to create a configuration file.
4. Run "doxygen Doxyfile" to generate documents.

**Version**
1.9.2 (31709395a2ab12ba137e87cf4d8b70f2cada8396*)

albert-github:
Code has been integrated in master on github (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.2.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #8337, opened 2021-01-15
issue #8336 Python class only with async methods support

yosshy:


src/pyscanner.l
```diff
@@ -261,6 +261,12 @@ STARTDOCSYMS      "##"
 			  searchFoundDef(yyscanner);
 			  BEGIN( FunctionDec );
       			}
+    ^{B}"async"{BB}"def"{BB} { // start of an async function/method definition with indent
+      			  DBG_CTX((stderr,"Found async def at %d\n",yyextra->yyLineNr));
+      			  yyextra->indent=computeIndent(yytext);
+			  searchFoundDef(yyscanner);
+			  BEGIN( FunctionDec );
+      			}
```
albert-github, line 269: - Can you give a reference to `async` in the python documentation (e.g. a pep)?
- You created a new rule analogous to `^{B}"def"{BB}` , why isn't there a rule analogous to `    "def"{BB}`?

albert-github (commented):
You added the `async` keyword to `pyscanner.l` but shouldn't it also be added to `pycode.l`?

src/pyscanner.l
doxygen: @albert-github The PEP can be found here https://www.python.org/dev/peps/pep-0492/ (feature seems to be available since Python 3.5)

albert-github: Thanks for the pep.
Shouldn't the word `await` be handled as well? I didn't test it, might not cause any problems in the pyscanner.l, but in the pycode.l it won't be marked as a keyword.

yosshy: @albert-github Because I don't know why "def"{BB} is also defined although ^{B}"def"{BB} is defined. BTW, you look right I should add a patch for pycode.l. I'll make another patchset and push it in a few days.

albert-github: Probably the `"def"{BB}` is also defined for cases where there are multiple statements on one line (not sure whether this is anyway possible though). There must have been a reason for it as the split has already been made in 2005.
I think it is wise to add the rule with `async` for this case as well.

albert-github: When looking at the documentation:

scanner1

I don't see any reference to the `async` status, shouldn't this me mentioned in the blue part as well?

src/pycode.l
```diff
@@ -190,7 +190,7 @@ SHORTSTRINGITEM   ({SHORTSTRINGCHAR}|{ESCAPESEQ})
 SHORTSTRINGCHAR   [^\\\n"]
 STRINGLITERAL     {STRINGPREFIX}?( {SHORTSTRING} | {LONGSTRING})
 STRINGPREFIX      ("r"|"u"|"ur"|"R"|"U"|"UR"|"Ur"|"uR")
-KEYWORD           ("lambda"|"import"|"class"|"assert"|"with"|"as"|"from"|"global"|"def"|"True"|"False")
+KEYWORD           ("lambda"|"import"|"class"|"assert"|"with"|"as"|"from"|"global"|"async"|"def"|"True"|"False")
```
albert-github, line 193: In the pep there is also the mention of `await`, shouldn't this be added as well here?

src/pycode.l
```diff
@@ -302,6 +302,12 @@ TARGET            ({IDENTIFIER}|"("{TARGET_LIST}")"|"["{TARGET_LIST}"]"|{ATTRIBU
                                       endFontClass(yyscanner);
                                       BEGIN( FunctionDec );
                                     }
+      "async"{BB}"def"{BB}          {
+                                      startFontClass(yyscanner,"keyword");
+                                      codify(yyscanner,yytext);
+                                      endFontClass(yyscanner);
+                                      BEGIN( FunctionDec );
+                                    }
```
albert-github, line 310: Not sure if this rule is really necessary, as `async` is already in the "KEYWORD" definition and we need the rule for `def` only as we need to go to a different state of the code scanner, the `async` does not result in a different state. (Though it is not that bad).

Issue #8350, opened 2021-01-22
Parent namespace in C++


pgn69:
**Describe the bug**
In C++ an identifier like `::name` indicates `name` in the parent namespace.

Example:
```
#include <iostream>
#include <typeinfo>

typedef int my_type;

namespace ns1 {
    typedef float my_type;

    typedef my_type T1; 
    typedef ::my_type T2;
}

int main()
{        
    ns1::T1 t1;
    ns1::T2 t2;
    
    std::cout << typeid(t1).name() << std::endl;
    std::cout << typeid(t2).name() << std::endl;
    
}
```

In the XML generated by Doxygen both `T1` and `T2` are of type `ns1::my_type` that is `float`.

In the generated XML both T1 and T2 have the same _refid_:

```
<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="1.8.17">
  <compounddef id="namespacens1" kind="namespace" language="C++">
    <compoundname>ns1</compoundname>
      <sectiondef kind="typedef">
      <memberdef kind="typedef" id="namespacens1_1a97e44b5c3b5e0ff2c6f5e10e2fb5e683" prot="public" static="no">
        <type>float</type>
        <definition>typedef float ns1::my_type</definition>
        <argsstring></argsstring>
        <name>my_type</name>
        <briefdescription>
        </briefdescription>
        <detaileddescription>
        </detaileddescription>
        <inbodydescription>
        </inbodydescription>
        <location file="example.cpp" line="8" column="19" bodyfile="example.cpp" bodystart="8" bodyend="-1"/>
      </memberdef>
      <memberdef kind="typedef" id="namespacens1_1ae68d60ab16e75a49595398395df15c71" prot="public" static="no">
        <type><ref refid="namespacens1_1a97e44b5c3b5e0ff2c6f5e10e2fb5e683" kindref="member">my_type</ref></type>
        <definition>typedef my_type ns1::T1</definition>
        <argsstring></argsstring>
        <name>T1</name>
        <briefdescription>
        </briefdescription>
        <detaileddescription>
        </detaileddescription>
        <inbodydescription>
        </inbodydescription>
        <location file="example.cpp" line="10" column="21" bodyfile="example.cpp" bodystart="10" bodyend="-1"/>
      </memberdef>
[...]

PR #8352, opened 2021-01-22
Incorrect code coloring of typedef

albert-github:
In case we have `typedef ::my_type T2;` the code coloring should be:
- typedef : keyword
- ::my_type : link
- T2 : link

though we see:
- typedef ::my_type : link
- T2 : link

this has been corrected.
Found by means of issue #8350
Note there might be other cases with the same problem as well.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/5856314/example.tar.gz)

### DIFF
--- a/src/code.l
+++ b/src/code.l
@@ -1102,6 +1102,7 @@ NUMBER {INTEGER_NUMBER}|{FLOAT_NUMBER}
                                           yyextra->name+=yytext;
                                         }
 <Body>{SCOPENAME}/{BN}*[:;,)\]]         { // "int var;" or "var, var2" or "debug(f) macro" , or int var : 5;
+                                          if (QCString(yytext).startsWith("typedef")) REJECT;
                                           addType(yyscanner);
                                           // changed this to generateFunctionLink, see bug 624514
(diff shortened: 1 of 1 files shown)


src/code.l
```diff
@@ -1102,13 +1102,15 @@ NUMBER {INTEGER_NUMBER}|{FLOAT_NUMBER}
                                           yyextra->name+=yytext;
                                         }
 <Body>{SCOPENAME}/{BN}*[:;,)\]]         { // "int var;" or "var, var2" or "debug(f) macro" , or int var : 5;
+                                          if (QCString(yytext).startsWith("typedef")) REJECT;
```
doxygen, line 1105: Shouldn't we use something like 

```
if (QCString(yytext)=="typedef" || QCString(yytext).startsWith("typedef "))
``` 
here otherwise a variable called `typedefinition` will also be matched?

albert-github: I oversaw this corner case, though I see that it might be that also "typedef:" (for e.g `typedef::my_type T5;`) has to be supported. There might be even more characters have to be supported, I'm not sure which characters would be possible though.

Please adjust as would be best.

PR #8361, opened 2021-01-28
Parser for verbatim including a (text) file.

albert-github:
In older projects it is quite common to have files like `README.txt` that don't contain code or markdown markup, but that would be nice to have verbatim into the documentation.
The same is true for other types of files like configuration files or sh / zsh / bat files that should not be processed in ant way.

For this the "language" `unparsed` is created.


albert-github:
Note: for the remapping to work for existing extensions like `txt`, it is also necessarythat PR #8360 is committed to the master branch.

PR #8369, opened 2021-02-01
Support vertical alignment in a multi-row HTML cell to LaTex

DuyDang007:
There is an issue of pdflatex where a rowspan cell value stay out of page margin (see the picture below)

doxygen_table_before

I created this request to support vertical alignment when using multi-row HTML cell as a solution.

User need to add "valign=..." attribute to \<td\> tag for vertical alignment. Supported values are "top", "bottom", otherwise the cell is middle aligned as default.

This can be combined with horizontal alignment ("align" attribute).

_Example_:

```html
<td rowspan=2 align=left valign=top>Cell value</td>
```

The result is looking better:

doxygen_table_after

This supports only HTML to LaTeX table. Markdown seems not have the same feature.

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -3408,6 +3408,8 @@ DocHtmlCell::Valignment DocHtmlCell::valignment() const
         return Top;
       else if (attr.value.lower()=="bottom")
         return Bottom;
+      else if (attr.value.lower()=="middle")
+        return Middle;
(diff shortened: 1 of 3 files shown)


DuyDang007:
That's OK. I added checking for middle value. But people can leave it empty, or any other value also makes it middle aligned as default.

PR #8382, opened 2021-02-13
add config option WARN_IF_INCOMPLETE_DOC

arwedus:
Add a new warning class, "INCOMPLETE_DOC", and warning option
"WARN_IF_INCOMPLETE_DOC", to control whether you get a warning
for only documenting some of your function parameters.

All users who want to retain the current behavior set
WARN_IF_INCOMPLETE_DOC to the value of WARN_IF_DOC_ERROR.

User story
----------

As a developer who follows clean code principles, 

I strive for self-explanatory function parameters (using specialized and constrained
data types, and using meaningful names). Some of my parameters need additional documentation, 
but not all.

I want to trigger a warning in my continuous integration system when a pull requests
leads to new Doxygen warnings, but I want to avoid boilerplate `@param` commands without content.

Problem
-------

Currently, Doxygen can only be configured to either issue no warnings about documentation errors
at all (WARN_IF_DOC_ERROR). I can tune Doxygen to not warn about completetly missing param docs
(WARN_NO_PARAMDOC=NO), but if I then *add* documentation for some parameters I get a warning.
This feels kinda weird.

Tests
-----

The problematic warning looks like this:

```
Generating docs for compound kdtree::KDTree...
kdtree/kdtree.h:89: warning: The following parameter of kdtree::KDTree::findNearestNeighbor(const PointT &query, SquaredDistanceFuncT squared_distance_function) const is not documented:
  parameter 'query'
```

With the new config option enabled, those warnings disappear. Ran it on 2 of my projects.
I did not find a warning-related automated test case in doxygen, so that's all I could do here.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -1342,7 +1342,7 @@ FILE_VERSION_FILTER = "cleartool desc -fmt \%Vn"
  documentation.
  If \ref cfg_extract_all "EXTRACT_ALL" is set to \c YES then this flag will
  automatically be disabled.
- See also \ref cfg_warn_if_incomplete_paramdoc "WARN_IF_INCOMPLETE_PARAMDOC"
+ See also \ref cfg_warn_if_incomplete_doc "WARN_IF_INCOMPLETE_DOC"
 ]]>
       </docs>
     </option>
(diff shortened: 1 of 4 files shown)


arwedus:
This is actually my re-implementation of a super-old PR https://github.com/doxygen/doxygen/pull/18 I unfortunately squashed my changed into the cherry-pick commit so I'm not the official author now but oh well.

PR #8383, opened 2021-02-13
Get preprocessor output without line numbers

albert-github:
For easier comparison of  the original source with doxygen preprocessed source the line numbers can be a bit of an obstacle.
To remove the line numbers the debug option  `PreprocessorNolineno` has been added.
(In case `Preprocessor` and `PreprocessorNolineno` the latest invoked one is used).

### DIFF
--- a/doc/preprocessing.doc
+++ b/doc/preprocessing.doc
@@ -259,6 +259,10 @@ you can run doxygen as follows:
 \verbatim
   doxygen -d Preprocessor
 \endverbatim
+or without line numbers:
+\verbatim
+  doxygen -d PreprocessorNolineno
+\endverbatim
 This will instruct doxygen to dump the input sources to standard output after
 preprocessing has been done (Hint: set <code>QUIET = YES</code> and 
 <code>WARNINGS = NO</code> in the configuration file to disable any other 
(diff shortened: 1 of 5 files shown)


doc/preprocessing.doc
```diff
@@ -259,6 +259,10 @@ you can run doxygen as follows:
 \verbatim
   doxygen -d Preprocessor
 \endverbatim
+or without line numbers:
+\verbatim
+  doxygen -d PreprocessorNolineno
```
doxygen, line 264: I think it would simplify things a bit if a `-d nolines` option is just added to the list, so without line numbers you'd call doxygen with `-d Preprocessor -d nolines`. Then the code below can just `|` the flag and in places where it makes a difference do a `Debug::isFlagSet(Debug::NoLines)`.

albert-github: No objections, should I update the PR or will you do it?

doxygen: If you can do it, then that would be nice 😃

albert-github: It is on my list for tomorrow

albert-github:
Withdrawn in favor of #8386

Issue #8390, opened 2021-02-17
Reusing documentation snippets


eseiler:
I am wondering if there is a way to reuse recurring documentation, e.g. in the form of doxygen snippets.

**Describe the bug**
For example, we have some documentation that looks like:

```cpp
/*!\brief A fine struct.
 *
 * \details
 *
 * Some descriptions.
 *
 * | Column | Column | Column |
 * |--------|--------|--------|
 * |  Cell  |  Cell  |  Cell  |
 * |  Cell  |  Cell  |  Cell  |
 * |  Cell  |  Cell  |  Cell  |
 */
struct fine_struct{};

/*!\brief Another struct.
 *
 * \details
 *
 * Some other description that differs from fine_struct, but uses the same table.
 *
 * | Column | Column | Column |
 * |--------|--------|--------|
 * |  Cell  |  Cell  |  Cell  |
 * |  Cell  |  Cell  |  Cell  |
 * |  Cell  |  Cell  |  Cell  |
 */
struct another_struct{};
```

Of course in different files, with different detailed description, but they share some, but not all, elements (e.g. only the table is common).

**Expected behavior**
We would like to be able to do something like this:

```cpp
/*!\brief Yet another struct.
 *
 * \details
 *
 * Yet another description.
 *
 * \include{doc} /dev/shm/tables/table.md
 */
struct yet_another_struct{};

/*!\brief One last struct.
 *
 * \details
 *
 * One last description that differs from yet_another_struct, but uses the same table.
 *
 * \include{doc} /dev/shm/tables/table.md
 */
struct one_last_struct{};
```

But this currently does not work:

**Screenshots**
![doxy report](https://user-images.githubusercontent.com/12967715/108201971-c270da00-7120-11eb-9b2c-47ffa4c95c1a.PNG)

**To Reproduce**
[example.tar.gz](https://github.com/doxygen/doxygen/files/5995485/example.tar.gz)
The `\include{doc} /dev/shm/tables/table.md` (`include.hpp`) is currently hardcoded and may need to be changed.

**Version**
`1.9.1 (ef9b20ac7f8a8621fcfc299f8bd0b80422390f4b)`

albert-github:
It looks like the included markdown commands are not going through the markdown processor and thus are literally included into the code. This is one of the (non mentioned) exceptions in the Note with the `\include{doc}` command.

You could use instead of the markdown code the corresponding HTML code.
In this case:
```
<table class="markdownTable">
[...]

Issue #7062, opened 2019-06-17
Markdown table not formatted when located in external file


skramm:
- Version: 1.8.15
- Platform: Ubuntu 16.04

When using `\include{doc}` to include an externally generated file holding Markdown content, the table formatting fails:

**Source:**

```
\section sec1 some direct markdown table

| col1 | col2 |
|------|------|
| ab | cd |
| ef | g1 |
| 23 | 45 |

\section sec2 some included markdown table

This below should appear formatted as above:

\include{doc} markdown_table.inc

```
**Result** (external file holds same table content):

![doxygen_bug](https://user-images.githubusercontent.com/6195178/59595339-e717e180-90f5-11e9-8080-3c754ed7ba23.png)

- no error/warning generated
- Steps to reproduce: extract attached file on Debian-based box, get down in folder, run `build.sh`, check `html/index.html`
[mardown_included_issue.zip](https://github.com/doxygen/doxygen/files/3296275/mardown_included_issue.zip)

[Source](https://stackoverflow.com/questions/56584239/)

albert-github:
In the documentation there is already the warning:

> Note
> Some that when using the {doc} option, commands like \cond and \if don't work with this command due to the moment of parsing.
> The included documentation should not have comment signs in it as they will appear in the documentation as well.

It looks like this is also one of these cases.

albert-github:
This issue has partly been solved through #8395 but this fix also removes everything after the `\section` command, so this issue is not yet completely solved.

JPenuchot:
Adding another example mentioned in https://github.com/doxygen/doxygen/pull/8395:

> If you `\include{doc} foo.md` with `foo.md` containing the following:
> 
> ```lisp
> 1. This will show up properly
> 1. This will show up properly
> 1. This will show up properly
> 1. This will show up properly
> 
> This will *show up* properly
> 
> > This will show up properly
> 
> This will show up properly
> 
> # This title won't show up
> 
> This won't show up either
> ```
> 
> or
> 
> ```lisp
> 1. This will show up properly
> 1. This will show up properly
> 1. This will show up properly
> 1. This will show up properly
> 
> This will *show up* properly
> 
> > This will show up properly
> 
> This will show up properly
> 
> @section This title won't show up
> 
> This won't show up either
> ```
> 
> Nothing will show up in the documentation starting from the title. I tried different title levels and got the same result.
> 
[...]

PR #8395, opened 2021-02-19
issue #8390 Reusing documentation snippets

albert-github:
When `MARKDOWN_SUPPORT` is enabled run the documentation of the included file / snippet through the markdown processor (analogous to comment blocks in the different scanners).

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -5153,10 +5153,15 @@ void DocPara::handleInclude(const QCString &cmdName,DocInclude::Type t)
        inc_line = lineBlock(inc_text, blockId);
        inc_text = extractBlock(inc_text, blockId);
      }
+
+     Markdown markdown(fileName,inc_line);
+     QCString strippedDoc = stripIndentation(inc_text);
+     QCString processedDoc = Config_getBool(MARKDOWN_SUPPORT) ? markdown.process(strippedDoc,inc_line) : strippedDoc;
+
      docParserPushContext();
      g_fileName = fileName;
      setDoctokinizerLineNr(inc_line);
-     internalValidatingParseDoc(this,m_children,inc_text);
+     internalValidatingParseDoc(this,m_children,processedDoc);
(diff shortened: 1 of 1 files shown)


JPenuchot:
Hi @albert-github, I've been testing your feature and it looks like content that follows the beginning of a title won't display. Here's an example:

If you `\include{doc} foo.md` with `foo.md` containing the following:

```md
1. This will show up properly
1. This will show up properly
1. This will show up properly
1. This will show up properly

This will *show up* properly

> This will show up properly

This will show up properly

# This title won't show up

This won't show up either
```

Nothing will show up in the documentation starting from the title. I tried different title levels and got the same result.

albert-github:
It might be one of the exceptions as (not mentioned explicitly) in the doxumentation:
> Some commands like \cond and \if don't work with this command due to the moment of parsing.
The included documentation should not have comment signs in it as they will appear in the documentation as
well.

But still best to create an issue for it.

albert-github:
I thought I had seen another issue as well and looks like I found it now: #7062 best is to add your example to that issue.

JPenuchot:
On 4/8/21 12:14 PM, albert-github wrote:
> I thought I had seen another issue as well and looks like I found it 
> now: #7062 <https://github.com/doxygen/doxygen/issues/7062> best is to 
> add your example to that issue.

Are you sure it's relevant to this issue? It points the lack of Markdown 
parsing in Doxygen 1.18.15 which was fixed, not the lack of Doxygen 
command parsing after the translation. I might be wrong but they seem to 
be separate issues to me.

Thanks for the Markdown fix however, I know someone will find that 
feature to be a total game changer. :)

albert-github:
The problem in the mentioned #7062 indeed looks like it signals the only the problem of the non markdown processing, but when applying that fix we miss the entire 2nd table (i.e. the table through the `\includedoc`) so the issue #7062 is only partly solved.

Issue #8375, opened 2021-02-09
Lowercase search does not find non-ASCII uppercase pages and vice versa


apolukhin:
**Describe the bug**
Searching "привет" while the page title is "Привет" does not work. 

**Expected behavior**
For ASCII everything works fine: searching for "faq" finds page "FAQ". The same behavior expected for non ASCII pages.

**Screenshots**
![seach_lowercase_ru](https://user-images.githubusercontent.com/1700098/107350048-fec98800-6ad9-11eb-8174-d72f4ec89a9b.png)

**To Reproduce**
1. Make some markdown pages starting with `# Привет`, `# Основные сведения`, `# прочее` and `# Введение`.
2. Generate docs from those pages using the `SEARCHENGINE=YES` and `SERVER_BASED_SEARCH=NO`
3. Search for `при`, `ПРИВ`, `основ`, `ПРОЧЕЕ`, `вВеден`

**Version**
1.9.1 and trunk.

albert-github:
Recently I've see a fix for a similar issue #5263 that might also might fix your problem (although you mentioned also trunk with the problem), though it is hard to tell without an example (I don't have a good example at hand with Cyrillic search terms).
- Please specify the trunk version you used (complete information of `doxygen -v`).
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

apolukhin:
```
doxygen -v
1.9.2 (46ffd77e8d3c15244732128caace858f2aa38d73)
```
Searching for "ОСН", "ПРИ", "При" and "проч" works (was fixed in #5263, thanks!)
Searching for "при", "вве", "ПРОЧ", "осн", does **not** work

Attaching a self contained example, all the info is duplicated inside in `Doxygen` config:
[doxygen_sample.zip](https://github.com/doxygen/doxygen/files/5952587/doxygen_sample.zip)

albert-github:
It was at first a bit confusing as the output language was still English, but the problem here lies with the names of the sections in the related pages.

albert-github:
The problem looks like a bit more a fundamental problem when writing (a.o. search/all*.js, search/pages*.js, search/searchdata.js) it looks like the translation to lowercase is **not** done as it is done for the Latin alphabet / ASCII

albert-github:
I've just pushed a proposed patch, pull request #8409

doxygen:
[...]

Issue #5263, opened 2018-07-02
Indexing and searching cannot treat non ASCII identifiers (Origin: bugzilla #705910)


doxygen:
status REOPENED severity *normal* in component *general* for *---*
Reported in version *1.8.6-GIT* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[20130812_fix_indexing_and_searching_to_enable_non_ascii_identifiers.diff](https://sourceforge.net/projects/doxygen/files/attachments/bug705910_251489_20130812_fix_indexing_and_searching_to_enable_non_ascii_identifiers.diff/download/download)_ (ID 251489)
- _[20130910_fix_indexing_and_searching_to_enable_non_ascii_identifiers.diff](https://sourceforge.net/projects/doxygen/files/attachments/bug705910_254582_20130910_fix_indexing_and_searching_to_enable_non_ascii_identifiers.diff/download/download)_ (ID 254582)
- _[bug705910_2013_10_21.zip](https://sourceforge.net/projects/doxygen/files/attachments/bug705910_257809_bug705910_2013_10_21.zip/download/download)_ (ID 257809)
- _[20131027_new_patch_for_bug_705910.zip](https://sourceforge.net/projects/doxygen/files/attachments/bug705910_258178_20131027_new_patch_for_bug_705910.zip/download/download)_ (ID 258178)
- _[bug705910_2013_10_27.zip](https://sourceforge.net/projects/doxygen/files/attachments/bug705910_258383_bug705910_2013_10_27.zip/download/download)_ (ID 258383)
- _[20131227_new_patch_for_issue705910.diff](https://sourceforge.net/projects/doxygen/files/attachments/bug705910_264918_20131227_new_patch_for_issue705910.diff/download/download)_ (ID 264918)

On 2013-08-13 13:24:02 +0000, Suzumizaki-Kimitaka wrote:

    Created attachment 251489
    fix indexing and built-in searching for non ASCII identifiers
    
    I made the patch already, please just apply it.
    The details and notes about regression test are below.
    
    Details:
    a1) Because the indices make groups against first byte(octet) of UTF-8, the
    entries are wrong grouped when the names start with non ASCII characters.
    Like U+0080 - U+00BF go to the group 0xC2 and U+0800 - U+0FFF go to the
    group 0xE0. Ofcourse they should go same as ASCII characters, like 'A' to
    'A', 'B' to 'B'.
    
[...]

PR #8409, opened 2021-03-04
issue #8375 Lowercase search does not find non-ASCII uppercase pages and vice versa

albert-github:
Implementation of a uppercase / lowercase conversion as needed by doxygen.
The standard tolower / toupper functions don't really work as they need a "locale" which in general is not necessary for Unicode / UTF8 conversions.
- caseconvert.cpp / caseconvert.h generated code based on the table from https://www.unicode.org/Public/13.0.0/ucd/UnicodeData.txt with some small modifications regarding uppercase values that shouldn't have a lowercase representation (Kelvin sign) or combined characters where there is no 100% one to one relation between uppercase and lowercase due to some mix (e.g.  DZ,  Dz and  dz).
- util.cpp / searchengine.cpp using the new functions
- search.js to old "workaround" is not necessary anymore (see issue #5263)

The used generation software and data: [codeconv.tar.gz](https://github.com/doxygen/doxygen/files/6084035/codeconv.tar.gz)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -220,6 +220,7 @@ add_library(doxymain STATIC
     ${GENERATED_SRC}/resources.cpp
     #
     arguments.cpp
+    caseconvert.cpp
     cite.cpp
     clangparser.cpp
     classdef.cpp
--- a/src/caseconvert.cpp
+++ b/src/caseconvert.cpp
@@ -1,0 +1,3804 @@
+// This is generated code
+// see github issue 8375 and the accompanying pull request for more information
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "caseconvert.h"
+int StrToUprExt(unsigned char* pString, CaseBuffer *cb)
+{
+  size_t pStrLen = strlen((char *)pString);
+  if (cb->curBufLen < pStrLen + 40)
+  {
+    cb->outBuf = (unsigned char *)realloc(cb->outBuf,pStrLen + 41);
+    cb->curBufLen = pStrLen + 40;
+  }
+
+  size_t curLen = 0;
+  unsigned char tmpBuf[4];
+  if (pString && *pString)
+  {
+    size_t replacementLen = 0;
+    unsigned char* p = pString;
+    while (*p)
+    {
+      replacementLen = 0;
+      switch(*(p+0))
+      {
+        case 0x61: tmpBuf[0] = 0x41; replacementLen = 1; break;
+        case 0x62: tmpBuf[0] = 0x42; replacementLen = 1; break;
+        case 0x63: tmpBuf[0] = 0x43; replacementLen = 1; break;
+        case 0x64: tmpBuf[0] = 0x44; replacementLen = 1; break;
+        case 0x65: tmpBuf[0] = 0x45; replacementLen = 1; break;
+        case 0x66: tmpBuf[0] = 0x46; replacementLen = 1; break;
+        case 0x67: tmpBuf[0] = 0x47; replacementLen = 1; break;
+        case 0x68: tmpBuf[0] = 0x48; replacementLen = 1; break;
+        case 0x69: tmpBuf[0] = 0x49; replacementLen = 1; break;
+        case 0x6a: tmpBuf[0] = 0x4a; replacementLen = 1; break;
+        case 0x6b: tmpBuf[0] = 0x4b; replacementLen = 1; break;
+        case 0x6c: tmpBuf[0] = 0x4c; replacementLen = 1; break;
+        case 0x6d: tmpBuf[0] = 0x4d; replacementLen = 1; break;
+        case 0x6e: tmpBuf[0] = 0x4e; replacementLen = 1; break;
+        case 0x6f: tmpBuf[0] = 0x4f; replacementLen = 1; break;
+        case 0x70: tmpBuf[0] = 0x50; replacementLen = 1; break;
+        case 0x71: tmpBuf[0] = 0x51; replacementLen = 1; break;
+        case 0x72: tmpBuf[0] = 0x52; replacementLen = 1; break;
+        case 0x73: tmpBuf[0] = 0x53; replacementLen = 1; break;
+        case 0x74: tmpBuf[0] = 0x54; replacementLen = 1; break;
+        case 0x75: tmpBuf[0] = 0x55; replacementLen = 1; break;
+        case 0x76: tmpBuf[0] = 0x56; replacementLen = 1; break;
+        case 0x77: tmpBuf[0] = 0x57; replacementLen = 1; break;
+        case 0x78: tmpBuf[0] = 0x58; replacementLen = 1; break;
+        case 0x79: tmpBuf[0] = 0x59; replacementLen = 1; break;
+        case 0x7a: tmpBuf[0] = 0x5a; replacementLen = 1; break;
+        case 0xc3:
+          switch(*(p+1))
+          {
+            case 0x9f: tmpBuf[0] = 0xe1; tmpBuf[1] = 0xba; tmpBuf[2] = 0x9e; replacementLen = 3; break;
+            case 0xa0: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x80; replacementLen = 2; break;
+            case 0xa1: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x81; replacementLen = 2; break;
+            case 0xa2: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x82; replacementLen = 2; break;
+            case 0xa3: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x83; replacementLen = 2; break;
+            case 0xa4: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x84; replacementLen = 2; break;
+            case 0xa5: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x85; replacementLen = 2; break;
+            case 0xa6: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x86; replacementLen = 2; break;
+            case 0xa7: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x87; replacementLen = 2; break;
+            case 0xa8: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x88; replacementLen = 2; break;
+            case 0xa9: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x89; replacementLen = 2; break;
+            case 0xaa: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x8a; replacementLen = 2; break;
+            case 0xab: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x8b; replacementLen = 2; break;
+            case 0xac: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x8c; replacementLen = 2; break;
+            case 0xad: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x8d; replacementLen = 2; break;
+            case 0xae: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x8e; replacementLen = 2; break;
+            case 0xaf: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x8f; replacementLen = 2; break;
+            case 0xb0: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x90; replacementLen = 2; break;
+            case 0xb1: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x91; replacementLen = 2; break;
+            case 0xb2: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x92; replacementLen = 2; break;
+            case 0xb3: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x93; replacementLen = 2; break;
+            case 0xb4: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x94; replacementLen = 2; break;
+            case 0xb5: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x95; replacementLen = 2; break;
+            case 0xb6: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x96; replacementLen = 2; break;
+            case 0xb8: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x98; replacementLen = 2; break;
+            case 0xb9: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x99; replacementLen = 2; break;
+            case 0xba: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x9a; replacementLen = 2; break;
+            case 0xbb: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x9b; replacementLen = 2; break;
+            case 0xbc: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x9c; replacementLen = 2; break;
+            case 0xbd: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x9d; replacementLen = 2; break;
+            case 0xbe: tmpBuf[0] = 0xc3; tmpBuf[1] = 0x9e; replacementLen = 2; break;
+            case 0xbf: tmpBuf[0] = 0xc5; tmpBuf[1] = 0xb8; replacementLen = 2; break;
+            default:
+              break;
+          }
+          break;
+        case 0xc4:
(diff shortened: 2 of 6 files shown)


doxygen:
@albert-github I like the idea of generating a mapping file for the case conversion, but I think it can be done more easily with a Python 3 script like below. Here I just iterate over all character codes and check if Python would give a different output when doing `.upper()` or `.lower()` thus avoiding processing of external tables altogether. The characters are stored as a utf8 string in the generated mapping file using the C++11 `u8""` string literal prefix

```python
import codecs

toupper = {}
tolower = {}

def writeMapping(file,mapping):
    for k,v in sorted(mapping.items()):
        file.write(u"  { u8\"%s\",u8\"%s\" }, // %s -> %s\n" % (k,v,hex(ord(k[0])),hex(ord(v[0]))))

# create mappings of characters whose upper and lower case differs
for codeValue in range(0,0x1FFFF):
        s = chr(codeValue)
        sl = s.lower()
        su = s.upper()
        if ord(s[0])!=ord(sl[0]):
            tolower[s]=sl
        if ord(s[0])!=ord(su[0]):
            toupper[s]=su

file = codecs.open("caseconvert_data.cpp", "w", "utf-8")
file.write(r'''/** This file is generated by caseconvert.py. DO NOT EDIT! */

#include <string>
#include <unordered_map>
std::unordered_map<std::string,std::string> toUpperMap = {
''');
writeMapping(file,toupper);
file.write(r'''};

std::unordered_map<std::string,std::string> toLowerMap = {
''');
writeMapping(file,tolower);
file.write(r'''};
''');
file.close()
```

Then the string conversion routines in C++ can be done using the generated mapping files one utf8 character at a time:

```c++
static void convert(const std::string &input,std::string &output,
                    const std::unordered_map<std::string,std::string> &map)
{
  output="";
  if (input.empty()) return;
  const unsigned char *p = (const unsigned char *)input.c_str();
  while (*p)
  {
    int len=0;
    const unsigned char uc = *p;
    unsigned char utf8char[5];
    if (uc<0xF7)
    {
      if      ((uc&0x80)==0x00) len=1;
      else if ((uc&0xE0)==0xC0) len=2;
      else if ((uc&0xF0)==0xE0) len=3;
      else if ((uc&0xF8)==0xF0) len=4;
    }
    bool valid = len>0;
    for (int i=0; i<len && valid; i++)
    {
      if (*p==0 || (i>0 && (*p&0xC0)!=0x80)) { valid = false; } else { utf8char[i]=*p++; }
    }
    if (!valid) break;
    utf8char[len]=0;
    auto it = map.find((const char *)utf8char);
    if (it!=map.end())
    {
      output+=it->second;
    }
    else
    {
      output+=(const char *)utf8char;
    }
  }
}

void strToUpperUTF8(const std::string &input,std::string &output)
{
  convert(input,output,toUpperMap);
}

void strToLowerUTF8(const std::string &input,std::string &output)
{
  convert(input,output,toLowerMap);
}

```

What do you think of this approach?

albert-github:
Promising approach
Some remarks and errors:
- I'm missing the handling of some special characters, like the Kelvin symbol. I personally think that e.g. the Kelvin symbol has no lowercase pendant.
- besides the Lower case and Upper case characters there are also the Title case characters, the later are most of the times equal to the Upper case characters. I don't think by the python version the Title case characters are handled.
  - handling of  combined characters where there is no 100% relation between uppercase / lowercase conversion e.g. the DZ/Dz/dz conversion (Unicode: 1F1 / 1F2/ 1F3) 
- part where the python conversion is wrong e.g 'GREEK SMALL LETTER ALPHA WITH PSILI AND YPOGEGRAMMENI' (U+1F80) <-> 'GREEK CAPITAL LETTER ALPHA WITH PSILI' (U+1F08)
- how will the speed of the map will be be compared to the nested case construct?

**Below a list of differences found**

Uppercase -> Lowercase

Extra in the python approach

 0x1c5 -> 0x1c6

 0x1c8 -> 0x1c9

 0x1cb -> 0x1cc

 0x1f2 -> 0x1f3

 0x2126 -> 0x3c9
 0x212a -> 0x6b
 0x212b -> 0xe5

Extra in the cpp approach 
 0xa7ba -> 0xa7bb
 0xa7bc -> 0xa7bd
 0xa7be -> 0xa7bf
 0xa7c2 -> 0xa7c3
 0xa7c4 -> 0xa794
 0xa7c5 -> 0x282
 0xa7c6 -> 0x1d8e
 0xa7c7 -> 0xa7c8
 0xa7c9 -> 0xa7ca
 0xa7f5 -> 0xa7f6

Lowercase -> Uppercase
Extra in the python approach
 python approach

 0xb5 -> 0x39c
 0xdf -> 0x53

 0x149 -> 0x2bc

 0x1f0 -> 0x4a


 0x345 -> 0x399

 0x390 -> 0x399

 0x3b0 -> 0x3a5

 0x587 -> 0x535

 0x1e96 -> 0x48
 0x1e97 -> 0x54
 0x1e98 -> 0x57
 0x1e99 -> 0x59
 0x1e9a -> 0x41

 0x1f50 -> 0x3a5

 0x1f52 -> 0x3a5

 0x1f54 -> 0x3a5

 0x1f56 -> 0x3a5

 0x1f80 -> 0x1f08
 0x1f81 -> 0x1f09
 0x1f82 -> 0x1f0a
 0x1f83 -> 0x1f0b
 0x1f84 -> 0x1f0c
 0x1f85 -> 0x1f0d
 0x1f86 -> 0x1f0e
 0x1f87 -> 0x1f0f
 0x1f88 -> 0x1f08
 0x1f89 -> 0x1f09
 0x1f8a -> 0x1f0a
 0x1f8b -> 0x1f0b
 0x1f8c -> 0x1f0c
 0x1f8d -> 0x1f0d
 0x1f8e -> 0x1f0e
 0x1f8f -> 0x1f0f
 0x1f90 -> 0x1f28
 0x1f91 -> 0x1f29
 0x1f92 -> 0x1f2a
 0x1f93 -> 0x1f2b
 0x1f94 -> 0x1f2c
 0x1f95 -> 0x1f2d
 0x1f96 -> 0x1f2e
 0x1f97 -> 0x1f2f
 0x1f98 -> 0x1f28
 0x1f99 -> 0x1f29
 0x1f9a -> 0x1f2a
 0x1f9b -> 0x1f2b
 0x1f9c -> 0x1f2c
 0x1f9d -> 0x1f2d
 0x1f9e -> 0x1f2e
 0x1f9f -> 0x1f2f
 0x1fa0 -> 0x1f68
 0x1fa1 -> 0x1f69
 0x1fa2 -> 0x1f6a
 0x1fa3 -> 0x1f6b
 0x1fa4 -> 0x1f6c
 0x1fa5 -> 0x1f6d
 0x1fa6 -> 0x1f6e
 0x1fa7 -> 0x1f6f
 0x1fa8 -> 0x1f68
 0x1fa9 -> 0x1f69
 0x1faa -> 0x1f6a
 0x1fab -> 0x1f6b
 0x1fac -> 0x1f6c
 0x1fad -> 0x1f6d
 0x1fae -> 0x1f6e
 0x1faf -> 0x1f6f

 0x1fb2 -> 0x1fba
 0x1fb3 -> 0x391
 0x1fb4 -> 0x386
 0x1fb6 -> 0x391
 0x1fb7 -> 0x391
 0x1fbc -> 0x391

 0x1fc2 -> 0x1fca
 0x1fc3 -> 0x397
 0x1fc4 -> 0x389
 0x1fc6 -> 0x397
 0x1fc7 -> 0x397
 0x1fcc -> 0x397

 0x1fd2 -> 0x399
 0x1fd3 -> 0x399
 0x1fd6 -> 0x399
 0x1fd7 -> 0x399

 0x1fe2 -> 0x3a5
 0x1fe3 -> 0x3a5
 0x1fe4 -> 0x3a1

 0x1fe6 -> 0x3a5
 0x1fe7 -> 0x3a5
 0x1ff2 -> 0x1ffa
 0x1ff3 -> 0x3a9
 0x1ff4 -> 0x38f
 0x1ff6 -> 0x3a9
 0x1ff7 -> 0x3a9
 0x1ffc -> 0x3a9

 0xfb00 -> 0x46
 0xfb01 -> 0x46
 0xfb02 -> 0x46
 0xfb03 -> 0x46
 0xfb04 -> 0x46
 0xfb05 -> 0x53
 0xfb06 -> 0x53
 0xfb13 -> 0x544
 0xfb14 -> 0x544
 0xfb15 -> 0x544
 0xfb16 -> 0x54e
 0xfb17 -> 0x544

Extra in the cpp approach 

 0xdf -> 0x1e9e

 0x282 -> 0xa7c5

 0x1d8e -> 0xa7c6

 0x1f80 -> 0x1f88
 0x1f81 -> 0x1f89
 0x1f82 -> 0x1f8a
 0x1f83 -> 0x1f8b
 0x1f84 -> 0x1f8c
 0x1f85 -> 0x1f8d
 0x1f86 -> 0x1f8e
 0x1f87 -> 0x1f8f
 0x1f90 -> 0x1f98
 0x1f91 -> 0x1f99
 0x1f92 -> 0x1f9a
 0x1f93 -> 0x1f9b
 0x1f94 -> 0x1f9c
 0x1f95 -> 0x1f9d
 0x1f96 -> 0x1f9e
 0x1f97 -> 0x1f9f
 0x1fa0 -> 0x1fa8
 0x1fa1 -> 0x1fa9
 0x1fa2 -> 0x1faa
 0x1fa3 -> 0x1fab
 0x1fa4 -> 0x1fac
 0x1fa5 -> 0x1fad
 0x1fa6 -> 0x1fae
 0x1fa7 -> 0x1faf

 0x1fb3 -> 0x1fbc

 0x1fc3 -> 0x1fcc

 0x1ff3 -> 0x1ffc

 0xa794 -> 0xa7c4

 0xa7bb -> 0xa7ba
 0xa7bd -> 0xa7bc
 0xa7bf -> 0xa7be
 0xa7c3 -> 0xa7c2
 0xa7c8 -> 0xa7c7
 0xa7ca -> 0xa7c9
 0xa7f6 -> 0xa7f5

doxygen:
> I'm missing the handling of some special characters, like the Kelvin symbol. I personally think that e.g. the Kelvin symbol has no lowercase pendant.

Formally you are correct of course, but if someone really used the Kelvin sign (and not the capital K) which is already rather unlikely, would it then be a problem if this result appeared when searching for `k`. Maybe it would even be expected.

> besides the Lower case and Upper case characters there are also the Title case characters, the later are most of the times equal to the Upper case characters. I don't think by the python version the Title case characters are handled.

These are also typically only used in historic literature, so I'm not sure how big of a problem this would be for the application domain. 

> part where the python conversion is wrong e.g 'GREEK SMALL LETTER ALPHA WITH PSILI AND YPOGEGRAMMENI' (U+1F80) <-> 'GREEK CAPITAL LETTER ALPHA WITH PSILI' (U+1F08)

I'm not so sure this is wrong. If I may believe this page https://graphemica.com/%E1%BE%80 then the upper case character is https://graphemica.com/%E1%BC%88 and https://graphemica.com/%E1%BE%88 is actually listed as the title case character.

> how will the speed of the map will be be compared to the nested case construct?

Using the map approach is probably a bit slower, but it is not like this conversion will be done a huge amount of times, so I don't expect it will have a measurable impact on the total runtime.

albert-github:
> > I'm missing the handling of some special characters, like the Kelvin symbol. I personally think that e.g. the Kelvin symbol has no lowercase pendant.
> 
> Formally you are correct of course, but if someone really used the Kelvin sign (and not the capital K) which is already rather unlikely, would it then be a problem if this result appeared when searching for k. Maybe it would even be expected.

When the conversion to lowercase is only used for the purpose of the search index it is probably the most logical and expected usage. It also depends on how the conversion in search.js is performed. So further investigation here is required.
(this accounts for: "2126" OHM SIGN, "212A" KELVIN SIGN and "212B" ANGSTROM SIGN).
For the "00B5" MICRO SIGN, it is a bit the other way around, here there is no real uppercase (suggested is the capital M), so when this is accepted as exception it is OK (i.e. that a Micro sign will never appear at the beginning of a line, somewhere I thought that the first character of a brief description was converted to uppercase, but didn't find it quickly).

> > besides the Lower case and Upper case characters there are also the Title case characters, the later are most of the times equal to the Upper case characters. I don't think by the python version the Title case characters are handled.
> 
> These are also typically only used in historic literature, so I'm not sure how big of a problem this would be for the application domain.

Probably for doxygen code purposes they would not be a big issue, though users might use these characters as they like them better than the normal uppercase letters so in searching they would be missing.

> > part where the python conversion is wrong e.g 'GREEK SMALL LETTER ALPHA WITH PSILI AND YPOGEGRAMMENI' (U+1F80) <-> 'GREEK CAPITAL LETTER ALPHA WITH PSILI' (U+1F08)
> 
> I'm not so sure this is wrong. If I may believe this page https://graphemica.com/%E1%BE%80 then the upper case character is https://graphemica.com/%E1%BC%88 and https://graphemica.com/%E1%BE%88 is actually listed as the title case character.

When looking at the pyton conversion we see (`grep -i -w 1F[08][08]`):
- { u8"ἀ",u8"Ἀ" }, // 0x1f00 -> 0x1f08
- { u8"ᾀ",u8"ἈΙ" }, // 0x1f80 -> 0x1f08
- { u8"ᾈ",u8"ἈΙ" }, // 0x1f88 -> 0x1f08
- { u8"Ἀ",u8"ἀ" }, // 0x1f08 -> 0x1f00
- { u8"ᾈ",u8"ᾀ" }, // 0x1f88 -> 0x1f80

when looking in UnicodeData.txt we see:
- 1F00;GREEK SMALL LETTER ALPHA WITH PSILI;Ll;0;L;03B1 0313;;;;N;;;1F08;;1F08
- 1F08;GREEK CAPITAL LETTER ALPHA WITH PSILI;Lu;0;L;0391 0313;;;;N;;;;1F00;
- 1F80;GREEK SMALL LETTER ALPHA WITH PSILI AND YPOGEGRAMMENI;Ll;0;L;1F00 0345;;;;N;;;1F88;;1F88
- 1F88;GREEK CAPITAL LETTER ALPHA WITH PSILI AND PROSGEGRAMMENI;Lt;0;L;1F08 0345;;;;N;;;;1F80;

(first column Unicode character,, last 3 columns: Uppercase, Lowercase, Titlecase, see also: https://www.unicode.org/reports/tr44/tr44-26.html#UnicodeData.txt)

when looking at :
- https://www.fileformat.info/info/unicode/char/1f00/index.htm
- https://www.fileformat.info/info/unicode/char/1f08/index.htm
- https://www.fileformat.info/info/unicode/char/1f80/index.htm
- https://www.fileformat.info/info/unicode/char/1f88/index.htm

From the 2 last sources I gather that 1F00 and 1F08 are coupled and 1F80 and 1F88 are coupled and thus python is not corect.

> > how will the speed of the map will be be compared to the nested case construct?
> 
> Using the map approach is probably a bit slower, but it is not like this conversion will be done a huge amount of times, so I don't expect it will have a measurable impact on the total runtime.

This depends on the number of times the routines for the character case conversion are called (search for , looks like it is concentrated in index.cpp and searchindex.cpp should probably also be considered in combination with the remarks about "KELVIN").

Some new insights:
- ligatures / 
  in the UnicodeData.txt we see:
  `FB05;LATIN SMALL LIGATURE LONG S T;Ll;0;L;<compat> 017F 0074;;;;N;;;;;`
  so no uppercase version
  in the python conversion we see:
  `{ u8"ﬅ",u8"ST" }, // 0xfb05 -> 0x53`
  so an uppercase version present consisting of 2 uppercase characters (where also last characters is missing in comment and probably in test).
  From https://www.fileformat.info/info/unicode/char/fb05/index.htm this is logical as here a compatible decomposition is given, it looks like this is a missing feature of the cpp version.

So in total:
- I see some wrong characters
- List has to be checked for "strange" conversions
- some further tests have to be done regarding the usage of the lowercase / uppercase usage.

doxygen:
@albert-github With commit a4ecbee86766b35d25d41d1a178806e1688485df I've kept the python conversion approach (if it is good enough for Python it should be good enough for doxygen, even if some esoteric cases may be incorrect). 

I've now optimized the code's performance by using a big switch case (Assuming the compiler is able to optimize this better than using a dynamic hash map) and I've now moved all UTF8 specific routines to an utf8.cpp file to promote code reuse.

albert-github:
> if it is good enough for Python it should be good enough for doxygen, even if some esoteric cases may be incorrect

I think it is a bad idea to take python as a reference for good code things as it has in my opinion some big very strange language constructs:
- a block is defined by the amount of space, so differences is space can mean a difference of the meaning of a program. Also mixing tabs and spaces (always a bad idea to use spaces anyway) might give very strange results
- no necessity for declaring variables
- no strong typing for arguments or local variables, so a local variable can mean in the same routine 2 different things, even on the same "block" level.

albert-github:
A different approach has been chosen for generating the conversion table.

PR #8417, opened 2021-03-07
UTF-8 characters in email address

albert-github:
When having:
```
# test email adddress

 mailru: Видео@Mail.Ru

```
we get a warning like:
```
.../bb.md:3: warning: Found unknown command '@Mail'
```

According to RFC6530 it is also possible to have UTF-8 (see also https://en.wikipedia.org/wiki/Email_address#Internationalization).

- extended lexers with range `\x80-\xFF` as done for other UTF-8 searches in lexers
- adjusted no spam filter so UTF-8 characters are kept together

(Found by Fossies in youtube-dl-2021.03.03)

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6096928/example.tar.gz)

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -124,7 +124,7 @@ static void clearCommentStack(yyscan_t yyscanner);
 
 %}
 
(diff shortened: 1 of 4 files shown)


doxygen:
Note that I changed the "no spam" filtering to use `writeUTF8Char()`

PR #8426, opened 2021-03-13
Incorrect line number from tag file

albert-github:
In the ACE TAO project we see a lot of warnings like:
```
.../ACE_TAO/ACE/html/libtao-doc/TAO.tag:1: error: Internal inconsistency: namespace in IDL not module, library or constant group
```
the line number 1 is here incorrect and should be 54613, 55152 or 55345.
The reason is that the real line number is not stored although it is available from the xml reader

(the reason for the warning is another issue and should be handled separately)

### DIFF
--- a/addon/doxysearch/doxyindexer.cpp
+++ b/addon/doxysearch/doxyindexer.cpp
@@ -175,7 +175,7 @@ class XMLContentHandler
     };
 
     /** Handler for a start tag. Called for <doc> and <field> tags */
-    void startElement(const std::string &name, const XMLHandlers::Attributes &attrib)
+    void startElement(const std::string &name, const XMLHandlers::Attributes &attrib, const int lineNr)
     {
       m_data="";
       if (name=="field")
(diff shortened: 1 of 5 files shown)


doxygen:
Hi Albert, can you attach the tag file that gives this error? I'd like to reproduce the problem and try to solve it using XMLLocator instead of passing an extra lineNr parameter.

albert-github:
I've made a small example showing the problem (artificial, based on the actual TAO code but a bit smaller).
In the old version we see
```
tst.tag:1: error: Internal inconsistency: namespace in IDL not module, library or constant group
```
though this is the incorrect line and should be (as it is in the corrected version):
```
tst.tag:28: error: Internal inconsistency: namespace in IDL not module, library or constant group
```

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6391345/example.tar.gz)

doxygen:
@albert-github I've just pushed a slightly adapted implementation using the `m_locator` object for the line number.
Please verify that it works the same as your proposal.

albert-github:
Looks OK to me and is a bit simpler as  now the info is available at the "source.

Wouldn't it be better to initialize the `int linenr` (e.g. to -1 or 0) in the tagreader.cpp` at the lines 93 and 108.

PR #8427, opened 2021-03-13
Fix bug linking C++ anonymous namespace

lcarlier:
When using the \copydoc or \copybrief command, a variable declared
inside an anonymous workspace couldn't be fixed.

This patch resolves this issue.

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -734,7 +734,21 @@ static bool findDocsForMemberOrCompound(const char *commandName,
   // for symbols we need to normalize the separator, so A#B, or A\B, or A.B becomes A::B
   cmdArg = substitute(cmdArg,"#","::");
   cmdArg = substitute(cmdArg,"\\","::");
-  cmdArg = substitute(cmdArg,".","::");
+  static bool extractAnonNs = Config_getBool(EXTRACT_ANON_NSPACES);
+  if (extractAnonNs &&
+      cmdArg.startsWith("anonymous_namespace{")
+      )
+  {
+    int rightBracePos = cmdArg.find("}", std::strlen("anonymous_namespace{"));
+    QCString leftPart = cmdArg.left(rightBracePos + 1);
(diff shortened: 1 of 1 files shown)


albert-github:
Is this related to: https://stackoverflow.com/questions/14166416/doxygen-c-how-to-link-to-anonymous-namespace-variables/66617323#66617323

lcarlier:
Hello,
yes it is.
I forgot to attach to this thread a small reproduction scenario whichs doesn't work without my patch.

[doxybug.zip](https://github.com/doxygen/doxygen/files/6136341/doxybug.zip)
Cheers,
Laurent

lcarlier:
I just noticed that version 1.8.13 didn't have that issue so this bug fix is actually a regression.

albert-github:
Based on the comment https://github.com/doxygen/doxygen/pull/8427#issuecomment-798978319 I checked the different versions and between 1.8.15 (no warning) and 1.8.16 (warning present) the problem occurred.

The first commit with the warning is from pull request #7042:
```
commit 67cd1a3bf4981fd755ada6eeb85622b46638a749
Date:   Fri Jun 7 16:34:52 2019 +0200
    Bug 632754 - The \copydoc Command Requires The Use of C++ Syntax in C# Code
    Replace the scope separators to the default scope separators (::) of doxygen.
```
(and merged as ade74102d869ce21bef28b60deead0bc4e63bb36, though the initial commit already gave the problem).

The problem looks like a side effect of the replacement, might happen at other, not yet found, places as well.

albert-github:
Interesting is also the fix for the issue #7692 regarding file names (and the resulting implementation and changes). It looks like that the anonymous namespace has not been considered here although it also contains a filename.

lcarlier:
Indeed, the issue is that `anonymous_namespace{file.h}::variable` is replaced to `anonymous_namespace{file::h}::variable` and is then not found in the set.

I didn't check all the use case but this patch can be handy for providing a beginning of fix.

albert-github:
Looks like another case is present #8437

PR #8429, opened 2021-03-14
Warnings regarding Internal inconsistency: namespace in IDL

albert-github:
In the ACE TAO package we find a number of warnings like:
```
.../ACE_TAO/ACE/html/libtao-doc/security/TAO_Security.tag:1: error: Internal inconsistency: namespace in IDL not module, library or constant group
.../ACE_TAO/TAO/orbsvcs/orbsvcs/CosEventChannelAdmin.idl:30: error: Internal inconsistency: namespace in IDL not module, library or constant group
```

There are a number of reasons for these warnings a.o. IDL modules not handled properly but seen as namespaces and when type is not an IDL modules a warning is given. This happens especially when the module is documented with `@namespace`, this can partly be solved by testing the file type but when in a general file (out of place documentation) this does not work.
- introducing new commands to document IDL specific "namespaces"
- introducing and handling of IDL specific "namespaces" in tag files.

(A problem remains when we have e.g. an IDL module and in a C++ file there is a namespace with the same name.)

### DIFF
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -118,7 +118,10 @@ documentation:
 \refitem cmdhideinitializer \\hideinitializer
 \refitem cmdhtmlinclude \\htmlinclude
 \refitem cmdhtmlonly \\htmlonly
+\refitem cmdidlconstants \\idlconstants
 \refitem cmdidlexcept \\idlexcept
+\refitem cmdidllibrary \\idllibrary
+\refitem cmdidlmodule \\idlmodule
 \refitem cmdif \\if
 \refitem cmdifnot \\ifnot
 \refitem cmdimage \\image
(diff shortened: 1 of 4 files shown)


albert-github:
It is to be decided whether the pull request #8431 should be honored or the pull request #8429.
I think the pull request #8429 is better and more complete but also more complex and introduces some new doxygen commands.
The HTML output for the ACE TAO project was for both versions identical, just the generated tag file were (as expected) different.

doxygen:
@albert-github I opted for the simpler fix. Don't think IDL is used enough to add extra commands for it.

albert-github:
Based on the decision to use the simple fix from #8431 this proposed pull request is obsolete.

PR #8442, opened 2021-03-19
Incorrect handling of JAVA style code statement in preprocessor

albert-github:
When we have a program like:
```
/**
 * {@code
 * uspoof_setChecks(USPOOF_ALL_CHECKS & ~USPOOF_CONFUSABLE);
 * }
 *
 */
U_CAPI void U_EXPORT2
uspoof_setChecks(USpoofChecker *sc, int32_t checks, UErrorCode *status);

/**
 * \cond
 */
U_DEFINE_LOCAL_OPEN_POINTER(LocalUSpoofCheckerPointer, USpoofChecker, uspoof_close);
/** \endcond */
```
In the console output this results in some empty lines from the commentcnv lexer.

The problem comes from the preprocessor that doesn't recognize `{@code` as a JAVA style code statement that should end at the corresponding `}` and thus we remain in a code section. and thus not handling the `\cond` statement.
In the commentcnv lexer the `{@code` is handled as a special case and this is done in the preprocessor now as well.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6172066/example.tar.gz)

(Found by Fossies in package icu4c-69)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -288,6 +288,7 @@ struct preYY_state
   bool               insideFtn      = false;
   bool               isSource       = false;
 
+  int                javaBlock      = 0;
   yy_size_t          fenceSize      = 0;
   bool               ccomment       = false;
   QCString           delimiter;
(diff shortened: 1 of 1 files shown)


albert-github:
Superseded by #8878

PR #8444, opened 2021-03-20
Automatically create multi-layer subdirectories

albert-github:
When have a Doxyfile like:
```
QUIET=YES
OUTPUT_DIRECTORY = tst/lev1
HTML_OUTPUT = html_1/h_lev1
LATEX_OUTPUT = latex_1/l_lev1/l_lev2
```
we get the warning:
```
error: tag OUTPUT_DIRECTORY: Output directory 'tst/lev1' does not exist and cannot be created
```

In the previously used QDir  software there was only a `mkdir` routine and not a `mkpath` routine (as in newer Qt versions).
With the new implementation of the QDir through the ghc filesystem it is, easily, possible to create a `mkpath` by means of a `create_subdirectories`

Note: it is not necessary to replace all `mkdir` by `mkpath` as at some places it is guaranteed that only one level of directories can be missing (but it won't hurt).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6175412/example.tar.gz)


Ryanf55:
I am a fan of this PR, it solves an issue with generating logs to certain directories when there are multiple directories. Anything holding up merge?

albert-github:
Based on https://github.com/doxygen/doxygen/issues/10866#issuecomment-2118710447 and https://github.com/doxygen/doxygen/issues/10866#issuecomment-2118713980 I've just pushed a proposed patch, pull request #10876

Also seen the above #8444 is obsolete.

PR #8452, opened 2021-03-22
Regression, incorrect creation of path

albert-github:
Running the generation of the doxygen internal documentation I noticed the message:
```
Notice: Output directory 'doxygen_docs' does not exist. I have created it for you.
```
although the directory did exist.
The result was `doxygen_docs/doxygen_docs` but should have been the absolute path.

### DIFF
--- a/src/dir.cpp
+++ b/src/dir.cpp
@@ -218,7 +218,7 @@ std::string Dir::filePath(const std::string &path,bool acceptsAbsPath) const
   }
   else
   {
-    result = (p->path / path).string();
+    result = currentDirPath() + "/" + path;
   }
   correctPath(result);
   return result;


doxygen:
@albert-github If I look at the `QDir::filePath()` implementation then it does not use `currentDirPath()` as the base... So I'm not sure that this fix is correct and will not cause other side effects.

albert-github:
I had a look at the code of `QDir.cpp` and `dir.cpp` and most likely the difference comes from:

**dir.cpp**
```
bool Dir::exists(const std::string &path,bool acceptsAbsPath) const
{
  std::string result = filePath(path,acceptsAbsPath);
  std::error_code ec;
  bool exist = fs::exists(fs::path(result),ec);
  return !ec && exist;
}

bool Dir::exists() const
{
  return exists(p->path.string());
}

```

**QDir.cpp**
```
bool QDir::exists() const
{
    QFileInfo fi( dPath );
    return fi.exists() && fi.isDir();
}

```

maybe the statemens
```
std::string result = filePath(path,acceptsAbsPath);
```
should not be present:
```
bool Dir::exists(const std::string &path,bool acceptsAbsPath) const
{
  std::error_code ec;
  bool exist = fs::exists(fs::path(path),ec);
  return !ec && exist;
}
```

doxygen:
@albert-github can you check if the problem is solved with https://github.com/doxygen/doxygen/commit/ff2b8d933b3adf499ddaea07b53a8db000611ab6

albert-github:
As far as I can see this solution does solve the problem.

PR #8460, opened 2021-03-27
Environment variabelen in changelog

albert-github:
The environment variables in the changelog were replaced by their values and this was not the intention in the changelog
By using the HTML Entity here for the dollar sign this can be prevented.


doxygen:
@albert-github I think a more readable alternative would be to use `\$` to escape the dollar sign. Any reason why this would not work?

albert-github:
The escape won't worki in this case:
- it is in `\htmlonly` so unprocessed
- when using it we will only see the `\`

PR #8463, opened 2021-03-28
Write out used settings when generating XML output

albert-github:
When generating XML output is is afterwards unknown which settings have been used to generate the XML output, e.g. whether `EXTRACT_PRIVATE` was set or not as the XML output tries to write out all information contained in the sources and not to limit it.
By writing out a Doxyfile.xml with all the used settings this can be overcome.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -61,6 +61,11 @@ namespace Config
    */
   void compareDoxyfile(std::ostream &t);
 
+  /*! Writes a the used settings of the current configuartion as XML format
+   *  to stream \a t.
+   */
+  void writeXMLDoxyfile(std::ostream &t);
+
   /*! Parses a configuration file with name \a fn.
    *  \returns TRUE if successful, FALSE if the file could not be
    *  opened or read.
--- a/src/configimpl.h
+++ b/src/configimpl.h
@@ -77,14 +77,16 @@ class ConfigOption
   protected:
     virtual void writeTemplate(std::ostream &t,bool sl,bool upd) = 0;
     virtual void compareDoxyfile(std::ostream &t) = 0;
+    virtual void writeXMLDoxyfile(std::ostream &t) = 0;
     virtual void convertStrToVal() {}
     virtual void emptyValueToDefault() {}
     virtual void substEnvVars() = 0;
     virtual void init() {}
+    virtual bool isDefault() {return true;}
 
-    void writeBoolValue(std::ostream &t,bool v);
-    void writeIntValue(std::ostream &t,int i);
-    void writeStringValue(std::ostream &t,const QCString &s);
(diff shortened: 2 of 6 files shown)


albert-github:
- Why should a switch be used to control this feature? There is just 2 extra files `Doxyfile.xml` and `doxyfile.xsd` written as reference.
- the XML output provides a broad overview about all that is available in the code that is documented and this means e.g. also the static functions are mentioned in the XML output even when `EXTRACT_STATIC=NO`, in this way the user has the possibility to filter the output to his liking (which is the purpose of the XML output).

tishion:
This change will confuse the users, because it broke the understanding of `config` and `output` of doxygen from the users.

Generally the workflow with doxygen looks like 

1. we feed the `input`(source code) to doxygen 
2. we compose a `config` file to instruct doxygen how to work(generating, rendering)
3. doxygen generates `content`  with the provided `config
4. doxygen render the `content` to different `output` format

I believe the configuration options EXTRACT_XXXXs were designed to work on the `content` generation phase. That means for a specified `config` it should generate the same `content` no matter which kind of `output` format it will be rendered to.

In short, this change coupled the `output` rendering and the `content` generation.

>the XML output provides a broad overview about all that is available in the code that is documented and this means e.g. also the static functions are mentioned in the XML output even when EXTRACT_STATIC=NO, in this way the user has the possibility to filter the output to his liking (which is the purpose of the XML output).

I totally understand what you are trying to do, you just treat the xml format as an intermediate payload of `content` I mentioned above (it is more like the intermediate language of the compiler). This is reasonable,  but why we must make it **unconfigurable**?

From the doxygen release with this feature we have to choose one of the two facts below:
1. xml is no longer a configurable output exposed by doxygen
2. the config options EXTRAC_XXXXX will no longer work on xml output

I still believe that without this feature we also could achieve the purpose proposed in this PR by composing a correct config file, but with this feature, we have to do some extra work to get desired xml output because it ignore the config file~

albert-github:
I think we have here a small interpretation problem. The XML output has always been intended as an intermediate format and "dump"  all possible information and not take all selection possibilities from doxygen (like `EXTRACT_...`) into consideration. In the end user version documentation format (like HTML, LaTeX, RTF, ...) these setting are taken into consideration.
With this pull request nothing has been changed for the different output formats, just that 2 extra files are provided to give the user an idea which options have been used (like `PREDEFINED, ALIASES` etc.).

What you would like to have, as far as I understand, is that the XML output has also the possibility to follow all the rules that are now just to select the information as done for e.g. HTML, LaTeX, RTF ... output. 
If that is the case we would indeed need a setting that would enable these kind of rules and this has to be implemented. For that a new, enhancement, issue is necessary.

PR #8481, opened 2021-04-06
Determine ghostscript directory (Windows)

albert-github:
Determine the ghostscript directory in a more system and version independent way for Windows.


albert-github:
Rationale: I was not happy with the need to change the GitHub actions script each time the version of ghostscript changed, so I came up with this idea.

PR #8484, opened 2021-04-07
Inconsistency between files with underscore or in sub directories

albert-github:
When having the files:
```
docs_dev_git.md
docs/dev/git.md
```
with at the beginning something like:
```
# The header
```
we get the waring:
```
docs_dev_git.md:1: warning: multiple use of page label 'md_docs_dev_git', (other occurrence: <md_docs_dev_git>, line: 1)
```

This due to the fact that the "/" is translated to a "\_" and thus clashes.
This is fixed by means of converting an existing "\_" in the filename etc. to "\_\_"

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6272394/example.tar.gz)


doxygen:
@albert-github Or maybe better to use `convertNameToFile` like is done for other symbols?

albert-github:
I don't think so as here we are talking over `Id`s and not about file names as in `convertNameToFile`. In `convertNameToFile` also things are handled regarding the settings like `SHORT_NAMES` and `CREATE_SUBDIRS`. Problem might be that we get very long `Id`s and this again would be reflected in the resulting file name, see the changed line in testing/055_markdown.md  and this might lead to problems (path length).
Furthermore I tried to stay as close as possible to the original implementation, it might though be an idea to use `escapeCharsInString` for this (not tested by me).

Issue #8511, opened 2021-04-20
Java: linebreak after @link can cause wrong parsing of subsequent doc


Stewori:
**Describe the bug**
Processing Java code
```
public class A {

    /**
     * A linebreak in a link {@link
     * A}
     *
     * @param  <S> can cause trouble
     */
    public static void test(Class<S> cls) {}
}
```
makes doxygen invent an ``<s>`` tag in xml output:
```
<detaileddescription>
<para>A linebreak in a link {<ref refid="classA" kindref="compound">}  <s> can cause trouble </s></ref></para>
</detaileddescription>
```
This is triggered by the linebreak in `@link`. If one writes it as a one-liner ``* A linebreak in a link {@link A}``
 one gets the expected behavior described below.

**Expected behavior**
I expect the xml output to be the same as if the `@link` was written as a one-liner:
```
<detaileddescription>
<para>A linebreak in a link <ref refid="classA" kindref="compound">A</ref></para>
<para><parameterlist kind="param"><parameteritem>
<parameternamelist>
<parametername>&lt;S&gt;</parametername>
</parameternamelist>
<parameterdescription>
<para>can cause trouble </para>
</parameterdescription>
</parameteritem>
</parameterlist>
</para>
</detaileddescription>
```

**To Reproduce**
Find a Java file and doxygen config attached. The issue is demonstrated like mentioned above.
After running doxygen on the example, the files in the subfolder ``xml`` contain the snippets shown above.
[issue_param_s.zip](https://github.com/doxygen/doxygen/files/6344531/issue_param_s.zip)

**Version**
Tested with doxygen clone from 2021-04-11:
```
/blah/doxygen/build/bin/doxygen -v
1.9.2 (e9185650216f3c3fa97759caf67ee79db66cb5e1)
```

`lsb_release -a`:
```
LSB Version:	core-9.20160110ubuntu0.2-amd64:core-9.20160110ubuntu0.2-noarch:security-9.20160110ubuntu0.2-amd64:security-9.20160110ubuntu0.2-noarch
Distributor ID:	LinuxMint
Description:	Linux Mint 18 Sarah
Release:	18
Codename:	sarah
```

**Additional context**
[...]

PR #8512, opened 2021-04-20
issue #8511 Java: linebreak after @link can cause wrong parsing of subsequent doc

albert-github:
Allowing also a `\n` after the @link.

### DIFF
--- a/src/doctokenizer.l
+++ b/src/doctokenizer.l
@@ -610,7 +610,7 @@ RCSID "$"("Author"|"Date"|"Header"|"Id"|"Locker"|"Log"|"Name"|"RCSfile"|"Revisio
                          g_token->indent = computeIndent(text,dotPos);
                          return TK_ENDLIST;
                        }
-<St_Para>"{"{BLANK}*"@link"/{BLANK}+ {
+<St_Para>"{"{BLANK}*"@link"/{WS}+ {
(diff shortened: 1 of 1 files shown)


Stewori:
I see you fixed it specifically for `@link`. Would it make sense to apply this adjustment to other (supported) `@xyz` tags similarly? At least to the inline ones, i.e. those of the form `text {@xyz value} more text`.

albert-github:
> Would it make sense to apply this adjustment to other (supported) `@xyz` tags similarly? At least to the inline ones, i.e. those of the form text `{@xyz value}` more text.

That would make sense, though at first glance I didn't see any others with signature `{@xyz value}`.

Stewori:
I used to think that e.g. `{@code ...}` is supported by doxygen. However I cannot spot it in `doctokenizer.l`. Are some javadoc tags implemented at a different place?

albert-github:
In `commentcnv.l` I found:
```
<CComment>"{@code"/[ \t\n] 
```
so this looks OK to me.

Maybe you can create a small example (with the original issue of the `{@code}`)

Stewori:
I modified the original example by replacing `@link` with `@code` and can confirm that the issue does not occur for `@code`. Further, I can confirm that this PR fixes the issue. Thanks for fixing! Get it merged please :)

Issue #8495, opened 2021-04-11
Java: members entirely missing if generic parameter extends another generic


Stewori:
**Describe the bug**
Processing a Java class
```
public class A<E extends Vector<String>> {
    public String name;
}
```
causes in xml members to be missing entirely:
```
<compounddef id="classA_3_01E_01extends_01Vector" kind="class" language="Java" prot="public">
    <compoundname>A&lt; E extends Vector</compoundname>
    <templateparamlist>
      <param>
        <type>String</type>
      </param>
    </templateparamlist>
    <briefdescription>
    </briefdescription>
    <detaileddescription>
    </detaileddescription>
    <location file="A.java" line="3" column="7" bodyfile="A.java" bodystart="3" bodyend="5"/>
    <listofallmembers>
    </listofallmembers>
  </compounddef>
```
This does also affect html output. (Other outputs not tested.)
Both in html and xml it only affects the class compound, not the file compound. However, using the file compounds has other disadvantages and collecting information from both is fiddly.

Note that this issue does not occur if one uses `Vector` without parameter: `public class A<E extends Vector>`.
It does occur e.g. for `public class A<E extends Vector<E>>`. Of course `Vector` is just an exemplary choice. Any other generic would cause the issue as well.
Also note that the compound name is messy.

**Expected behavior**
I expect the xml output to contain members much in the same way as the output of
```
public class A<E extends Vector> {
    public String name;
}
```
would do, yet listing `Vector<String>` as type constraint:
```
<compounddef id="classA" kind="class" language="Java" prot="public">
    <compoundname>A</compoundname>
    <templateparamlist>
      <param>
        <type>E</type>
        <typeconstraint>Vector&lt;String&gt;</typeconstraint>
      </param>
    </templateparamlist>
      <sectiondef kind="public-attrib">
      <memberdef kind="variable" id="classA_1a8f0b657e06a68f6f696026386a5828dd" prot="public" static="no" mutable="no">
        <type>String</type>
        <definition>String A&lt; E extends Vector&lt;String&gt; &gt;.name</definition>
        <argsstring></argsstring>
        <name>name</name>
        <briefdescription>
        </briefdescription>
        <detaileddescription>
        </detaileddescription>
        <inbodydescription>
        </inbodydescription>
        <location file="A.java" line="4" column="19" bodyfile="A.java" bodystart="4" bodyend="-1"/>
      </memberdef>
      </sectiondef>
    <briefdescription>
    </briefdescription>
[...]

PR #8534, opened 2021-05-04
issue #8495 Java: members entirely missing if generic parameter extends another generic

Stewori:
Proposed fix for #8495. See discussion therein.

### DIFF
--- a/src/defargs.l
+++ b/src/defargs.l
@@ -471,6 +471,7 @@ CPPC  "/\/"
                                            {
                                              yyextra->curTypeConstraint.resize(0);
                                              yyextra->lastExtendsContext=YY_START;
+                                             assert(yyextra->argSharpCount == 0);
                                              BEGIN(ReadTypeConstraint);
                                            }
                                         }
@@ -500,9 +501,25 @@ CPPC  "/\/"
 <CopyArgRound,CopyArgRound2,CopyArgSharp,CopyArgCurly>.  {
 					  *yyextra->copyArgValue += *yytext;
 					}
-<ReadTypeConstraint>[,)>]               {
+<ReadTypeConstraint>[,)]               {
                                           unput(*yytext);
                                           BEGIN(yyextra->lastExtendsContext);
+                                       }
+<ReadTypeConstraint>">"                {
+                                          if (yyextra->argSharpCount > 0)
+                                          {
+                                            yyextra->curTypeConstraint+=yytext;
+                                            --yyextra->argSharpCount;
+                                          }
+                                          else
+                                          {
+                                            unput(*yytext);
+                                            BEGIN(yyextra->lastExtendsContext);
+                                          }
+                                        }
+<ReadTypeConstraint>"<"                 {
+                                          yyextra->curTypeConstraint+=yytext;
+                                          ++yyextra->argSharpCount;
                                         }
 <ReadTypeConstraint>.                   {
                                           yyextra->curTypeConstraint+=yytext;
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -1049,7 +1049,7 @@ static void addClassToContext(const Entry *root)
       }
     }
     std::unique_ptr<ArgumentList> tArgList;
-    if ((root->lang==SrcLangExt_CSharp || root->lang==SrcLangExt_Java) && (i=fullName.findRev('<'))!=-1)
+    if ((root->lang==SrcLangExt_CSharp || root->lang==SrcLangExt_Java) && (i=fullName.find('<'))!=-1)
     {
       // a Java/C# generic class looks like a C++ specialization, so we need to split the
       // name and template arguments here


src/doxygen.cpp
```diff
@@ -1049,7 +1049,7 @@ static void addClassToContext(const Entry *root)
       }
     }
     std::unique_ptr<ArgumentList> tArgList;
-    if ((root->lang==SrcLangExt_CSharp || root->lang==SrcLangExt_Java) && (i=fullName.findRev('<'))!=-1)
+    if ((root->lang==SrcLangExt_CSharp || root->lang==SrcLangExt_Java) && (i=fullName.find('<'))!=-1)
     {
```
albert-github, line 1053: Is this fix also valid for `C#` or are apply there different rules?

src/defargs.l
```diff
@@ -500,9 +501,25 @@ CPPC  "/\/"
...
+                                          }
+                                          else
+                                          {
+                                            unput(*yytext);
+                                            BEGIN(yyextra->lastExtendsContext);
+                                          }
+                                        }
+<ReadTypeConstraint>"<"                 {
+                                          yyextra->curTypeConstraint+=yytext;
+                                          ++yyextra->argSharpCount;
```
albert-github, line 522: What happens when the uses has an error in the number of `<` `>` pairs e.g. maore `>` than `<` (not a valid program but still ...)?

src/defargs.l
```diff
@@ -471,6 +471,7 @@ CPPC  "/\/"
                                            {
                                              yyextra->curTypeConstraint.resize(0);
                                              yyextra->lastExtendsContext=YY_START;
+                                             assert(yyextra->argSharpCount == 0);
```
albert-github, line 474: Not sure why an assert is necessary here? or does this have to do with my remark around the lines 503.? Wouldn't it be better to give a real doxygen warning?

src/defargs.l
Stewori: What behavior would you recommend in this case? Note that the status quo is rather random behavior for examples described in the issue. I guess the behavior of doxygen on invalid programs is in general undefined. However I admit that we may come up with a friendly behavior for this particular type of syntax issue. E.g. raise a warning if a premature `[(){}]` is encountered (notice that the entrance to `<ReadTypeConstraint>` is already guarded as Java-only)?

src/doxygen.cpp
Stewori: Honestly, I hoped you would know. I guess the ``findRev`` was there for a reason and this might require to be guarded as Java-only. What I am rather confident about is that this is the proper way to do it for Java. I can guard it if you think that's required.

src/defargs.l
Stewori: I believe ``argSharpCount != 0`` can never happen here. Just placed the ``assert` to see whether it would go through the ci. The line can be removed.

src/doxygen.cpp
albert-github: Probably safest is to guard it (I'm not a C# expert either)

Stewori: Unfortunately I am not much familiar with ``C#``. I am aware that it has somewhat similar generics like Java but probably with some differences. I already suspected the issue might apply to ``C#`` as well but do not know for sure. The opinion of someone familiar with ``C#`` would be helpful. After a quick search it appears to me that it is not totally trivial to comprehend the full set of rules from ``C#`` documentation.

albert-github: The latest standard,is to the best of my knowledge, ECMA-334, 5th Edition / December 2017. After this there are changes to the standard as Specification Proposals but they have not bee officially standardized (see e.g. https://en.wikipedia.org/wiki/C_Sharp_(programming_language)).

src/defargs.l
albert-github: Actually I'm a bit confused as I see that `ReadTypeConstraint` is entered as soon as the rule `<ReadFuncArgType,ReadFuncArgPtr>"extends"` is satisfied (for Java) and the count is increased as soon as `<` is seen and decreased when `>` is seen, but the count is at that moment still > 0 and when no `,` or `)` is found the user remains in the `ReadTypeConstraint` and will leave it as soon as it sees a `,` or `)` but what will happen when the user is in a `<` ..`>` part and sees a `,` or `)` or is this not possible?

Stewori: (for reference: discussion on this continued in #8495)

src/defargs.l
```diff
@@ -140,6 +141,9 @@ CPPC  "/\/"
 %x	FuncQual
 %x	ReadDocBlock
 %x	ReadDocLine
+%x  ReadAnnotation
+%x  ReadStringLiteral
+%x  ReadEscape
```
Stewori, line 146: Sorry for the misalignment; did you notice the blank between ``%x`` and the identifier above my edits is actually a tab, which is why it looked good in my editor. Would it be okay if I aligned the whole block consistently? Or was there a reason for the original misalignment and for using the tabs?

albert-github: In principle there should not be any tabs in the code but they are a remainder of the past (same accounts for white space at the end of a line, shouldn't be present). In a lot of files this has already been corrected by @doxygen with a special procedure. Unfortunately it is still present in a number of files (and giving headaches over and over again due to inconsistent tab settings, especially in Python where whitespace has, unfortunately, a special meaning for defining blocks).

src/defargs.l
```diff
@@ -501,9 +517,54 @@ CPPC  "/\/"
...
+                                       }
+<ReadAnnotation>"\""                   {
+                                          BEGIN(ReadStringLiteral);
+                                       }
+<ReadAnnotation>.                      {
+                                          // skip it for now
+                                       }
+<ReadStringLiteral>"\\"                {
+                                          BEGIN(ReadEscape);
+                                       }
```
albert-github, line 543: Why not combine this with the rule `<ReadEscape>.` to `<ReadStringLiteral>"\\."` (like it is also done for `<CopyArgString>` )?

src/defargs.l
```diff
@@ -501,9 +517,54 @@ CPPC  "/\/"
...
+<ReadAnnotation>{B}+                   {
+                                          if (yyextra->argRoundCount == 0)
+                                          {
+                                            BEGIN(yyextra->readAnnotationContext);
+                                          }
+                                       }
+<ReadAnnotation>"\""                   {
+                                          BEGIN(ReadStringLiteral);
+                                       }
```
albert-github, line 537: Does for annotations also the single quote possibility exist?

src/defargs.l
Stewori: Okay. Good idea.

src/defargs.l
Stewori: Turns out there are still *plenty* of tabs in ``defargs.l``. I think replacing all of them would pretty much contaminate the changeset of this PR and should be a separate effort. So, for now I just fixed the alignment of the lex declaration block.

src/defargs.l
Stewori: No. The string literals in annotations are just usual Java string literals. Java *only* uses double quotes as string boundaries, much like C and friends do (and unlike Python does).

src/defargs.l
Stewori: I found the "\\." slightly misleading for escape sequences because they can be longer than one character, e.g. for unicode ``\uXXXX``. So I went with ``<ReadStringLiteral>"\\\""`` which I think is more on spot for the purpose of that rule. Anyway, just a matter of taste I guess.

src/defargs.l
albert-github: Correct, the big change should be a separate changeset and not be part of a "real" fix.

src/defargs.l
albert-github: Probably yes, but I don't really see the problem with the `\uXXXX` as here the `\\.` rule would eat the `\u` and the normal `.` rule `X`s. I think both solutions would be valid in this case.

Stewori: Yes, sure both would work. That's why I wrote it's a matter of taste. I just found the `\\\"` rule more explicit for it's purpose.

src/defargs.l
Stewori: Oh sorry, I forgot there can also be character literals, which indeed use a single quote. However, there can only be one single character in between (or an escape sequence), so it should be sufficient to handle the rules ``<ReadAnnotation>"'('"`` and ``<ReadAnnotation>"')'"``, ``<ReadAnnotation>"'\"'"`` and probably also ``<ReadAnnotation>"'\\\"'"``. It should be possible to combine the first three into ``<ReadAnnotation>"'"[()\"]"'"``. I will add them soon.

albert-github: Not sure but don't forget about the `'\''` i.e. "the escaped single quote character literal".

Stewori: I thought about that, but it should be covered by the general ``<ReadAnnotation>.`` rule. Since we do not enter a new state on ``'`` (unlike we do for ``"``), no extra care should be required to avoid premature return from such a state. The ``"`` must be considered to avoid entering ``<ReadStringLiteral>`` falsely, which cannot happen for ``'``.

albert-github: But couldn't in that case `'"`'` be problematic?

Stewori: That's covered by ``<ReadAnnotation>"'\"'"``. The overall rule I added just now is

```<ReadAnnotation>"'"[()\"]"'"|"'\\\"'"```. It should catch the Java literals ``'('``, ``')'``, ``'"'``, ``'\"'`` and I confirmed this with some simple tests.

asteele0:
@albert-github what ended up happening here? It looks like the PR was completed?
I'm also running into this issue, so it would be nice to have fixed.

Issue #8541, opened 2021-05-10
Javadoc: external links to classes in java.lang are not resolved


Stewori:
**Describe the bug**
Assume a tagfile of the module ``java.base`` is located at ``/path_to_doxygen_output_of_java.base_1.14/java.base_1.14_tagfile.xml`` and declared in the Doxygen config file under ``TAGFILES``.
Processing Java code
```
import java.util.Vector;

public class Test {
    public String a;
    public Vector b;
}
```
results in the following xml:
```
...
<compounddef id="classTest" kind="class" language="Java" prot="public">
    <compoundname>Test</compoundname>
      <sectiondef kind="public-attrib">
      <memberdef kind="variable" id="classTest_1a4dfe3f6d6cbfd2814da6d802e9d916c3" prot="public" static="no" mutable="no">
        <type>String</type>
        <definition>String Test.a</definition>
        <argsstring></argsstring>
        <name>a</name>
        <briefdescription>
        </briefdescription>
        <detaileddescription>
        </detaileddescription>
        <inbodydescription>
        </inbodydescription>
        <location file="Test.java" line="4" column="19" bodyfile="Test.java" bodystart="4" bodyend="-1"/>
      </memberdef>
      <memberdef kind="variable" id="classTest_1a184487560da0c075f32b0cbd2261afe2" prot="public" static="no" mutable="no">
        <type><ref refid="classjava_1_1util_1_1Vector" kindref="compound" external="/path_to_doxygen_output_of_java.base_1.14/java.base_1.14_tagfile.xml">Vector</ref></type>
        <definition>Vector Test.b</definition>
        <argsstring></argsstring>
        <name>b</name>
        <briefdescription>
        </briefdescription>
        <detaileddescription>
        </detaileddescription>
        <inbodydescription>
        </inbodydescription>
        <location file="Test.java" line="5" column="19" bodyfile="Test.java" bodystart="5" bodyend="-1"/>
      </memberdef>
      </sectiondef>
...
```
Notice how ``Vector`` is properly linked to its external resource while ``String`` is not. This is because Doxygen does not consider that the package ``java.lang`` is always implicitly imported. The same issue occurs without external linking if one processes the source of the ``java.base`` module with Doxygen. No internal links to classes in ``java.lang`` are created although there should be plenty of them.

**Expected behavior**
I expect the xml output to be the same as if one had prepended the line ``import java.lang.*`` to every ``.java`` file.
[...]

PR #8545, opened 2021-05-11
issue #8541 Javadoc: external links to classes in java.lang are not resolved

albert-github:
Adding the default packages from "java.lang"
From the standard "The Java® Language, Specification, Java SE 16 Edition" ,Chapter 7, Packages and Modules:

> Code in a compilation unit automatically has access to all classes and interfaces
> declared in its package and also automatically imports all of the public classes and
> interfaces declared in the predefined package java.lang.

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -7563,8 +7563,8 @@ static void parseMain(yyscan_t yyscanner,
   }
   else if ( yyextra->insideJava )
   {
-    QCString scope="java.lang.*";
-    yyextra->current->name=removeRedundantWhiteSpace(substitute(scope.left(scope.length()-1),".","::"));
+    QCString scope="java.lang";
+    yyextra->current->name=removeRedundantWhiteSpace(substitute(scope,".","::"));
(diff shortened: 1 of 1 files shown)


src/scanner.l
```diff
@@ -7561,6 +7561,16 @@ static void parseMain(yyscan_t yyscanner,
   {
     BEGIN( FindMembersPHP );
   }
+  else if ( yyextra->insideJava )
+  {
+    QCString scope="java.lang.*";
+    yyextra->current->name=removeRedundantWhiteSpace(substitute(scope.left(scope.length()-1),".","::"));
```
doxygen, line 7567: Since scope is a constant this can be optimized to `yyextra->current->name="java::lang";` I think. 
I don't think a trailing `::` at the end is needed.

albert-github: A `::` at the end is not necessary, intention was to use make explicitly, by ".*` that all subpackages were used but that is not necessary.

Updated pull request

PR #8551, opened 2021-05-14
No specific way to set class attributes of a treeview link.

albert-github:
A link in a navtree / treeview has not a general class attached to it so its properties can only be changed by means of the global `a` tag.
There is a class attached to it but this contains the link reference so it is possible to highlight an item in the treeview when clicking in the text part of the documentation on a link.
A general class has been added to the navtree link.

### DIFF
--- a/templates/html/navtree.css
+++ b/templates/html/navtree.css
@@ -65,6 +65,9 @@
   padding:0px;
 }
 
+#nav-tree .link {
+}
+
 #nav-tree {
   padding: 0px 0px;
   background-color: #FAFAFF; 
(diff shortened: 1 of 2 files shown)


templates/html/navtree.css
```diff
@@ -65,6 +65,9 @@
   padding:0px;
 }
 
+#nav-tree .link {
+}
+
```
doxygen, line 70: I think the same can be already be achieved via a descendant selector, e.g. using

```.css
#nav-tree span.label a {
        background: red;
}
```
which will style all `a` tags under `span.label` inside `#nav-tree`.

albert-github: This is indeed the case, it is a little bit less clear (to me) but it works.

PR #8554, opened 2021-05-15
Extra settings for MathJax V3

albert-github:
For support of the `\eqref` command (used in CGAL) in MathJax V3 a small change has to be made (for a rationale see the discussion on https://groups.google.com/g/mathjax-users/c/oS0yQLb5BMk)

### DIFF
--- a/templates/html/header.html
+++ b/templates/html/header.html
@@ -2,7 +2,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
 <meta http-equiv="Content-Type" content="text/xhtml;charset=UTF-8"/>
-<meta http-equiv="X-UA-Compatible" content="IE=9"/>
+<meta http-equiv="X-UA-Compatible" content="IE=11"/>
 <meta name="generator" content="Doxygen $doxygenversion"/>
 <meta name="viewport" content="width=device-width, initial-scale=1"/>
 <!--BEGIN PROJECT_NAME--><title>$projectname: $title</title><!--END PROJECT_NAME-->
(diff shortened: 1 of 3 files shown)


templates/html/header.html
```diff
@@ -2,7 +2,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
 <meta http-equiv="Content-Type" content="text/xhtml;charset=UTF-8"/>
-<meta http-equiv="X-UA-Compatible" content="IE=9"/>
+<meta http-equiv="X-UA-Compatible" content="IE=11"/>
```
doxygen, line 5: @albert-github did you test the effect of this change on the HtmlHelp output? (remember the IE=9 line was added to make CHM output look properly, maybe IE=11 works as well, but it is good to check this).

albert-github: I took the current master and changed the header.html and htmlbase.tpl according to this PR and r5ecreated the chm documentation (from scratch i.e removing the chm, html and latex directories.
I didn't observer a problem.

@doxygen do you remember specifically where / why the `IE=9` was required?

doxygen: > @doxygen do you remember specifically where / why the `IE=9` was required?

Without it the output was processed in IE 7 mode if I recall correctly, which caused some CSS formatting to be ignored resulting in wrong/ugly output.

albert-github: In that case I don't think a higher version probably won't hurt.

As side notes:
- CHM generator is since a long time not maintained so we should think about what to do with this output format
- IE 11 is also nearly at the end of its life time (recently saw: Starting June 15, 2022, the IE11 desktop application will no longer be supported. See: https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge)

doxygen: I expect that we indeed need to drop support for CHM at some point in the not too distant future.

Issue #8563, opened 2021-05-21
Link not created and the Doxygen comment is placed in the output improperly


N0NB:
Our project https://github.com/Hamlib/Hamlib has long used Doxygen.  Earlier this year I took on the task of updating our Doxygen configuration and source file comments.  I have tried to post to the doxygen-user mailing list, but never receive a confirmation, so I am posting this here.

**Describe the bug**
As I was documenting some functions I noticed that I had some spurious text appearing. The text rendered in HTML is:

< '' – None

**Expected behavior**

That text not appear!

**Screenshots**
This online copy of the manual from the latest release shows the [output in the rig_get_mode() function](http://hamlib.sourceforge.net/manuals/4.2/group__rig.html#ga1281f19b50f2401fbfd69e7ddc561aa0).

Here is the [source code comment](https://github.com/Hamlib/Hamlib/blob/86d2b4ee6c07bb7dd2af70a51567f2bc86adf4e6/src/rig.c#L2207) to the above output.  Note, that this is not the section of code I've been working with, just the most egregious example.

I found the [offending text in one of our main header files](https://github.com/Hamlib/Hamlib/blob/86d2b4ee6c07bb7dd2af70a51567f2bc86adf4e6/include/hamlib/rig.h#L1163):

```
#define    RIG_MODE_NONE      0                         /*!< '' -- None */
```

Which immediately precedes a series of defines of bit shifted values such as:

```
#define    RIG_MODE_AM        CONSTANT_64BIT_FLAG (0)   /*!< \c AM -- Amplitude Modulation */
#define    RIG_MODE_CW        CONSTANT_64BIT_FLAG (1)   /*!< \c CW -- CW "normal" sideband */
#define    RIG_MODE_USB       CONSTANT_64BIT_FLAG (2)   /*!< \c USB -- Upper Side Band */
```

**To Reproduce**
I will see if I can work up an example in the coming days.  This project is large and complex and something elsewhere could be working in concert to cause this.

**Version**
1.9.1 packaged in the forthcoming Debian 11 release.

**Additional context**
I will assist in all ways possible.

albert-github:
>  I have tried to post to the doxygen-user mailing list, but never receive a confirmation, so I am posting this here.

wise decision

You already mentioned a few things:
- offending define
- the part where the problem occurs

this looks all OK.

- what happens when you use (as a workaround):  /*!< \c '' -- None */
- can you post already the Doxyfile you used?

albert-github:
[...]

PR #8566, opened 2021-05-24
issue #8563 Link not created and the Doxygen comment is placed in the output improperly

albert-github:
A small warning about the usage of "Putting documentation after members" in case of a define.

### DIFF
--- a/doc/docblocks.doc
+++ b/doc/docblocks.doc
@@ -288,10 +288,15 @@ Here is an example of the use of these comment blocks:
 
 \warning These blocks can only be used to document \e members and \e parameters.
          They cannot be used to document files, classes, unions, structs,
-         groups, namespaces and enums themselves. Furthermore, the structural 
+         groups, namespaces, defines and enums themselves. Furthermore, the structural 
          commands mentioned in the next section 
          (like <code>\\class</code>) are not allowed 
          inside these comment blocks.
+
+\warning Don't use this construct with a define as in that cases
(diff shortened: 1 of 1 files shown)


doc/docblocks.doc
```diff
@@ -288,11 +288,16 @@ Here is an example of the use of these comment blocks:
...
          They cannot be used to document files, classes, unions, structs,
-         groups, namespaces and enums themselves. Furthermore, the structural 
+         groups, namespaces, defines and enums themselves. Furthermore, the structural 
          commands mentioned in the next section 
          (like <code>\\class</code>) are not allowed 
          inside these comment blocks.
 
+\warning Don't use this construct with a define as in that cases
+         (when \ref cfg_macro_expansion "MACRO_EXPANSION" is set to `YES`)
+         at places where the define is used the defines is replaced including the comment
```
doxygen, line 298: I think we should use the term 'macro' not 'define' and 'the defines is' should then also be 'the macro is' (so without the s)

albert-github: No problem with the change, do you give it a go or should I do it (tomorrow)?

doxygen:
@albert-github I've changed the text and changed 'don't use' by 'beware', as for instance the following is a valid use case (but not for documenting `FIELD` itself):

```c
#define FIELD(x) int x; /**< docs for x */

/** An S */
struct S
{
  FIELD(x)
  FIELD(y)
  FIELD(z)
};
```

Issue #8574, opened 2021-05-28
Javadoc: no warning for unclosed quotations/unformatted text


Stewori:
**Describe the bug**
Processing a Java class with a missing ``"``
```
public class Test {

    /**
     * Some example text
     * "A, B "C" and "D", blah blah
     *
     * <p><a id="my_anchor" href="http://something">.
     *
     */
    public void x() {}
}
```
may cause messy doc with possibly vast ranges of verbatim html:
```
...
        <detaileddescription>
<para>Some example text &quot;A, B &quot;C&quot; and &quot;D&quot;, blah blah

&lt;p&gt;&lt;a id=&quot;my_anchor&quot; href=&quot;<ulink url="http://something">http://something</ulink>"&gt;. </para>
        </detaileddescription>
...
```
I am aware that this won't be considered a bug as there is a syntax flaw in the doc. However, there is currently no good way to deal with such a syntax issue; not even a warning is raised so one could detect it. AFAIK there is no flag to turn off the interpretation of ``"..."`` as unformatted text (please correct me if I overlook some solution).

**Expected behavior**

I expect Doxygen to raise some warning and ideally to offer some way to opt out of the interpretation of ``"..."`` as unformatted text.

**To Reproduce**
Find a Java file and doxygen config attached.
[example.zip](https://github.com/doxygen/doxygen/files/6561322/example.zip)

**Version**
Tested with doxygen clone from 2021-05-26:
```
/blah/doxygen/build/bin/doxygen -v
1.9.2 (da0a6b42c349bc06b2804a4b0e8f55c7690f5cf9)
```

`lsb_release -a`:
```
LSB Version:	core-9.20160110ubuntu0.2-amd64:core-9.20160110ubuntu0.2-noarch:security-9.20160110ubuntu0.2-amd64:security-9.20160110ubuntu0.2-noarch
Distributor ID:	LinuxMint
Description:	Linux Mint 18 Sarah
Release:	18
Codename:	sarah
```

**Additional context**
Typical Javadoc authors do not expect ``"`` to have a special syntactical meaning when used in text bodies, i.e. outside of html tags or code snippets. So they are not particularly alerted to always close a quotation properly (in doc comments).
Then, Javadoc is rather forgiving about a non-closed ``"`` in a text body. It will just be rendered as a plain ``"``, no closing is expected and the subsequent text is just normal. (Also keep in mind that -- unlikely though -- someone might use ``"`` to denote inch.)
[...]

PR #8581, opened 2021-06-01
Address #8574 by adding a warning and an opt-out flag for (unclosed) quoted text

Stewori:


src/doctokenizer.l
```diff
@@ -57,8 +57,10 @@ static yy_size_t g_inputPos = 0;
 static const char *g_inputString;
 static QCString g_fileName;
 static bool g_insidePre;
+static bool g_insideQuot;
 static int g_sharpCount=0;
-static bool g_markdownSupport=TRUE;
+static bool g_markdownSupport = TRUE;
+static bool g_quotSupport = FALSE;
```
Stewori, line 63: I guess, both ``g_markdownSupport`` and ``g_quotSupport`` shouldn't be initialized here since they are initialized in  ``doctokenizerYYinit``.

Stewori:
I can rename the new option ``QUOT_SUPPORT`` to a better name if you like (and come up with a better name).

src/docparser.cpp
```diff
@@ -1800,7 +1800,8 @@ static int internalValidatingParseDoc(DocNode *parent,DocNodeList &children,
 
   if (doc.isEmpty()) return retval;
 
-  doctokenizerYYinit(doc.data(),g_fileName,g_markdownSupport);
+  doctokenizerYYinit(doc.data(),g_fileName,g_markdownSupport,
+                     Config_getBool(QUOT_SUPPORT));
```
Stewori, line 1804: Alternatively I can handle ``quotSupport`` like ``markdownSupport`` all the way through, i.e. create ``g_quotSupport`` and so on. However, that would involve various further changes throughout the code base. So far I did it the easy way to await your recommendation.

PR #8610, opened 2021-06-21
Fix variable assignment to itself in util.cpp

ralexe:
Issue was found by PVS-Studio static analysis tool.

Error reported:
/home/ralexe/Desktop/doxygen/src/util.cpp:2065: error: V570 The 'dstA.name' variable is assigned to itself.
/home/ralexe/Desktop/doxygen/src/util.cpp:2120: error: V570 The 'dstA.name' variable is assigned to itself.

The fix is simple, assigning srcA.name to dstA.name as it was the intention.

Thank you,
Razvan Alexe


albert-github:
@ralexe Thanks, looks like to make sense and it looks like it is in the code for a while.

Do you have also an example showing the problem or is it just the report from PVS-Studio?

ralexe:
@albert-github Only got the report from PVS-Studio.

Issue #8615, opened 2021-06-23
Markdown **emphasis** at the beginning of the line is not rendered.


dhebbeker:
### Describe the bug

Markdown emphasis (using `**word**`) at the beginning of the new line does not render correctly.

For example this code:

```cpp
/*! \file 

not bold

**should be bold**

not bold

this **should be bold**

not bold
*/
```

### Expected behavior
I would expect the following output:

> not bold
> 
> **should be bold**
> 
> not bold
> 
> this **should be bold**
> 
> not bold

Instead the output is:

> not bold
> 
> should be bold**
> 
> not bold
> 
> this **should be bold**
> 
> not bold

### Screenshots
![grafik](https://user-images.githubusercontent.com/1594340/123086082-4adeee80-d423-11eb-80e6-19459eefaa6e.png)

### To Reproduce

Use the code above with default configuration (empty Doxyfile). This is also provided here: [doxy-markdown-emph.zip](https://github.com/doxygen/doxygen/files/6701274/doxy-markdown-emph.zip)

### Version

 - doxygen 1.9.1 (ef9b20ac7f8a8621fcfc299f8bd0b80422390f4b)
 - Windows 10 64 bit

### Additional context

May be related to #7786.

According to the [manual emphasis should be parsed](https://www.doxygen.nl/manual/markdown.html#mddox_emph_spans) if

> it is preceded by a [...] newline

albert-github:
This is not nice, quick and dirty workaround:
```
&zwnj;**should be bold**
```


(note: problem is probably located in the C scanner, scanner.l, as when I use a markdown file it is rendered correctly and ans in version 1.8.3, very old, it is OK)

albert-github:
The mentioned quote:
> it is preceded by a space, newline, or one the following characters <{([,:;

is valid for markdown, but here we have to do with C / Cpp type of code and one of the properties of the C / Cpp comments is that it can have leading asterisks (0 or more) at the beginning of the line that are ignored by doxygen.
In the scanner (`scanner.l`)  this is handled by means by the rule:
```
<DocBlock>^{B}*"*"+/[^/]
```
it could be argued:
- when on previous / next lines there are no leading asterisks these 2 asterisks should be seen as "bold" indicator, but nowhere it is required that the number of asterisks is consistent and also what to do with the, unlikely block:
  ```
  /**
    ** bold item ** text ... <br>
    ** bold item ** text ... <br>
    ** bold item ** text ... <br>
  */
 ```
[...]

PR #8619, opened 2021-06-24
issue #8615 Markdown **emphasis** at the beginning of the line is not rendered.

albert-github:
The mentioned restrictions in the documentation are valid for markdown. When using C-type comment and no "lineup"  character, the start emphasis is seen as "lineup".
There is no real solution for this problem so a note in the documentation.

### DIFF
--- a/doc/markdown.doc
+++ b/doc/markdown.doc
@@ -600,6 +600,21 @@ An emphasis or a strikethrough ends if
 
 Lastly, the span of the emphasis or strikethrough is limited to a single paragraph.
 
+Note When using emphasis by means of `*`s at the beginning of a line and a
+C-type doxygen comment block (`/&zwj;** ... *&zwj;/`) and not using an `*` as
+"lineup" doxygen will seen the sequence of `*`s at the beginning of the line as
+"lineup" and not as emphasis. So the following will not render as bold:
+\verbatim
+/**
+ **this is not bold**
+ */
+\endverbatim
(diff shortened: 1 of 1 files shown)


doc/markdown.doc
```diff
@@ -600,6 +600,21 @@ An emphasis or a strikethrough ends if
...
+Note When using emphasis by means of `*`s at the beginning of a line and a
+C-type doxygen comment block (`/&zwj;** ... *&zwj;/`) and not using an `*` as
+"lineup" doxygen will seen the sequence of `*`s at the beginning of the line as
+"lineup" and not as emphasis. So the following will not render as bold:
+\verbatim
+/**
+ **this is not bold**
+ */
+\endverbatim
+but will render as bold:
```
dhebbeker, line 612:
```suggestion
But this will render as bold:
```

doc/markdown.doc
```diff
@@ -600,6 +600,21 @@ An emphasis or a strikethrough ends if
 
 Lastly, the span of the emphasis or strikethrough is limited to a single paragraph.
 
+Note When using emphasis by means of `*`s at the beginning of a line and a
```
dhebbeker, line 603:
```suggestion
Note that when using emphasis by means of `*`s at the beginning of a line and a
```

dhebbeker (approved):
IMHO this does resolve #8615 :heavy_check_mark:

PR #8624, opened 2021-06-27
bug_735146 MSC images not included in QHP index

albert-github:
The problem mentioned in this issue, missing MSC file in index.qhp, was not present anymore, but a few other problems showed:
- QHP should also not use `GENERATE_TREEVIEW` and `SEARCHENGINE`
- quite a few file names were placed into the `index.qhp` and `index.hhp` although not relevant for these "HTML type " output as they came from other output formats, resulting in:
  - double entries
  - non existing files as the files are not present on the `HTML_OUTPUT` directory
- some paths were shown with (full-) path due to a wrong used name (see msc.cpp: imgName -> outFileName).
- in case of `CREATE_SUBDIRS` the path was stripped (for "dot" files) due to the usage of `fileName()` (see htmlhelp.cpp)

A bit a more extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6721439/example.tar.gz)

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -9362,7 +9362,7 @@ static void copyStyleSheet()
   }
 }
 
-static void copyLogo(const QCString &outputOption)
+static void copyLogo(const QCString &outputOption, const QCString docFormat)
 {
   QCString projectLogo = Config_getString(PROJECT_LOGO);
   if (!projectLogo.isEmpty())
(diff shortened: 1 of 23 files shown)


src/doxygen.cpp
```diff
@@ -9362,7 +9362,7 @@ static void copyStyleSheet()
   }
 }
 
-static void copyLogo(const QCString &outputOption)
+static void copyLogo(const QCString &outputOption, const QCString docFormat)
```
doxygen, line 9366: Any reason to pass a "magic" string value here instead of a boolean/enum indicating whether this item needs to be added to the index?

albert-github: A "magic" string is not the best idea.
I think I couldn't find a suitable enum for it. 
The closest what I see, now, is the enun Type in the class DocImage (in docparser.h), I don't think this is the best choice either (the copyLogo is about an image but not related to DocImage so taking it is not good). Probably best is to define an new enum class

PR #8628, opened 2021-06-29
Semantic syntax highlighting

tttapa:
This PR adds additional CSS classes to the symbols in the generated code snippets, in order to support more expressive and semantic syntax highlighting. For example, using the VSCode color theme, you can see the differentiation between class and function names:
Screenshot from 2021-06-29 14-27-43
Compared to the original, where all identifiers are the same blue color (using the master version of Doxygen, but with the same custom style sheet):
Screenshot from 2021-06-29 15-00-38

To this end, I've passed a pointer to the symbol's `Definition` to the `writeCodeLink` function, so this function has access to its `Definition::definitionType()` function. Based on the value of `definitionType()`, it then adds a CSS class in the generated HTML. It works and it's not too intrusive, but if someone who is more familiar with the architecture of Doxygen has a better idea, I am of course open to that.

Some questions I have regarding further improvements:

1. Right now, most of the symbols are classified as `Definition::DefType::TypeMember`, there is no distinction between free functions, member functions, global variables, member variables, enumeration types, enumerator names, type aliases, etc. Doxygen knows which category each symbol belongs to, because it correctly classifies them in the generated documentation, but is there any way to get this information to the HTML formatter for code snippets? Maybe a new data member `enum SemanticHighlightClass` could be added to the `Definition` class, but where would this be initialized to the correct value?
2. Some keywords like `constexpr` are not highlighted. I've tried adding them to the [list of keywords in the grammar definition](https://github.com/doxygen/doxygen/blob/53f2a068a53dbc62bb82712e238788d93e6b76a9/src/code.l#L280) and that seems to work, but I'm not sure if there's anything else that has to be updated. (This change is not included in this PR.)
3. There appear to be some tokenization/parsing problems with type names in template arguments, the `int` substring in `template <uint8_t>` is highlighted differently (this issue is not caused by this PR, it's present v1.8.20 and v1.9.1 as well): 
Screenshot from 2021-06-29 14-52-59
4. Would it be possible to highlight the `#include` keyword differently from the filename that follows it? This would be similar to what most editors do, e.g. 
Screenshot from 2021-06-29 14-39-42

(2-4) are minor details, but I'd really like to fix (1). I think differentiating between functions and variables is useful, and the fact that type aliases and enum types are drawn in a different color than normal classes is a bit misleading.  
I'd be grateful if anyone could point me in the right direction.


src/htmlgen.cpp
```diff
@@ -720,12 +720,28 @@ void HtmlCodeGenerator::writeLineNumber(const QCString &ref,const QCString &file
...
+        case Definition::DefType::TypeFile: return "highlightfile code";
+        case Definition::DefType::TypeNamespace: return "highlightnamespace code";
+        case Definition::DefType::TypeMember: return "highlightmember code";
+        case Definition::DefType::TypeGroup: return "highlightgroup code";
+        case Definition::DefType::TypePackage: return "highlightpackage code";
+        case Definition::DefType::TypePage: return "highlightpage code";
+        case Definition::DefType::TypeDir: return "highlightdir code";
+        case Definition::DefType::TypeConcept: return "highlightconcept code";
+        default: return "code";
+      }
```
albert-github, line 741: At first glance I'm missing the default setting in template/doxygen.css (could be empty, but would signal a user wheer to add something.

albert-github (commented):
- having a fine grain of coloring possibilities can sometimes be of an advantage (during debugging and then coloring different item with a different color, though for a normal the css file has already a lot of possibilities and I'm not sure whether adding extra classes would add to the understandability of the css file.
- I'm not sure whether to pass the pointer to the complete definition to writeCodeLink or just the `definitionType()` would be better.

tttapa:
> though for a normal the css file has already a lot of possibilities and I'm not sure whether adding extra classes would add to the understandability of the css file.

I'm not sure if I understand this correctly: are you concerned about the readability of template/doxygen.css?

The main aim for me was to improve the readability of the code snippets in the generated documentation pages: right now most of it are just blue links, and it really helps if you can just see at a glance whether something is a function, a type name, a constant, an enumerator, etc.  
This would be similar to the semantic syntax highlighting used by most IDEs and editors.

For example, Doxygen produces the following blue links:
Screenshot from 2021-06-29 16-27-16
Whereas VSCode highlights types, variables and functions in different colors:
Screenshot from 2021-06-29 16-26-45
In my opinion, the latter is much easier to read.

I'm not advocating that this should become the default for Doxygen, but by at least adding the extra class names to the HTML, this allows users to enable more expressive syntax highlighting in their own custom stylesheets.

src/htmlgen.cpp
tttapa: Thanks, I've added some empty rules for the most important ones (I don't think group, package, page and dir occur in source code listings).

templates/html/doxygen.css
```diff
@@ -228,6 +228,22 @@ a.codeRef, a.codeRef:visited, a.lineRef, a.lineRef:visited {
...
+}
+a.code.highlightnamespace {
+    /* style for links to namespace names in code snippets */
+}
+a.code.highlightmember {
+    /* style for links to member variable/function names in code snippets */
+}
+a.code.highlightconcept {
+    /* style for links to concept names in code snippets */
+}
```
albert-github, line 245: Just 5 ?

tttapa: I left out the others because I don't think they are ever produced in code snippets, or is that incorrect? Can a Doxygen “group” (TypeGroup) ever be linked to in a code snippet? Maybe `TypePackage` could occur in Java or Python code?

I'll add them just in case, if you think that's best :)

PR #8630, opened 2021-06-29
Allow overwriting scope of an object from comment block (via new command)

ignis-sec:
Right now, in my opinion, documenting C++/Python mixed code has some shortcomings, one of which being the limited namespace controls for python.

In my example, i wanted to add all my python classes in a namespace called Python::Module, while keeping C++ code the same.

I've implemented a small comment command, `\belongsto {SCOPENAME}`

This allows overwriting the scope of an object from the related comment block (from class docstrings)

Consider the following example:

```py
#example_py.py
##
# @namespace Python::Example
# @brief Python portion of a mixed-code project

class ExampleClass:
    """!
         @belongsto Python::Module::example_py
         @brief A python class
    """
```

```cpp
//example_cpp.cpp

namespace ExampleCPP{
    class ExampleClass{};
};
```

Without belongsto command, this will generate the following hierarchy:

```
/
├─ ExampleCPP
│  ├─ ExampleClass
├─ example_py
   ├─ ExampleClass
```

As far as i know, it is not possible to change the scope/namespace of a Python class in Doxygen without changing the filenames.

This command allows moving the class out of example_py namespace created from the filename.
```

/
├─ ExampleCPP
│  ├─ ExampleClass
├─ Python
   ├─ Module
      ├─ example_py
          ├─ ExampleClass
```

Please let me know if there was another way of doing this without implementing the new command.
I've been going over the documentation and pretty much trying anything with package, relates, memberof commands but nothing worked.

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -94,6 +94,7 @@ static bool handleXRefItem(yyscan_t yyscanner,const QCString &, const StringVect
 static bool handleRelated(yyscan_t yyscanner,const QCString &, const StringVector &);
 static bool handleRelatedAlso(yyscan_t yyscanner,const QCString &, const StringVector &);
 static bool handleMemberOf(yyscan_t yyscanner,const QCString &, const StringVector &);
+static bool handleBelongsTo(yyscan_t yyscanner,const QCString &, const StringVector &);
 static bool handleRefItem(yyscan_t yyscanner,const QCString &, const StringVector &);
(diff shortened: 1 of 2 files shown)


albert-github (changes requested):
At first glance I'm missing the documentation of the new command in doc/commands.doc

ignis-sec:
I will get on it at the first opportunity, wanted to see what you'd think about this before adding docs and additional lex checks to not break anything.
Thanks for the heads up

Issue #5266, opened 2018-07-02
Multi-line "after the member" ///< comment works incorrectly (Origin: bugzilla #706084)


doxygen:
status NEEDINFO severity *normal* in component *general* for *---*
Reported in version *1.8.4* on platform *Other*
Assigned to: Dimitri van Heesch

On 2013-08-15 19:23:17 +0000, Aleksey Vaneev wrote:

> We are writing a lot of code and find it convenient to comment it the following way. Unfortunately, Doxygen does not parse such comments in a desired way, it does not want to concatenate the detailed description spanning several lines. What is the reason for such behavior? Is it a bug? (note that we use tab character before the ///< markup).
> 
> double Len2; ///< Required half kernel length in samples (can be a
>     ///< fractional value). Final physical kernel length will be provided
>     ///< in the KernelLen variable. Len2 should be >= 2.
> double Freq1; ///< Required corner circular frequency 1 [0; pi]. Used only
>     ///< in the generateBand() function.
> double Freq2; ///< Required corner circular frequency 2 [0; pi]. Used only
>     ///< in the generateBand() function. The range [Freq1; Freq2] defines
>     ///< a pass band for the generateBand() function.

On 2013-08-15 19:38:01 +0000, Aleksey Vaneev wrote:

> The commenting in such way seems to be working IF an additional EMPTY comment line is added. Unfortunately, this is quite counter-intuitive and takes valuable screen space.
> 
> static const int FuncLen = N; ///< Fractional delay function length in
>     ///< samples (taps).
>     ///<

On 2013-08-15 20:13:26 +0000, Aleksey Vaneev wrote:

> Strangely enough, mixing two ways of commenting does not work:
> 
> static const int FuncLenD2Minus1 = FuncLenD2 - 1; ///< = FuncLenD2 - 1.
>     ///< This value also equals to filter's latency in samples (taps).
>     ///<
> static const int FuncLenD2Plus1 = FuncLenD2 + 1; ///< = FuncLenD2 + 1.
> static const int FuncFrac = Fracs; ///< The number of fractional sample
>     ///< positions to use.
>     ///<
> 
> This will leave the FuncLenD2Plus1 member without comments.

On 2015-06-21 16:20:13 +0000, Dimitri van Heesch wrote:

> Hi Aleksey,
> 
[...]

PR #8656, opened 2021-07-08
Feature: IGNORE_CPP_ALIGNMENT

johnny-mc:
This feature is a workaround for the following problem (although it can conceivable be used on its own).

## PROBLEM DESCRIPTION:
**Incorrect brief/detailed blocks in ///< comments on macros**

Doxygen (on default config settings) is presently interpreting two-line comments like this:

```
#define SOME_MACRO 42  ///< This is the documentation for SOME_MACRO
                       ///< where this second line appears as brief description
```

taking the first line as "detailed" description, and the second as "brief" description.
(It you were to show "brief first, detailed after", the two lines would be in reverse order.)

This is especially a problem for people who use the XML output.

**Related issue:**
  https://github.com/doxygen/doxygen/issues/5266

## CAUSE AND WORKAROUND:

The reason for this behavior is in the `pre.l` pass, where
the body of the macro ("42" plus blanks in this example)
is removed from the input text, and as a result the vertical
alignment of the comments is broken.

A proper solution would have involved preserving the vertical alignment
of C++ comments in `pre.l`. Since I had no time for it, I am submitting
a workaround: make vertical alignment optional via a configuration keyword,
"IGNORE_CPP_ALIGNMENT".

Note that Doxygen doesn't seem to mind if ///< and /// are intermixed, as long
as they are vertically aligned (is this another bug?), so our workaround explicitly
avoids mixing ///< and /// in the same block when IGNORE_CPP_ALIGNMENT is YES.

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -40,7 +40,8 @@
 
 #include <assert.h>
 
-#define YY_NO_INPUT 1
+// yyinput used to assign afterMarker
+//#define YY_NO_INPUT 1
 #define YY_NO_UNISTD_H 1
 
 #define ADDCHAR(c)    yyextra->outBuf->addChar(c)
@@ -83,6 +84,7 @@ struct commentcnvYY_state
   int      lastCommentContext = 0;
   bool     inSpecialComment = FALSE;
   bool     inRoseComment= FALSE;
+  bool     afterMarker = FALSE;
   int      stringContext = 0;
   int      charContext = 0;
   int      javaBlock = 0;
@@ -311,6 +313,12 @@ SLASHopt [/]*
 				       yyextra->inSpecialComment=TRUE;
 				       //BEGIN(SComment); 
 				       yyextra->readLineCtx=SComment;
+
+				       // remember the after-mark, compare later on replaceComment()
+				       int nextChar = yyinput(yyscanner);
+				       yyextra->afterMarker = nextChar == '<';
+				       unput(nextChar);
+
 				       BEGIN(ReadLine);
 				     }
                                    }
@@ -1140,7 +1148,15 @@ static void replaceComment(yyscan_t yyscanner,int offset)
   {
     //printf("replaceComment(%s)\n",yytext);
     int i=computeIndent(&yytext[offset]);
-    if (i==yyextra->blockHeadCol)
+
+    // ignore C++ comment alignment
+    // plus extra check for the '<' after-marker (do not pair ///< with plain ///)
+
+    bool afterMarker = yytext[yyleng-1] == '<';
+    bool ignoreCppAlignment = Config_getBool(IGNORE_CPP_ALIGNMENT)
+                              && afterMarker == yyextra->afterMarker;
+
+    if (ignoreCppAlignment || i==yyextra->blockHeadCol)
     {
       replaceCommentMarker(yyscanner,yytext,(int)yyleng);
     }
--- a/src/config.xml
+++ b/src/config.xml
@@ -499,6 +499,16 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+    <option type='bool' id='IGNORE_CPP_ALIGNMENT' defval='0'>
+      <docs>
+<![CDATA[
+  When the \c IGNORE_CPP_ALIGNMENT tag is set to \c YES, doxygen will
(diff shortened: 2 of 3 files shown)


albert-github:
Thanks for the patch, though I don't think this is the right approach though.
You try, as far as I understand, to solve 2 problems with one patch?
- the spacing problem
  I think it would be better to call, where it isn't yet done, for the state `DefineText` (and the states it calls) in `pre.l` the routine `outputChar(yyscanner,...);` to replace the original characters with space characters, so the indentation is kept correct. Some quick tests look promising with this approach.
- the `IGNORE_CPP_ALIGNMENT`
   I don't yet get it, can you give an example. When I have something like:
   ```
   #define SOME_MACRO 42  ///< This is the documentation for SOME_MACRO
                          /// where this second line appears as brief description
   ```
  doxygen will see the `///<` line as documentation for the `define` and the line `///` as start of the documentation for the next item in the file, which is in principle correct although maybe a bit a bad coding style.

johnny-mc:
Thanks for your reply, Albert.

I did mention that a proper solution would imply preserving the alignment in `pre.l`; I just do not have the time.

As for your second observation, in my example the two comment lines use ///< ; you omitted the "<" on the second line.

albert-github:
I don't think I got the proper intention for the `IGNORE_CPP_ALIGNMENT`, so it is just for overcoming the alignment problem?

johnny-mc:
(apologies for the misunderstadings - I was editing the previous comment when you replied.)

On my side the only reason for adding IGNORE_CPP_ALIGNMENT was to work around the alignment problem.

As for what Doxygen does when mixing /// and ///<, I may have not tested this extensively. I do know that, after implementing IGNORE_CPP_ALIGNMENT, then it became necessary to check for '<', otherwise the code would take subsequent lines without '<' as part of the same block. On the original, unmodified  code, some cases do mix both ///< and ///; for instance,
```
#define OTHER_MACRO 77 ///< first line
///< second line
/// Lines intended for the following object
```
will make all three lines part of the same block. (Not sure why the second line is accepted as part of the same block, even on a different indent; but once this happens, the third is also added to the same block.)
Thanks again.

albert-github:
Note that in the master documentation there is now the warning (in the section "Putting documentation after members"):

>Warning
These blocks can only be used to document members and parameters. They cannot be used to document
files, classes, unions, structs, groups, namespaces, macros, and enums themselves. Furthermore, the structural
commands mentioned in the next section (like \class) are not allowed inside these comment blocks.
>
>Be careful using this construct as part of a macro definition, because when MACRO_EXPANSION is set to
YES at the places where the macro is applied, also the comment will be substituted and this comment is then
used as documentation for the last item encountered and not for the macro definition itself!

(this is introduced by means of pull request #8566 and some rewording).

johnny-mc:
For the record, in our actual environment we use EXPAND_ONLY_PREDEF to reduce the list of expanded macros to a short list in PREDEFINED, so the PR works for us.

aundro:
@albert-github thank you very much for your help! 

If I understand correctly, the fact that the ` 42 ` isn't replaced by white spaces is a bug — regardless of whether `///<` should be used for macros or not.

Our concern (I'm @johnny-mc 's colleague) is that our header files have a *ton* of such constructs, so if fixing that problem can avoid us days of comments reshaping, it would be welcome.

Do you happen to know where we should be looking in order to fix this? Thanks!

albert-github:
@aundro The fix @johnny-mc was something that would have probably worked for your cases, but as he said it was just a quick workaround. The real problem is that that the part that is taken out is indeed not replaced with spaces. I'm currently looking into the problem and it is quite a nasty problem. It looks like I covered so far the part of the defines without arguments, looking into the part with arguments now. A small complicating factor is that all characters can be replaced by a spaces except for the `<tab>` character (I would never advise this in code but it pops up often, especially in older code), that has to remain.
The fixes have to be made in `pre.l` as this is the place where the define values "disappear".

So no need for you to look into the problem.

albert-github:
I've just pushed a proposed patch, pull request #8667 This proposed patch should fix the spacing issues without the need for a configuration setting.

johnny-mc:
Hi, Albert,
I gave #8667 a try, and it seems to work fine! It appears to solve all of our problems with macros.
Thanks for the work!

albert-github:
The code of #8667 has been integrated in master on github, so you can close this PR.

Issue #8657, opened 2021-07-09
Plantuml ditaa diagram is not visible in refman.pdf


broucke:
**Describe the bug**

The plantuml ditaa diagram is not visible in LaTeX ouput, but it is visible in html output.

**Expected behavior**

I expect the diagram to be visible in `./latex/refman.pdf`.

**To Reproduce**

The example consists out of 2 files: 
- the doxygen configuration file `./doxygen.conf`
- the input file `./mainpage.dox`

``` sh
# Example structure
$ tree
.
├── doxygen.conf
└── mainpage.dox
```

```
/** \mainpage ditaa example
\startuml {example.png} "ditaa example"
ditaa
    +--------+   +-------+    +-------+
    |        | --+ ditaa +--> |       |
    |  Text  |   +-------+    |diagram|
    |Document|   |!magic!|    |       |
    |     {d}|   |       |    |       |
    +---+----+   +-------+    +-------+
        :                         ^
        |       Lots of work      |
        +-------------------------+
\enduml
 */
```

``` sh
# Config file generated with: doxygen -g doxygen.conf

# Manual changes to doxygen.conf:
INPUT=mainpage.dox
PLANTUML_JAR_PATH=/opt/plantuml/plantuml.1.2021.8.jar
```
Commands used to generate the html and LaTeX output:
``` sh
doxygen doxygen.conf && make -C ./latex
```

**Screen shots**
Html output
![html_output](https://user-images.githubusercontent.com/51946272/125081341-3018a500-e0c6-11eb-9606-5dd47efe572d.png)

LaTeX output
![latex_output](https://user-images.githubusercontent.com/51946272/125081349-31e26880-e0c6-11eb-9c51-9d0f351bd492.png)

**Version**

``` sh
# Doxygen compiled from source
$ doxygen --version 
1.9.1 (ef9b20ac7f8a8621fcfc299f8bd0b80422390f4b)
```

``` sh
$ lsb_release -a
No LSB modules are available.
Distributor ID:	LinuxMint
Description:	Linux Mint 19.3 Tricia
Release:	19.3
Codename:	tricia
```

``` sh
$ java -jar /opt/plantuml/plantuml.1.2021.8.jar -version
PlantUML version 1.2021.8 (Sat Jun 26 10:20:59 CEST 2021)
(GPL source distribution)
Java Runtime: OpenJDK Runtime Environment
JVM: OpenJDK 64-Bit Server VM
Default Encoding: UTF-8
Language: en
[...]

PR #8658, opened 2021-07-10
startuml documentation

albert-github:
- small extra notes in usage for engines: ditaa (see issue #8657) , latex and math
- some textual improvements


albert-github:
In case the proposed pull request is #8661 this proposed pull request should **NOT** be integrated but the proposed pull request #8662

PR #8661, opened 2021-07-11
issue #8657 Plantuml ditaa diagram is not visible in refman.pdf

albert-github:
The current version of plantuml does not support EPS as output format for `ditaa` (Diagram through ASCII art), so we have to revert to PNG in this case.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6796798/example.tar.gz)

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -5567,8 +5567,30 @@ int DocPara::handleCommand(const QCString &cmdName, const int tok)
         defaultHandleTitleAndSize(CMD_STARTUML,dv,dv->children(),width,height);
         doctokenizerYYsetStatePlantUML();
         retval = doctokenizerYYlex();
-        int line=0;
-        dv->setText(stripLeadingAndTrailingEmptyLines(g_parserContext.token->verb,line));
+        int line = 0;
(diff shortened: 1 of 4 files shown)


albert-github:
In case this proposed pull request is integrated the proposed pull request #8658 should **NOT** be integrated but the proposed pull request #8662

src/docparser.cpp
```diff
@@ -5567,8 +5567,30 @@ int DocPara::handleCommand(const QCString &cmdName, const int tok)
...
+        if (engine == "ditaa")
+        {
+          dv->setUseBitmap(true);
+        }
+        else if (engine == "uml")
+        {
+           // check on ditaa in first line
+           if (!trimmedVerb.isEmpty())
+           {
+             const char *p = trimmedVerb.data();
```
doxygen, line 5581: this loop seems like a difficult way to do `int i = trimmedVerb.find('\n')`

albert-github: indeed probably will be easier to use find, I'll have a look.

albert-github:
This pull request is merged on July 15, 2021 and the last release is 1.9.4 from May 5, 2022. This release contains this patch.
- Where do you see problems?

stevan80236:
Sorry again, just update the plantuml.jar file to a newer version, and it works.

PR #8662, opened 2021-07-11
startuml documentation

albert-github:
- small extra notes in usage for engines: latex and math
- some textual improvements


doxygen:
@albert-github is this different from https://github.com/doxygen/doxygen/pull/8658?

albert-github:
In #8658 there is an extra:
```
  \note The use of the `<engine>` `ditaa` is not possible in \LaTeX as PlantUML only
  supports the `png` format and doxygen requires, temporary, `eps` output.
```
but that is not valid anymore #8661 has been integrated,

Issue #8679, opened 2021-07-21
No link to the source code anymore


couet:
I have installed the master version of doxygen (1.9.2) to take advantage of a [bug fix not yet released](https://github.com/doxygen/doxygen/issues/8620). 

With that new version I saw that the C++ methods having only a `\brief` as documentation did not have a detailed description anymore. Adding `ALWAYS_DETAILED_SEC` fixed this problem.

But there is still an issue for methods having no help at all, not even a `\brief` . Such methods do not have a detailed description anymore. One can say it's normal as there is no doc why having a detailed description ?

Indeed the detailed description is needed in a such case because it is the only way to access the source code. Seeing the source code for such methods can be very important to understand the algorithm or the logic used. Specially in a Physics Data analysis system like [ROOT](https://root.cern). Such methods do not need a `\brief` or a help because their names are explicit enough and a `\brief` will be a copy of the method name.

So, we do not need the detailed description for such methods, but we need a direct link to the code. 
Is there a way to do it ?

albert-github:
- Which version of 1.9.2 did you install (i.e. complete output of `doxygen -v` is relevant)?
- You probably have `EXTRACT_ALL` set, but the relevant functions don't have a detailed description, so no need for a detailed section. The `EXTRACT_ALL`, incorrectly, triggered a detailed section in the past.

Without an example it is hard to tell. So:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

couet:
Thanks for your fast reply. Sorry to not have been precise enough. Let me complete my post.

  - `doxygen -v` gives me:

```
% doxygen -v                    
1.9.2 (97dffa0977cc41729942cf3a8c87232eb7edc09d)
```

   - Yes I have `EXTRACT_ALL` set.

  - A small reproducer is:
```
/// \class A
/// Class A
class A {
public:
   /// \brief method 1 in A
   void method1(int i){}
   void method2(int i){}
};
```

  it gives the following output:
<img width="953" alt="Screenshot 2021-07-21 at 13 00 12" src="https://user-images.githubusercontent.com/4697738/126478665-289ee899-845d-4426-baf8-c5d17e291206.png">

As can you see: 
   - `method1` has a `\ brief` and we can access to the code, 
   - `method2` has no `\brief` we can see it in the doc but we cannot access the code.

albert-github:
[...]

PR #8681, opened 2021-07-21
A fix to make sure code for all methods is accessible ...

couet:
.. even those without documentation.
Description here: https://github.com/doxygen/doxygen/issues/8679

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -2542,13 +2542,15 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
 
 bool MemberDefImpl::isDetailedSectionLinkable() const
 {
-  static bool alwaysDetailedSec = Config_getBool(ALWAYS_DETAILED_SEC);
-  static bool repeatBrief       = Config_getBool(REPEAT_BRIEF);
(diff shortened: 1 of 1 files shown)


src/memberdef.cpp
```diff
@@ -2584,7 +2586,9 @@ bool MemberDefImpl::isDetailedSectionLinkable() const
          _isReimplements() ||
          (_countReimplementedBy() > 0) ||
          // has user comments
-         Doxygen::userComments
+         Doxygen::userComments ||
+         // has source files and all entities are assumed to be documented
+         (sourceBrowser && extractAll)
```
albert-github, line 2591: As indicated in the issue already:
- this will only be OK for HTML
- I don't think the `EXTRACT_ALL` is necessary
- I'm currently working on a correct solution

couet:
Thanks. Meanwhile I may use this temporary fix to make our documentation work.

doxygen:
@couet @albert-github This should now work again with the latest series of commits. Let me know if you still see issues.

couet:
@albert-github : I took the latest series of commits and, yes, I confirm it is working again. Many thanks !
(sorry for the late reply, I was on holidays).

PR #8683, opened 2021-07-22
Source code requires in a number of cases `EXTRACT_ALL`

albert-github:
Looking at issue #8679 we detected that the source code was not always shown when `EXTRACT_ALL` was not set. This happened with LaTeX, RTF and Docbook.
To be able to really address the problem of #8679 it is necessary that the source code is correctly included when it is required, otherwise there will be a lot of undefined references.
In this pull request the handling of the source code for the mentioned output formats has been corrected. Due to the extra flag to show the source code in the output for these format some code duplication was necessary.

Further improvements:
- debug statement in util.cpp
- creation of anchor in LaTeX

### DIFF
--- a/src/docbookgen.cpp
+++ b/src/docbookgen.cpp
@@ -377,6 +377,8 @@ DB_GEN_C
 void DocbookGenerator::startIndexSection(IndexSections is)
 {
 DB_GEN_C2("IndexSections " << is)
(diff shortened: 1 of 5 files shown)


doxygen:
@albert-github I've now solved it by obsoleting the output specific options. Let me know if you see any issues.

PR #8478, opened 2021-04-04
Extending startuml with extra figure types

albert-github:
Not all diagrams can be created with the PlantUML `@startuml` command but need another
PlantUML `@start...` command. This wil look like `@start<engine>` where currently supported are
the following `<engine>`'s: `uml`, `bpm`, `wire`, `dot`, `ditaa`, `salt`, `math`, `latex`,
`gantt`, `mindmap`, `wbs`, `yaml`, `creole`, `json` and `flow`. By default the `<engine>` is
`uml`. The `<engine>` can be specified as an option.

Explicitly the option variant has been chosen so we won't get an explosion of extra commands.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6254988/example.tar.gz)

PR #8684, opened 2021-07-22
Mermaid diagrams

djbwork:
This attempts to add support for Mermaid diagrams by cloning the approach for PlantUML diagrams and sending markup to the mermaid-cli tool for processing into images. Let me know if there are issues with how I've done the pull request; first time contributor.

Mermaid code like this works:

* @startmermaid "LRM chain flow"
* graph TD
*     id1[(L1b)] --> id2[[MSSG]]
*     id2 --> id3[[DISC]]
*     id3 --> id4[(L2i)]
* @endmermaid


albert-github:
A few quick remarks (in random order):
- looks like a bit of a competitor of PlantUML (https://mermaid-js.github.io/mermaid/#/n00b-gettingStarted)
- Can you make the patch against the current master
- Recently a change regarding the use of the global cache file has been integrated in the master version of doxygen. It looks like that the Mermaid implementation still uses this way of caching (I see  in the mermaid.h the define: MERMAID_CACHE_FILENAME and in mermaid.cpp this define is used).
- the Mermaid executable is hard coded in the code as mmdc, PlantUml uses here a setting for that can be tested.
- Where can the executable be found and is it available for multiple platforms / can be generated for platforms where no executable is available? I can find some js code and on the https://github.com/mermaid-js/mermaid-cli I cannot find a way to create an executable.
- Where can we find the source code of Mermaid (I can only find some js code)
- I'm missing doxygen documentation for this new command.
- Would be nice to have a set of, separate,  examples (like I did with #8478)
- A minor fact, the fact that the spaces are removed is nice, but unfortunately distracts quite a bit from the real implementation

djbwork:
Hello

Answers inline:

On 22 Jul 2021, at 16:53, albert-github ***@***.******@***.***>> wrote:



A few quick remarks (in random order):

  *   looks like a bit of a competitor of PlantUML (https://mermaid-js.github.io/mermaid/#/n00b-gettingStarted<https://eur01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fmermaid-js.github.io%2Fmermaid%2F%23%2Fn00b-gettingStarted&data=04%7C01%7C%7C46edab3d7f584406b29708d94d28e765%7C1faf88fea9984c5b93c9210a11d9a5c2%7C0%7C0%7C637625660342729032%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=KRYxJp4cRr5M25ZMpWlgFMkGhYSC%2FFeLj0dtxCsSE58%3D&reserved=0>)

It is - the two overlap, but both have abilities that the other doesn't. I find the old-style flowchart symbols in Mermaid more useful for documenting procedural C than UML.

  *   Can you make the patch against the current master

I certainly can. Will take me a while due to work pressure but I’ll redo it against whatever is current when I can.

  *   Recently a change regarding the use of the global cache file has been integrated in the master version of doxygen. It looks like that the Mermaid implementation still uses this way of caching (I see in the mermaid.h the define: MERMAID_CACHE_FILENAME and in mermaid.cpp this define is used).

I’m not knowingly using the cache, this is just from cloning everything in the PlantUML code. I’ll look into this.

  *   the Mermaid executable is hard coded in the code as mmdc, PlantUml uses here a setting for that can be tested.

Will do.

  *   Where can the executable be found and is it available for multiple platforms / can be generated for platforms where no executable is available? I can find some js code and on the https://github.com/mermaid-js/mermaid-cli I cannot find a way to create an executable.

I’ll look into this. I’ve installed via Homebrew on OSX.

  *   Where can we find the source code of Mermaid (I can only find some js code)

I’m not sure, but I think the JS might _be_ the implementation. You can use Mermaid on a web page just by pointing to the JS.

  *   I'm missing doxygen documentation for this new command.

Yes, not documented yet. Will fix.

  *   Would be nice to have a set of, separate, examples (like I did with #8478<https://github.com/doxygen/doxygen/pull/8478>)

I’ll look at that change and come up with some examples.

  *   A minor fact, the fact that the spaces are removed is nice, but unfortunately distracts quite a bit from the real implementation

Not sure what this means - spaces where?

Thanks, David

albert-github:
>  *   Where can we find the source code of Mermaid (I can only find some js code)
> I’m not sure, but I think the JS might _be_ the implementation. You can use Mermaid on a web page just by pointing to the JS.

Doxygen has the ability to output to different output formats,HTML is one of them (and probably the widest used one), though there are others as well like LaTeX / PDF, RTF, Docbook. Probably these will work as it looks like you are outputting the mermaid code to png, svg, eps / pdf as well (like PlantUML does).

You also wrote:
> You can use Mermaid on a web page just by pointing to the JS.

does this mean that you can integrate Mermaid code on the HTML page and use a Mermain native mode(i.e. without converting to to and image)? Might be interesting, something similar is done for formulas where code can be converted to an image, for e.g HTML, RTF and Docbook (LaTeX will always use native mode of course and the formulas are in LateX) but for HTML there is the extra functionality through MathJax.

>  *   A minor fact, the fact that the spaces are removed is nice, but unfortunately distracts quite a bit from the real implementation
> Not sure what this means - spaces where?

When looking at e.g the `commentscan.l`, we see that some spaces at the end of the line disappeared (the slightly darker red parts):

image

tttapa:
If anyone else is interested in getting Mermaid diagrams to work, you can do this relatively easily without changes to doxygen itself by introducing an alias and adding the Mermaid JavaScript to the HTML header of your project:

https://github.com/tttapa/doxygen-mermaid

djbwork:
Thanks for the tip - I’ve not had a lot of luck finding a block of time to get back to this.

On 18 Jun 2022, at 23:28, Pieter P ***@***.******@***.***>> wrote:


⚠ Caution: External sender



If anyone else is interested in getting Mermaid diagrams to work, you can do this relatively easily without changes to doxygen itself by introducing an alias and adding the Mermaid JavaScript to the HTML header of your project:

https://github.com/tttapa/doxygen-mermaid

KangLin:
Whether there is progress?

albert-github:
This PR is obsolete as a new implementation has been made (#12069) and has been integrated into the master of doxygen

PR #8691, opened 2021-07-26
Equation reference  for formulas missing for HTML / Docbook / RTF

albert-github:
When having a simple example:
```
# Math references

\f{equation}{
\alpha = \beta * \gamma
  \label{eq:my_system}
\f}

The reference: \f$\eqref{eq:my_system}\f$
```
and
```
EXTRA_PACKAGES = amsmath amssymb
```

The reference will be `??` instead of `1` due to the fact that that `latex` command is only run once.
Note: When using MathJax with (but this is not a solution for e.g RTF):
```
USE_MATHJAX = YES
EXTRA_PACKAGES = amsmath amssymb
MATHJAX_EXTENSIONS     = amssymb amsmath
MATHJAX_CODEFILE       = mycode.js
```
and `mycode.js` like
```
MathJax.Hub.Config({
  TeX: { equationNumbers: { autoNumber: "AMS" } }
});
```
the reference is shown again.

Old result:

image

New result:
image

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6877705/example.tar.gz)

The problem was found based on: https://stackoverflow.com/questions/68521882/autonumbering-equations-in-doxygen/

### DIFF
--- a/src/formula.cpp
+++ b/src/formula.cpp
@@ -195,7 +195,7 @@ void FormulaManager::generateImages(const QCString &path,Format format,HighDPI h
     Portable::sysTimerStart();
     char args[4096];
     sprintf(args,"-interaction=batchmode _formulas.tex >%s",Portable::devNull());
-    if (Portable::system(latexCmd,args)!=0)
+    if ((Portable::system(latexCmd,args)!=0) || (Portable::system(latexCmd,args)!=0))
     {
       err("Problems running latex. Check your installation or look "
           "for typos in _formulas.tex and check _formulas.log!\n");


doxygen:
@albert-github I've made the number of reruns dependent on the contents of the `_formulas.log` generated by LaTeX just like we do in the Makefile that is generated in the latex output directory. Let me know if you see any problems with this approach.

albert-github:
I don't really see a problem in this approach.
The potential number of reruns of 8 is a bit high ( I think that just 1 rerun is sufficient).
Advantage is the flexible number of reruns, disadvantage is that the text (like "Rerun to get cross-references right")is now on multiple places i.e. in formulas.cpp and in latexgen.cpp.

Issue #3850, opened 2018-07-02
Expanded seealso tags create multiple "See also" sections (Origin: bugzilla #624366)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.8.2-SVN* on platform *Other*
Assigned to: Dimitri van Heesch

On 2010-07-14 18:29:44 +0000, Ray Lefuel wrote:

> I'm not sure if this is by design, but this affects HTML and XML outputs. (I have not tried other outputs)
> 
> ---
> See also sections are generated correctly when I use seealso tags in the following context:
> 
> <seealso cref="MemberNameA"/>
> <seealso cref="MemberNameB"/>
> 
> In this example, the generated XML output is as follows:
> 
> <simplesect kind="see"><para><ref refid="class_foo" kindref="member">MemberNameA</ref>, <ref refid="class_foo" kindref="compound">MemberNameB</ref></para></simplesect>
> 
> ---
> Multiple See also sections are generated when I use seealso tags in the following context:
> 
> <seealso cref="MemberNameA"></seealso>
> <seealso cref="MemberNameB"></seealso>
> 
> In this example, the generated XML output is as follows:
> 
> </para><para><simplesect kind="see"><para><ref refid="class_foo" kindref="member">MemberNameA</ref></para></simplesect>
> </para><para><simplesect kind="see"><para><ref refid="class_foo" kindref="compound">MemberNameB</ref></para></simplesect>

On 2012-11-18 11:07:36 +0000, Dimitri van Heesch wrote:

> Changed version 'latest' to '1.8.2-SVN' so I can remove 'latest' as an option as it is a moving target.

albert-github:
I've just pushed a proposed patch, pull request #8701

PR #8701, opened 2021-08-02
bug_624366 Expanded seealso tags create multiple "See also" sections

albert-github:
Also known as issue #3850

Although in the original specs of the documentation tags it would be better when both both types of closing tags (`<.../>` and `<...></..>`) are supported.
In newer versions of the Microsoft specifications (see e.g. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#seealso) also a text is supported in the seealso tag.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/6917400/example.tar.gz)

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -4405,14 +4405,15 @@ int DocSimpleSect::parseXml()
     if (m_children.empty())
     {
       par->markFirst();
+      par->markLast();
+      m_children.push_back(std::unique_ptr<DocPara>(par));
     }
     else
     {
       ASSERT(m_children.back()->kind()==DocNode::Kind_Para);
       ((DocPara *)m_children.back().get())->markLast(FALSE);
+      par = (DocPara *)m_children.back().get();
     }
-    par->markLast();
-    m_children.push_back(std::unique_ptr<DocPara>(par));
 
     // parse the contents of the paragraph
     retval = par->parse();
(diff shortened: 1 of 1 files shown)


albert-github:
The problem for the `<exception>` tag was already present before the `<seealso>` patch.

Example: [example2.tar.gz](https://github.com/doxygen/doxygen/files/6923075/example2.tar.gz)

doxygen:
@albert-github The spec isn't very clear but I would expect that `<seealso cref="target">Link text</seealso>` would produce a link with "Link Text" pointing to "target" (e.g. similar to `<a href="target">Link Text</a>` would). With your change both the target and the Link text appear in the output.

albert-github:
The spec is indeed not clear.
This is indeed the case, at the moment of developing I didn't find a possibility to suppress the `cref` mentioning and following text, but maybe I can find a solution with the hint `<a href="target">Link Text</a>`.

For reference with the code (part of the example):
```
/// new para
///
/// <seealso cref="MemberNameA">MemberNameA extra </seealso>
/// <seealso cref="MemberNameB">MemberNameB extra </seealso>
///
int tst2() { }

/// \see MemberNameA extra
/// \see MemberNameB extra
int tst3() { }
```
image

Note the difference between the `C#` syntax  and the `\see` result.

albert-github:
The current implementation results for:
```
/// <seealso cref="MemberNameA">MemberNameA extra</seealso>
```
in the equivalent:
```
/// \see MemberNameA %MemberNameA extra
```
but should look like:
```
/// \see \ref MemberNameA "MemberNameA extra"
```

PR #8727, opened 2021-08-20
Fix unknown pragma warning in MinGW

Biswa96:
The #pragma warning directive is specific to MSVC only.
So, MinGW gcc compiler shows a unknown pragma warning.

### DIFF
--- a/src/definition.h
+++ b/src/definition.h
@@ -23,8 +23,8 @@
 #include "types.h"
 #include "reflist.h"
 
-#if defined(_WIN32) && !defined(__CYGWIN__)
-// To disable 'inherits via dominance' warnings.
+#ifdef _MSC_VER
+// To disable 'inherits via dominance' warnings with MSVC.
 // See also https://stackoverflow.com/a/14487243/784672
 #pragma warning( disable: 4250 )
 #endif


src/definition.h
```diff
@@ -23,8 +23,8 @@
 #include "types.h"
 #include "reflist.h"
 
-#if defined(_WIN32) && !defined(__CYGWIN__)
-// To disable 'inherits via dominance' warnings.
+#ifdef _MSC_VER
+// To disable 'inherits via dominance' warnings with MSVC.
```
albert-github, line 27: This fix looks like to work without problems. Though the underlying question is is the `#pragma` still necessary?
I tried it with:
- Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27034 for x86
- Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27034 for x64

and it looks like the entire `#pragma` is not necessary anymore.

In the GitHub Action the compiler used is:
- The C compiler identification is MSVC 19.29.30133.0

so I don't expect any problems there either (compiler looks even newer).

Biswa96: Everyone may not use latest Visual Studio and Windows SDK toolchain.

albert-github: That is for sure true, though I don't know when this problem has been fixed in the compiler.
(The mentioned compiler is not the latest, the latest is Visual Studio 2019 and this is Visual Studio 2017. Furthermore there is a Visual Studio 2022 Preview available (see: https://visualstudio.microsoft.com/downloads/). Though it is hard to keep track on the numbers used by Microsoft ...).

Issue #8755, opened 2021-09-01
:: lost in nested names when using markdown


jwakely:
**Describe the bug**

A comment containing <code>\`foo\<bar\>::baz\`</code> gets rendered as `foo<bar>baz` when AUTOLINK_SUPPORT=NO is used.

The output is correct when using `@c foo<bar>::baz` instead of using markdown.

**Expected behavior**

It should be `foo<bar>::baz` whether using markdown backticks or `@c`.

**To Reproduce**
[doxygen-colons.tar.gz](https://github.com/doxygen/doxygen/files/7093271/doxygen-colons.tar.gz)

**Version**

1.9.3 (4bf942faa3128099a4fbc47162e11b0545487fd5)

albert-github:
The complete example is a bit large but when we create, for tests, as smaller example like:
```
/// \file

/**
  * `make_unsigned<difference_type>::type`
  *
  * \c make_unsigned<difference_type>::type
  */
  using size_type = typename _Size<_Alloc, difference_type>::type;
```
and we run this with `doxygen -d markdown`, we see:
```
======== Markdown =========
---- input -------

`make_unsigned<difference_type>::type`

\c make_unsigned<difference_type>::type

---- output -----

<tt>make_unsigned\<difference_type\>::type</tt>

\c make_unsigned<difference_type>::type

=========
```
Note: the escaped `<` and `>` in the case of the back tacks / code span.
when this is run through the `doctokenizer.l` we see (only relevant parts):
```
--accepting rule at line 611 ("<tt>")
--accepting rule at line 630 ("make_unsigned")
--accepting rule at line 521 ("\<")
--accepting rule at line 630 ("difference_type")
--accepting rule at line 521 ("\>")
--accepting rule at line 630 ("::type")
--accepting rule at line 611 ("</tt>")
--accepting rule at line 669 ("
")
--accepting rule at line 520 ("\c")
--accepting rule at line 668 (" ")
--accepting rule at line 630 ("make_unsigned<difference_type>::type")
```

In other words in case of the back ticks the `::type` is seen as a new word and later on in the processing as a potential global class name etc.

I've just pushed a proposed patch, pull request #8756

jwakely:
> The complete example is a bit large

Sorry, it came from an even bigger example, and it took me a while to figure out that the combination of markdown and AUTOLINK_SUPPORT=NO was the trigger.

> I've just pushed a proposed patch, pull request #8756

Thanks, I'll give this a try.

[...]

PR #8756, opened 2021-09-02
issue #8755 :: lost in nested names when using markdown

albert-github:
During the translation of the back ticks the double colons following a `>` should be escaped as well ass otherwise the `::` might be seen as an indication for a global class name etc.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -280,8 +280,23 @@ static QCString escapeSpecialChars(const QCString &s)
     switch (c)
     {
       case '"':  if (pc!='\\')  { insideQuote=!insideQuote; } growBuf.addChar(c);   break;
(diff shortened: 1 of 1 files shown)


jwakely:
Should there be a regression test for this, to ensure it stays fixed?

albert-github:
Would be nice.
In the xml output we see without the patch:
```
<para><computeroutput>make_unsigned&lt;difference_type&gt;type</computeroutput></para>
<para><computeroutput>make_unsigned&lt;difference_type&gt;::type</computeroutput> </para>
```
and with the patch:
```
<para><computeroutput>make_unsigned&lt;difference_type&gt;::type</computeroutput></para>
<para><computeroutput>make_unsigned&lt;difference_type&gt;::type</computeroutput> </para>
```

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

@doxygen thanks for adding the regression tests

Issue #8768, opened 2021-09-08
Doxygen Fortran: prevent automatic coversion of names to all lower-case names


shahmoradi:
**Describe the bug**
Fortran code becomes all lower-case in HTML files. This includes procedure names, and the internals of code sections (anything that appears between `\code{fortran} ... \endcode` other than the procedure interface).

**Expected behavior**
Preserve the case sensitivity in the output HTML files. This is highly desired because some projects follow the CamelCase coding convention, which makes the Doxygen documentation unreadable.

**Screenshots**
Here is an example namespace header generated by Doxygen for a module named `DistMultiNormLoggammaMix`: 
`distmultinormloggammamix Module Reference`

**To Reproduce**
[test.zip](https://github.com/doxygen/doxygen/files/7125503/test.zip)

**Version**
Doxygen version 1.9.2 (caa4e3de211fbbef2c3adf58a6bd4c86d0eb7cb8)    

**Additional context**

albert-github:
I think this issue is a duplicate of issue #5114

shahmoradi:
I managed to fix this "all-to-lower-case" behavior in a fork of Doxygen last night. I can make a PR for this if it is likely to be accepted, but I think it would be better to understand why such "all-to-lower-case" behavior was enforced in the first place. I think the reason might be related to the fact that Fortran is case-insensitive, so different writings of the same keyword or object name should be regarded as the same, but that is not the case in HTML and this causes problems for hyper-linking and references. This was likely a bit of work to implement, so the reason why the original authors decided to lower the case everywhere as a quick solution.

albert-github:
Unfortunately without knowing what you changed to solve the problem it is hard to tell whether or not it will be accepted, so you should make a PR out of it.

The exact reasoning for the usage of the choice of Fortran lowercase is not known to me, but to solve the problem one of course needs a standardized form of the different names and here there was chosen for the lowercase version and one has to store 2 versions of the name i.e. the name to display as used in the definition (e.g the definition of a subroutine i.e. the CamelCase name in your case) and the name to use in (internal) searches (i.e. the all lowercase name).
- What happens when the case of the definition is different from the case in e.g. a call (it is a violation of your coding conventions, but it can happen and is OK for Fortran)
- did you take the ISO C binding (i.e. bind(C,...) into consideration?

shahmoradi:
[...]

PR #8774, opened 2021-09-13
Revives Fortran case-sensitivity + full Fortran 2018 syntax highlighting

shahmoradi:
While Fortran is case-insensitive, modern-Fortran programmers (and most old-school Fortranners) rarely modify the case of Fortran entities throughout their codebase. So, if a decision is to be made on whether to keep case sensitivity, the readability of the generated documentation should be given more weight than the remote scenario where a Fortran developer uses mixed-case coding. In particular, there are ample Fortran codebases that use the CameCase style which becomes unreadable when Doxygen lowers the case of all letters. This PR of the Doxygen repository resolves this issue by removing all case-lowering instances of the Fortran entities in Doxygen. This effectively means that Doxygen treats Fortran code as case-sensitive, and Fortran users of Doxygen should respect case sensitivity in their codebase if fully-functional hyperlinking is desired. This does not affect other aspects of the documentation such as syntax highlighting and the functionalities related to the `iso_c_binding` module and C-Fortran interoperations. If any, it should only enhance C-Fortran interoperations as it interoperation rules defined by the Fortran standard are case-sensitive.

This PR also implements the following major enhancements to the Fortran syntax highlighting rules of Doxygen:

+   This PR implements a clean classification of (almost) all keywords in recent Fortran standards (2018/2008/2003/1995/1990/1977), fully consistent with the formal classification of keywords presented in [Fortran 2018 standard](https://j3-fortran.org/doc/year/18/18-007r1.pdf). Among the major categories are:
    1. Fortran attributes
    1. Fortran constructs
    1. Fortran statements
    1. Fortran intrinsic elemental functions
    1. Fortran intrinsic inquiry functions
    1. Fortran intrinsic transcendental functions
    1. Fortran intrinsic impure subroutines
    1. Fortran intrinsic pure subroutines
    1. Fortran intrinsic atomic subroutines
    1. Fortran intrinsic Coarray collective subroutines
    1. Fortran intrinsic IEEE modules and their entities
    1. Fortran intrinsic `iso_c_binding` module and its entities
    1. Fortran intrinsic `iso_fortran_env` module and its entities
    1. Fortran intrinsic specifiers (intrinsic-procedure argument names)
    1. Fortran intrinsic derived types

+ Over 600 new names of Fortran intrinsic functions, subroutines, modules, named constants, and specifiers (e.g., intrinsic procedure argument names) are now recognized and properly highlighted by Doxygen. The new set effectively covers all Fortran keywords in the latest Fortran standard 2018 and covers all keywords in 5 older standards F2008, F2003, F95, F90, F77.

+   Syntax recognition for MPI subroutines and entities, OpenMP directives, and Intel Fortran Compiler
    directives are now added.

+   The `real` and `logical` intrinsic functions are now properly recognized and highlighted as intrinsic functions or type declarations where appropriate.

+   Syntax recognition for Fortran `submodule`s and `module (function|subroutine|procedure)` now added.

+   Syntax recognition for Fortran intrinsic modules, including IEEE modules and their entities, `iso_fortran_enc` and `iso_c_binding` and their entities are now added.

+   Syntax recognition for new and modern Fortran constructs now added, for example, `block` and `associate` constructs.

+   Syntax recognition for all Fortran intrinsic Coarray parallel procedures and entities now added.

+   syntax recognition for new procedure attributes added (e.g., IMPURE, `(non_)?overridable`, ...).

+   The Fortran component separator symbol `%` is now recognized and highlighted. This is
    very useful for enhanced readability of Fortran codes as `%` is a rather busy symbol.
    
+   All Fortran intrinsic operators are now properly recognized and highlighted.

[Here is an example of Fortran documentation built with the modified Doxygen-Fortran implemented in this PR](https://cdslaborg.github.io/paramonte-kernel-doc/html/Array__mod_8f90_source.html).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+cd build && cmake -G "Unix Makefiles" .. && make || {
+    echo
+    echo "Doxygen build failed."
+    echo
+}
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+if ! [ -d build/ ]; then
+    mkdir build
+fi
 cd build && cmake -G "Unix Makefiles" .. && make || {
     echo
     echo "Doxygen build failed."
--- a/install.sh
+++ b/install.sh
@@ -2,7 +2,7 @@
 if ! [ -d build/ ]; then
     mkdir build
 fi
-cd build && cmake -G "Unix Makefiles" .. && make || {
+cd build && cmake -G "Unix Makefiles" .. && make -j || {
     echo
     echo "Doxygen build failed."
     echo
--- a/src/fortrancode.l
+++ b/src/fortrancode.l
@@ -244,10 +244,14 @@ CHAR      (CHARACTER|CHARACTER{BS}"*"({BS}[0-9]+|{ARGS}))
 /* The patterns ({NUM_TYPE}{KIND}) and ({NUM_TYPE}({BS}"*"{BS}[0-9]+)?) are inconsistent with Fortran syntax and as such, was removed from TYPE_SPEC */
 TYPE_SPEC ({NUM_TYPE}|DOUBLE{BS}COMPLEX|DOUBLE{BS}PRECISION|{CHAR}|TYPE|CLASS|PROCEDURE|ENUMERATOR)
 
-INTENT_SPEC         INTENT{BS}"("{BS}(IN|OUT|IN{BS}OUT){BS}")"
 ACCESS_SPEC         (PROTECTED|PRIVATE|PUBLIC)
+INTENT_SPEC         INTENT{BS}"("{BS}(IN|OUT|IN{BS}OUT){BS}")"
 
+/* Fortran attributes except BIND and C which are treated separately */
 ATTR_SPEC           (ALLOCATABLE|ASSIGNMENT|ASYNCHRONOUS|ABSTRACT|BIND|CODIMENSION|CONTIGUOUS|DEFERRED|DIMENSION|ELEMENTAL|EXTENDS|EXTERNAL|IMPURE|INTENT{BS}"("{BS}(IN|OUT|IN{BS}OUT){BS}")"|INTRINSIC|LOCAL|LOCAL_INIT|NON_INTRINSIC|NON_OVERRIDABLE|NON_RECURSIVE|NOPASS|OPERATOR|OPTIONAL|OVERRIDABLE|PARAMETER|PASS|POINTER|PRIVATE|PROTECTED|PUBLIC|PURE|RECURSIVE|SAVE|SEQUENCE|SHARED|TARGET|VALUE|VOLATILE)
+
+/* attributes that take arguments */
+ATTR_SPEC_WARG      (CODIMENSION|DIMENSION|EXTENDS|OPERATOR|PASS|BIND)
 
 /* Assume that attribute statements are almost the same as attributes. */
 ATTR_STMT           {ATTR_SPEC}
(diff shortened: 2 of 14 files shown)


albert-github:
I didn't really do a review yet but I did a small test.

With the example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7154547/example.tar.gz)
I.e.
**aa.f90**
```
!> \file


!> \brief The module
!> \details The module
module tst
contains
  !> \brief the first subroutine
  !> \details the first subroutine
  subroutine SUB1()
    call SUB2()
    call sub2()
    call Sub2()
  end subroutine Sub1
  !> \brief the second subroutine
  !> \details the second subroutine
  subroutine Sub2()
    write(*,*) "Hello"
  end subroutine SUB2
end module
```
**main.f90**
```
!> \file


!> \brief The main program
!> \details The main program
program pgm
  use tst
  call sub1()
end program
```
and *Doxyfile**
```
QUIET = YES
OPTIMIZE_FOR_FORTRAN = YES
CALL_GRAPH=YES
CALLER_GRAPH=YES
HAVE_DOT=YES
REFERENCED_BY_RELATION = YES
REFERENCES_RELATION    = YES
SOURCE_BROWSER = YES
STRIP_CODE_COMMENTS = NO
DOT_CLEANUP = NO
MAX_DOT_GRAPH_DEPTH = 1000
```

This gives in the 1.9.2 version:
image

and

image


With the patches as proposed in this pull request:

image

and

image


(I did manipulate the colors of the links to make them clearer).

On directly sees:
- not all calls to sub1 are linked in the code (due to the inconsistent spelling but it is legal in Fortran and happens quite often in old and legacy code)
- the call / caller graphs are not correct

shahmoradi:
Thanks for testing. The broken links are not surprising to me (it would be the expected behavior with this PR). But I understand your concern and the necessity of supporting the language the way it is. Two alternatives that I see are the following,
1. Going the way you suggested in issue [#7868](https://github.com/doxygen/doxygen/issues/8768#issue-990644045) (Unfortunately, I do not have the knowledge of implementing it yet).
2. implementing an extra option for Doxygen to respect case sensitivity when requested by the user at runtime on the command line.

The latter seems to be easier to implement than the former, but again I do not currently have a deep understanding of the inner workings of Doxygen to judge what is possible? Let me know please if you believe any one of the above solutions should become an issue and I will open one for it, or if there are people already working on these or similar ideas that I could contribute.

The other parts of the PR (syntax highlighting) could be likely salvaged if the case-sensitivity is not desired. The two are easily separable.

albert-github:
A command line option is not really an alternative, it would be an entry in the doxygen settings file (`Doxyfile`) but I think this is still a workaround and it is better to really solve the problem and store the display name and the search name (as I indeed suggested before).

install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+cd build && cmake -G "Unix Makefiles" .. && make || {
+    echo
+    echo "Doxygen build failed."
+    echo
+}
```
albert-github, line 6: In my opinion this is script should not be in the repository, it is a private script used by the user.

src/fortrancode.l
```diff
@@ -932,14 +932,14 @@ INTRINSIC_DERIVED_TYPE          (C_FUNPTR|C_PTR|EVENT_TYPE|LOCK_TYPE|TEAM_TYPE)
...
-
+ /*
 <*>^[Cc*].*                                             {   // normal comment
                                                             if(! yyextra->isFixedForm) YY_FTN_REJECT;
                                                             startFontClass(yyscanner,"fortran_comment");
                                                             codifyLines(yyscanner,yytext);
                                                             endFontClass(yyscanner);
                                                         }
-
+ */
```
albert-github, line 942: The old style comment is still standard Fortran.

albert-github (changes requested):

install.sh
doxygen: I agree

rajeeja: Is this good to go: on removal of the above script?

albert-github: I don't think so see my comment about the old style comment.

src/fortrancode.l
rajeeja: What is the fix for this?

albert-github: Most  likely the not inclusion of the lines 935 and 942, see also https://github.com/doxygen/doxygen/pull/8774#issuecomment-918355501

shahmoradi: The F77 comment style can be revived by removing the lines 935 and 942 as Albert says. However, adding it will mess up with Fortran90 source code lines that start with C or c. One option is to limit this comment style to .f and .for file extensions since this F77 comment style is not recognized and is illegal in the modern free-format Fortran files with extension .f90.
Another option is to modify the rule to better recognize the F77 comment-style where they appear without interfering with the syntax highlighting of free-format f90 source files.

albert-github: @shahmoradi 
Can you give an example where normally the free formatted code is messed up due to this rule?
I can imagine that this is happening in the code coloring as I see in the lexical scanner (to determine the syntax) there is an attempt to convert from fixed formatted code to free formatted code, for source code output this conversion cannot be done and something different has to be done and is done as well but maybe not everywhere.

rajeeja:
@shahmoradi I tried to get the version from your repo: 
```
(base) jain @ jain ~/projects/doxygen/build (master)
└─ $ ▶ git remote -v
origin	git@github.com:cdslaborg/doxygen.git (fetch)
```

My doxyfile has some dot and graph features, but they end up being self loops, when in reality they are referenced. (different from one pictured below)
image

shahmoradi:
> @shahmoradi I tried to get the version from your repo:
> 
> ```
> (base) jain @ jain ~/projects/doxygen/build (master)
> └─ $ ▶ git remote -v
> origin	git@github.com:cdslaborg/doxygen.git (fetch)
> ```
> 
> My doxyfile has some dot and graph features, but they end up being self loops, when in reality they are referenced. (different from one pictured below) image

@rajeeja Thanks for pointing this out. The graphs should look fine as long as case sensitivity is respected in the codebase. However, the dot-graph self-reference was a bug that I have now fixed in the lastest commit to the fork (not the oxygen main branch). The issue of submodules remains unresolved in this fork. 

Here is an example dot-graph generated with the latest commit:

image

src/fortrancode.l
shahmoradi: Yes, I do not have a minimal reproducible sample right now since I removed the F77 style comment rule in the fork longtime ago. But I remember this happened frequently in code sections within the documentation where Doxygen highlighted a `call xxx` statement starting at the first column as a comment line.

rajeeja:
cool, thanks for the fix.
 
The original version had caller graph (see pic below) also "Referenced by:" functions are missing, the version you have doesn't show caller graph: 

image

shahmoradi:
@rajeeja If you could post a minimal example that reproduces the bug, I'd appreciate it very much. Without the specific case, it is hard to figure out what is wrong.

rajeeja:
> @rajeeja If you could post a minimal example that reproduces the bug, I'd appreciate it very much. Without the specific case, it is hard to figure out what is wrong.
Understandable: How about this DOXYFILE: 
https://gist.github.com/rajeeja/5c785a17ef9ec59945f79ec9c16985b2

as I cannot share the private code, 
below is what the grep for the call of the function looks like (for the above picture):

./IO/IOMain/IO_output.F90:  !call IO_writeParticles.  If particles are not included in a simulation then
./IO/IOMain/IO_output.F90:  if(outputParticleFile) call IO_writeParticles( .false.)
./IO/IOMain/IO_writeCheckpoint.F90:  call IO_writeParticles( .true.)

shahmoradi:
@rajeeja Here is an example snapshot of the output by the most recent commit.

image

I am not sure what is missing here. Perhaps you can elaborate further.

PR #6422, opened 2018-08-03
Bug 774166 - Access to configuration variables in html

albert-github:
A more general implementation has been made where it is possible to access the configuration values everywhere.
The configuration values can be accessed by means of `$[...]`

Issue #6106, opened 2018-07-02
Access to configuration variables in html (Origin: bugzilla #774166)


doxygen:
status NEEDINFO severity *enhancement* in component *general* for *---*
Reported in version *1.8.12* on platform *Other*
Assigned to: Dimitri van Heesch

On 2016-11-09 21:15:02 +0000, Bob D. wrote:

    I think it would be nice to access to the doxygen configuration varibles in
    the html header, through $VARIABLE. In such way, we could have configuration
    specific modification of the header.
    
    That is only a suggestion.

On 2016-11-13 10:39:18 +0000, albert wrote:

    Can you be a bit more clear about which part you are speaking about/ what
    kind of changes you propose?
    
    Can you accomplish this with an own html header / footer ?
    (see "15.1.4 Header, Footer, and Stylesheet changes" of the doxygen manual).

On 2016-11-14 15:02:26 +0000, Bob D. wrote:

    Yep, sure! Sorry, I wasn't clear enough.
    
    What I was trying to do is to access a Doxygen variable (let say for example
    SHOW_INCLUDE_FILES) in the html file (my own html header).
    By modifying the html header, I was able to add a couple of new HTML tags.
    In one of it (a hidden div), I wanted to keep a project specific information
    that I could later process in a custom javascript.
    
    In my particular situation, I share a lot of the Doxygen configuration
    across my projects, including the custom header.html. But I needed to access
    a specific project value, that I couldn't share among projects. Thus, I
    tried to add a variable PROJ_VAR in the Doxygen configuration and I tried to
    access it by $PROJ_VAR or $(PROJ_VAR) in the header html. But in the html,
    it seems that every Doxygen variable are mapped differently ($projectname
    for PROJECT_NAME or $extrastylesheet for HTML_EXTRA_STYLESHEET). Thus, the
    variable was kept plain text (it was not processes). And since the flag that
    I use in the Doxygen configuration file is unknown to Doxigen, I have a
    warning saying that the flag is ignored. If there was a mapping there, where
    the value was replaced by what I entered in the configuration for (or a
    custom XML aside for instance), it would have been convenient.
    
    I hope that it is clearer. If not, feel free to ask more explanations!

albert-github:
I've just pushed a proposed patch, pull request #6422

albert-github:
An alternative pull request has been integrated into the doxygen 1.9.7 version ( #8779 ), so I think this problem has been solved in the mean time.

Issue #5517, opened 2018-07-02
Markdown - scope delimiter '::' disappears when using backticks with AUTOLINK_SUPPORT=NO (Origin: bugzilla #732922)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.8.7* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[doxysample.tar.gz](https://sourceforge.net/projects/doxygen/files/attachments/bug732922_280190_doxysample.tar.gz/download/download)_ (ID 280190)

On 2014-07-08 21:21:53 +0000, Pawel Tomulik wrote:

    The following README.md:
    
    
    TEST
    > ====
    > 
    > Plain:
    > 
    > - klass1<T>::foo
    > - klass1::bar
    > 
    > Single backticks:
    > 
    > - `klass2<T>::foo`
    > - `klass2::bar`
    > 
    > Double backticks:
    > 
    > - ``klass3<T>::foo``
    > - ``klass3::bar``
    > 
    > 
    > Source code:
    > 
    >     // Sample code
    >     klass4<T>::foo
    >     klass4::bar
    
    When run doxygen with AUTOLINK_SUPPORT=NO in Doxyfile, the "Single
    backticks" and "Double backticks" show as klass1<T>foo and klass2<T>foo
    respectivelly instead of klass1<T>::foo and klass2<T>::foo. When turn on
    autolink support, they're generated properly. It seems to be issue only with
    template classes. All the '::bar' nested compounds display fine.

On 2014-07-08 21:23:18 +0000, Pawel Tomulik wrote:

    Created attachment 280190
    A minimal sample project to replicate the issue

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.3.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #8779, opened 2021-09-15
Making settings from settings file available.

albert-github:
Making the settings from the doxygen settings / configuration file directly available inside the documentation.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7169045/example.tar.gz)


albert-github:
During debugging #5517 I found a use case, here it would be nice to have in the output the value of `AUTOLINK_SUPPORT`

albert-github:
I now found again the issue regarding the configuration values: #6106 (Access to configuration variables in html (Origin: bugzilla #774166)) apparently I already made another implementation for this is as well: #6422.
I have no real preferences for either one.

doxygen:
@albert-github I renamed `\doxysetting` to `\doxyconfig` and kept the values of obsolete and disabled options empty (does generate a warning).

Issue #8785, opened 2021-09-17
Markdown runs on <PRE> if class attribute is specified


elfprince13:
**Describe the bug**
`<PRE>` tags are supposed to be ignored by Markdown, to preserve ASCII art; however, [`<PRE class="markdeep">`](https://casual-effects.com/markdeep/#doxygen) is hit by the Markdown parser, thus breaking ASCII art.

**Expected behavior**
ASCII art should be preserved.

**To Reproduce**
Default configuration, see "Describe the bug".

**Version**
Doxygen version 1.9.2 (but compiled from source release to run on Ubuntu 18.04)

Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.5 LTS
Release:	18.04
Codename:	bionic

albert-github:
According to the https://spec.commonmark.org/0.30/#html-block it is true tat `<pre>` blocks should be treated as raw-HTML (and nit interpreted) and according to the original specs from https://daringfireball.net/projects/markdown/syntax#html there are some extra restrictions.


- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

albert-github:
I've just pushed a proposed patch, pull request #8786

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.3.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #8786, opened 2021-09-18
issue #8785 Markdown runs on <PRE> if class attribute is specified

albert-github:
A tag can have attributes and these were not taken into consideration as there was a hard test for a `>`.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7189692/example.tar.gz)


elfprince13:
I would propose making the check `std::isspace` to avoid other people running into variations of the same bug down the line.

albert-github:
Problem here will be, in my opinion, a character as `\n` I wouldn't like to see this is ine.g. a `pre` tag,  the   `<tab>` should have been replaced, internally, already with a space.

PR #8800, opened 2021-09-28
Add a warning message for an unbalanced \endcond

tomchy:


albert-github:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that shows the problem that should be fixed with the proposed patch? Please don't add external links as they might not be persistent.

src/pre.l
```diff
@@ -3182,6 +3182,7 @@ static void endCondSection(yyscan_t yyscanner)
   if (state->condStack.empty())
   {
     state->skip=FALSE;
+    warn(g_yyFileName,g_yyLineNr,"the \\endcond does not have a corresponding \\cond in this file");
   }
```
albert-github, line 3186: Your warning statement should probably read something like:
```
        warn(state->yyFileName,state->yyLineNr,"the ....
```

albert-github (changes requested):

src/pre.l
tomchy: Huh, you were 40 seconds faster!
I tried to describe everything in the issue, linked to the PR.

albert-github:
Seen the example you showed (similar to mine) with issue #8802 I don't think this solves the issue as this is correct code and correct nesting of the `\cond` / `\endcond`. It just throws a warinng but doesn't solve the nesting problem. I think the nesting can only be solved by means of checking the implementation of the stack in `pre.l`. 
When looking at `commentcnv.l` there is also `\cond` handling done but this is OK, this `\cond` handling can be reached by setting `ENABLE_PREPROCESSING=NO`. (Note: in `commentcnv.l`

I see also some small problems regarding regarding `\cond` when they are in one comment block, but this is logical as `\cond`  / \endcond` should not be in the same block (the `\if` is existing for this purpose), so I slightly modified my example and attached it with #8802

tomchy:
That's correct. This PR does not solve the issue, it simply throws a warning if it is reproduced, which allows for finding a nested `\cond`s in existing code and to look for possible workarounds (i.e. using a pure C `#ifdef`s).
Such warning is already present in the `commentcnv.l`.

What I was trying to add was an additional case, such as:
```
<SkipCond>[\\@]"cond"[!()&| \ta-z_A-Z0-9\x80-\xFF.\-]+  {
                                          startCondSection(yyscanner," ");
                                        }
```

Unfortunately my understanding of the toll's internals is not good enough to be 100% sure that it will not break something else -or- handle all cases.

tomchy:
@albert-github : What do you think about the fix?

When it comes to the question - why the regular processing logic does not match the `\cond` command, my theory is that it is skipped by the following piece of logic (there is a complementary piece for `SkipCPPComment` too):
```
<*>"/*"[*]?				{
                                          if (YY_START==SkipVerbatim || YY_START==SkipCond)
                                          {
                                            REJECT;
                                          }
                                          else
                                          {
                                          ...
                                          BEGIN(SkipCComment);
                                          ...
```

The idea behind the fix is that:
 -  You have to be sensitive to `\endcond` command to be able to get out of the skipping state. Since the command itself does not contain information about the expression that was checked it has to be every `\endcond` in the file.
  - There is no other way to say if we are getting out of the desired `\cond` section without counting `\cond` <-> `\endcond` pairs
   - It does not matter what section is pushed onto the stack as long as it has the correct `state->skip` value and there will be an item for each `'\endcond`.

albert-github:
I'm currently looking into the problem. It looks like that the `\endcond` is detected properly but in some cases the `\cond` isn't.
I hope I can find a solution soon.

albert-github:
I've just pushed a proposed patch, pull request #8804.
This patch is now obsolete as I also included the warning from this patch.

PR #8807, opened 2021-10-01
value attribute for <li> tag

albert-github:
With the `<li>` tag in HTML it is possible to have a `value=` attribute, this was only working for the HTML output, but has now been added to other output formats as well.
Also fixed small counting issue in case we have more than the supported number of levels.

(For docbook there is no solution yet)

### DIFF
--- a/src/latexdocvisitor.cpp
+++ b/src/latexdocvisitor.cpp
@@ -43,6 +43,8 @@ const int maxLevels=5;
 static const char *secLabels[maxLevels] =
    { "doxysection","doxysubsection","doxysubsubsection","doxyparagraph","doxysubparagraph" };
 
+LatexListItemInfo latex_listItemInfo[latex_maxIndentLevels];
+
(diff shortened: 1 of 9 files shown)


albert-github:
In case the pull request #6602 (or at least the `doxygen.sty` part has been merged beforehand the following has to be changed:
- `latexdocvisitor.h`
  change the line::
  ```
  const int latex_maxIndentLevels = 5;
  ```
  into:
  ```
  const int latex_maxIndentLevels = 13;
  ```
- `latexdocvisitor.cpp`
  change the line
  ```
  m_t << "\n\\setcounter{enum"<< integerToRoman(m_indentLevel,false).data() << "}{" << val - 1 << "}";
  ```
  into"
  ```
  m_t << "\n\\setcounter{DoxyEnumerate"<< integerToRoman(m_indentLevel,false).data() << "}{" << val - 1 << "}";
  ```

PR #8808, opened 2021-10-02
value attribute for <li> tag for docbook

albert-github:
In the docbook listitem tag there is no option for the implementation of the value attribute.
We have to use a workaround as suggested from the support person of XMLMind: close the current ordered list and start a new list (i.e. at the same level).
As we need to know the type of list (Arabic numbers, lower / upper case Roman) we have to "reparse" the attributes of the List again.
Also special care has to be taken when we have a value attribute with the first list item (we have to prevent that we get an empty ordered docbook list (this results in the m_isFirst field with the list items).

### DIFF
--- a/src/docbookvisitor.cpp
+++ b/src/docbookvisitor.cpp
@@ -936,33 +936,7 @@ void DocbookDocVisitor::visitPre(DocHtmlList *s)
 {
 DB_VIS_C
   if (m_hide) return;
-  if (s->type()==DocHtmlList::Ordered)
-  {
-    m_t << "<orderedlist";
-    for (const auto &opt : s->attribs())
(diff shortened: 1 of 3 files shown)


albert-github:
It looks like there is a solution after all (the override attribute with the listitem tag, see https://github.com/docbook/docbook/issues/222), I will have to check this out properly and check with the people of XMLMind as well.

albert-github:
The standard for the docbook is quite vague about the `override` and it states that it is just for the current item, so not really appropriate for the do0xygen / HTML value attribute usage.
The current implementation is the best that can be done at the moment.
See also https://github.com/docbook/docbook/issues/222

PR #8812, opened 2021-10-04
bug_684050 line continuation in \copydoc

albert-github:
Conform the documentation of `\copydoc` it cannot have an embedded linefeed and thus the `\copydoc` is terminated here when converted to `\copybrief` and `\copydetails`.
The rest of the "signature") is automatically added at the `\copydetails` leaving the `\copybrief` with just half a line (all can be observed with `doxygen -d commentscan`).
The round brackets were not handled and now a rudimentary  version has been implemented.

(Note the original issue contains 3 problems this is the last one of the 3 and about copydoc)

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -554,6 +554,7 @@ STopt  [^\n@\\]*
 %x      SkipLang
 %x      CiteLabel
 %x      CopyDoc
+%x      CopyDocExpr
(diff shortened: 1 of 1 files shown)


albert-github:
The functionality of this proposed pull request has been made, in a slightly different way,  available through #8977, so this pull request is obsolete.

Issue #8724, opened 2021-08-19
@ingroup not working with concepts


r-neal-kelly:
**Describe the bug**
When using the `@ingroup`, `@{`, and `@}` commands, only one concept is added to the group, even if there are several in the braces.

**Expected behavior**
Any concept in between the braces should be added to the group.

**To Reproduce**
[ingroup concepts.zip](https://github.com/doxygen/doxygen/files/7011106/ingroup.concepts.zip)

**Version**
Windows v1.9.2, using doxywizard.exe

**Additional context**
This is only a minor issue. One can use `@addtogroup` instead of `@ingroup` and that works correctly.

Thank you for Doxygen!

albert-github:
Looks like a regression from 1.9.1 -> 1.9.2

doxygen:
@albert-github Concept support was first added in 1.9.2, so I'm not sure about that.

albert-github:
I just looked at the HTML output, so it might have been giving, by accident, "good" looking information:

**1.9.1**
![image](https://user-images.githubusercontent.com/5223533/130058858-8c192a4d-f681-4591-ac27-da115818e367.png)

**1.9.2**
![image](https://user-images.githubusercontent.com/5223533/130058920-9c3c388d-8253-4db0-acf8-ba0e6b7ce053.png)


With a closer look we now see a heading `Concepts` in 1.9.2 whilst in  1.9.1 it is more a variable, so not really a regression.

albert-github:
When I modify the code a little bit to:
```
/**
* @ingroup concepts_ingroup
*/
template <typename>
concept concept_a = true;
/**
* @ingroup concepts_ingroup
*/
template <typename>
concept concept_b = true;
```

I get:
![image](https://user-images.githubusercontent.com/5223533/130060014-f5a6236d-9892-4092-b2b0-1798a92d2d7c.png)

Which looks OK to me.

The workaround is though a bit contradicting the way it is working for:
```
/**
* @ingroup functions_ingroup
*
* @{
*/
void function_a();
void function_b();
/**
* @}
*/
```

doxygen:
Please verify if the referenced commit fixes the problem.

albert-github:
I just added some documentation to the concepts:
```
/**
* @ingroup concepts_ingroup
*
* @{
*/
/**
 * \brief a concept
 * \details a concept
 */
template <typename>
concept concept_a = true;
/**
 * \brief b concept
 * \details b concept
 */
template <typename>
[...]

PR #8827, opened 2021-10-12
fix group close

### LINKED ISSUE #8815: Issue with "/// @}"
Sometimes single-line instructions 
```
/// @}
/// \}
/**  @} */
```
don't close groups as expected.

Multi-line one
```
/**
@}
*/
```
works as expected.

Reproduced and worked around in https://github.com/makelinux/examples/commit/9f6e350d8b4438d4de7c7d607ada8a5db2a7b549

makelinux:
I don't know if it is the right solution. At least it has new unit test and all tests are passed.

single-line instructions
```
/// @}
/// \}
/**  @} */
```
don't close groups as expected.

fix #8815

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -3428,6 +3428,7 @@ NONLopt [^\n]*
                                           {
                                             yyextra->mtype = Method;
                                             yyextra->virt = Normal;
+                                            yyextra->current->groups.clear();
                                             initEntry(yyscanner);
                                           }
                                         }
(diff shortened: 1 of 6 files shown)


makelinux:
@albert-github, what do you propose to do?

albert-github:
The fix should be in such a way that both cases i.e. the issue  #8815 and #8724 both work, probably @doxygen should have a look at it.

doxygen:
@albert-github Fixing this without breaking #8724 turned out to be quite tricky. Actually, #8724 used a construct for concepts that depended on functionality implemented for MemberDef's only, so my original fix was wrong. I now made it work for concepts as well. Maybe need to generalize this further so it also works for classes, structs and namespaces later on.

PR #8836, opened 2021-10-20
bug_774167  Left nav pane

albert-github:
In case a higher level visibility is set to `no` also the children should, automatically, not be visible.

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -122,7 +122,14 @@ LayoutNavEntry *LayoutNavEntry::find(LayoutNavEntry::Kind kind,
   return result;
 }
 
-
+void LayoutNavEntry::adjustVisibility()
+{
+  if (parent()) setVisible(visible() & parent()->visible());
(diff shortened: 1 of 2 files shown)


src/layout.cpp
```diff
@@ -122,7 +122,14 @@ LayoutNavEntry *LayoutNavEntry::find(LayoutNavEntry::Kind kind,
   return result;
 }
 
-
+void LayoutNavEntry::adjustVisibility()
```
doxygen, line 125: I would expect that instead of fixing the visibility after processing, it can also be set while parsing by changing

```c++
bool isVisible = elemIsVisible(attrib);
```

by e.g.

```c++
bool isVisible = elemIsVisible(attrib) && (parent()==0 || parent()->visible());
```

at various places. Did you try or consider this?

albert-github: No I didn't consider this, but I think that in that case also the recursive part is missing i.e. checking grandparent etc.

Issue #8528, opened 2021-04-30
Markdown links to Markdown pages with explicit page command are broken


arwedus:
Hi @albert-github , I have now created a separate task for the bug in #8378, I hope this can be fixed quickly... I didn't find a quick solution.

**Bug description**

If you start a markdown page with a `\page` command, Markdown links to the page lead to a non-existing page instead.

See also: #8378 

Example page:

```
\page p_input_handling Input Handling

# Input Handling

...
```

Example page 2:

```
subpages:
- \subpage p_input_handling input handling

Please refer to [input handling](doc/input_handling.md) documentation for details.
```

If a markdown page starts with just the L1 heading, links in Markdown style like `[output handling](doc/output_handling.md)` lead to the right page.

**Expected behavior:**

- Generated HTML for a relative markdown link to another markdown page works as expected and opens the page.

**To Reproduce:**

See attached example (much reduced version of #8378).

**Version:**

Tested on Doxygen 1.9.1, but present in older versions too.

[example_broken_md_link.tar.gz](https://github.com/doxygen/doxygen/files/6406309/example_broken_md_link.tar.gz)

albert-github:
I reduced the small problem even a little bit further to:

doc/input_handling.md
```
@page input-handling Input Handling

# Input Handling

Data is further processed by the [output handling](output_handling.md).
```

and doc/output_handling.md
```
# Output Handling

- runs after [input handling](input_handling.md) (broken link)
- format output data after triggered by \ref input-handling (this link works)
```

To understand what happen we have to run doxygen so it outputs as well the result of the internal markdown conversion (so `doxygen -d markdown`):

For input_handling.md this is:
```
======== Markdown =========
---- input -------
@page input-handling Input Handling

# Input Handling

Data is further processed by the [output handling](output_handling.md).

---- output -----
@page input-handling Input Handling

@section autotoc_md0 Input Handling

Data is further processed by the @ref .../issue_8528/example_broken_md_link/doc/output_handling.md "output handling".

=========
```
and for output_handling.md:
```
======== Markdown =========
---- input -------
@page md_doc_output_handling Output Handling\ilinebr

- runs after [input handling](input_handling.md) (broken link)
- format output data after triggered by \ref input-handling (this link works)

---- output -----
@page md_doc_output_handling Output Handling\ilinebr

- runs after @ref .../issue_8528/example_broken_md_link/doc/input_handling.md "input handling" (broken link)
[...]

PR #8896, opened 2021-11-17
issue #8528 Markdown links to Markdown pages with explicit page command are broken

albert-github:
In case a  `@page` command is the first line in a markdown file it is used as  "label" but also to construct the name of the output file and hence the regular markdown reference (like `[...](..)`) does not work anymore as here an internal label and filename is used  (`md_...`).
To overcome this problem the `@page <label> (title)` is changed into `@page md_<filename> (title)\ilinebr @anchor <label>`.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -195,6 +195,12 @@ int Trace::s_indent = 0;
    data[i]=='\\' || \
    data[i]=='@')
 
+enum class ExplicitPage
+{
+  notExplicit,
+  explicitPage,
+  explicitMainPage
+};
 //----------
 
 struct TableCell
@@ -2849,8 +2855,11 @@ QCString Markdown::processBlocks(const QCString &s,const int indent)
   return m_out.get();
 }
 
-/** returns TRUE if input string docs starts with \@page or \@mainpage command */
-static bool isExplicitPage(const QCString &docs)
+/** returns ExplicitPage::explicitPage in case input string docs starts with page command
+ *  returns ExplicitPage::explicitMainPage in case input string docs starts with mainpage comand
+ *  returns ExplicitPage::notExplicit otherwise
+ */
+static ExplicitPage isExplicitPage(const QCString &docs)
 {
   TRACE(docs);
   int i=0;
@@ -2867,12 +2876,20 @@ static bool isExplicitPage(const QCString &docs)
         (qstrncmp(&data[i+1],"page ",5)==0 || qstrncmp(&data[i+1],"mainpage",8)==0)
        )
     {
-      TRACE_RESULT(TRUE);
-      return TRUE;
+      if (qstrncmp(&data[i+1],"page ",5)==0)
+      {
+        TRACE_RESULT(ExplicitPage::explicitPage);
+        return ExplicitPage::explicitPage;
+      }
+      else
+      {
+        TRACE_RESULT(ExplicitPage::explicitMainPage);
+        return ExplicitPage::explicitMainPage;
+      }
     }
   }
-  TRACE_RESULT(FALSE);
-  return FALSE;
+  TRACE_RESULT(ExplicitPage::notExplicit);
+  return ExplicitPage::notExplicit;
 }
 
 QCString Markdown::extractPageTitle(QCString &docs,QCString &id, int &prepend)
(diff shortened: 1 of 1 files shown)


arwedus:
@albert-github: I'm very happy that you took the time to fix this now. I tested it locally on my example I created for this issue and the output works as expected. If I get around to it, I will test on a bigger project tomorrow.

src/markdown.cpp
```diff
@@ -53,6 +53,13 @@
 #include "fileinfo.h"
 #include "utf8.h"
 
+enum class ExplicitPage
+{
```
arwedus, line 57:
```suggestion
/** Used to differentiate between markdown page handling variants.
 */
enum class ExplicitPage
{
```

src/markdown.cpp
```diff
@@ -135,6 +142,11 @@ class Trace
       m_resultSet = true;
       m_resultValue = b ? "true" : "false";
     }
+    void setResult(ExplicitPage ep)
```
arwedus, line 145: inconsistent code formatting: In other source files, class methods are separated by a single empty line each.

src/markdown.cpp
```diff
@@ -3107,30 +3129,65 @@ void MarkdownOutlineParser::parseInput(const QCString &fileName,
...
-    for (int i = 0; i < prepend; i++) docs.prepend("\n");
+  static const reg::Ex re(R"([\\@]page[ \t]*)");
+  static const reg::Ex re0(R"(^[a-z_A-Z\x80-\xFF])"); // from LABELID, see commentscan.l
+  static const reg::Ex re1(R"([a-z_A-Z\x80-\xFF][a-z_A-Z0-9\x80-\xFF\-]*)"); // LABELID, see commentscan.l
+  static const reg::Ex re2(R"(\n)");
+  reg::Match match;
+  reg::Match match1;
+  std::string t;
+  std::string t1;
+  switch (isExplicitPage(docs))
```
arwedus, line 3140: you may want to move the variable definitions to the innermost scope where they are used (in this case, the body of the switch... case explicitPage branch could be put into a block scope {} ). see also:  https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rnr-top

albert-github: Thanks for the feedback.
Indeed, I had the code there before and got warnings like:
```
...\doxygen\src\markdown.cpp(3191): error C2360: initialization of 't1' is skipped by 'case' label
...\doxygen\src\markdown.cpp(3167): note: see declaration of 't1'
```

but that was because I forgot to put the case part here in a block scope.

arwedus:
I tried it now in one of our big projects and the broken links are fixed there, too. Links to pages which have a page reference marker on the same line still work as well.

One thing I found weird though, is this:

Given a file names awesome.md with following content:

```
# Awesome Page Title  {#p_awesome_page}
```

This gives a page name like p_awesome_page.html

but a file named awesome.md with:

```
\page p_awesome_page "Awesome Page Title"

# Awesome Page Title
```

gives a page name like md_path_to_filename__awesome.html, i.e. md_<path>__<filename>.html

Is this intentional? I would Other than that it looks good to me and it works, so I'm happy with the change.

albert-github:
Yes this is done on purpose. The request was to be able to address the page by means of the `p_awesome_page `.
To accomplish this an extra anchor has to be created with this name but at that moment the `@page` label / ID cannot have the same name anymore (page will not be shown), so for the page the label/ID is based on the file name.
When you run with `doxygen -d markdown` you will see:
```
======== Markdown =========
---- input -------
\page p_awesome_page "Awesome Page Title"

# Awesome Page Title

---- output -----
\page md_awsome "Awesome Page Title"\ilinebr @anchor p_awesome_page

@section autotoc_md0 Awesome Page Title

=========
```
and
```
======== Markdown =========
---- input -------
# Awesome Page Title  {#p_awesome_page}

---- output -----
@page p_awesome_page Awesome Page Title\ilinebr @anchor md_awesome\ilinebr

=========
```

arwedus:
@doxygen ready to merge? Or new automated test required?

doxygen:
@albert-github I squashed the individual commits onto one commit, and added a new one to simplify the regex logic a bit, but now GitHub doesn't recognize it anymore as a merge. Please verify if all is ok, and then close the pull request.

albert-github:
Pity that the merge didn't succeed, but no big issue. The PR could do with a squash as there were some intermediate compilation problems on my side (maybe I should have created a new PR.

Regarding the fix. It looks OK, I tried in respect to the labels close to the LABEL ID as specified in commentscan.l so that there would not be unexpected anchors etc. later on.

Issue #7807, opened 2020-05-27
`*` in line in \code block causes `*` on subsequent lines


paul-j-lucas:
Doxygen 1.8.18, when run on this file [bug.doxy.gz](https://github.com/doxygen/doxygen/files/4691282/bug.doxy.gz) with this configuration file [Doxyfile.gz](https://github.com/doxygen/doxygen/files/4691286/Doxyfile.gz), produces this incorrect HTML file [_bug.html.gz](https://github.com/doxygen/doxygen/files/4691290/_bug.html.gz) that looks like this:
<img width="632" alt="screen_shot" src="https://user-images.githubusercontent.com/7366785/83064030-f990df80-a015-11ea-9f54-428398e96c4f.png">
That is, the `*` on the first line inside the `\code` block incorrectly causes `*` to be placed at the beginning of all subsequent lines within that block.

paul-j-lucas:
I think it's because it parses `/*` as the start of a comment (which it shouldn't be doing inside a `\code` block.

albert-github:
The problem is indeed due to the `/*. This is due to the fact that there is, after an internal doxygen comment conversion, a nested comment and this is not allowed.

Possible solutions (the first 2 are the best ones in my opinion):
- why use cpp as "language" and not just markdown (i.e. `md`), has the advantage that you can also remove the comment signs.
- don't use `code` / `\endcode` but ` ``` `
- don't use `\code` but put (at least) 4 spaces befor `rm` and `mkdir`:
   ```
   ///     rm -rf /root/chroot/*
   ///     mkdir -p /root/chroot
   ```
- convert the comment the `/**` comment without preceding `*` on the following lines and use `\verbatim` instead of `\code`

paul-j-lucas:
> This is due to the fact that there is, after an internal doxygen comment conversion, a nested comment and this is not allowed.

`@code` should pass through everything as-is until `@endcode`. It shouldn't be trying to parse internal comments at all.

> why use cpp as "language" and not just markdown

Because the larger body of documentation makes reference to C++ symbols in the codebase that is also documented with Doxygen. If the Doxygen file were Markdown, it wouldn't auto-link to those symbols.

> don't use `\code` / `\endcode` but ```

That works (so thanks for the workaround), but those should be treated exactly the same. ``` is just the Markdown syntax for `@code`.

> use `\verbatim` instead of` \code`

But `\verbatim` doesn't put the box around the text (which I like).

[...]

PR #8908, opened 2021-11-23
issue #7807 Incorrect handling of triple backticks with specifying language.

albert-github:
When having:
~~~
/** \file
 *
 * ```{.c}
 *   // backticks c ext
 *   void fie_b4(void);
 * ```
*/
~~~
the triple backticks with language are handled properly, though when having:
~~~
/** \file

```{.c}
  // backticks c ext
  void fie_b4(void);
```
*/
~~~
this is not the case.

Note: when using triple tildes instead of backticks it does work as well.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7587860/example.tar.gz)

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -6625,7 +6625,7 @@ NONLopt [^\n]*
                                             REJECT;
                                           }
                                         }
-<DocBlock>[^@*~\/\\\n]+                 { // any character that isn't special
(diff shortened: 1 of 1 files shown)


albert-github:
Gives some problems with having an inline backticked block, like
~~~
some text `<pre>`
~~~
or
~~~
some text ```<pre>```
~~~

PR #8912, opened 2021-11-25
Give markdown verbatim  and fenced code blocks their own command.

albert-github:
When we have e.g. ` ``` ` doxygen translates this to a `\code` .. `\endcode` block and this gives some hard to understand warnings when we have in the fenced block the text `\endcode`.
(analogous for a 4 space extra indentation that is seen as a `\verbatim` block). By defining new, internal, commands for this the problem can be overcome.

In Python there is also an under the hood conversion and this benefits also from these new commands.

(Checked against CGAL, found some problems there as well)


albert-github:
Ran again into the problem for the tor package with Fossies

Issue #8921, opened 2021-11-29
Overloaded function (with templates as change) is not displayed in html page


zchrissirhcz:
**Describe the bug**
I' like to display the following 3 functions in doxygen web page, the 2nd and the 3rd are overloaded each other:
```c++
void nice();

void hello(const char* name);

template<typename T>
void hello(const char* name);
```

**Expected behavior**
The web page should display two `hello()` function.

**Screenshots**
![图片](https://user-images.githubusercontent.com/3831847/143830913-96f4000b-12ce-4501-9d88-86e8d5357162.png)

As illustrated in the screenshot, there is only one `hello()` function.

**To Reproduce**

[test.tar.gz](https://github.com/doxygen/doxygen/files/7616581/test.tar.gz)

**Version**
```
doxygen --version

1.9.3 (c83813394ceb75aff58341f70f86fad87f8c97d0)
```

**Stack trace**

**Additional context**
The C/C++ compiler I'm using is `clang version 13.0.0 (https://github.com/llvm/llvm-project/ 24c8eaec9467b2aaf70b0db33a4e4dd415139a50)
`

zchrissirhcz:
So let's make it more suitable to debug. Based on the previously pasted C++ code, now adding `@brief` for the tempalted `hello()`, i.e.

```c++
#pragma once

void nice();

void hello(const char* name);

/// @brief the templated one
template<typename T>
void hello(const char* naming);
```

Then the generated html would absorb the templated `hello()`'s brief message to the non-templated one:

![图片](https://user-images.githubusercontent.com/3831847/144428702-c61bf622-b78f-4430-9cd1-8ced0d09103f.png)

zchrissirhcz:
Now turn on the debug print for functions:
```c++
//Debug::DebugMask Debug::curMask = Debug::Quiet;
Debug::DebugMask Debug::curMask = Debug::Functions;
```

And get:
```
FUNCTION_SEC:
  'void' ''::'nice' '()' relates='' relatesType='0' file='/home/zz/work/test/doxygen_example3/src/hello.h' line='3' bodyLine='-1' #tArgLists=0 mGrpId=-1 spec=0 proto=1 docFile=
  --> new function nice found!
  Global Function:
    'void' ''::'nice' '()' proto=1
    def='void nice'
FUNCTION_SEC:
[...]

PR #8930, opened 2021-12-02
Fix: display template function when there's same param list none temp…

zchrissirhcz:
Hi, Doxygen developers

This PR try to fix https://github.com/doxygen/doxygen/issues/8921 . This works for my provided example in the issue.
Hope this works fine for other cases.


albert-github:
The problem has been fixed in a slightly different fix than this pull request (see d80f4e4)

PR #8931, opened 2021-12-02
Incorrect markdown links.

albert-github:
When having a space between  the `]` and `(` of a potential markdown link this is not a markdown link.
Example:
```
[is a link](aa.html)
[is not a link] (aa.html)
```

This was found in the plantuml package  where the following shows as a link:
```
[3, 3, 3, 3, 3, 3, 3, 3] (result: A=000, B=001, C=010, ..., H=111)
```

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7642867/example.tar.gz)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -972,10 +972,12 @@ int Markdown::processLink(const char *data,int,int size)
   if (!isImageLink && content.isEmpty()) { TRACE_RESULT(0); return 0; } // no link text
   i++; // skip over ]
 
+  bool whiteSpace = false;
   // skip whitespace
-  while (i<size && data[i]==' ') i++;
+  while (i<size && data[i]==' ') {whiteSpace = true; i++;}
   if (i<size && data[i]=='\n') // one newline allowed here
   {
+    whiteSpace = true;
     i++;
     nl++;
     // skip more whitespace
@@ -983,6 +985,7 @@ int Markdown::processLink(const char *data,int,int size)
   }
   nlTotal += nl;
   nl = 0;
+  if (whiteSpace && i<size && (data[i]=='(' || data[i]=='[')) return 0;
 
   bool explicitTitle=FALSE;
   if (i<size && data[i]=='(') // inline link


src/markdown.cpp
```diff
@@ -972,17 +972,20 @@ int Markdown::processLink(const char *data,int,int size)
...
   {
+    whiteSpace = true;
     i++;
     nl++;
     // skip more whitespace
     while (i<size && data[i]==' ') i++;
   }
   nlTotal += nl;
   nl = 0;
+  if (whiteSpace && i<size && (data[i]=='(' || data[i]=='[')) return 0;
```
doxygen, line 988: It looks like a space is not allowed for an inline link but it is for reference link, see https://www.markdownguide.org/basic-syntax/#reference-style-links, so I think check for `data[i]=='['` should be removed. To complicate things a newline is not allowed, so probably spaces and newlines need to be tracked separately.

albert-github: I think this depends a bit on what the outcome should be. When having:
```
Square bracket
[hobbit-hole][1]

Square bracket, now with space
[hobbit-hole] [1]

End of text

[1]: https://en.wikipedia.org/wiki/Hobbit#Lifestyle
```
in the implementations on:
- https://spec.commonmark.org/dingus/
- https://markdownlivepreview.com/
- Here on GitHub:
   Square bracket
   [hobbit-hole][1]

   Square bracket, now with space
   [hobbit-hole] [1]

   End of text

   [1]: https://en.wikipedia.org/wiki/Hobbit#Lifestyle
- stackexchange

we get (I made the link red for clarity):

image

with my proposed patch we get the same, when allowing the space in it we would get:

image


So I'm not sure whether or not the current implementation isn't correct.

doxygen: hmm, the commonmark spec mentions a [shortcut reference link](https://spec.commonmark.org/0.30/#shortcut-reference-link). This is what the detached `[1]` is then interpreted as, so you are right if we use this spec as the reference.

Issue #8933, opened 2021-12-03
Return type "unsigned int constexpr" not parsed properly


ralfulrich:
**Describe the bug**

In a valid and working C++17 header file, whenever I use return types of type `unsigned int constexpr` this results in sphinx WARNINGs since apparently it mixes up while parsing my code (see output below). For example, the perfectly fine C++17 function:
```
/**
   * Get the nucleus mass number A.
   *
   * @param code for nucleus
   * @return unsigned int constexpr
   */
  inline unsigned int constexpr get_nucleus_A(Code const code) {
    return (static_cast<CodeIntType>(code) % 10000) / 10;
  }
```
produces the following sphinx WARNING:
```
Running Sphinx v3.5.3

 WARNING: Error when parsing function declaration.
If the function has no return type:                                                                                                                                                                                                                             
  Error in declarator or parameters-and-qualifiers                                                                                                                                                                                                              
  Invalid C++ declaration: Expected identifier in nested name, got keyword: unsigned [error at 8]                                                                                                                                                               
    unsigned constexpr int get_nucleus_A (Code const)                                                                                                                                                                                                           
    --------^                                      
```

If I replace `unsigned int` with `size_t` this warning disappears, but isn't there another less intrusive way to fix this? 


**Expected behavior**
No warning, and no "unsigned constexpr int"

**Version**
This is on ubuntu 20.04 using doxygen 1.8.17 

Originally asked at: https://stackoverflow.com/questions/69882899/unsigned-int-constexpr-in-doxygen-sphinx-c17

albert-github:
I've just pushed a proposed patch, pull request #8939

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.3.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #8939, opened 2021-12-03
issue #8933 Return type "unsigned int constexpr" not parsed properly

albert-github:
Handle `constexpr` similar to `const`

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -2310,7 +2310,7 @@ NONLopt [^\n]*
                                                 }
                                                 yyextra->current->name= yyextra->current->name.mid(7);
                                               }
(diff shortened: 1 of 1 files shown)


src/scanner.l
```diff
@@ -2315,6 +2310,18 @@ NONLopt [^\n]*
                                                 }
                                                 yyextra->current->name= yyextra->current->name.mid(7);
                                               }
+                                              else if (yyextra->current->name.left(10)=="cconstexpr ")
```
doxygen, line 2313: I think there is a typo here (extra `c` before `constexpr`), right?

albert-github: 100% right, fixed

PR #8953, opened 2021-12-08
Improve markdown image handling

bollenn:
# Add check for non-inlined images
Images which do not have text before and after the image command will be treated as non-inline images and as such can for example have image numbers in Latex...

Inlined would be:
```md
This ![doxygen logo](https://www.doxygen.nl/images/doxygen.png) is the doxygen logo
```

While non-inlined is considered:

```md
Below logo is the doxygen logo
![doxygen logo](https://www.doxygen.nl/images/doxygen.png)
```

# Add option to add attributes to images

Images can be written in markdown as
```md
![image caption](link){attributes}
```
`Attributes` will be appended to the Doxygen image command as in

```
@image format link "image caption" attributes
```

A useful example would be:

```md
![doxygen logo](https://www.doxygen.nl/images/doxygen.png){width=5cm}
```


bollenn:
@doxygen @albert-github I'm starting to wonder if we would need something like:
```md
![image caption](link){format: attributes, format: attributes}
```

As `width=5cm` fits latex but does not fit html; and vice versa `width=50%` fits html but not latex...

albert-github:
In principle it looks like a useful addition, but I have some doubts seen the following points:
- Is there a reference to a standard where the optional `{...}` is mentioned? (doxygen tries a bit to follow: https://daringfireball.net/projects/markdown/syntax and https://spec.commonmark.org/)
- it looks to me a doxygen specific extension so it should also be mentioned in the paragraph for Markdown Extension
- Indeed HTML / LaTeX / RTF / Docbook ... all have their own way of specifying width / height etc. and it is often quite hard to convert from cm to e.g. pixels etc as this is not always known at the moment to the generating code (e.g. the dpi might be set somewhere else). It looks indeed that something like `{format: attributes, format: attributes}` is needed (a start about how this is done, for a completely other purpose and use case type though, can be found with the implementation of the `\tableofcontents` command)

bollenn:
@albert-github I updated the MR. For me this is ready for review/merge...

src/markdown.cpp
```diff
@@ -352,6 +352,37 @@ static Alignment markersToAlignment(bool leftMarker,bool rightMarker)
...
+    {
+      // multiple formats
+      QCString format = attr.left(i).stripWhiteSpace();
+      if (!format.isEmpty())
+      {
+        if (format == fmt)
+        {
+          retval = attr.right(attr.length() - i - 1).data();
+          break;
+        }
```
albert-github, line 373: Wouldn't this prevent something like:
```
...{html: width=5cm, html: height=50%}
```

src/markdown.cpp
```diff
@@ -937,6 +982,22 @@ int Markdown::processLink(const char *data,int,int size)
       TRACE_RESULT(0);
       return 0;
     }
+    // check if there is text before '![' to know if the image needs to be inlined
```
albert-github, line 985: I don't get this, do you imply here that only when the `![` is not at the beginning of of line it will be in-lined? What happens when you have it in  block of text and due to layouting the image would start at the beginning of a line.?

albert-github (commented):
Some remarks.

src/markdown.cpp
bollenn: I've done some updates in 715298173ac90d9d60516a1a711c575f688a6a54.

So at least 2 newline before and after image tag, of course whitespaces are ignored when checking...

src/markdown.cpp
bollenn: should we consider such situation or just document first item is taken?

albert-github: I think we should consider this situation.
I think that it should be possible to have e.g.
```
{html: width=5cm, html: height=50%, latex: width=4cm, latex: height=40%}
```
I don't think something like:
```
{html: width=5cm, height=50%, latex: width=4cm, height=40%}

```
should be supported (is more compact but will be harder to parse and would probably be less flexible).

bollenn: At this moment the format would be 

```
![caption](image.png){width=50% height=10%}
![caption](image.png){html: width=100px height=250px, latex: width=5cm height=10cm}
```

So I consume all attributes after the first valid format definition and pass them on to the image command which is defined and used as:

```
\image['{'option[,option]'}'] <format> <file> ["caption"] [<sizeindication>=<size>]
\image html image.png width=100px height=250px
\image latex image.png width=5cm height=10cm
```

I'm not sure adding format definition for each sizeindication is making things clearer but can be done...

doxygen:
@bollenn @albert-github I've squashed the commits into one, and polished the results a bit. Please check if I didn't break anything in the process.

albert-github:
I think this issue can be closed see comment https://github.com/doxygen/doxygen/pull/8953#issuecomment-999056752.
@bollenn please test @doxygens fix.

Letme:
Why did we close this and not merge?

albert-github:
@Letme see the discussion beforehand and especially https://github.com/doxygen/doxygen/pull/8953#issuecomment-999394213

Letme:
You realize you have just deprived author of this merge request for his contribution as now there is no commit about this feature in his name https://github.com/doxygen/doxygen/commits?author=bollenn ? I mean open source and a lot of various tools work on the Signed-off-by or author tags to determine activity, yet now it sees like his feature was rejected (not merged) and the commit on the main branch is not by the author of this feature. Do you think that motivates people to contribute if you steal contributions?

albert-github:
@Letme The effort by @bollenn is highly valued by me but the owner / merger of the repository found it better to squash the contribution due to the changes and some improvements necessary (I also think that the owner / merge values the contribution of @bollenn as otherwise he would not have used it).

I think it still motivates the right people as they see that their work, they probably need for their work / hobby, is in the product.

Letme:
You said, it motivates people that need the fix for their work/hobby, but that is not motivation but necessity. Motivation is that your contribution is in one of the most known tools in the world.

Imagine there is a bug in this feature, so bisect shows commit https://github.com/doxygen/doxygen/commit/5bb622158e33aac19d23068c08f434aaced60b4c which in no way mentions author. It is also a fairly big block of code, so it will take more time than finding an exact change from this commit from this PR, not to mention get the reasoning if it came from review or original author (again @bollenn is not mentioned anywhere).

I suggest you review your merging strategy as Closed Pull Requests instead of Merged Pull requests show that you declined the contribution, not accepted it. And as you can see this Pull request is marked as "Closed". Contribution of author is also not visible on his GitHub profile page, it is not visible on the Contributions page of this repository, so effectively for all reviews this feature was done by @doxygen .

doxygen:
@letme I didn't consider this when merging in the changes. Since there was a lot of rework/iterations done I had problems overseeing the end result and didn't want to pollute the commit history, so I decided to look at the changes after squashing the commits and I also made some changes myself. 

I probably should have asked the author to do the squashing for me instead and use a different merge strategy or keep all the intermediate commits. I did reference to the pull request in my commit message, so it is still traceable and it for sure it not an attempt to hide the source or deprive the author of his valuable contribution. 

I would like to hear from @bollenn if he is upset/offended by this, since I did not know if you speak on his behalf. And if this is the case, I apologize and try to do a better job next time.

Letme:
@doxygen I am just a person interested in the feature/user, so just passing by. Still, I was surprised by the actions taken. I would suggest you either keep commits for later debugging or give a possibility for the author to squash his/her own commits in the future.

doxygen:
@letme I used the "easy" approach outlined [here](https://gist.github.com/patik/b8a9dc5cd356f9f6f980) for the first time. Didn't think about the side-effects of this.

bollenn:
@doxygen I would have to admit that I was a bit disappointed seeing the process in this PR.

Of course I'm happy that besides my previous 2 contributions this one might make it into a release soon so that we can keep using this great tool...

For me contributing to open source is 2 folded
* Helping the community to maintain even greater tools
* Showing of skills to the world

The first is fulfilled. The second is not. Or at least not easily recognizable...

If it's about having a clean git history there should be an option in the PR to enable squash (and rebase) when pressing merge.

bollenn:
> ommits into one, and polished the results a bit. Please check if I didn't break anything in the process.

@doxygen I tested the updates and made https://github.com/doxygen/doxygen/pull/8978

PR #8958, opened 2021-12-09
Latex writer shall not prepend labels with filename

bollenn:
Latex writer shall not prepend labels with filename


bollenn:
Probably this prepending is intentional to overcome duplicate links in single pdf...

PR #8960, opened 2021-12-10
Latex: allow using percentage numbers for images dimensions

bollenn:
Today it is not possible to use `%` attributes to images when the output format is latex since the `%` is considered a comment in Latex and as such the generated tex code is invallid.

With this MR I convert the `%` into something understandable for Latex.

### DIFF
--- a/src/latexdocvisitor.cpp
+++ b/src/latexdocvisitor.cpp
@@ -52,6 +52,32 @@ static const char *getSectionName(int level)
   return secLabels[std::min(maxLevels-1,l)];
 }
 
+static void insertDimension(TextStream &t, QCString dimension, const char *percentString)
+{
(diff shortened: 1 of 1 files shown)


src/latexdocvisitor.cpp
```diff
@@ -52,6 +52,32 @@ static const char *getSectionName(int level)
...
+  int i = dimension.find('%');
+  if (i > 0)
+  {
+    QCString value = dimension.left(i).stripWhiteSpace();
+    int number = 0;
+    if (sscanf(value.data(),"%d",&number) != 0)
+    {
+      t << (double)number / 100;
+      t << "\\text";
+      t << percentString;
```
albert-github, line 68: Maybe good to mention in the documentation (with the `\image` command) that in case of a percentage the percentage is taken relative to the current textwidth / textheight

PR #8969, opened 2021-12-15
bug_365053 Wrong reference to ::classname

albert-github:
Identified few places where the global class prefix (`::`) was not handled.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -889,12 +889,12 @@ void linkifyText(const TextGeneratorIntf &out, const Definition *scope,
     bool keepSpaces,int indentLevel)
 {
   if (text.isEmpty()) return;
-  //printf("linkify='%s'\n",text);
+  //printf("linkify='%s'\n",qPrint(text));
   std::string txtStr=text.str();
   size_t strLen = txtStr.length();
   if (strLen==0) return;
 
-  static const reg::Ex regExp(R"(\a[\w~!\\.:$]*)");
+  static const reg::Ex regExp(R"(:?:?\a[\w~!\\.:$]*)");
   reg::Iterator it(txtStr,regExp);
   reg::Iterator end;
 
--- a/src/code.l
+++ b/src/code.l
@@ -1534,7 +1534,7 @@ ENDQopt ("const"|"volatile"|"sealed"|"override")({BN}+("const"|"volatile"|"seale
                                           yyextra->code->codify(yytext);
                                           endFontClass(yyscanner);
                                         }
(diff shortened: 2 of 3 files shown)


src/util.cpp
```diff
@@ -889,12 +889,12 @@ void linkifyText(const TextGeneratorIntf &out, const Definition *scope,
...
 {
   if (text.isEmpty()) return;
-  //printf("linkify='%s'\n",text);
+  //printf("linkify='%s'\n",qPrint(text));
   std::string txtStr=text.str();
   size_t strLen = txtStr.length();
   if (strLen==0) return;
 
-  static const reg::Ex regExp(R"(\a[\w~!\\.:$]*)");
+  static const reg::Ex regExp(R"(:?:?\a[\w~!\\.:$]*)");
```
bangerth, line 897: I have to admit that I don't know what the regex is actually used for, but since you're proposing this in the context of global scope qualifiers, shouldn't this be
```suggestion
  static const reg::Ex regExp(R"((::)?\a[\w~!\\.:$]*)");
```

albert-github: I do agree that `(::)?`  instead of `:?:?` would have been a better solution, though the `reg:Ex` (see regex.cpp and regex.h in the doxygen src directory) does not handle this (properly), hence I had to choose the less good solution.
In theory it would now be possible to write something like `:A` but this would not be a valid syntax (and doxygen assumes valid code, in case of invalid code the behavior is undefined). Before committing the pull request I already did some tests and doxygen just ignored the single colon to be part of a link and, more important, did not crash .

When `(::)?` is supported by the `reg:Ex` the mentioned test should be rewritten.

(I should have mentioned this all before in the code / this pull request).

doxygen: Indeed. Support for optional ranges is missing in `reg::Ex` at the moment. I only implemented what I needed to migrate away from `QRegExp` and using `std::regex` caused a too big performance hit. I will add this.

bangerth: I'm not familiar with the doxygen code base, so this may be a dumb question: Are you saying that the `reg::Ex` class does not support `(...)?` expressions, only individual characters followed by `?`?

albert-github: As far as I could see / tested yes. The class is tailored to the usage in doxygen, so apparently till now we didn't need it.

albert-github: As far as I could see / tested yes. The class is tailored to the usage in doxygen (like @doxygen already said), so apparently till now we didn't need it.

albert-github: As far as I can see the pull request has been integrated in master on GitHub including a change for the `:?:?` to `(::)?`.

PR #8970, opened 2021-12-16
bug_771606 'doxygen -w' should not require existence of header.html/footer.html even if listed in an options file

albert-github:
In case of usage of the `-w` option with doxygen don't check on the existence of files mentioned in the Doxyfile

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -83,8 +83,9 @@ namespace Config
 
   /*! Check the validity of the parsed options and correct or warn the user where needed.
    * \param quiet setting for the QUIET option (can have been overruled by means of a command line option)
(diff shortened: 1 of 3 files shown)


src/config.h
```diff
@@ -83,8 +83,9 @@ namespace Config
 
   /*! Check the validity of the parsed options and correct or warn the user where needed.
    * \param quiet setting for the QUIET option (can have been overruled by means of a command line option)
+   * \param check check  HTML / LaTeX header file etc. on existance (and terminate when not present)
```
bangerth, line 86:
```suggestion
   * \param check check  HTML / LaTeX header file etc. on existence (and terminate when not present)
```

albert-github: Thanks.

PR #8971, opened 2021-12-17
Specifying strings to be excluded from markdown processing

albert-github:
When having code like
```
/** @struct _MODEL_DICT_
 *  @brief This structure contains models.dat info
 *  @var _MODEL_DICT_::s_tech_class
 *  Member 's_tech_class' contains tech_class info
```
we get warnings like:
```
 warning: the name 'MODEL_SMALL_DICT' supplied as the argument of the \class, \struct, \union, or \include command is not an input file
 warning: Member s_tech_class (variable) of struct _MODEL_SMALL_DICT_ is not documented.
```
even though the member is properly documented. This is due to the fact that the `_MODEL_SMALL_DICT_` is in the markdown processing seen as an italic string `MODEL_SMALL_DATA`.

To overcome this problem in a general way an new setting had to be introduces: `MARKDOWN_EXCLUDE`

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7734952/example.tar.gz)

(Found by fossies for the hplip package)

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -668,6 +668,18 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+    <option type='list' id='MARKDOWN_EXCLUDE' format='string' depends='MARKDOWN_SUPPORT'>
+      <docs>
+<![CDATA[
+ The strings as supplied in the \c MARKDOWN_EXCLUDE tag will be excluded from the
+ markdown processing. It might be necessary to exclude strings from the processing
+ as they are e.g. the name of a structure but start and end with an underscore,
+ meaning that they would by handled by the markdown processor as an italic string.
+ \note The order of the excluded patterns is of importance. When one needs to
+ exclude `_A_` and `__A__` one has to place the `__A__` before `_A_` in the list.
+]]>
+      </docs>
+    </option>
     <option type='int' id='TOC_INCLUDE_HEADINGS' minval='0' maxval='99' defval='5' depends='MARKDOWN_SUPPORT'>
       <docs>
 <![CDATA[
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -36,6 +36,7 @@
 #include <unordered_map>
 #include <functional>
 #include <atomic>
+#include <mutex>
 
 #include "markdown.h"
 #include "growbuf.h"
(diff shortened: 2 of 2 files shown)


doxygen:
@albert-github I don't think this is the right solution direction. I would prefer that `@struct XXX` command is ignored by the markdown processing (or better: not passing it to the Markdown processor in the first place). 

Also note that macros starting with underscore and then a capital are reserved for the language implementation in C/C++ (see  https://stackoverflow.com/a/228797/784672)

albert-github:
I think the given solution gives a general possibility to exclude strings from markdown processing (and is independent of the used computer language).

The example might not be the best example as it is geared towards C++ and the `struct` command (as here I saw the problem surface), but I think the problem is broader.
- variables like these might not be reserved words, but valid identifiers in other languages.
- ignoring `@struct XXX` would mean that a lot of commands should be added here (I think probably all commands that use IDs).
- not passing `@struct XXX`  to markdown, in principle this would for part of the problem be an idea, the ID parts should be excluded.

However,
The problem still remains as shown with the `\param` commands in the example, but this could also be in some normal running text. With the proposed pull request it would also be possible to have the excluded reserved words handled properly (a user is free to give a description / hint to the reserved entities). In running text the user could use the escaped `_` i.e `\_` but this would make it less readable. Also there might be problems with links (e.g. in backtick parts).

doxygen:
I don't think we should push the problems in the parser to the user. A lot of these problems are introduced by doing the markdown processing before the comment scan phase. So maybe it is indeed a better solution to make the markdown processor aware of commands and parameters that it may not touch.

albert-github:
Wouldn't that mean that each time a word is found this should be compared against the complete list of variables / classes / namespaces / citations / ..., I think this would be quite slow.

doxygen:
> Wouldn't that mean that each time a word is found this should be compared against the complete list of variables / classes / namespaces / citations / ..., I think this would be quite slow.

We already have `Markdown::processSpecialCommand` so this would be an extension of that I suppose. Of course a map should be used to speed up the lookup and probably all commands can be divided in a few categories, where each category has a dedicated handler to skip over the text. I'll do a quick investigation to see if this is feasible.

albert-github:
I don't think the extension of the commands list in markdown.cpp would help, the "offending" strings can also be in running text, like:
```
/// Running text:
/// the modeldata structure data from backticks `_MODEL_DICT_`<br>
/// the modeldata structure data from cmd c \c _MODEL_DICT_<br>
/// the modeldata structure data from bare _MODEL_DICT_
```

albert-github:
Another approach has been taken which gives good results though not all patterns are matched but can, probably, be overcome by means of escaping some characters (especially the leading underscore).

Related item is also #8973

PR #8973, opened 2021-12-20
Incorrect citations for entries with double dash and for cross references

albert-github:
When having a simple citation problem like:
```
/** \mainpage
- \\cite s--92 \cite s--92
- \\cite csos-frcms-95 \cite csos-frcms-95
*/
```
with the bib entries (a bit strange here but it is because of the example):
```
@phdthesis{s--92
, author =      "M. F. Sanner"
, title =       "??"
, school =      "UHA"
, address =     "France"
, year =        1992
, keywords =    "doctoral thesis"
, update =      "98.03 bibrelex"
}
@inproceedings{csos-frcms-95
, author =      "Michel F. Sanner and Arthur J. Olson and Jean-Claude Spehner"
, title =       "Fast and Robust Computation of Molecular Surfaces,"
, booktitle =   "Proc. 11th Annu. ACM Sympos. Comput. Geom."
, year =        1995
, pages =       "C60--C70"
, keywords =    "molecules, biology, visualization"
, crossref =    "s--92"
, update =      "98.03 bibrelex, 95.09 mitchell"
}
```
and the settings:
```
CITE_BIB_FILES  = mybib.bib
QUIET = YES
OUTPUT_DIRECTORY = docs
```
we get the warning like:
```
aa.h:2: warning: \cite command to 's' does not have an associated number
```

This is due to the fact that the `--` has been converted to `&ndash;`  in the cite IDs and the detection of the cite IDs was not prepared for this.
Also the output of the cross-references would contain the unknown / by doxygen not supported `<cite>` tag.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7745749/example.tar.gz)

(Found by doing some tests on CGAL citations).

### DIFF
--- a/src/cite.cpp
+++ b/src/cite.cpp
@@ -24,6 +24,7 @@
 #include "debug.h"
 #include "fileinfo.h"
 #include "dir.h"
+#include "markdown.h"
 
 #include <map>
 #include <string>
@@ -135,7 +136,7 @@ void CitationManager::insertCrossReferencesForBibFile(const QCString &bibFile)
     {
       // assumption entry like: "@book { name," or "@book { name" (spaces optional)
       int j = line.find('{');
-      // when no {, go hunting for it
+      // when no '{', go hunting for it
       while (j==-1 && getline(f,lineStr))
       {
         line = lineStr;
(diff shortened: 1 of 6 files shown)


doxygen:
This is also an example where markdown processing before the comment scan phase causes problems. Also here, a better solution would be to avoid that markdown processing is done on the cite parameter, rather than detecting and undoing the damage later.

albert-github:
This is indeed also a problem as stated in #8971 but I think that it is very hard to overcome the problem here as also in the `bib2xhtml.pl` some processing is done. Note that in `cite.cpp` there was already an attempt to undo some damage on the line `label = substitute(substitute(label,"&ndash;","--"),"&mdash;","---");`

albert-github:
In case #8974 has been integrated it is possible to re-introduce the `<cite>` again in the doxygen.bst file

doxygen:
which means it should be fixed in doxygen version 1.9.3.

PR #8974, opened 2021-12-20
Adding HTML tag <cite>

albert-github:
Adding the HTML `<cite>` tag as a style change tag.


albert-github:
In case #8973 has been integrated it is possible to re-introduce the `<cite>` again in the doxygen.bst file

Issue #4887, opened 2018-07-02
line continuation in preprocessor in documented text (Origin: bugzilla #684050)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.8.2* on platform *Other*
Assigned to: Dimitri van Heesch

On 2012-09-14 21:46:52 +0000, Matthew Woehlke wrote:

    I am trying to define some standard methods with common documentation, like
    so:
    
    #define MY_STANDARD_METHODS(name) \
      /** Brief description of name. \
          Detailed description of name. \
       */ \
      void my_standard_method()
    
    This *almost* works, except that doxygen emits the line continuation (\)
    characters verbatim in the documentation, which does not seem correct.
    
    Is this a bug (seems like it...)? Is there a way around this that does not
    require using ALIASES? (I have a single doxyfile template I use across
    multiple libraries, and would prefer to keep this sort of thing per-
    library.)

On 2012-09-14 21:52:33 +0000, Matthew Woehlke wrote:

    So, to answer part of my own question...
    
    #define MY_STANDARD_METHODS(name) \
      /** Brief description of name.  Detailed description of name.  */ \
      void my_standard_method()
    
    ...which is a fascinating hack, but still wondering if it should be
    necessary.

On 2017-10-14 16:46:36 +0000, albert wrote:

    In https://stackoverflow.com/questions/46744573/does-doxygen-have-a-line-
    continuation-feature of user AcId on stackoverflow it was mentioned that in
    the following example:
    /**
     * \file mwe.cpp
     * 
     * \details
     * MWE::MWE(int a, int b, int c)
     * MWE::MWE(int a,
     *           int b, int c)
     */
    
    
    class MWE
    {
        public:
            /** 
             * \brief constructor with one parameter
             */
            MWE(int a);
    
            /** 
             * \brief constructor with three parameters
             **/
            MWE(int a, int b, int c);
    }
    
    When generating the documentation, the first link (MWE::func(int a, int b,
    int c)) correctly points to the constructor taking three integer parameters.
    However, the second link, where there is a line-break, points to the
    constructor taking only a single integer parameter (also the parameter list
    does not become part of the link, only the function name).
    
    A "solution" for this was, similar to the solution of Matthew Woehlke:
    
    /**
     * \file mwe.cpp
     * 
     * \details
     * MWE::MWE(int a, int b, int c)
     * MWE::MWE(int a,          int b, int c)
     */
    
    
    class MWE
    {
        public:
[...]

PR #8977, opened 2021-12-22
bug_161451 @copydoc cannot span multiple lines

albert-github:
Analogous to the  `\fn` command also for the `\copydoc` command the round braces are implemented.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7764003/example.tar.gz)

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -1940,9 +1940,7 @@ STopt  [^\n@\\]*
 
   /* ----- handle argument of the copydoc command ------- */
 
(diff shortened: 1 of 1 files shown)


albert-github:
Looks like this problem is related to the issue #4887 and I already tried to solve this before see pull request #8812

albert-github:
I think the fix #8814 is still necessary.
I think #8812 is not necessary anymore due to this pull request.

PR #8978, opened 2021-12-23
Fix inline check of markdown images

bollenn:
Fixing the updated content of https://github.com/doxygen/doxygen/pull/8953 in https://github.com/doxygen/doxygen/commit/5bb622158e33aac19d23068c08f434aaced60b4c.

Offset is the offset in the original array but not the index in data[]


doxygen:
Indeed. Good catch.

PR #8992, opened 2021-12-30
Make email obfuscation CSP friendly

AMDmi3:
Obfuscation makes email addresses harder to harvest by inserting into them HTML elements which are supposed to be invisible in the rendered HTML (display: none). The style is specified inline, and inline styles may be prohibited by content-security-policy HTTP header, breaking the markup if generated html is served from the server which sets such header. So move the style to CSS and refer to it via `class` in HTML code to fix this.

For content served from a server which sets CSP header, here's how the current doxygen genereted HTML look:
2021-12-30-143020_maim
After the fix:
2021-12-30-143110_maim


AMDmi3:
Thank you!

Issue #1311, opened 2018-07-01
INPUT_FILTER mechanism broken if filtered file has a space in path on Windows (Origin: bugzilla #156184)


doxygen:
status NEW severity *minor* in component *general* for *---*
Reported in version *1.3.x* on platform *Other*
Assigned to: Dimitri van Heesch

On 2004-10-22 21:08:58 +0000, Keith Gunderson wrote:

> I have a filter made to change my TSL (i.e. WinRunner) source/comments into C.  
> But when the location of my INPUTS is in a path with a space (c:\Documents and 
> Settings\...) the filter step isn't done right.  
> 
> When FILTER_SOURCE_FILES is on, the source is shown to be 1 line long, with the 
> error message shown.  (c:\Documents is not a something program or whatever.)
> 
> Moving things to a more suitable location will make things work just fine.  
> Thanks a bunch.  I'll share my filter code once its got a few miles under its 
> belt.

On 2005-07-22 16:15:14 +0000, Max A. Shonichev wrote:

> yep, that's really annoying bug.

On 2005-07-24 14:01:23 +0000, Dimitri van Heesch wrote:

> yep, so is dealing with spaces in the path. So just don't do it or expect problems.

Issue #7351, opened 2019-10-26
Doxygen does not return error code if INPUT_FILTER fails


boostmultifruit:
INPUT_FILTER can be, e.g., a Python script. If it fails, it means that documentation is corrupted.
Currently there are no checking of INPUT_FILTER return code:
https://github.com/doxygen/doxygen/blob/master/src/util.cpp#L8167-L8181

albert-github:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

Hi-Angel:
@albert-github thank you,

## Steps to reproduce *(in terms of terminal commands)*:

```
$ mkdir bar
$ cd bar
$ doxygen -g doxy-config


Configuration file 'doxy-config' created.

Now edit the configuration file and enter

  doxygen doxy-config

to generate the documentation for your project

$ sed -i 's:INPUT_FILTER.*=:INPUT_FILTER = /bin/false:g' doxy-config
$ doxygen doxy-config
Doxygen version used: 1.8.18
Searching for include files...
Searching for example files...
Searching for images...
Searching for dot files...
Searching for msc files...
Searching for dia files...
Searching for files to exclude
Searching INPUT for files to process...
Searching for files in directory /tmp/bar
Reading and parsing tag files
Parsing files
Building group list...
Building directory list...
Building namespace list...
Building file list...
Building class list...
Computing nesting relations for classes...
Associating documentation with classes...
Building example list...
Searching for enumerations...
Searching for documented typedefs...
Searching for members imported via using declarations...
Searching for included using directives...
Searching for documented variables...
Building interface member list...
Building member list...
Searching for friends...
Searching for documented defines...
Computing class inheritance relations...
Computing class usage relations...
Flushing cached template relations that have become invalid...
Computing class relations...
Add enum values to enums...
Searching for member function documentation...
Creating members for template instances...
Building page list...
Search for main page...
Computing page relations...
Determining the scope of groups...
Sorting lists...
Determining which enums are documented
Computing member relations...
Building full member lists recursively...
Adding members to member groups.
Computing member references...
Inheriting documentation...
Generating disk names...
Adding source references...
Adding xrefitems...
Sorting member lists...
Setting anonymous enum type...
Computing dependencies between directories...
Generating citations page...
Counting members...
Counting data structures...
Resolving user defined references...
Finding anchors and sections in the documentation...
Transferring function references...
[...]

PR #9017, opened 2022-01-07
Bugfix for issue #1311 (Prevent problems with spaces in paths in Windows popen)

ftvkyo:
Yet to check if it breaks anything, but doesn't look like it does at the moment.

Recommendation to add quotes taken from here:
https://www.reddit.com/r/learnpython/comments/ml2h7v/cprogram_is_not_recognized_as_an_internal_or/

### DIFF
--- a/src/portable.cpp
+++ b/src/portable.cpp
@@ -447,10 +447,16 @@ bool Portable::fileSystemIsCaseSensitive()
 
 FILE * Portable::popen(const QCString &name,const QCString &type)
 {
-  #if defined(_MSC_VER) || defined(__BORLANDC__)
-  return ::_popen(name.data(),type.data());
+  #if defined(_WIN32)
+  QCString newname = "\"" + name + "\"";
   #else
-  return ::popen(name.data(),type.data());
+  QCString newname = name;
+  #endif
+
+  #if defined(_MSC_VER) || defined(__BORLANDC__)
+  return ::_popen(newname.data(),type.data());
+  #else
+  return ::popen(newname.data(),type.data());
   #endif
 }
 


albert-github:
Thanks for the patch though I have some doubts.
- do you have an example where it goes wrong in the current master version of doxygen? (i.e. Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent)

I see in a number of files that `popen` is used
- src/definition.cpp
- src/filedef.cpp
- src/util.cpp

In `src/definition.cpp` I see:
```
        QCString cmd=filter+" \""+fileName+"\"";
        Debug::print(Debug::ExtCmd,0,"Executing popen(`%s`)\n",qPrint(cmd));
        f = Portable::popen(cmd,"r");
```
so it is likely to go wrong here when explicitly quotes are added. I think it would be better to use at this place the `quoted()`  function on the `filter` variable.
Similar is true for `src/filedef.cpp` and `src/util.cpp`.

In case the user has problems he can also set in the Doxyfile  the fileer command like `...FILTER... = '\"<filer name>\""`

ftvkyo:
Yes, I will provide actual reproduction steps as soon as I can get back to this issue.

Personally I had problems with windows doxygen 1.9.3 from the website. I used a filter that was a python script, it looked like
```
INPUT_FILTER           = \"C:/Program Files/Python310/python.exe\" \".../filter.py\" \"FLAG\" \"ARGUMENT\"
```
And the script was expecting a concatenated path to the processed file (because that's how input filters work).

I would just use a workaround adding extra quotes there right in config, but as this string gets concatenated (right in the bit of `src/definition.cpp` you posted) it's not a good idea.

So that was causing the `"C:/Program" is not recognized...` error, I'll post the error message with the reproduction steps. After patching the error disappeared, even though the idea of adding extra quotes that are semantically incorrect is wild to me. I'll also check other places where popen is used to see if it can cause problems, yes, my main concern is for it not to break existing filters. I guesss there's no tests for it so will need to check manually.

ftvkyo:
Hi @albert-github, here's the archive: [files.zip](https://github.com/doxygen/doxygen/files/7838550/files.zip)
Contents:
```
- "./folder with spaces/"
  - "./folder with spaces/filter.bat"
- "./src/"
  - "./src/main.cpp"
- "./Doxyfile"
```
`filter.bat` is empty (it doesn't matter); `main.cpp` has 2 functions and one of them has a doc comment (it doesn't matter); `Doxyfile` is default except these things:
```
INPUT                  = src
INPUT_FILTER           = "\"folder with spaces/filter.bat\" /f /l /a /g /s"
OUTPUT_DIRECTORY       = build
```

I built the b0a910e3d1e1e5e2d1c8e47f93fa9230c6d91d4f (current main branch) and my versions of doxygen and named the binaries `doxygen-1311-original` and `doxygen-1311-fixed`.

Now, running `../bin/doxygen-1311-original -d extcmd Doxyfile` in the project folder has this in the output:
```
...
Parsing files
Preprocessing D:/Proj/doxygen-1311/src/main.cpp...
Executing popen(`"folder with spaces/filter.bat" /f /l /a /g /s "D:/Proj/doxygen-1311/src/main.cpp"`)
'folder' is not recognized as an internal or external command,
operable program or batch file.
Parsing file D:/Proj/doxygen-1311/src/main.cpp...
...
```
Running `../bin/doxygen-1311-fixed -d extcmd Doxyfile` has different output:
```
...
Parsing files
Preprocessing D:/Proj/doxygen-1311/src/main.cpp...
Executing popen(`"folder with spaces/filter.bat" /f /l /a /g /s "D:/Proj/doxygen-1311/src/main.cpp"`)
Parsing file D:/Proj/doxygen-1311/src/main.cpp...
...
```

As you can see, windows says `folder` is not a program which happens because it parses the command incorrectly. The fact it doesn't stop when it happens is a [separate problem](https://github.com/doxygen/doxygen/issues/7351), not talking about it here.

albert-github:
I tried your example as well and I can see that there is a small problem in the original situation (all tests with 1.9.4 (070bbdcd7523dcbe88d8131556635717f82a937d)):
When having the original situation of the Doxyfile:
```
INPUT_FILTER           = "\"folder with spaces/filter.bat\" /f /l /a /g /s"
```
and running `doxygen -q -d extcmd` on Windows we get:
```
Executing popen(`"folder with spaces/filter.bat" /f /l /a /g /s ".../src/main.cpp"`)
'folder' is not recognized as an internal or external command,
operable program or batch file.
```
and under Cygwin:
```
Executing popen(`"folder with spaces/filter.bat" /f /l /a /g /s ".../src with spaces/main.cpp"`)
```
The Cygwin version gives the correct result and the Windows version gives an error.

When adding a set of escaped quotes in the Doxyfile:
```
INPUT_FILTER           = "\"\"folder with spaces/filter.bat\" /f /l /a /g /s\""
```
we see under Windows:
```
Executing popen(`""folder with spaces/filter.bat" /f /l /a /g /s" ".../src/main.cpp"`)
```
and under Cygwin:
```
Executing popen(`""folder with spaces/filter.bat" /f /l /a /g /s" ".../src with spaces/main.cpp"`)
/bin/sh: folder: command not found
```
so here the Cygwin version gives and error and the Windows version gives the correct result.

So based on my, limited, tests the results with the patch looks OK.
(the tests may be a bit confusing as in `portable.cpp` there is a switch and effectively uses for the Windows version my version with escaped quotes and for the Cygwin version the original version of the Doxyfile).

ftvkyo:
I see, I didn't think about comparing with Cygwin.
Yes, indeed, the idea behind the patch is so that we could format the Doxyfile's INPUT_FILTER string on Windows the same way we would do it on Linux and it should work. And no special formatting on windows could help work around the problem because of the concatenation of the string and the path.

Meanwhile, what is that `.../src with spaces/` directory in your Cygwin examples? That directory doesn't exist, so if it generates such popens -- something is probably wrong, isn't it?

albert-github:
I think you misunderstood my remarks, your patch works for Windows as well as for Cygwin.

The `.../src with spaces/` is the name of the directory I used instead of the `src` directory (also as a test as directory name as well as in the Doxyfile). I don't think `popen` has any problems here as it is properly quoted
I should have shown the results just for the `src` directory.

ftvkyo:
> The .../src with spaces/ is the name of the directory I used

Ah I see, I got confused because I didn't think you made another folder.
Great then. So, for the patch, tell me if you want me to check anything or extend the patch so it can be merged for the next version.

Issue #9007, opened 2022-01-06
Using DOT_PATH with a symlink for dot does not always work


kzrkk:
**The bug**
Ubuntu 20.04 LTS, set the path to dot in the configuration:

DOT_PATH = /usr/bin

Now call doxygen to create the documentation, dot execution fails, with error "/usr/sbin/dot: not found"

If DOT_PATH is not set and dot is only found via PATH, everything works as expected.

**Expected behaviour**
The error in case of an explicit DOT_PATH should not occur.

**Problem details**
The problem is a result of the fact, that in the mentioned Ubuntu version, /usr/bin/dot is a symlink to /usr/sbin/libgvc6-config-update. Now, within src/configimpl.l DOT_PATH is processed such that `FileInfo::dirPath` is used to get the path to the executable. This funtion in turn calls in the end `ghc::filesystem::canonical()` which in turn follows the symlinks and returns the link destination.

Now we have *DOT_PATH=/usr/bin* vs. the resolved path */usr/sbin*. In the end *dot* as executable name is appended again to the resolved path, but in */usr/sbin* there is no *dot* so the execution fails.

What is the reason for resolving the symlinks here? This way it is impossible to have an actual executable for *dot* that is named differently and that does not reside in the same directory as the *dot* symlink. This looks like an unnecessary restriction to me. Especially because by not setting DOT_PATH at all and letting *dot* to be taken from PATH it works. If someone stumbles across this problem he is unable to understand the root cause without debugging the doxygen code. There is no message whatsoever that the provided path /usr/bin is changed for some reason, you only see that eventually /usr/sbin is used.

The same problem might also exist for other path settings, I have not checked that.

**Version**
Ubuntu 20.04.3 LTS, Doxygen version 1.9.4 built from git:
1.9.4 (dc2950946033e9dab125e389af6eafc3d1f73eac)

albert-github:
The `DOT_PATH` signals the path where the `dot` executable can be found, there has not been the intention that another "dot" can be used, even worse when an executable is given and it is different from `dot` it might be able to find it but later on the other executable name is stripped and an explicit `dot` is appended.
The reason for checking whether or not the `dot` executable exists is that a proper warning is given like:
```
warning: the dot tool could not be found at ...\mydot.exe
```

When you run `doxygen -d extcmd` what do you see when you have `DOT_PATH ` set to `/usr/bin`?
[...]

PR #9018, opened 2022-01-08
issue #9007 Using DOT_PATH with a symlink for dot does not always work (plantuml)

albert-github:
Regression as mentioned in https://github.com/doxygen/doxygen/issues/9007#issuecomment-1007651759
The `-graphvizdot` cannot just contain the name of an executable somewhere in the path but needs a relative or full path.
So just when having no explicit `DOT_PATH` set just setting the command `dot` won't work as plantuml won't resolve the path.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -92,6 +92,11 @@ namespace Config
 
   /*! Clean up any data */
   void deinit();
+
+  /*! Signal whether or not the DOT_PATH is explicitly set. plantuml needs either a full or
+   * relative path to the dot executable
+   */
+  bool getPlantumlDotPathSet();
 }
 
 #endif
--- a/src/configimpl.h
+++ b/src/configimpl.h
@@ -583,6 +583,20 @@ class ConfigImpl
       return substitute(result,"\r","");
     }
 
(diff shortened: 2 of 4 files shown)


doxygen:
@albert-github I decided to go for a slightly different implementation where `HAVE_DOT` returns the uncorrected value and `Doxygen::verifiedDotPath` holds the corrected version. That why the data does not have to be passed via the Config class.

albert-github:
I did a quick test and it didn't reveal problems.

To have a clean implementation is a good thing though I decided for the implementation I had:
- the `m_plantumlDotPathSet` variable and the getter / setter functions were related directly to a configuration setting and could be see as some sort of configuration setting as well
- the checking of the configuration setting `DOT_PATH` is something that I think should be done in the config implementation and not in a different place (in the very unlikely event that the meaning of `DOT_PATH` changes its implementation has to be changed with the config reading and the checking elsewhere).

doxygen:
@albert-github I understand your line of thinking. In this case I don't see it as a new setting, more as two flavors of a setting, the raw value and the corrected (or derived) value. If we see more cases where this is needed/useful, I propose we store for every setting both the original value and the updated value, and have a getter for each.

Issue #9011, opened 2022-01-06
Feature request: Generate images for XML output


canton7:
I have a generator which takes Doxygen's XML output, and produces a Word document. This all works fine, and the XML output works really well for this.

One slightly tricky part is generating images from formulae, `\dot`, `\dotfile`, `\msc`, `\mscfile`, `\diafile`, `\plantuml` (and others that I might have missed!) to insert into the document. This is do-able, and I have the information to do this in the XML, but it does mean replicating Doxygen's image generating logic, including using configuration options such as `FORMULA_MACROFILE`, `HAVE_DOT`, `DOT_PATH`, `PLANTUML_INCLUDE_PATH`, etc.

Since Doxygen already produces these images for the other output formats, my request is to support generating these for the XML output as well, and providing a file path in the corresponding `<formula>`/`<dotfile>`/etc tag.

I appreciate that this probably has some hidden complexity, and I won't be sad if you reject this! But it's worth asking.

Thanks!

albert-github:
Indeed an interesting thought.
We shouldn't only look at the commands as mentioned by you but also to the more automatically generated images by doxygen like call / caller , inheritance / collaboration, include, directory, ... graphs.
Doxygen creates the images for the different output formats separately and it might (will) take also some extra time to generate it for XML.


As side notes:
- why using the XML output for generation of the word documentation and not the rtf version?
- do you have a script / procedure to generate the word document from the doxygen XML output that you could share?

canton7:
> why using the XML output for generation of the word documentation and not the rtf version?

I'm heavily customising the Word output (this is to provide as a regulatory submission). We used to use the PDF output, but massaging it into the form we needed was getting really quite painful (lots of manipulating the generated tex files). Using the XML as input lets us generate exactly the output we need.

> do you have a script / procedure to generate the word document from the doxygen XML output that you could share?

Sure, https://github.com/canton7/Dox2Word

albert-github:
I've just pushed a proposed patch, pull request #9024

2bndy5:
Just my two cents: Static images may not work well if your document viewer (Word in this case) is set to a dark theme, especially if the images use a transparent background. I'd be happier if the default mechanism remains the same (an XML structure that represents the dot syntax used to generate the image) and let this feature be optional.

canton7:
It is optional? You're not forced to read or use the images generates by doxygen's xml output, and the xml output will continue to contain the structure that represents the dot syntax.

PR #9024, opened 2022-01-11
issue #9011 Feature request: Generate images for XML output

albert-github:
Implemented calculation and storage of the resulting images for the XML_OUTPUT and having a reference in the xml file.
Implementation for the the formats `\dotfile`, `\dot`, `\msc`, `\mscfile`, `\startuml` (plantuml), `\diafile` and the formula commands

Examples: [examples.tar.gz](https://github.com/doxygen/doxygen/files/7847851/examples.tar.gz)


albert-github:
The problem with the tests here is that we now would require that LaTeX is also installed as soon as formulas are present and the XML output is chosen.
Unfortunately, probably best to "hide" the figures (for all mentioned output formats) behind a setting: `XML_INCORPORATE_FIGURES` wit de fault value `NO`.

canton7:
Amazing, thank you!

Being opt-in is fine for my use-case.

What's the behaviour if e.g. `dot` isn't installed? I notice that the `figure` attribute isn't optional in the XSD.

albert-github:
@canton7 I'm working on these issues. In case dot is not installed you will have `HAVE_DOT` set to no and this will give a warning.

canton7:
Just tested: the `figure` gets added, but the file it refers to doesn't exist on disk. Otherwise works well in my test, thank you!

albert-github:
- Did you get a warning?
- Do you mean you have `HAVE_DOT=NO`?

canton7:
I was playing with formulae, but without latex in my PATH. I do get warnings, but my XML-processing tool doesn't necessary know whether the user got any warnings but decided to run the tool anyway. 

It's not a problem though - checking whether the formula figure file exists is easy. Much easier than implementing each of those bits of image-generation logic myself, so I'm very grateful for your work on this!

src/xmldocvisitor.cpp
```diff
@@ -84,6 +95,75 @@ static void visitPreStart(TextStream &t, const char *cmd, bool doCaption,
     visitCaption(parent, children);
     t << "\"";
   }
+  if ((cmd == "dotfile") || (cmd =="dot"))
+  {
+    QCString baseName=name;
```
doxygen, line 102: I count 4 copies of the code from lines 100-109. Better to factor out this part.

src/xmldocvisitor.cpp
```diff
@@ -323,19 +403,79 @@ void XmlDocVisitor::visit(DocVerbatim *s)
       m_t << s->text();
       break;
     case DocVerbatim::Dot:
-      visitPreStart(m_t, "dot", s->hasCaption(), this, s->children(), QCString(""), FALSE, DocImage::Html, s->width(), s->height());
-      filter(s->text());
-      visitPostEnd(m_t, "dot");
+      {
+        static int dotindex = 1;
```
doxygen: Static counters will give unpredictable output if used in a multi-threaded context (i.e. `NUM_PROC_THREADS>1`).

albert-github: I took this code straight away from the other `*docvisitor.cpp` code, (see e.g `htmldocvisitor.cpp` and search for `static int dotindex`.  It is a for the output format specific "global" counter.

@doxygen What do you propose?

src/xmldocvisitor.cpp
albert-github: At which commit did you look?

It looks like you looked at the commit:
```
Commit: 723c6eabbb48ce43c203044ca0fde132a0c7249b [723c6ea]
Date: Tuesday, January 11, 2022 4:41:07 PM

issue #9011 Feature request: Generate images for XML output

Implemented calculation and storage of the resulting images for the XML_OUTPUT and having a reference in the xml file.
Implementation for the the formats `\dotfile`, `\dot`, `\msc`, `\mscfile`, `\startuml` (plantuml), `\diafile` and the formula commands
```
as in the commit:
```
Commit: c47d80d1f810aef9270c1a884099ab0a9a6f8b6d [c47d80d]
Date: Wednesday, January 12, 2022 12:06:10 PM

issue #9011 Feature request: Generate images for XML output

- Introduced setting `XML_INCORPORATE_IMAGES`
- better support for `CREATE_SUBDIRS`
```

it look to me that this has been "factored out".

gaodingpan:
Any idea on the schedule of this pull being merged?

PR #9025, opened 2022-01-12
Fix for size_t treated as signed rather than unsigned

neilsimon:
Small fix that sometimes results in searches failed due to `Unknown Exception!`.
Since `bi` is of type `size_t` it can never be negative so a less than comparison with 0 will never be true.


doxygen:
@neilsimon I found it clearer to show that variable `bi` can be negative and made it an `int`. Please verify if this also solves the problem for you.

Issue #9023, opened 2022-01-11
markdown bold and ref


dlyr:
**Describe the bug**
In markdown source, if a ref to a page is used inside a bold line, the resulting dox is not bold.

**Expected behavior**
using \ref should not break markdown bold typesetting 

**Screenshots**

![Screenshot from 2022-01-11 11-53-57](https://user-images.githubusercontent.com/11328509/148929941-04710f13-dbc5-4ec0-a79c-712028834584.png)

**To Reproduce**
Run doxygen for attached file [Test.zip](https://github.com/doxygen/doxygen/files/7845878/Test.zip)

**Version**
1.9.0 (77a37d141d3603c0089bc4a231c9381c1bb7d707)

albert-github:
Besides the **bold** problem I see in the current doxygen version (1.9.3 (c0b9eafbfb53286ce31e75e2b6c976ee4d345473)) and master (1.9.4 (1f1ac2838cfcc8e4bd95501df4d7376d35a0848e)) a problem with the `\ref` command.
The link text for the `\ref` command is with the 1.9.0 version "hello world" but for the "1.9.3 version "Test" (it looks like to be introduced in version 1.9.3 as in the 1.9.2 version the text is still "hello world".

dlyr:
damn, I thought I use last master, but I forgot to pull ;) right, there is another bug in ref label.

albert-github:
I did a quick bisect on the problem and found:
```
a36c9943ffe2e64954c1370a6bd9ce33aaf54cec is the first bad commit

Date:   Wed Nov 17 18:05:46 2021 +0100

    issue #8528 Markdown links to Markdown pages with explicit page command are broken

    In case a  `@page` command is the first line in a markdown file it is used as  "label" but also to construct the name of the output file and hence the regular markdown reference (like `[...](..)`) does not work anymore as here an internal label and filename is used  (`md_...`).
    To overcome this problem the `@page <label> (title)` is changed into `@page md_<filename> (title)\ilinebr @anchor <label>`.
```

so a fix with a small side-effect, have to look into it.


In the correct version the line:
```
\page test hello world
```
is left unchanged during the markdown processing, but in case of the incorrect version this line is changed into:
```
\page md_test hello world\ilinebr @anchor test
```

albert-github:
For th eproblem in respect to the text in the link I've just pushed a proposed patch, pull request #9035

PR #9032, opened 2022-01-13
bug_711041 Undocumented "enum" values in TOC of qhp files causes tag-mismatch and corrupt TOC.

albert-github:
The problem regarding the extra `</section>` in the case of a `user` section in the doxyfile could be solved by means of better calculating the "diff", though this lead to bad index sequences in the qhp navtree.
To solve the bad tree entries, for the enum values, a function `closeContentsItem` is introduced.
(in `layout_default.xml` some white space at the end of the line is removed.)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -3540,9 +3540,7 @@ void MarkdownOutlineParser::parseInput(const QCString &fileName,
           docs = docs.left(labelStartPos)+                     // part before label
                  newLabel+                                     // new label
                  docs.mid(labelEndPos,lineLen-labelEndPos-1)+  // part between orgLabel and \n
-                 "\\ilinebr @ianchor "+orgLabel+               // add original anchor
-                 docs.mid(labelEndPos,lineLen-labelEndPos-1)+  // part between orgLabel and \n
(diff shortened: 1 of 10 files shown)


src/markdown.cpp
```diff
@@ -3540,7 +3540,9 @@ void MarkdownOutlineParser::parseInput(const QCString &fileName,
           docs = docs.left(labelStartPos)+                     // part before label
                  newLabel+                                     // new label
                  docs.mid(labelEndPos,lineLen-labelEndPos-1)+  // part between orgLabel and \n
-                 "\\ilinebr @anchor "+orgLabel+"\n"+           // add original anchor
+                 "\\ilinebr @ianchor "+orgLabel+               // add original anchor
```
doxygen, line 3543: is the `i` in `@ianchor` a typo?

albert-github: oops this file should not have landed into this PR (it was a start for #9023, next issue I probably work on).
I'll remove the file from the PR

ferdymercury:
I am building documentation using this patch and it's working totally stable now, thanks a lot!

Issue #9051, opened 2022-01-19
Wrong "Related page" generating in 1.9.3 (works good in 1.9.1!) from *.md files


Beszt:
**Describe and reproduce the bug**
When i try generate HTML output from *.MD files with /page header i have a some bugs. For Example my code in tests.md:

```
# Unit tests Requirements
\page test_requirements Unit tests Requirements
```

In 1.9.3 my "Related pages" link for this file generated like "Unit tests Requiremen" without "t" and "s" is in body of page! In 1.9.1 it's all ok.

albert-github:
The problem with the `s` missing at the end of the word "Requirements" I can reproduce. In 1.9.1 I see:
![image](https://user-images.githubusercontent.com/5223533/150127230-879f507c-74af-42e7-b97d-d171b756546a.png)

and in 1.9.2 (and current master 1.9.4 (99482978eb791a6159c4e409c42ec95acbbd2c1d)) I see:
![image](https://user-images.githubusercontent.com/5223533/150127350-580425ec-0eda-410b-a3cc-7e6a7678b335.png)

The problem with the `t` I don't see, please indicate and also (so we are tallking about the same problem):
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

albert-github:
Regarding the problem with the `s` I've just pushed a proposed patch, pull request #9053

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release)

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.4.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9053, opened 2022-01-19
issue #9051 Wrong "Related page" generating in 1.9.3 (works good in 1.9.1!) from *.md files

albert-github:
Small problem in respect to the definition of the end position of the label and subsequently the number of characters to be copied.

Note: will also effect the proposed pull request  #9035


doxygen:
@albert-github Found a slightly simpler solution that makes better use of the capture ranges of the regular expression.

albert-github:
Indeed making use of more "named" ranges is better, I just kept the old style.

Issue #9057, opened 2022-01-21
Allow external mscgen


tmpgnh:
**Describe the bug**
Prior to 1.9 I could use an external alternative to the -- now inlined -- MSC generator, and make use of a much extended chart syntax.

**Expected behavior**
With a configuration option I would like to tell doxygen to try using an external program for MSCs -- the now obsoleted MSCGEN_PATH would be a perfect fit.

**Version**
I'm using 1.9.1 on Debian/bullseye and onwards, and Ubuntu/Hirsute.

I try to provide a patch in a PR.

albert-github:
I think this problem might be related to #7994

albert-github:
Ps. Please don't use "Linked pull requests " as this would disrupt the doxygen workflow when the pull request would be integrated.

tmpgnh:
> Ps. Please don't use "Linked pull requests " as this would disrupt the doxygen workflow when the pull request would be integrated.

Sorry; should I try to remove the link (I guess the 'fixxes NN' in the commit message)?

albert-github:
Yes please.

See for a workflow / formulation I always use e.g. https://github.com/doxygen/doxygen/issues/9027#issuecomment-1012412610 and when the proposed pull request is integrated in the master I always try to set the fixed label and use a text like in https://github.com/doxygen/doxygen/issues/9027#issuecomment-1013646990

tmpgnh:
Propose patch in https://github.com/tmpgnh/doxygen/pull/1

albert-github:
@tmpgnh When you have a finished patch you should make a pull request to the doxygen repository.

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.7.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9059, opened 2022-01-21
issue #9057 Allow calling external mscgen

tmpgnh:
Resurrect pre-1.9 functionality to allow calling external mscgen for producing MSCs. In particular, this enables to use msc-generator[1] again for such charts -- which is a commandline-compatible alternative MSC generator with much extended syntax than the original mscgen.

[1] https://sourceforge.net/projects/msc-generator

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3514,6 +3514,18 @@ where `loc1` and `loc2` can be relative or absolute paths or URLs.
       </docs>
     </option>
   </group>
+  <group name='Msc' docs='Configuration options related to message sequence charts (MSC)'>
+    <option type='string' id='MSCGEN' format='file' defval=''>
+      <docs>
+<![CDATA[
+ You can define MSCs within doxygen comments using the \ref cmdmsc "\\msc"
+ command. Doxygen will either use an internal tool to produce the chart, or if
+ specified the <code>MSCGEN</code> tag allows you to use an external program
+ for that.
+]]>
+      </docs>
+    </option>
+  </group>
   <group name='Dot' docs='Configuration options related to the dot tool'>
     <option type='string' id='DIA_PATH' format='dir' defval=''>
       <docs>
--- a/src/msc.cpp
+++ b/src/msc.cpp
@@ -89,6 +89,44 @@ static bool convertMapFile(TextStream &t,const QCString &mapName,const QCString 
   return true;
 }
 
+bool do_mscgen_generate(const QCString& inFile,const QCString& outFile,mscgen_format_t msc_format,
+                     const QCString &srcFile,int srcLine)
+{
+  QCString external_mscgen = Portable::getenv("MSCGEN_PATH");//TODO use ConfigValues::instance().MSCGEN_PATH() instead
+  if (!external_mscgen.isEmpty()) {
+    QCString type;
+    switch (msc_format)
+    {
+      case mscgen_format_png:
+        type = "png";
+        break;
+      case mscgen_format_eps:
+        type = "eps";
+        break;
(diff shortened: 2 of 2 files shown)


src/msc.cpp
```diff
@@ -89,6 +89,44 @@ static bool convertMapFile(TextStream &t,const QCString &mapName,const QCString
...
+    }
+    int exitcode = Portable::system(external_mscgen,"-T"+type+" -o "+outFile+" "+inFile);
+    if (exitcode==0)
+      return true;
+    warn(srcFile,srcLine,"Problems running %s given with MSCGEN (exit status: %d); check your installation. Trying with internal mscgen instead.",
+        external_mscgen.data(),exitcode);
+  }
+  int code = mscgen_generate(inFile.data(),outFile.data(),msc_format);
+  if (code!=0)
+  {
```
albert-github, line 122: This indicate that when `MSCGEN` is set and the generation of the image fails it attempts to create it with the build in generator? I think that this is unwanted behavior, I think it should be either the external generator or the internal generator but not both.

tmpgnh: Yes, that was the idea but I can see you point and 

1. Remove this behavior, make the warn into an err, no problem.
2. I can add another config option like MSCGEN_FALLBACK_TO_INTERNAL, and check for it on line 117.

Which would you prefer?

albert-github: No I wouldn't prefer this, the external mscgen should anyway be a superset of the internal function.

tmpgnh: OK, do you consider this resolved then?

albert-github:
On note that probably has to be added to the `config.xml` file is that the program to be used has to follow the rules regarding type of the conversion and input  / output  file on the command line as indicated in:
```
int exitcode = Portable::system(external_mscgen,"-T"+type+" -o "+outFile+" "+inFile)
```

tmpgnh:
Added an explicit description on the cmdline switches.

src/config.xml
```diff
@@ -3510,6 +3510,20 @@ where `loc1` and `loc2` can be relative or absolute paths or URLs.
...
+    </option>
+  </group>
+  <group name='Msc' docs='Configuration options related to message sequence charts (MSC)'>
+    <option type='string' id='MSCGEN' format='file' defval=''>
+      <docs>
+<![CDATA[
+ You can define MSCs within doxygen comments using the \ref cmdmsc "\\msc"
+ command. Doxygen will either use an internal tool to produce the chart, or if
+ specified the <code>MSCGEN</code> tag allows you to use an external program
+ for that. In the latter case, the external program shall be callable as
```
albert-github, line 3524: Better to use `\c MSCGEN` for consistency.

tmpgnh: Done

tmpgnh:
Hm, this build failure must be some transient thing only (installing 3pp). Can or should I re-trigger it, or no bother?

albert-github:
Indeed looks like some mirrors have regularly problems with installing Qt / Ghostscript, restarting doesn't always help (as often the same host is used).
As the other builds succeeded I would just ignore it in this case.

tmpgnh:
> As the other builds succeeded I would just ignore it in this case.

OK, can I do anything more with this request?

laurensmiers:
*friendly bump*
Are there any plans on merging this or are there still things to be done? Cause I'd love this feature to be in.

tmpgnh:
> _friendly bump_ Are there any plans on merging this or are there still things to be done? Cause I'd love this feature to be in.

Thanks for asking, but I don't have any info.

Maybe @albert-github ?

Issue #5940, opened 2018-07-02
Look for [[deprecated]] markers (Origin: bugzilla #761131)


doxygen:
status NEW severity *enhancement* in component *general* for *---*
Reported in version *unspecified* on platform *Other*
Assigned to: Dimitri van Heesch

On 2016-01-26 13:22:21 +0000, Stephan Beyer wrote:

> Hi,
> 
> there are several ways to mark deprecation in C++:
> 
> C++14:
>    [[deprecated("reason")]]
> 
> MSVC-style:
>    __declspec(deprecated("reason"))
> 
> GNU-style:
>    __attribute__((deprecated("reason")))
> 
> It would be a great enhancement if Doxygen replaced these attributes
> internally by "/** \deprecated reason */" to show the deprecations in the documentation. (I have simulated this using a hack for my project but I guess doxygen should simply do it.)
> 
> Note that it is also possible to mark deprecations without reason in code. It would also be nice if this case is handled better by Doxygen (although IMHO one should always mention a reason or an alternative for a deprecated entity).
> 
> Cheers
>   Stephan

FlorianFranzen:
Are there any plans to pursue this feature?

wjwwood:
This would be great to have. It's annoying to have to duplicate this information in the doc block and the attribute.

Dschaehn:
We would also really like to have this feature. We have a workaround using our own deprecated marco, but it does not work well (we could not figure out how to make doxygen link symbols inside the deprecate-reason text).

sl42:
I also need this feature for evaluating the C++14 attributes '`[[deprecated]]`' and  '`[[deprecated("<string-literal>")]]`' badly since nowadays almost all compilers support this feature and people are using now this usefull feature.

First I thought, I can implement myself a short workaround because within '`scanner.l`' there is a parsing construct named '`SkipC11Attribute`'. Therefore I created a small test example which uses the new C++ 14 attribute '`deprecated`' in the most common forms.

[doxygen_deprecated_C++14_attributes.zip](https://github.com/doxygen/doxygen/files/7834877/doxygen_deprecated_C%2B%2B14_attributes.zip)

After I tried to debug doxygen using this example, it turned out that the debugger never reached any code, where the '`SkipC11Attribute`' was called or set. 
[...]

PR #9064, opened 2022-01-23
Added a proposal for '#5940 Look for [[deprecated]] markers' to evalu…

sl42:
This is a small suggestion of taking over the '`[[deprecated]]`' and '`[[deprecated("Comment")]`' attributes automatically into the Doxygen documentation. The simple test example within '[https://github.com/doxygen/doxygen/files/7918504/doxygen_deprecated_C%2B%2B14_attributes_tests.zip]' works with that solution.

### DIFF
--- a/addon/doxywizard/config_doxyw.l
+++ b/addon/doxywizard/config_doxyw.l
@@ -19,7 +19,7 @@
 %{
 
 /*
- *	includes
+ * includes
  */
 #include "config.h"
 #include "input.h"
@@ -46,7 +46,7 @@
 
 /* -----------------------------------------------------------------
  *
- *	static variables
+ * static variables
  */
 
 struct ConfigFileState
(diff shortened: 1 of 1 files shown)


albert-github:
I didn't yet review the proposed patch, but I checked why my version didn't want to compile. I found out that I was using, on Windows) flex 2.5.37 and that this didn't work. On Cygwin I was using 2.6.4 and here it compiled. The latest win_flex executable is 2.6.4 as well, so I upgraded to this version.

@doxygen should it still be possible to use flex 2.5.37 or should we set as a minimum requirement 2.6.4?

albert-github:
It doesn't look to bad, though I found a few problems, with the following code:
```
[[deprecated]] int b_aa;
[[deprecated]] int b_aa1, b_aa2;
[[deprecated("dep_var")]] int v_aa;
[[deprecated("dep_var_both")]] int v_aa1, v_aa2;
```
this results in:
image

- the extra `"` (not only with variables, but also with other entities.
- missing `b_aa1`, `b_aa2` and `v_aa2` as deprecated

albert-github:
The problems as indicated previously have been solved, though I found some other. When having:
```
/// \file


    using PS_aa [[deprecated("dep_using")]] = S_aa*;
    [[deprecated("dep_var")]] int v_aa;
```

We get as output:

image


So the deprecated message of the `using` is not recorded with the using, but lands with the variable.

When using:
```
/// \file


    /// @deprecated dep_using
    using PS_bb = S_bb*;
    /// @deprecated dep_var
    int v_bb;
```
all works normal.

Example: [example_2.tar.gz](https://github.com/doxygen/doxygen/files/7971229/example_2.tar.gz)

PR #9067, opened 2022-01-23
Reference to non existing anchor with "More..."

albert-github:
When having an anonymous struct, the `More ...` in the detailed description points to a non existing page and anchor. We can see this with e.g.:
```
/// \file

class test {
 public:
    struct {
        double s2_time;  //!< \brief brief time in secs
                         //!< \details details time in secs
    } s2_temp;
};
```

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/7920648/example.tar.gz)

the correction is analogous to the correction for writing the name.

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -2470,7 +2470,19 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
         ol.pushGeneratorState();
         ol.disableAllBut(OutputGenerator::Html);
         ol.docify(" ");
-        ol.startTextLink(getOutputFileBase(),anchor());
+        MemberDefMutable *annMemb = NULL;
+        if (!isAnonymous()) // hide anonymous stuff
+        {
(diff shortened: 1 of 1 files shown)


src/memberdef.cpp
```diff
@@ -2470,7 +2470,19 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
         ol.pushGeneratorState();
         ol.disableAllBut(OutputGenerator::Html);
         ol.docify(" ");
-        ol.startTextLink(getOutputFileBase(),anchor());
+        MemberDefMutable *annMemb = NULL;
+        if (!isAnonymous()) // hide anonymous stuff
```
doxygen, line 2474: shouldn't this be `if (isAnonymous())`?

albert-github: I was wondering as well, though in the "write name" there was also `!isAnonymous()`, see line 2267.

Issue #8784, opened 2021-09-17
CLANG_ASSISTED_PARSING tries to parse markdown files


reddwarf69:
doxygen 1.9.2 on Fedora 34 (built from https://src.fedoraproject.org/rpms/doxygen/tree/5671b82de32b5e12b78b8f420196fda5a7272750) (same thing with the default doxygen 1.9.1 package)

Running `doxygen -g`, enabling CLANG_ASSISTED_PARSING, and running `doxygen Doxygen` causes doxygen to complain with
```
error: /home/reddwarf/tmp/ppp1/test/test.md:1:1: error: C++ requires a type specifier for all declarations [clang]
```
when the following file is used
```
Test {#mainpage}
====
```
[test.tar.gz](https://github.com/doxygen/doxygen/files/7186566/test.tar.gz)

albert-github:
Same accounts also for Fortran and Python file (I didn't try other types).

Bagira80:
Similar, Doxygen tries to parse `*.d` files (probably expecting them to be source for the [D language](https://dlang.org/)) and `*.m` and `*.mm` files (probably expecting them to be Objective-C source files). However, at least in my (C++) codebase they have a different meaning.

Is there maybe some way to restrict the language or file-extensions that may be parsed by (Clang assisted) Doxygen?

albert-github:
@Bagira80
- Please also specify the **full** doxygen version used (`doxygen -v`).
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

- did you have a look at the doxygen setting `EXTENSION_MAPPING` (https://www.doxygen.nl/manual/config.html#cfg_extension_mapping)

Bagira80:
Answering my own question: I must restrict the list of file-extension in the `FILE_PATTERNS` configuration value to not include these unwanted file-extensions.

However, I still would prefer if Doxygen automatically generates the full list but lets me specify some exceptions that I do not want to parse...

Bagira80:
> * Please also specify the full doxygen version used (doxygen -v).

I now tried with Doxygen built with the latest commit from `master` (432814d75249f89b7e324f3e5653ceb5292e7139), still linking against a static libclang 13.0.0.

> * Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

[...]

PR #9073, opened 2022-01-25
issue #8784 CLANG_ASSISTED_PARSING tries to parse markdown files

albert-github:
Restrict the handling by CLang to only C / Cpp type of files and process the other files in the usual way.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -7995,7 +7995,7 @@ static void generateFileSources()
       {
         for (const auto &fd : *fn)
         {
-          if (fd->isSource() && !fd->isReference() &&
+          if (fd->isSource() && !fd->isReference() && getLanguageFromFileName(fd->docName())==SrcLangExt_Cpp &&
               ((fd->generateSourceFile() && !g_useOutputTemplate) ||
                (!fd->isReference() && Doxygen::parseSourcesNeeded)
               )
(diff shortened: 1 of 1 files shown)


Bagira80:
I can confirm that this PR works as expected and prevents the clang-parser from parsing files that are not marked/known to contain C++ code.

src/doxygen.cpp
```diff
@@ -8057,24 +8057,51 @@ static void generateFileSources()
...
-              // we needed to parse the sources even if we do not show them
+            else
             {
-              auto clangParser = ClangParser::instance()->createTUParser(fd.get());
-              msg("Parsing code for file %s...\n",qPrint(fd->docName()));
-              clangParser->parse();
-              fd->writeSourceHeader(*g_outputList);
-              fd->writeSourceBody(*g_outputList,clangParser.get());
-              fd->writeSourceFooter(*g_outputList);
+              for (const auto &fn : *Doxygen::inputNameLinkedMap)
```
doxygen, line 8084: This doesn't look good. We are already in a for loop with the same iterator.

doxygen:
@albert-github I reduced the amount code duplication, and fixed the issue mentioned above. Also found that the source code didn't have anchors (was already missing), so fixed that as well.

albert-github:
I see the improvement regarding the anchors, now we jump from the description (e.g. in aa_8h.html) to the right place in the source file (aa_8h_source.html).
Though I see that the "highlighting " of the line number is still missing:
Version without CLANG usage:
image

Version with CLANG usage:
image

note the line 11.
This problem is also present in the 1.9.3 version.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8079956/example.tar.gz)

albert-github:
I've just pushed a proposed patch, pull request #9126 for the problem mentioned in my comment: https://github.com/doxygen/doxygen/pull/9073#issuecomment-1041459405

PR #9084, opened 2022-01-29
Extend CREATE_SUBDIRS to be able to control the number of created sub directories

makoehr:
This allows to control the maximum number of sub-folders that
will be created with the CREATE_SUBDIRS options.


src/config.xml
```diff
@@ -275,11 +275,26 @@ Go to the <a href="commands.html">next</a> section or return to the
...
 <![CDATA[
  If the \c CREATE_SUBDIRS tag is set to \c YES then doxygen will create
- 4096 sub-directories (in 2 levels) under the output directory of each output
+ up to 4096 sub-directories (in 2 levels) under the output directory of each output
  format and will distribute the generated files over these directories.
  Enabling this option can be useful when feeding doxygen a huge amount of source
  files, where putting all generated files in the same directory would otherwise
  causes performance problems for the file system.
+ Adapt \c CREATE_SUBDIRS_LEVEL to control the number of sub-directories.
+]]>
```
albert-github, line 284: `\c CREATE_SUBDIRS_LEVEL` should be more like  `\ref cfg_create_subdirs_level "CREATE_SUBDIRS_LEVEL"`

src/config.xml
```diff
@@ -275,11 +275,26 @@ Go to the <a href="commands.html">next</a> section or return to the
...
+      </docs>
+    </option>
+    <option type='int' id='CREATE_SUBDIRS_LEVEL' minval='0' maxval='8' defval='8'>
+      <docs>
+<![CDATA[
+ Controls the number of sub-directories that will be create when \c CREATE_SUBDIRS tag is set to \c YES.
+ The number is computed by the formula 16 * (2 ^ CREATE_SUBDIRS_LEVEL).
```
albert-github, line 291: - see above more like a `\ref`
- `16 * (2 ^ CREATE_SUBDIRS_LEVEL)` should be more like a formula like `\f$16 * (2^{CREATE_SUBDIRS_LEVEL})\f$`

src/util.cpp
```diff
@@ -3602,6 +3602,8 @@ static std::unordered_map<std::string,int> g_usedNames;
 static std::mutex g_usedNamesMutex;
 static int g_usedNamesCount=1;
 
+
+
```
albert-github, line 3606: Why?

src/util.cpp
```diff
@@ -3645,12 +3647,17 @@ QCString convertNameToFile(const QCString &name,bool allowDots,bool allowUndersc
   if (createSubdirs)
   {
     int l1Dir=0,l2Dir=0;
+    static int createSubdirsLevel = std::min(8, Config_getInt(CREATE_SUBDIRS_LEVEL));
+    static const uchar createSubdirsBitmaskL2[] =
```
albert-github, line 3651: Shouldn't this be protected against values `< 0` as well?

albert-github (changes requested):
- Initially doxygen creates 4096 sub directories and later on removes the not used subdirectories (still can lead to quite a few directories though.
- Another thought is to use for the `CREATE_SUBDIRS` an `enum` variable with the values `NO`, `YES`, `0 ... 8`, so we won't get a new settings variable.

makoehr:
For what ever reason the Windows builds did not pass...

src/util.cpp
```diff
@@ -3602,6 +3602,35 @@ static std::unordered_map<std::string,int> g_usedNames;
...
+/*! This function converts the value of CREATE_SUBDIRS
+ *  into the number of relevant bits for the number of
+ *  created directories in the 2nd level.
+ */
+static int convertCreateSubDirs2Level(CREATE_SUBDIRS_t createSubDirs)
+{
+  switch (createSubDirs)
+  {
+    case CREATE_SUBDIRS_t::D16:
+      return 0;
```
albert-github, line 3614: Doesn't this mean that in this case there is only one sub-directory at level 2, I think this is a bit overdone (even `D32`  and `D64` might be a bit overdone).

src/util.cpp
```diff
@@ -3602,6 +3602,35 @@ static std::unordered_map<std::string,int> g_usedNames;
...
+    case CREATE_SUBDIRS_t::D256:
+      return 4;
+    case CREATE_SUBDIRS_t::D512:
+      return 5;
+    case CREATE_SUBDIRS_t::D1024:
+      return 6;
+    case CREATE_SUBDIRS_t::D2048:
+      return 7;
+    default: /* NO, YES, D4096 */
+      return 8;
```
albert-github, line 3630: I understand that you set here also in case of `NO` to 8 from a convenience point of view, so when entering by accident the array `createSubdirsBitmaskL2` one doesn't get out of bounds but seen the code `convertNameToFile` this cannot happen maybe a better way would be to have the value `-1` and place a warning in this routine that we should never get in the `NO` part by means of an `assert`.

albert-github (approved):

doxygen:
Sorry for my late response. I think that combining a boolean option (`CREATE_SUBDIRS`) and an integer option (`CREATE_SUBDIRS_LEVEL`) into an enum option is not the prettiest solution and will lead to downgrade issues (e.g. reading a Doxyfile with `D64` set with a doxygen version that still expects a boolean). 

So I prefer the original solution (new option).

@makoehr can you make a pull request with just the first two commits (and maybe some typo fixes you did along the way)? I can also do it myself, but then you will not be listed as the author of this improvement.

albert-github:
@doxygen 
I only partly agree
- downgrading is not a normal situation
- in case a downgrade the values other that `YES` and `NO` will just give a warning and the default will be used (i.e. "unfortunately" `NO`).

makoehr:
> Sorry for my late response. I think that combining a boolean option (`CREATE_SUBDIRS`) and an integer option (`CREATE_SUBDIRS_LEVEL`) into an enum option is not the prettiest solution and will lead to downgrade issues (e.g. reading a Doxyfile with `D64` set with a doxygen version that still expects a boolean).
> 
> So I prefer the original solution (new option).
> 
> @makoehr can you make a pull request with just the first two commits (and maybe some typo fixes you did along the way)? I can also do it myself, but then you will not be listed as the author of this improvement.

I have created another pull request (https://github.com/doxygen/doxygen/pull/9280) that consists only of the first two commits from this pull request.

I have no preferences which of the two proposed pull requests is more suitable...

Regards

Mathias

doxygen:
This pull request has been replaced by #9280

PR #9079, opened 2022-01-27
Drop the support for Qt4

albert-github:
The last Qt4 version (4.8.7) was released on May 26, 2015

PR #9085, opened 2022-01-29
Adding Qt6 support, dropping Qt4 support

albert-github:
- Adding the support for the Qt6 version (The first Qt6 version was released December 8, 2020)
- Dropping support for the Qt4 version (The last Qt4 version (4.8.7) was released on May 26, 2015)


albert-github:
Note this proposed pull request makes the proposed pull request #9079 obsolete

CMakeLists.txt
```diff
@@ -29,9 +29,11 @@ option(use_libclang    "Add support for libclang parsing." OFF)
 option(static_libclang "Link to a statically compiled version of LLVM/libclang." OFF)
 option(win_static      "Link with /MT in stead of /MD on windows" OFF)
 option(english_only    "Only compile in support for the English language" OFF)
-option(force_qt4       "Forces doxywizard to build using Qt4 even if Qt5 is installed" OFF)
 option(enable_coverage "Enable coverage reporting for gcc/clang [development]" OFF)
 
+set(force_qt CACHE INTERNAL "Forces doxywizard to build using the specified major version, this can be Qt5 or Qt6")
```
doxygen, line 34: When removing the `force_qt` option, how does one force the Qt version to use from the command line?

addon/doxywizard/config.h
```diff
@@ -18,14 +18,22 @@
...
+#else
+class QStringEncoder;
+#endif
 
 bool parseConfig(
       const QString &fileName,
       const QHash<QString,Input *> &options
     );
 
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
```
doxygen, line 32: I see a lot of these `#if` constructs being introduced. Wouldn't it be better to introduce one or two adapter functions that internally do the conditional switching and use these adapters at the places where there is now a `#if`?

CMakeLists.txt
albert-github: Only the `force_qt4`  is gone, with `force_qt` one can do:
```
-Dforce_qt=Qt5
```
or
```
-Dforce_qt=Qt6
```

or did I misunderstand the question?

doxygen: No, I indeed see that this works. Was just expecting the option to show up in the list when using `cmake -L` like before.

addon/doxywizard/config.h
albert-github: Indeed there quite a few `#if` constructs and it is not very flexible (e.g. when newer versions have again different requirements.
Adapter functions might be a good idea, though I'm not good in creating adapter functions.

CMakeLists.txt
albert-github: Indeed I was hoping for it as well  but didn't find a way to do it.

addon/doxywizard/config.h
doxygen: Ok, I'll give it a try. 

I also see some errors when launching the Qt6 build of doxywizard from the command line regarding missing signals:

```
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QComboBox::activated(const QString &)
qt.core.qobject.connect: QObject::connect: No such signal QButtonGroup::buttonClicked(int)
qt.core.qobject.connect: QObject::connect: No such signal QButtonGroup::buttonClicked(int)
qt.core.qobject.connect: QObject::connect: No such signal QButtonGroup::buttonClicked(int)
qt.core.qobject.connect: QObject::connect: No such signal QButtonGroup::buttonClicked(int)
qt.core.qobject.connect: QObject::connect: No such signal QButtonGroup::buttonClicked(int)
```

So I think some signals have been renamed or gotten a different signature with Qt6.

albert-github: Strange, I didn't get any message. Which version of Qt did you try? I used the Windows 6.1.2/msvc2019_64 version, though as this version is a dynamic version I needed to add `...\6.1.2\msvc2019_64\bin` to the `PATH`.

doxygen: I used 6.1.3, but I think you would only see these messages on Windows if you launch the executable with console attached (i.e. run `editbin /SUBSYSTEM:CONSOLE,6.00 /OSVERSION:5.1 doxywizard.exe`)

albert-github: Indeed after the `editbin` I do see the messages as well.

- Should we add the `editbin` to the creation of the `doxywizard.exe` on Windows?

PR #9098, opened 2022-02-06
Automatically show correct year in about box

albert-github:
Automatically show copyright year of build in the about box

### DIFF
--- a/addon/doxywizard/doxywizard.cpp
+++ b/addon/doxywizard/doxywizard.cpp
@@ -14,6 +14,7 @@
 #include "version.h"
 #include "expert.h"
 #include "wizard.h"
+#include <ctime>
 
 #include <QMenu>
 #include <QMenuBar>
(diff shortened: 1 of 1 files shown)


addon/doxywizard/doxywizard.cpp
```diff
@@ -263,6 +264,12 @@ void MainWindow::manual()
 
 void MainWindow::about()
 {
+  auto now = std::chrono::system_clock::now();
+  std::time_t time = std::chrono::system_clock::to_time_t(now);
+  auto current = *localtime(&time);
+  char tmp[10];
+  sprintf(tmp,"%d",current.tm_year+1900);
```
doxygen, line 271: Since we have a dependency on Qt here, I think it better to use `int year = QDate::currentDate().year();`

albert-github: I wanted to have it as close as possible to the version used in the doxygen main program so I used the construct as proposed, but I see the point. Adjusted.

doxygen:
even the `tmp` variable was not needed.

albert-github:
Indeed as the construct changed from one line multiple lines and the `<<` operator also supports the `int` like a `%d`.

PR #9103, opened 2022-02-07
Automatically set right executable properties for windows

albert-github:
In case the `editbin` executable is available automatically set the right properties  of the created executables.
For de bugging purposes it might be useful to have the `doxywizard` not as a "windows" type of executable but as "console" executable, this can be accomplished by means of the setting of `enable_console=ON`.
The `enable_console` only operates on the `doxywizard` executable. Usage like: `cmake -Denable_console=ON ..`


CMakeLists.txt
```diff
@@ -80,6 +81,18 @@ if (CMAKE_SYSTEM MATCHES "Darwin")
     set(EXTRA_LIBS ${CORESERVICES_LIB})
 endif()
 
+find_program(EDITBIN editbin)
```
doxygen, line 84: shouldn't this be limited to Windows only. Suppose an `editbin` tool exists on Linux, it probably does something different...

albert-github: It is a possibility, I thought about it, I didn't think it was necessary, but it might happen that other OS have also an editbin.

addon/doxyapp/CMakeLists.txt
```diff
@@ -11,6 +11,14 @@ include_directories(
 add_executable(doxyapp
 doxyapp.cpp
 )
+if(EDITBIN)
```
doxygen, line 14: I wonder if it is it really need to explicitly call `editbin` on the executables other than doxywizard? 
I know I did that in the installer script, but I don't think it really makes a difference, or did you see an effect?

albert-github: It is indeed difficult to decide about it, one of the problems is that the, in my personal opinion, the Microsoft documentation is not clear about the consequences of the settings of `/OSVERSION` (which I could not find at all) and the version number in `/SUBSYSTEM`. From other sources I got that:
- `\OSVERSION` is the minimum OS version the product can run on.
- the version numbers:
  - 5.1: Windows XP
  - 6.0: Windows Vista
  - 6.1: Windows 7
  - 10.0: Windows 10

I see that on my system (Windows 10) the OS version in the executables is set to 6.0, so my executables could run from Windows Vista and newer.
Currently Microsoft only supports Windows 8.1, Windows 10 and Windows 11.
I think that a number of users still have (standalone) systems of Windows 7 and maybe even Windows Xp or Windows Vista).

Sources:
- https://docs.microsoft.com/en-us/cpp/build/reference/editbin-options?view=msvc-170
- https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
- https://www.gaijin.at/en/infos/windows-version-numbers
- https://docwiki.embarcadero.com/RADStudio/XE8/en/Linking (delphi related, but only place I found so far where the OS version is a little bit explained, found through https://stackoverflow.com/questions/30336886/how-to-set-the-minimum-required-windows)

So with this possibility it is easily possible to set minimum required OS version.

Should we use editbin on all executables? I think yes.
- gives a possibility to set the minimal version of the OS to be used, even though we cannot test it
- gives users that build the executables themselves an easy way to have the executables available on older system versions
- gives an easy way to overrule certain settings through the command line without the need to bother to find and run the `editbin` executable and the executables to update.
- gives also the possibility, for the future, to place the doxygen version in the executable (I did some experiments with `/VERSION` and this does look a good possibility as well, even though it is a bit limited e.g. (only 2 levels i.e. major.minor) , but when I do `/VERSION:19412.12345` I see in the executable: "19412.12345 image version" so this should give some possibilities (quickly looks like major and minor are each limited to 5 digits).

albert-github: As a small side note: I found the file winnt.h at D:\Windows Kits\10\include\10.0.19041.0\um  and around line 17090 I saw:
```
    WORD    MajorOperatingSystemVersion;
    WORD    MinorOperatingSystemVersion;
    WORD    MajorImageVersion;
    WORD    MinorImageVersion;
    WORD    MajorSubsystemVersion;
    WORD    MinorSubsystemVersion;
    DWORD   Win32VersionValue;
```
Unfortunately all without any explanation...
I think , by head, that a WORD is 16-bit and DWORD is 32-bit.


(For GCC under Cywin I found the same definitions)

albert-github:
Looks OK to me.

PR #9109, opened 2022-02-10
Fix build with MinGW and Ninja

HazardyKnusperkeks:
I don't think there should be a check on the generator but on the
compiler, because my gcc complains rightfully:
c++.exe: warning: /bigobj: linker input file unused because linking not done
c++.exe: error: /bigobj: linker input file not found: No such file or directory

And this in turn failed the configuration:
-- Performing Test ICONV_COMPILES - Failed

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,9 +81,7 @@ if (CMAKE_SYSTEM MATCHES "Darwin")
 endif()
 
 if (WIN32)
-    if ((NOT CMAKE_GENERATOR MATCHES "MinGW Makefiles") AND 
-        (NOT CMAKE_GENERATOR MATCHES "MSYS Makefiles") AND
-        (NOT CMAKE_GENERATOR MATCHES "Unix Makefiles"))
+    if (MSVC)
         if (NOT ICONV_DIR)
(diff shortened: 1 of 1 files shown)


albert-github:
Solution looks OK though:
- What is the CMake command that you give, i.e the `cmake -G ...`?
- can yo give the complete log file that you get when you run original CMake  ?

HazardyKnusperkeks:
``cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../doxygen -Duse_libclang=ON -Dstatic_libclang=ON -DLANG_CODES=DE;EN ../doxygen.git``

For the log file I can check later. I'm currently migrating my whole environment to newer versions and tools.

albert-github:
Thanks for the answer, the cmake line already give a clear indication.
The Ninja compiler can be run with the gcc compiler but also with the Microsoft compiler and the options are specific for the later one. A test based on the compiler might indeed be better than a test on the underlying make tool that is used.

HazardyKnusperkeks:
So, we could merge that, or do you still want a cmake log?

albert-github:
No I don't need the log anymore.

Issue #9108, opened 2022-02-10
collaboration diagram with template relations is broken if typedef is used (in some cases)


dhebbeker:
### Describe the bug
With `TEMPLATE_RELATIONS = YES` *some* relations are not drawn if a `typedef` is used instead of the "bare" template specialization.

In this example the combination of a reference to a `typedef` combined with a reference or `const` qualifier leads to the template relation not to be drawn. Although the same relation is drawn if the "bare" template specialization is used instead of the `typedef` reference.

#### Example

```c++
struct S1 {};
struct S2 {};
struct S3 {};
template <typename T> struct Temp1 {};
template <typename T> struct Temp2 {};

typedef Temp1<S1> TD1;
typedef Temp1<S2> TD2;
typedef Temp1<S3> TD3;

struct Collection
{
  TD1 as1;
  const TD2 as2;
  TD3& as3;
  
  Temp2<S1> bs1;
  const Temp2<S2> bs2;
  Temp2<S3>& bs3;
};
```

### Expected behavior

I would expect that the template relations which are shown for `bs2` and `bs3` would be draw for `as2` and `as3` as well.

### Screenshots

![grafik](https://user-images.githubusercontent.com/1594340/153439081-a15833f2-39ca-45cf-b773-b248d62f92ba.png)

### To Reproduce

Full example: [test.zip](https://github.com/doxygen/doxygen/files/8042185/test.zip)

### Version

 * doxygen version: `1.9.3 (c0b9eafbfb53286ce31e75e2b6c976ee4d345473)`
 * Windows 10 64bit

PR #9115, opened 2022-02-12
issue #9108 collaboration diagram with template relations is broken if typedef is used (in some cases)

albert-github:
To resolve typedefs the `const`, `&` etc. are not relevant and should be stripped during the resolve process.

(some routines had to be moved in the source to make them accessible)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -361,6 +361,66 @@ int guessSection(const QCString &name)
   return 0;
 }
 
+static void stripIrrelevantString(QCString &target,const QCString &str)
+{
+  if (target==str) { target.resize(0); return; }
+  int i,p=0;
+  int l=str.length();
+  bool changed=FALSE;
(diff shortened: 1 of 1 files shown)


albert-github:
I have to dive into this a bit deeper.

The problem with the `ATTRIBUTE` amazes me a bit as it shows the information for the explicit writing but not for the typedef version, but can be fixed for the `as3` by means of the setting `MACRO_EXPANSION=YES`.

dhebbeker:
I do not understand - from the user point of view - why it should make a difference for the "`ATTRIBUTE` problem", whether a `typedef` or the explicit type is used. The difference here is the need for setting `MACRO_EXPANSION=YES`.

dhebbeker:
The current state of the merge request is a significant improvement. I advocate to merge at least this state. :rocket:

Issue #5300, opened 2018-07-02
CHM treeview structure changed from v 1.7.6.1 to v 1.8.5 (Origin: bugzilla #708712)


doxygen:
status NEW severity *normal* in component *build* for *---*
Reported in version *1.8.5* on platform *Other*
Assigned to: Dimitri van Heesch

On 2013-09-25 06:02:25 +0000, Christian Schmidt wrote:

> I'm facing the following problem. I'm using the HtmlHelpCompiler with doxygen to produce a CHM output. When I use the lastest version of doxygen 1.8.5 the CHM treeview structure changed to the structure before when using 1.7.6.1 (I'm using the same settings file with Doxywizard). The main thing is that the documented classes are shown with full scope (namespace and class) in the treeview. What I want is to show the documented classes without having a node for the namespace and so on.
> The treeview structure of the mainpage documentation changed as well. The first
> node does not show the mainpage name it actually shows a section name of the
> mainpage.
> 
> Best regards,
> Chris

PR #9137, opened 2022-02-21
Incorrect representation of some items in HTMLHELP

albert-github:
When trying to generate the doxygen internal documentation with htmlhelp we see a number of items that are shown as pages but actually have sub pages and should thus be shown as books.
- general documentation
- namespace members
- file members
**Old**
image

**New**
image


ferdymercury:
Maybe related: https://github.com/doxygen/doxygen/issues/5300

Issue #9138, opened 2022-02-21
QHP index contains duplicated file entries


ferdymercury:
**Describe the bug**
As discussed in https://github.com/doxygen/doxygen/issues/9094#issuecomment-1046799213, the doxygen-generated QHP file contains duplicated entries in the file name list. (This QHP generation is disabled by default, but one can activate it to create a QtHelp book.)

**Expected behavior**
No duplicates are found in index.qhp

**Screenshots**
```
      <file>functions.html</file>
      <file>functions.html</file>
      <file>functions.html</file>
      <file>functions.html</file>
      <file>functions.html</file>
      <file>functions_func.html</file>
      <file>functions_func.html</file>
      <file>functions_func.html</file>
      <file>functions_func.html</file>
      <file>functions_func.html</file>
      <file>files.html</file>
      <file>globals.html</file>
      <file>globals.html</file>
      <file>globals.html</file>
      <file>globals_func.html</file>
      <file>globals_func.html</file>
      <file>globals_func.html</file>
```

**To Reproduce**
See reproducer at https://github.com/doxygen/doxygen/issues/9094

**Version**
1.9.4 (15dae504a167d838873053cc96417d8beae78df7)

albert-github:
I've just pushed a proposed patch, pull request #9139

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.4.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9139, opened 2022-02-21
issue #9138 QHP index contains duplicated file entries

albert-github:
Analogous to the implementation in `htmlhelp.cpp` place the files in a `StringSet` so they are unique (and sorted)

### DIFF
--- a/src/qhp.cpp
+++ b/src/qhp.cpp
@@ -148,7 +148,7 @@ class Qhp::Private
     std::ofstream docFile;
     TextStream doc;
     TextStream index;
-    TextStream files;
+    StringSet files;
     QhpSectionTree sectionTree;
 };
 
@@ -266,8 +266,6 @@ void Qhp::initialize()
 
   writeIndent(p->index,2);
   p->index << "<keywords>\n";
-  writeIndent(p->files,2);
-  p->files << "<files>\n";
 }
 
 void Qhp::finalize()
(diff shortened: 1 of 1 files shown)


ferdymercury:
Thanks! I just compiled it and it solves it :)

doxygen:
This works around the issue nicely, but why are the same entries added multiple times in the first place?

albert-github:
Indeed it works nicely around the issue and I used this way as it is consistent with other places where the addIndexFile is is implemented (htmlhelp). In docsets, eclipsehelp and ftvhelp the addIndexFile is defined as an empty function.

I did some research regarding the question "why are the same entries added multiple times in the first place".

In the file index.cpp we see in the function `static void writeClassMemberIndexFiltered(OutputList &ol, ClassMemberHighlight hl)` the lines:
```
...
  for (const auto &kv : g_classIndexLetterUsed[hl])
  {
...
    ol.startFile(fileName+extension,QCString(),title);
...
```
analogous constructs we find in:
```
writeFileMemberIndexFiltered
writeNamespaceMemberIndexFiltered
```

Note that in `renderMemberIndicesAsJs` there is also a similar construct:
```
 for (const auto &kv : memberLists[i])
 {
```

Furthermore we have the function `sortMemberIndexList`, maybe here some unique operation might take place as well.

Issue #9145, opened 2022-02-22
ALIASES in Python docstrings with ''' are not expanded.


mcpat:
**Describe the bug**
Doxygen seems to handle Python docstrings started with `"""` differently compared to those started with `'''`. In the first case, all aliases in the Doxyfile will be correctly expanded. In the latter case no alias seems to be found but instead something like the following is issued

> warning: Found unknown command '@myfancyalias'

**Expected behavior**
Both docstring styles behave the same.

**Version**
I checked with version 1.8.17 (Ubuntu Focal) and version 1.9.1 (most recent Ubuntu version).

albert-github:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

Note that in the documentation (see: https://www.doxygen.nl/manual/docblocks.html#pythonblocks) there:
```
Note
When using """ none of doxygen's special commands are supported and the text is shown as verbatim text
see \verbatim. To have the doxygen's special commands and have the text as regular documentation instead
of """ use """! or set PYTHON_DOCSTRING to NO in the configuration file.
Instead of """ one can also use '''.
```

But to be able to give a better answer a small example is needed.

mcpat:
Sure, here you are:
[bug.tar.gz](https://github.com/doxygen/doxygen/files/8117166/bug.tar.gz)

![bug](https://user-images.githubusercontent.com/225660/155146804-2958f5e1-8995-4969-bec9-d2d30a38895b.png)


```
class SomeClass(object):
    def func1(self):
        """!
        @brief Docstring with `3x"`

        @myfancyalias
        """
        pass

    def func2(self):
        '''!
        @brief Docstring with `3x'`

        @myfancyalias
        '''
        pass
```
And yes, I saw the documentation entry

albert-github:
I've just pushed a proposed patch, pull request #9147

ddeepwel:
Please merge this pull request! I will find it very useful

albert-github:
This problem has also been solved by means of the PR #10495 and all looks OK.

I think the problem has been solved in the mean time.

PR #9147, opened 2022-02-22
issue #9145 ALIASES in Python docstrings with ''' are not expanded.

albert-github:
When handling `"""!` also the `'''!` should be handled in, internal, comment conversion


albert-github:
This PR is obsolete as the problem has also been solved by means of the PR #10495

PR #9161, opened 2022-02-25
Reinstate warnings about stale translations

olaf-mandel:
While replacing string comparisons with enums in commit
a0fb56e0560d657d6fde8ade9e74624c9412eb4d, the warning about stale
translations was also removed:

    The selected output language "foo" has not been updated since
    release x.y.z.  As a result some sentences may appear in English.

Reinstate the warning as that is useful and unrelated to the original
focus of the removing commit.

### DIFF
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -301,4 +301,7 @@ void setTranslator(OUTPUT_LANGUAGE_t langName)
 #endif
 #endif
   }
+
+  QCString msg = theTranslator->updateNeededMessage();
+  if (!msg.isEmpty()) warn_uncond("%s", qPrint(msg));
 }
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -303,5 +303,5 @@ void setTranslator(OUTPUT_LANGUAGE_t langName)
   }
 
   QCString msg = theTranslator->updateNeededMessage();
-  if (!msg.isEmpty()) warn_uncond("%s", qPrint(msg));
+  if (!msg.isEmpty()) ::msg("%s", qPrint(msg));
 }


src/language.cpp
```diff
@@ -301,4 +301,7 @@ void setTranslator(OUTPUT_LANGUAGE_t langName)
 #endif
 #endif
   }
+
+  QCString msg = theTranslator->updateNeededMessage();
+  if (!msg.isEmpty()) warn_uncond("%s", qPrint(msg));
```
albert-github, line 306: As is it is good that the user is informed about a not updated language, but doing this through the `warn...` mechanism is not good as when the user runs a CI job and has set `WARN_AS_ERROR` to `YES` or `FAIL_ON_WARNINGS` the job will stop.

It is probably netter to use the `msg` mechanism (though the user will not see the message when `QUIET=YES`).

albert-github (changes requested):

src/language.cpp
olaf-mandel: This change simply reinstates the previous behavior, i.e. versions before 1.9.3 *did* error out for the described settings (I presume).

If that is to be changed, should it be done here or in a separate MR?

albert-github: I don't know why @doxygen did take it out, but when we would reinstate it I think we should take into consideration the effect effect regarding the `warn` versus `msg` (I think the old behavior has some negative side effects, especially for languages that are nearly up to date). 
I think @doxygen should decide whether it should be reinstated and when reinstated in which form.

doxygen: I think this warning was accidentally removed. I also have a preference for using `msg` for the reason @albert-github  mentioned.

olaf-mandel: Added a second commit that downgrades the severity.

olaf-mandel: Whoops, now it compiles...

olaf-mandel: I should have said "now it compiles for me". I can't see why the [miktexsetup_standalone](https://github.com/doxygen/doxygen/runs/5346726527?check_suite_focus=true#step:11:2) command in the [Windows MSVC Debug](https://github.com/doxygen/doxygen/runs/5346726527) pipeline failed. I think this is a sporadic failure, unrelated to the actual code change. I don't know how to trigger a rerun, though.

albert-github: As the problem is just at one platform don't worry about it it is indeed a sporadic failure.

PR #9162, opened 2022-02-25
Modernize LANGUAGE.HOWTO

olaf-mandel:
* replace qmake references with cmake instructions
* replace latin-1 references with utf-8 ones (including the umlaut)
* update list of fields in Translator class to be edited

### DIFF
--- a/LANGUAGE.HOWTO
+++ b/LANGUAGE.HOWTO
@@ -6,40 +6,48 @@ Just follow these steps:
    is already working on support for that language, you will be 
    assigned as the maintainer for the language. I'll create a 
    list on Doxygen's homepage, so everyone knows who is doing what.
-2) Create a copy of translator_en.h and name it 
+2) Edit src/config.xml:
+   - find the <option> with id='OUTPUT_LANGUAGE'
+   - add a new value with your language and an optional description:
+     <value name='YourLanguage' desc='(only for disambiguation)'/>
+3) Create a copy of translator_en.h and name it
    translator_<your_2_letter_country_code>.h
    I'll use xx in the rest of this document.
-3) Edit language.cpp:
-   - Add a #include<translator_xx.h> 
+4) Edit language.cpp:
+   - Add an include block:
+
+     #ifdef LANG_XX
+     #include "translator_xx.h"
+     #endif
+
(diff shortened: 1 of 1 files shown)


LANGUAGE.HOWTO
```diff
@@ -6,40 +6,42 @@ Just follow these steps:
...
-           leave the Html and man output for what it is (which is fine, if
-           idLanguageCharset() is set correctly).
-        b) Use html codes like &auml; for an a with an umlaut (i.e. �).
+        a) Enter them directly and store the files using UTF-8 encoding.
+        b) Use html codes like &auml; for an a with an umlaut (i.e. ä).
            See the HTML specification for the codes.
-6) Run configure and make again from the root of the distribution, 
-   in order to regenerate the Makefiles.
+6) Delete the CMakeCache.txt file and run cmake and make again from the root of
+   the distribution, in order to regenerate the binary.
```
albert-github, line 44: There is a internal discussion going on regrading the fact whether or not all languages should by default be supported or have the selection mechanism. @doxygen what should we do?

I'm not sure whether or not the complete removal of the `CMakeCache.txt` is the best choice as you have to run the initial cmake command again with all its options, maybe better to delete the line with `LANG_CODES` and run just `cmake ..`

olaf-mandel: Rewrote the part about deleting `CMakeCache.txt` to run `cmake -U LANG_CODES .`

albert-github: I didn't know about the `-U` option of `Cmake`, but it looks like this nicely solves the issues we have. 

A bit nitpicking but this is of course only necessary when we update an existing build directory).

doxygen: > There is a internal discussion going on regrading the fact whether or not all languages should by default be supported or have the selection mechanism. @doxygen what should we do?

I think we should  remove the compile time language selection, and just always support all languages.

albert-github: @doxygen I agree, this would also mean to remove remove the English only option.
I think it would be good to have this issue integrated, so the documentation is better, afterwards I can have a go at removing the mentioned dependency.

Issue #9159, opened 2022-02-25
Doxygen produces ugly empty space in LaTeX and PDF output because of hypertargets


heikki-heikkila:
**Describe the bug**
When doxygen produces LaTeX files, it includes hyperlinks and hypertargets in *.tex files.  Occasionally a `\Hypertarget` command is inserted inside a TeX box, as follows:
```
\mbox{\Hypertarget{...
```
By the way, this behaviour apparently has something to do with file https://github.com/doxygen/doxygen/blob/master/src/latexgen.cpp

Occasionally this `\mbox` is inserted in the main vertical list of TeX typesetting between two heading macros.  **This is a bug.**  In the particular case at hand, a file named test_8h.tex (which doxygen produces as output) contains something like this:
```
\doxysubsubsection{Function Documentation}
\mbox{\Hypertarget{test_8h_a024e8f3ced8cfd2c34f7cfd254093e5e}\label{test_8h_a024e8f3ced8cfd2c34f7cfd254093e5e}}
\index{test.h@{test.h}!example@{example}}
\index{example@{example}!test.h@{test.h}}
\doxyparagraph{\texorpdfstring{example()}{example()}}
```
Note two headings, `\doxysubsubsection` and `\doxyparagraph`, and a `\mbox` between them.  The result, in the main vertical list of TeX, contains, if penalties are excluded: 
1. a box for "Function Documentation"; 
2. vskip to separate heading from the following item (which would often be a paragraph of text);
3. mbox, which is the next item;
4. another vskip to the separate the next heading from the previous item; 
5. a box for heading "example()".

In the final output, vskips (2) and (4) are visible after one another.  Then the final printable output contains **double** the amount of normal vertical space between the headings "Function Documentation" and "example()".

This is clearly visible in file refman.pdf, on page 2, between headings 2.1.3 and 2.1.3.1. 

**Expected behavior**
It is expected that hypertargets are adjacent to, or contained in, the relevant TeX vbox and never produce any superfluous empty space in the final LaTeX or PDF document.

**Screenshots**
None (but see the PDF file)

**To Reproduce**
This can be reproduced **exactly** as another issue.  See https://github.com/doxygen/doxygen/issues/9028.  The same files, Doxyfile and test.h, and the same PDF file, refman.pdf, show also the behaviour described in this issue

**Version**
Doxygen 1.8.17.  Linux 652723cd6b4f 3.10.0-1062.9.1.el7.x86_64

**Stack trace**

**Additional context**

[...]

Issue #6093, opened 2018-07-02
Underscores in type or member name cause unwanted hyphenation in PDF output Data Fields (Origin: bugzilla #773231)


doxygen:
status RESOLVED severity *normal* in component *general* for *---*
Reported in version *unspecified* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[ugly_tabu_hyph.tar.gz](http://ftp.stack.nl/pub/users/dimitri/bugzilla/attachments//bug773231_338042_ugly_tabu_hyph.tar.gz)_ (ID 338042)

On 2016-10-19 18:27:42 +0000, kees@ijzerbout.nl wrote:

> Created attachment 338042
> Example to show hyphenation in type name
> 
> This shows up in the PDF (LaTeX) output. Type names like uint32_t are hyphenated in a table for the fields of a struct. The table takes less than half of a page width, so it is not caused by lack of width.
> 
> I've attached a small example to show what I mean.
> Untar and run "doxygen && make -C latex", and look at the last page of the PDF.

On 2016-10-23 11:21:05 +0000, Dimitri van Heesch wrote:

> Latex hyphenation is very difficult to get right and interacts badly with the tabu package.
>  
> I'll suppress the hypenation hints in the 'type' and 'name' columns. Hopefully this doesn't lead to too many ill side-effects.

On 2016-12-29 18:45:37 +0000, Dimitri van Heesch wrote:

> This bug was previously marked ASSIGNED, which means it should be fixed in
> doxygen version 1.8.13. Please verify if this is indeed the case. Reopen the
> bug if you think it is not fixed and please include any additional information 
> that you think can be relevant (preferably in the form of a self-contained example).

PR #9163, opened 2022-02-25
issue #9159 Doxygen produces ugly empty space in LaTeX and PDF output because of hypertargets

albert-github:
The fix as done for issue #6093
```
Commit: d571efb062fbe17d7257f3971e3db6c9cba833d0 [d571efb]
Date: Sunday, October 23, 2016 1:15:37 PM
```
introduced this side effect, this has been corrected.

### DIFF
--- a/src/docbookgen.cpp
+++ b/src/docbookgen.cpp
@@ -837,7 +837,7 @@ DB_GEN_C
 }
 void DocbookGenerator::startDoxyAnchor(const QCString &fName,const QCString &,
                                  const QCString &anchor,const QCString &,
-                                 const QCString &)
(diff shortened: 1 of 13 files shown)


doxygen:
@albert-github I've implemented a more local solution, see https://github.com/doxygen/doxygen/commit/0a881fd19e0f22a1990bd0d387f77e79bcc60db7

albert-github:
Indeed is a bit overdone to add an extra argument, but looked like the easiest way and gave flexibility for the future when in other output formats something similar would be needed. Looks like that the more local solution fixes the problem as well (and when something is needed in another format probably a local fix would also be better).

PR #9165, opened 2022-02-25
add PREDICTABLE_URLS configuration option

ilmari-lauhakangas:
This allows to predict the generated URL for classes and members from
only the class and member name.

Enumerations get their own page too and are no longer documented on
the namespace page.


ilmari-lauhakangas:
We need this in order to generate URLs to the LibreOffice API docs from our wiki.

ilmari-lauhakangas:
Sorry for cold pinging, but @doxygen any chance to review this? This patch is [originally from 2013](https://sourceforge.net/p/doxygen/mailman/doxygen-develop/thread/5183D2D1.7000000%40redhat.com/#msg30802342), so would be nice to get it in finally.

PR #9166, opened 2022-02-26
Moving punctuation dependency to translator

albert-github:
Moving a dependency of the used `OUTPUT_LANGUAGE` into the respective translators.

### DIFF
--- a/src/translator.h
+++ b/src/translator.h
@@ -143,7 +143,7 @@ class Translator
     virtual QCString getLanguageString() = 0;
 
     /** add punctuation at the end of a brief description when needed and supported by the language */
-    virtual const inline bool needsPunctuation() {return true;}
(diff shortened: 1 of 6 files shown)


src/translator.h
```diff
@@ -143,7 +143,7 @@ class Translator
     virtual QCString getLanguageString() = 0;
 
     /** add punctuation at the end of a brief description when needed and supported by the language */
-    virtual const inline bool needsPunctuation() {return true;}
+    virtual inline bool needsPunctuation() {return true;}
```
doxygen, line 146: `inline` is also not useful for virtual functions especially when called via the base class, adding `const` after `()` could be done as this method will not modify the state of the object.

albert-github: I've found some more problems, so I'm looking into the issue.

Issue #9188, opened 2022-03-07
Python: module-doc-string not recognized if module starts with a comment


thom-sch:
If a python module starts with comment lines (lines beginning with an "#") subsequent module doc string is not recognized.
An exception is a shebang comment as first line.
This behavior is independent from configuration option PYTHON_DOCSTRING

**Expected behavior**
Python allows a coding string (also preceeded by "#") as first or second line; thus Dexygen shold at least recognize this as valid comment before the module doc string, too.
Additionally it may be useful to recognize the module doc string regardless of the number of comment lines ("#" or "##") preceeding it.

**To Reproduce**
s. attached ZIP with default configuration option file:
[Test17_python__doc__string.zip](https://github.com/doxygen/doxygen/files/8196608/Test17_python__doc__string.zip)

**Version**
- Win10/64
- configuration: INPUT                  = .
- doxgen version: 1.9.3 (c0b9eafbfb53286ce31e75e2b6c976ee4d345473)

Reagards from Meerbusch
Thomas

albert-github:
Looks like that the shebang is explicitly ignored when it is on the first line of the code (from pyscanner.l):
```
    {SCRIPTCOMMENT}   { // Unix type script comment
                        if (yyextra->yyLineNr != 1) REJECT;
                      }
```
For normal `#` comments the following rule applies:
```
    {POUNDCOMMENT}    { // normal comment
                        yyextra->packageCommentAllowed = FALSE;
                      }
```
but probably a more sophisticated detection of the line has to take place (i.e. as long as we only have empty or single (!) `#` lines don't set the `yyextra->packageCommentAllowed = FALSE;`.

albert-github:
I've just pushed a proposed patch, pull request #9193

thom-sch:
sorry for late answer - I had to interrupt my tentative steps on Python due to other work...

Now I've tested your patch and it works fine for me.
Seem's that Doxygen now simply accepts any number of comment lines preceeding the docstring?

Regards from Meerbusch
Thomas

albert-github:
Looks like the problem has been solved by an even simpler fix see the Issue #9672 / PR #9673

thom-sch:
Hello Albert,

thanks for the info.
I've just done a short test - looks well on my example and on my real life project, too.

Feel free to close this issue.

Regards from Meerbusch
Thomas

PR #9193, opened 2022-03-08
issue #9188 Python: module-doc-string not recognized if module starts with a comment

albert-github:
Counting single `#` comment lines and empty lines, so that when this number is equal to the current line number a package comment is still possible


albert-github:
obsolete as the problem has also been solved by means of the Issue #9672 / PR #9673

Issue #9182, opened 2022-03-04
Better support for procedural C


Wurzelkoch:
In procedural languages like C, functions often are more important and sometimes more complicated than data structures and should have an analogous index page.

A first step however would be to make `MAX_ITEMS_BEFORE_MULTIPAGE_INDEX` and `MAX_ITEMS_BEFORE_QUICK_INDEX` available for config at runtime and document those. Right now I am running a custom build only to get a complete function index page.

(This is related to, but imho not a duplicate of #9107)

Issue #9107, opened 2022-02-10
How to generate a function index?


paddor:
I'm generating documentation from a C library. I see a page of structs, but no page of functions. I'd have to go to the list of files and then click on the specific file to see all its functions.

Ideally I'd see all functions with the same prefix as the struct I'm looking at. E. g. on the page for `struct XY_MyThing` (typedef'd to `XY_MyThing`) I'd not only see all its field, but also see a list of functions like `XY_MyThing_init(XY_MyThing *thing)` and `XY_MyThing_setFoo(XY_MyThing *thing, const char *foo)`.

I can't find an option for this in the FAQ nor in the generated Doxyfile.

albert-github:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.
- Please also specify the **full** doxygen version used (`doxygen -v`).

albert-github:
Problem might be. but without example it is all a bit guessing:
- missing documentation for the structs or an encompassing part (try `EXTRACT_ALL=YES`)

You could wrap up an example with just one struct in it and your used Doxyfile and show this as an example (after, of course, locally testing it).

paddor:
I do have `EXTRACT_ALL=YES` set. The structs are documented just fine. But there's no function index. I'd like to have an index of all functions.

albert-github:
Example please so we can se what you are missing.

paddor:
@ferdymercury Thanks for the help but that page only lists functions starting with "a". Yes, I do have ALPHABETICAL_INDEX enabled.

albert-github:
@paddor to be able to help you properly we **need** a small example (at least the Doxyfile and some images showing what you see and what you would expect, also a complete example would be beneficial as well):
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

paddor:
I'd appreciate it if you could just tell me whether Doxygen has this feature or not. I'm starting to think that it's lacking this very basic feature. It's a **function index of all function**.

albert-github:
I think that this is available:

![image](https://user-images.githubusercontent.com/5223533/154041964-2abf000d-75b6-47b5-b6eb-30da873a4617.png)

this image is taken for the documentation of the doxygen internal docs.

Or as showing the functions:
[...]

PR #9194, opened 2022-03-08
Make MAX_ITEMS_BEFORE_X_INDEX subject to runtime config

Wurzelkoch:
For large c projects, a function index is more urgently needed then a struct index. This is a quick fix to make globals_func.html a usable index.


src/config.xml
```diff
@@ -1809,6 +1809,22 @@ to disable this feature.
...
  (or a list of prefixes) that should be ignored while generating the index headers.
+]]>
+      </docs>
+    </option>
+    <option type='int' id='MAX_ITEMS_BEFORE_QUICK_INDEX' minval='0' maxval='3000' defval='30'>
+      <docs>
+<![CDATA[
+ MAX_ITEMS_BEFORE_QUICK_INDEX defines, when to set the global lists of functions, 
+ variables, typdefs, enums, macros grouped into sections by starting letter. Default value: 30
+]]>
```
albert-github, line 1820: Default value should not be mentioned (is automatically added when building the documentation etc.)

The formulation of the sentence is not correct, to be honest I don't understand it. Furthermore the `MAX_ITEMS_BEFORE_QUICK_INDEX` should be preceded by `The \c `. See formulations of other items.

A reference to `MAX_ITEMS_BEFORE_MULTIPAGE_INDEX` should be present.

albert-github (changes requested):

albert-github:
Related issue #9107 and #9182 and the comment https://github.com/doxygen/doxygen/issues/9107#issuecomment-1040284046

src/index.cpp
```diff
@@ -49,8 +49,11 @@
 #include "tooltip.h"
 #include "utf8.h"
 
-#define MAX_ITEMS_BEFORE_MULTIPAGE_INDEX 200
-#define MAX_ITEMS_BEFORE_QUICK_INDEX 30
+//#define MAX_ITEMS_BEFORE_MULTIPAGE_INDEX 2000
+//#define MAX_ITEMS_BEFORE_QUICK_INDEX 30
+
+#define MAX_ITEMS_BEFORE_MULTIPAGE_INDEX (int) Config_getInt(MAX_ITEMS_BEFORE_MULTIPAGE_INDEX)
+#define MAX_ITEMS_BEFORE_QUICK_INDEX (int) Config_getInt(MAX_ITEMS_BEFORE_QUICK_INDEX)
```
albert-github, line 56: Why using `(int)` here? The `ConfigGetInt` returns an `int`.

albert-github (changes requested):

src/index.cpp
Wurzelkoch: It solved some compiler warnings about comparing unsigned long to int.

Wurzelkoch: I thought it solved some compiler warnings about comparing int to unsigned long, that turned out to be unrelated.

src/config.xml
```diff
@@ -1815,16 +1815,19 @@ to disable this feature.
     <option type='int' id='MAX_ITEMS_BEFORE_QUICK_INDEX' minval='0' maxval='3000' defval='30'>
       <docs>
 <![CDATA[
- MAX_ITEMS_BEFORE_QUICK_INDEX defines, when to set the global lists of functions, 
- variables, typdefs, enums, macros grouped into sections by starting letter. Default value: 30
+ If the number of items on al list of functions, variables, typdefs, enums and macros 
+ exceeds \c MAX_ITEMS_BEFORE_QUICK_INDEX, it will be divided into multiple sections
+ to form a quick index. If the number of items exceeds \c MAX_ITEMS_BEFORE_MULTIPAGE_INDEX, 
+ those sections will be put on a single output page each.
```
albert-github, line 1821: Better, though:
- first line `al` should be `a`
- first line `and` shouldn't this be `or` or should there be something like the word `combined` in the sentence.
- `\c MAX_ITEMS_BEFORE_MULTIPAGE_INDEX` should be `\ref cfg_max_items_before_multipage_inxdex "MAX_ITEMS_BEFORE_MULTIPAGE_INDEX"`
- last sentence, I think the word `each` shouldn't be present.

src/config.xml
```diff
@@ -1815,16 +1815,19 @@ to disable this feature.
...
       </docs>
     </option>
     <option type='int' id='MAX_ITEMS_BEFORE_MULTIPAGE_INDEX' minval='0' maxval='30000' defval='200'>
       <docs>
 <![CDATA[
- MAX_ITEMS_BEFORE_MULTIPAGE_INDEX defines, when to set the global lists of functions, 
- variables, typdefs, enums, macros grouped into multiple pages by starting letter. Default value: 200
+ If the number of items on a list of functions, variables, enums, typedefs or macros 
+ exceeds \c MAX_ITEMS_BEFORE_MULTIPAGE_INDEX, it will be 
+ divided into multiple pages, each page containing a section of a quick index. (see \c MAX_ITEMS_BEFORE_QUICK_INDEX)
```
albert-github, line 1830: - the see should be reformulated, see remark with the other item.
- last sentence is a bit confusing, we are in the page for e.g. `functions` and when the number of functions exceeds the number given this specific page.
  maybe instead of `each page containing a section of a quick index.` something like `each sub-page containing a part of the items`.

albert-github (changes requested):

martymichal:
Recently I came to wonder whether there is such a feature in Doxygen. Looks like this is still in the making. I wonder whether it would make sense to make this the default behaviour when using `OPTIMIZE_OUTPUT_FOR_C = YES` instead having an option requiring some magic number that "works".

doxygen:
@martymichal can you explain what behavior you are looking for? One (potentially huge) page with all functions, or something else?

martymichal:
> One (potentially huge) page with all functions

That's exactly what I'm looking for. In our project the vast majority of functions share the same prefix so they end up in the same letter category while all of the other functions that do not have the common prefix end up creating other letter sub-pages that are incomparably shorter in length.

I've seen this pattern of having a single list of functions in a module in API documentations generated by [gi-docgen](https://gitlab.gnome.org/GNOME/gi-docgen) or [gtk-doc](https://gitlab.gnome.org/GNOME/gtk-doc). Some examples of this are in the documentations of [Cairo](https://www.cairographics.org/manual/api-index-all.html) or [GLib](https://docs.gtk.org/glib/#functions).

The categorized index is good but I'd like to have the option to see all of the functions at once.

Issue #7543, opened 2020-01-27
Support PlantUML diagrams inside plantuml code blocks


TheDallenger:
Many makdown renderes that supports PlantUML, requires the PlantUML code to be inside a clode block/fence to work. Like this (execpt the '!'):
! ~~~plantuml
@startuml
class Test
@enduml
! ~~~

But doxygen will show this like a regular code block without language syntax, and does not render the diagram:
~~~
@startuml
class Test
@enduml
~~~

The result is that normal markdown renderes and doxygen are incompatible regarding PlantUML diagrams.

albert-github:
Looks like this issue has a proposed pull request #6953 already.

TheDallenger:
Thank you for your response, and sorry for not discovering the pull request myself.

albert-github:
There is no need to close the issue as it signals the need of other people for the use of this feature and also prevents others to open new issues as long as the proposed pull request is not released.

mwerner96:
> Looks like this issue has a proposed pull request #6953 already.

Hi, ist there any chance this PR will be integrated soon? It looks like the PR is stale, even though it would be a great feature to have.

albert-github:
There has been a new pull request (#9196 superseding but identical to #6953) that has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release)

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.4.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9196, opened 2022-03-10
Support PlantUML in Markdown

mwerner96:
GitHub and GitLab render PlantUML diagrams in Markdown documents inline and don't display the code. This PR makes Doxygen behave in the same way.
This PR is an adaption of @ArjanKn's pull request #6953 to compile with the currend HEAD.
It addresses issue #7543.


albert-github:
This pull request is geared toward `plantuml` but shouldn't a similar approach be taken for: `dot`, `msc` ?

Issue #9192, opened 2022-03-08
"Packages" has a "Packages" in the tree


Codeberg-AsGithubAlternative-buhtz:
In the tree of my generated documentation the node/entry `Packages` appears two times.

![image](https://user-images.githubusercontent.com/11861496/157206006-33bc39d9-7d50-4ddd-a340-a0414febfd20.png)

Sidenode: This is not about `mypackage` which I know why it appears two times in the tree.

The full project including generated html is in that archive: [project.zip](https://github.com/doxygen/doxygen/files/8204516/project.zip).
.

I use doxygen `1.9.4 (57efad44bfc02744ba5fd8ea820b54a2443e3771*)` from git on Debian 11 (`Linux TONNE 5.10.0-11-amd64 #1 SMP Debian 5.10.92-1 (2022-01-18) x86_64 GNU/Linux`) with Python 3.9.

This is the file and folder structure of the project:
```
project
├── docs
│   ├── Doxyfile
│   └── html
│       ├── ...
│       └── ...
├── LICENSE
├── README.md
├── src
│   ├── mypackage
│   │   ├── __init__.py
│   │   └── mypackage.py
│   ├── setup.cfg
│   └── setup.py
└── tests
```

This is `doxygen -x`:
```
# Difference with default Doxyfile 1.9.4 (57efad44bfc02744ba5fd8ea820b54a2443e3771*)
PROJECT_NAME           = Project
OPTIMIZE_OUTPUT_JAVA   = YES
EXTRACT_ALL            = YES
EXTRACT_PRIVATE        = YES
EXTRACT_STATIC         = YES
EXTRACT_LOCAL_METHODS  = YES
SORT_MEMBER_DOCS       = NO
WARN_AS_ERROR          = YES
INPUT                  = ../src/mypackage
FILE_PATTERNS          =
RECURSIVE              = YES
EXCLUDE                = ./_trash
EXAMPLE_PATTERNS       =
HTML_TIMESTAMP         = YES
GENERATE_TREEVIEW      = YES
MATHJAX_RELPATH        = https://cdn.jsdelivr.net/npm/mathjax@3
GENERATE_LATEX         = NO
HAVE_DOT               = YES
UML_LOOK               = YES
DOT_UML_DETAILS        = YES
```

albert-github:
Which version did you use, 1.9.4 is the current master so the full id is necessary from `doxygen -v`

Codeberg-AsGithubAlternative-buhtz:
[...]

PR #9197, opened 2022-03-10
issue #9192 "Packages" has a "Packages" in the tree

albert-github:
The problem here is the setting `OPTIMIZE_OUTPUT_JAVA   = YES, when  not having this setting we see "Namespaces" and "Namespace list" though for the "Package list" there is no translation function.
A function `trPackageList` has been introduced.


albert-github:
- did you apply this proposed patch? As it is mot in master only in the branch of this patch.

Codeberg-AsGithubAlternative-buhtz:
> * did you apply this proposed patch? As it is mot in master only in the branch of this patch.

I tried. I followed the link to `albert-github:feature/issue_9192`  which is `https://github.com/albert-github/doxygen/tree/feature/issue_9192` and used the Clone-Repo URL there.
`https://github.com/albert-github/doxygen.git`

I have not checked that URL but it seems like it is wrong. I would call this a Bug on GitHub but MS is not interested in my use opinion. :D

So how do I clone explicit a branch/patch?

albert-github:
I never have to do this for my own repository, but I gave it a try and I think the following steps are requited:
- `git clone https://github.com/albert-github/doxygen`
- `git checkout feature/issue_9192`

and now you can build the mentioned patch.

From the email I got for the proposed patch you might also try to apply, locally, the patch onto the current master:
-    https://github.com/doxygen/doxygen/pull/9197.patch
-    https://github.com/doxygen/doxygen/pull/9197.diff

Codeberg-AsGithubAlternative-buhtz:
OK, now it is working. The second `Packages` is now `Package List`. But why? :D
It still does not makes much sense for me.

Why not move the content under _Package List_ to _Packages_. Would make sense to me, save one click and one tree level?

Are there any situations where other entries appear under _Packages_ instead of _Package List_ and _Package Functions_?

And by the way (maybe a new Issue): Under _Package Functions_ there are more then just functions but also _Variables_.

albert-github:
> Are there any situations where other entries appear under Packages instead of Package List and Package Functions?

By default no, but:
- for consistency with other tree items (like `Class`, `Interfaces` ....)
- the user can add add user defined sub item, e.g. `<tab type="user" url=".." title="..."/>`, see: https://www.doxygen.nl/manual/customize.html#layout)

> And by the way (maybe a new Issue): Under Package Functions there are more then just functions but also Variables.

When not using `OPTIMIZE_OUTPUT_JAVA   = YES` the displayed text is `Namespace Members` so the translation is probably not the best (as reference: the translator function is called: `trPackageMembers`)

albert-github:
Regarding the remark about "Package Functions" I've just pushed a proposed patch, pull request #9206

albert-github:
The pull request regarding the "Package Functions" (#9206) has been integrated in master on GitHub

Codeberg-AsGithubAlternative-buhtz:
I would say we can merge this PR now.

PR #9198, opened 2022-03-11
Extended support for language specified in markdown fenced code blocks

albert-github:
Extended support for image formats to be rendered in markdown fenced code blocks (see also #9196, #7543):
- plantuml
  - render as image in case plantuml support is enabled otherwise just as plain code
  - add plantuml commands in case they are missing.
- dot, msc added analogous to plantuml

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8232065/example.tar.gz)

### DIFF
--- a/doc/markdown.doc
+++ b/doc/markdown.doc
@@ -459,6 +459,25 @@ Another way to denote fenced code blocks is to use 3 or more backticks (```):
     also a fenced code block
     ```
 
+For the image formats `dot`, `msc` and `plantuml` the fenced block will be shown as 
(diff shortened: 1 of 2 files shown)


MrNice2001:
> Did you try it? I tried it as I thought this should work, though it didn't. Can you please create a new issue for this usage (with `@startjson`), including as small self contained example (source+configuration file in a, compressed, tar or zip file) that allows us to reproduce the problem and attach this to the new issue? Please don't add external links as they might not be persistent.

Yes, I tried it out and opened a new [issue](https://github.com/doxygen/doxygen/issues/9950) which includes a small example. 

Thank you for your swift reply!!!

Issue #9200, opened 2022-03-11
Doxygen cannot resolve link to HTML anchor page


lcarlier:
**Describe the bug**
Doxygen cannot resolve link to HTML anchor page in the same document.
![image](https://user-images.githubusercontent.com/5570374/157910517-1e257188-dba3-4fa2-a80e-2f957d10793d.png)

**Expected behavior**
The link should be generated
![image](https://user-images.githubusercontent.com/5570374/157910867-0dae39a5-b32c-4dc9-8f4a-c68cba31e567.png)

**Screenshots**
/home/lcarlier/Projects/bugDoxygen/README.md:6: warning: unable to resolve reference to 'user-content-as' for \ref command
/home/lcarlier/Projects/bugDoxygen/README.md:6: warning: unable to resolve reference to 'user-content-as' for \ref command

**To Reproduce**
Untar the attached tar ball and execute doxygen using the contained doxyfile.
[bugDoxygen.tar.gz](https://github.com/doxygen/doxygen/files/8233783/bugDoxygen.tar.gz)

**Version**
This is a regression.
The patch introducing the problem is 3a4db857340aeff18dcd57291e036809a0273449
The parent version is working fine.

**Stack trace**
NA

**Additional context**
I was trying to find a fix but it isn't clear to me what could be the right solution. I thought of adding the section manually via SectionManager::instance().add but the name of the output HTML isn't known inside markdown.cpp so I'm not sure whether that would be the right thing to do.

Looking at the updated documentation, it seems that some inconsistencies have been introduced.

This is the patch I was trying but it isn't working
```
diff --git a/src/markdown.cpp b/src/markdown.cpp
index bacefc85..f8ddeeb5 100644
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -52,6 +52,7 @@
 #include "regex.h"
 #include "fileinfo.h"
 #include "utf8.h"
+#include "pagedef.h"
 
 enum class ExplicitPageResult
 {
@@ -1225,7 +1226,8 @@ int Markdown::processLink(const char *data,int,int size)
         m_out.addStr(" \"");
         m_out.addStr(substitute(content.simplifyWhiteSpace(),"\"","&quot;"));
         m_out.addStr("\"");
-
+        SectionManager::instance().add(&link.data()[1],
+                                       m_fileName,-1,title,SectionType::Anchor,0,"");
       }
       else
       {
```

albert-github:
[...]

PR #9204, opened 2022-03-13
issue #9200 Doxygen cannot resolve link to HTML anchor page

albert-github:
Convert HTML type of anchors to doxygen type of anchors as long as this is possible (`{LABELID}`).

(Tested against CGAL as well)


lcarlier:
This patch fixes the regression https://github.com/doxygen/doxygen/issues/9200

Thanks a lot.

PR #9213, opened 2022-03-17
Global enum incorporates values from an enum class

albert-github:
When having:
```
/// the class enum
enum class F1 {
  /*! first */
  FFAST = 1
};

/// the enum
enum F3
{
  /// third
  FFAST = 103
};
```

we see in the output (a.o.):
```
enum   F3 { FFAST = 103 , FFAST = 1 }
```
so in the global enum the `FFAST` from the class enum is incorporated as well (based on the name).
When having first the `F3` enum followed by the `F1` enum the results are OK.

Adding of an enum value to the global enum should only take place when we have an enum value that is not from an `enum class`.

(In the CGAL original the `enum`s were embedded in a namespace).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8288308/example.tar.gz)

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -7372,20 +7372,26 @@ static void addEnumValuesToEnums(const Entry *root)
                       //    qPrint(fmd->name()),qPrint(fmd->getOuterScope()->name()));
                       if (nd && !nd->isAnonymous())
                       {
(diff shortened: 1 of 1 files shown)


albert-github:
Looks like there are some, not doxygen related, problems with installing Qt.

Issue #9223, opened 2022-03-23
Feature request: take param/return docu in account from the header file when generatig docu for the source file (C language)


PVeps:
**Scenario description**
Some function including all parameters and return type is already documented in a header file xyz.h. The source code of the same function (file xyz.c) contains some additionally documentation (e.g., documentation of the implementation-specific details only). The warnings "_parameters of member ... are not (all) documented_" or/and "_return type of member ... is not documented_" are being emitted when generating docs for the file xyz.c.

**Expected behavior**
Variant 1: no such warning(s) emitted at all - unconditionally (preferred).
Variant 2: introduce a global option to suppress such warnings (e.g., CONSIDER_HEADER_DOCU = TRUE).
Variant 3: introduce some file-local doxy-directive activating considering of the already existing documentation parts.

**To Reproduce**

[param_already_documented_issue.zip](https://github.com/doxygen/doxygen/files/8333304/param_already_documented_issue.zip)

~~~~
/** \file param_already_documented_issue.h
 */
/** \brief Convert a \b char to \b int.
 * \param x char value to be converted.
 * \return Representation of \b x as \b int.
 */
int f(char x);

/** \file param_already_documented_issue.c
 */
/** \details Implemented as simple cast from char to int. */
int f(char x) { return (int)x; }
~~~~

The doxy project contains following options:
WARN_IF_UNDOCUMENTED   = YES
WARN_IF_INCOMPLETE_DOC = YES
WARN_NO_PARAMDOC       = YES

**Version**
doxygen version:
1.9.3 (c0b9eafbfb53286ce31e75e2b6c976ee4d345473)
Windows 10

**Additional context**
Note that the `\brief` part of the documentation containing in the header file _will be included_ into HTML-documentation of the .c-file. That means for me that some information about the doxy comments will be joined from both .h- and .c-files (how expected), but unfortunately not all such information (e.g., information about param/return docu).

albert-github:
I've just pushed a proposed patch, pull request #9225

albert-github:
Related issue #1390

Issue #1390, opened 2018-07-01
Usability issues caused by single detailed comment requirement (Origin: bugzilla #163033)


doxygen:
status NEW severity *major* in component *general* for *---*
Reported in version *1.5.3-SVN* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[doxytest.zip](http://ftp.stack.nl/pub/users/dimitri/bugzilla/attachments//bug163033_35488_doxytest.zip)_ (ID 35488)

On 2005-01-05 17:36:07 +0000, Chris Reed wrote:

> Doxygen supports only one detailed comment per object (function, struct, etc). This causes some 
> serious usability issues.
> 
> For instance, we (my workplace) use a detailed comment describing all functions in each of our header 
> files. But we would also like to be able to sprinkle \todo commands within the source, near the line(s) 
> that needs work. But this causes Doxygen to drop the entire previous detailed comment. Of course, we 
> can work around this by always putting the \todo commands in the header file. But that is clearly not 
> how that command is intended to work.
> 
> As another example, we would like to break function documentation into two pieces. First, public 
> documentation for external consumption by customers (in the header files). And second, internal 
> documentation for ourselves, in the source files, controlled by \if-\endif commands. 
> But this cannot be done because of this restriction. There are other special commands that also make 
> sense to separate from the main detailed comment, such as \note.
> 
> Although it may make perfect logical sense to require a single detailed comment, in practice it is a 
> rather user-hostile requirement. Coming from the point of view of a user rather than the author of 
> Doxygen, this restriction simply doesn't make sense. We view it as the single real issue with Doxygen. 
> Fixing this would greatly improve the flexibility of Doxygen, in ways that  would be hard to compare 
> with other new features.
> 
> The attached example files demonstrate how we would expect Doxygen to work. The .h file contains a 
> function that has a \brief command and some public information about the function. The .c file 
> contains more internal-type documentation of the function. Of course, when you generate the 
> documentation, only the comments in the source file are applied. We would expect everything in the 
> source file to be appended after that in the header files.
> 
> Thanks.

On 2005-01-05 17:37:31 +0000, Chris Reed wrote:

> Created attachment 35488
> doxytest.zip
> 
> Contains a .c, .h, and Doxyfile showing problems caused by the single detailed
> comment requirement.

On 2007-10-15 19:40:41 +0000, Dimitri van Heesch wrote:

[...]

PR #9225, opened 2022-03-24
issue #9223 take param/return docu in account from the header file when generating docu for the source file

albert-github:
- correct determination when to overwrite which part of the documentation, e.g. replace
  ```
        if (!mdec->briefDescription().isEmpty())
  ```
  with
  ```
        if (mdef->briefDescription().isEmpty() && !mdec->briefDescription().isEmpty())
  ```
  so only empty parts are overwritten.
- merging documentation of declaration and definition
- fixing problem with warnings like:
  ```
  <f_brief_c>:1: warning: parameters of member f_brief_c are not documented
  ```
  instead if the better version with filename and line number.
  In case detailed documentation is missing set the line number etc. of the detailed documentation to the line number of the corresponding brief documentation.
  (when brief documentation is with declaration and definition it is not necessary to do this as on of the brief documentations will become detailed documentation.

Examples: [example1.tar.gz](https://github.com/doxygen/doxygen/files/8343934/example1.tar.gz) and [example2.tar.gz](https://github.com/doxygen/doxygen/files/8343935/example2.tar.gz)

### DIFF
--- a/doc_internal/Doxyfile.in
+++ b/doc_internal/Doxyfile.in
@@ -51,7 +51,7 @@ INLINE_GROUPED_CLASSES = NO
 INLINE_SIMPLE_STRUCTS  = NO
 TYPEDEF_HIDES_STRUCT   = NO
 LOOKUP_CACHE_SIZE      = 0
-NUM_PROC_THREADS       = 1
+NUM_PROC_THREADS       = 0
 #---------------------------------------------------------------------------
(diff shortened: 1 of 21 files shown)


albert-github:
Added some corrections based on #1390 
An extended example has been made: [example.tar.gz](https://github.com/doxygen/doxygen/files/10148497/example.tar.gz)

Issue #9221, opened 2022-03-22
Tag File - option to Strip path


ferdymercury:
**Describe the bug**
Our doxygen-generated tag files contain internal paths specific to the build machine that are not relevant to the public.
https://root.cern/doc/master/ROOT.tag.gz

**Expected behavior**
It would be nice if an option to strip this path from the output tag file is provided.
(I tried adding /opt/root_src to  STRIP_FROM_PATH but it does not seem to work).

**Screenshots**
```
<?xml version='1.0' encoding='UTF-8' standalone='yes' ?>
<tagfile doxygen_version="1.9.4" doxygen_gitid="70d1c6baf5107dcfaae6fee02ea6b3ef93cea61c*">
  <compound kind="file">
    <name>fillpatterns.C</name>
    <path>/opt/root_src/core/base/doc/macros/</path>
    <filename>fillpatterns_8C.html</filename>
```

**To Reproduce**
[tagpath.zip](https://github.com/doxygen/doxygen/files/8326127/tagpath.zip)
cd into "streamer" and call doxygen there.
Open resulting tag file and the full path to docu will be there, despite STRIP_FROM_PATH being `../`.

**Version**
`1.9.4 (70d1c6baf5107dcfaae6fee02ea6b3ef93cea61c*)`

**Additional context**
https://github.com/root-project/root/issues/10059

albert-github:
I can imagine that it is not desired that an internal path is shown.

I see that the path is set for files and directories (see `dirdef.cpp` and `filedef.cpp`) and that it is read and used in `tagreader.cpp`.
In the tag reader it is used to do a match for include files and to create a filename and a tagid. Most likely usage is to distinguish between files with the same name but in differt directories. I'm not sure whether the  or not the path is used further on.

Maybe a solution would be that in the `dirdef.cpp` and `filedef.cpp` the path is converted to a md5sum.

@doxygen do you see or know whether or not this approach would have have any side effects?

albert-github:
I've just pushed a proposed patch, pull request #9226, though with the pull request I also made a small compatibility remark so it is not 100%clear whether this will be the final solution or that we might need to hide the replacement behind a setting.
@ferdymercury maybe you could give the patch a test with your package.

ferdymercury:
Thanks, I run it and I see the paths being converted to md sums. :)
I do not use the Tag file right now externally, thus I cannot check about the potentially broken links.

albert-github:
[...]

PR #9226, opened 2022-03-25
issue #9221 Tag File - option to Strip path

albert-github:
Replace the path with a md5sum

### DIFF
--- a/src/dirdef.cpp
+++ b/src/dirdef.cpp
@@ -472,7 +472,13 @@ void DirDefImpl::writeTagFile(TextStream &tagFile)
 {
   tagFile << "  <compound kind=\"dir\">\n";
   tagFile << "    <name>" << convertToXML(displayName()) << "</name>\n";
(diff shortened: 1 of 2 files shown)


albert-github:
The proposed patch solves for simple problems the problem, though in case a reference in a source code part is made to an include file referenced through a, partial, path and the included file was documented the link that was present in the past might disappear.
We see this effect in CGAL where we have in the fresult file doc_output/Algebraic_foundations/Algebraic_foundations_2algebraic_structure_dispatch_8cpp-example.html a reference to the file `int.h` from the `Number_types` package.
Originally the reference was:
```
<div class="line"><span class="preprocessor">#include &lt;<a class="codeRef" href="../Number_types/int_8h.html">CGAL/int.h</a>&gt;</span></div>
```
and now it is
```
<div class="line"><span class="preprocessor">#include &lt;CGAL/int.h&gt;</span></div>
```

albert-github:
Superseded by #9227 that does not have the indicated side offects

Issue #9135, opened 2022-02-20
Qhelpgenerator debug mode


ferdymercury:
**Describe the bug**
Not a bug.

**Expected behavior**
It would be nice if, when `-d qhp` is enabled, qhelpgenerator is run first (twice) with the flags `-v -c`, which print the version and check that all files from the file list are found. (I think they cannot be used simultaneously, nor with -o).

**To Reproduce**
See the reproducer at https://github.com/doxygen/doxygen/issues/9094

**Screenshots**
```
Usage:

qhelpgenerator <help-project-file> [options]

  -o <compressed-file>   Generates a Qt compressed help
                         file called <compressed-file>.
                         If this option is not specified
                         a default name will be used.
  -c                     Checks whether all links in HTML files
                         point to files in this help project.
  -v                     Displays the version of 
                         qhelpgenerator.
```

**Version**
1.9.4 (15dae504a167d838873053cc96417d8beae78df7*)

**Additional context**
Included HTML path problems found in https://github.com/doxygen/doxygen/issues/9094

albert-github:
As is the idea is sympathetic (and not difficult to implement) though I tried to run the `-c` option on the mentioned case and it takes ages (runs already over 20 minutes at my place).

The documentation states: 
> Checks whether all links in HTML files point to files in this help project.

I think one can either interpret this as:
- checks whether all files mentioned in the `index.qhp` exist 

or
- checks whether all files mentioned in the `index.qhp` exist **and** all links in the mentioned files exist.

It looks to me that the later is done here. 

So I don't think it is a good idea to run the `-c` even under debug conditions.

Regarding the `-v` I'm not sure this is something doxygen should output either, though with the `-d qhp` this would be easy again.

- how fast did you see the results of the `-c`?
- @doxygen what to do?

Did you miss some files in the example besides the files in  the "user" : of the layout files (i.e. "group__Tutorials.html" and "release-notes.html")

albert-github:
I tried:
- 6.1.2 and this has the behavior as indicated
  -  did also do the test without the `index.html` and in that case it quickly shows  that the `index.html` is missing and after this it looks like it is in an end less loop or doing some very long lasting operations
- 6.2.2 ant this also takes ages
- 5.15.2 runs fast though
  -  as you mentioned doesn't warn at all for the groupTutorials and release-notes, so looks indeed like it is not looking at the `<toc>` part. 

So I'm still think it is not wise to enable the `-c` flag in debug mode.

[...]

PR #9228, opened 2022-03-26
issue #9135 Qhelpgenerator debug mode

albert-github:
Adding in the qhp debug mode the output of the `-v` and `-c` output.
The `-c` version didn't work for the Qt version up to version 6.2.5 resulting in some extra test (see: https://bugreports.qt.io/browse/QTBUG-101070)

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -12555,7 +12555,74 @@ void generateOutput()
     std::string oldDir = Dir::currentDirPath();
     Dir::setCurrent(Config_getString(HTML_OUTPUT).str());
     Portable::sysTimerStart();
-    if (Portable::system(Config_getString(QHG_LOCATION).data(), args.data(), FALSE))
+
+    QCString qhgLocation=Config_getString(QHG_LOCATION);
+    if (Debug::isFlagSet(Debug::Qhp))
+    {
+      QCString cmd=(qhgLocation+ " -v 2>&1").data();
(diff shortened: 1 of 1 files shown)


src/doxygen.cpp
```diff
@@ -12555,7 +12555,74 @@ void generateOutput()
...
+          std::string sub = match.str();
+          reg::Iterator it1(sub,marker1);
+          reg::Iterator end1;
+          for ( ; it1!=end1 ; ++it1)
+          {
+            const auto &match1 = *it1;
+            ver = ver * 100 + QCString(match1.str()).toInt();
+          }
+        }
+        if (ver && ( ver < 60000 || ver >= 62500))
```
doxygen, line 12600: I would expect version 6.2.5 to be encoded as 60205 given to loop above and do all versions <6.0.0 work?.

albert-github: It is not sure whether or not all versions <6.0.0 do work and which "6" version was the first version to fail. Due to the lack of information available available about this, so this is a best guess. 
Regarding versions <6.0.0 it is also not sure when the `-c` is introduced, but seen the fact that all versions <5.15 are already a few year not maintained, I tested it with the 5.4.1 version (test `-c` is OK), 4.7.4 also handles the `-c`, 4.6.1. does not yet handle `-c`.
I think the test is safe enough (we could refine it to 4.7.4 though).

Issue #9229, opened 2022-03-26
Crash with Segmentation fault in ClassDefImpl::mergeMembers


wfjm:
**Describe the bug**
The creation of the `VHDL` documentation of the [w11 project](https://github.com/wfjm/w11) crashes for 1.9.2 and 1.9.3 with
```
.....
Creating members for template instances...
Building page list...
Search for main page...
Computing page relations...
Determining the scope of groups...
Sorting lists...
Determining which enums are documented
Computing member relations...
Building full member lists recursively...
Segmentation fault (core dumped)
```
More details are described under _Stack Trace_.

**Expected behavior**
Works fine with 1.9.1 and previous versions

**To Reproduce**
The `VHDL` source tree is in [w11 project](https://github.com/wfjm/w11).  
The `Doxyfile` is attached.  
A full reproducer script can be provided upon request. 

**Version**
Using Ubuntu 2020.04 LTS. Both 1.9.2 and 1.9.3 crash. 1.9.1 works.

**Stack trace**
I've build 1.9.2 with debug
```
git checkout tags/Release_1_9_2
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Debug -G "Unix Makefiles" ..
```
and got the backtrace
```
#0  0x0000557d3ad3d390 in ClassDefImpl::mergeMembers (this=0x557d521fd100)
      at /mnt/data/opt/github/doxygen/src/classdef.cpp:3441
#1  0x0000557d3a7cd61c in buildCompleteMemberLists ()
      at /mnt/data/opt/github/doxygen/src/doxygen.cpp:7772
#2  0x0000557d3a7e4b8c in parseInput ()
      at /mnt/data/opt/github/doxygen/src/doxygen.cpp:11739
#3  0x0000557d3a79c5b1 in main (argc=2, argv=0x7ffd7a959578)
      at /mnt/data/opt/github/doxygen/src/main.cpp:37
```
The code around `classdef.cpp:3441` is
```
    3439        const ClassDef *dstCd = dstMd->getClassDef();
    3440        /...
    3441        if (srcCd==dstCd || dstCd->isBaseClass(srcCd,TRUE))
```
An `info locals` gives
```
dstCd = 0x0
srcCd = 0x557d50f138f8
```
The problems is that
- `srcCd` is not equal `dstCd`
- that's why `dstCd->isBaseClass(srcCd,TRUE)` is executed
- dstCd is `nullptr` --> **Segmentation fault**

[...]

PR #9233, opened 2022-03-28
issue #9229 Crash with Segmentation fault in ClassDefImpl::mergeMembers

albert-github:
Incorrect translation of regular expression:
```
  static const std::regex reg("[[:space:]:|]",std::regex::optimize);
```
to
```
  static const reg::Ex reg(R"([\s|])");
```
this should have been:
```
  static const reg::Ex reg(R"([\s:|])");
```


wfjm:
I tried the feature branch against my vhdl and c++ code base, and it works as 1.9,1 did.  
Thanks a lot, I'm looking forward to the 1.9.4 release with this bugfix integrated.

Issue #9234, opened 2022-03-28
NUM_PROC_THREADS not being honored when processing input.


ferdymercury:
**Describe the bug**
In an 8-thread computer, I enable NUM_PROC_THREADS to 4 when building my documentation. However, in some steps, doxygen uses 8 even if I told him to only use 4. In other steps, it correctly respects the maximum.

**Expected behavior**
It always respects the maximum.

**Screenshots**
doxygen's output is very clear about this:
```
...
Processing input using 8 threads.
...
Generating code files using 4 threads.
...
```

**To Reproduce**
Just turn in any project NUM_PROC_THREADS to half of your max number of threads, and check the output.

**Version**
`1.9.4 (91095fb62f0bc2c624b0e8e49d9316f51be76559*)`

ferdymercury:
I proposed a patch: https://github.com/doxygen/doxygen/pull/9235

albert-github:
I've just pushed a proposed patch, pull request #9237, which in my opinion supersedes the proposed pull request #9235

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.4.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9235, opened 2022-03-28
issue #9234 NUM_PROC_THREADS not being honored when processing input

ferdymercury:
Read num proc threads also for processing


albert-github:
I think correct as in the part above where the CLANG parsing is used the same logic is employed.

albert-github:
I've just pushed a proposed patch, pull request #9237 that superseded, in my opinion, this proposed pull request.

PR #9240, opened 2022-03-31
Fixing installation problem regarding Qt on MacOs

albert-github:
Not sure whether / when the action will be updated, the given solution is in my opinion a bit of a workaround) See also:  https://github.com/jurplel/install-qt-action/issues/142 / https://github.com/jurplel/install-qt-action/issues/142#issuecomment-1083151012

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -233,8 +233,15 @@ jobs:
         refreshenv
       if: matrix.config.os == 'windows-latest'
     
-    - name: Install Qt
+    - name: Install Qt (MacOs)
       uses: jurplel/install-qt-action@v2
+      with:
+          setup-python: 'false'
+      if: matrix.config.os == 'macos-latest'
+    
+    - name: Install Qt (Linux / Windows)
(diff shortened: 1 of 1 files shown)


albert-github:
It looks like this "MacOS" problem has been solved somewhere upstream as in the latest GitHub action (https://github.com/doxygen/doxygen/actions/runs/2076679592 for issue #9241 Undocumented external namespaces from .tags files are shown in output Namespace index CMake Build for Doxygen 1674) was OK, though the one before (https://github.com/doxygen/doxygen/actions/runs/2076226943 for issue #9243 GHC_NO_DIRENT_D_TYPE should be defined on Solaris CMake Build for Doxygen 1673) still had problems.

doxygen:
Does it mean this pull request can be closed or does it still make sense to integrate it?

albert-github:
I left this proposed pull request explicitly open to see whether or not the problem would pop-up again in the next few days, but in the mean time some new problems appeared but theses seem to have been solved now as well.

So yes I think that this proposed pull request can be closed.

Issue #9243, opened 2022-04-01
GHC_NO_DIRENT_D_TYPE should be defined on Solaris


lukyhruska96:
**Describe the bug**
Doxygen fails to compile (`filesystem/filesystem.hpp`) on Solaris because of missing `d_type` in `dirent` structure.

**Version**
1.9.3

albert-github:
A proposed patch is available as #9243

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.4.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9244, opened 2022-04-01
issue #9243 GHC_NO_DIRENT_D_TYPE should be defined on Solaris

lukyhruska96:


filesystem/filesystem.hpp
```diff
@@ -65,6 +65,8 @@
 #elif defined(_WIN32)
 #define GHC_OS_WINDOWS
 #define GHC_OS_WIN32
+#elif defined(__sun) && defined(__SVR4)
+#define GHC_NO_DIRENT_D_TYPE
```
albert-github, line 69: I would also add:
```
#define GHD_OS_SUN
```
here.

albert-github (changes requested):
Thanks for the patch.

Besides adding the `GHC_OS_SUN` this issue should also be reported to the upstream repository at: https://github.com/gulrak/filesystem

filesystem/filesystem.hpp
albert-github: The name in the fix is better than my suggestion.

albert-github (approved):

Issue #9241, opened 2022-03-31
Undocumented external namespaces from .tags files are shown in output Namespace index


oblivioncth:
No one seems to have any comment about this on stackoverflow so I'm filling a report since I believe this to be a bug; however, this is my first time using Doxygen so I'm not sure.

**Describe the bug**
The output documentation for my project shows namespaces from external tag files (Qt documentation) within the Namespace view as if they were my own namespaces. They appear without a hyperlink and no description, and there is definitely no documentation for them within my project, nor are there members added to them within my code.

**Expected behavior**
Namespaces within external documentation that are included via .tags files don't appear in my output documentation unless I explicitly add documentation for those namespaces via the \namespace command.

Perhaps what I'm experiencing is intended behavior, but I'm not sure why that would be.

**Screenshots**
![OGecB](https://user-images.githubusercontent.com/24661585/161124361-24c3a6ac-7ebb-490b-a6c1-b6b35e461ccc.png)

The only namespace here that is actually in my project is 'Qx'.

**To Reproduce**
[Doxy_External_Namespace_Issue_MRE.zip](https://github.com/doxygen/doxygen/files/8392139/Doxy_External_Namespace_Issue_MRE.zip)

The tags in question are from the docs of Qt 6.2.3 and are provided with the above example.

**Version**
Latest stable release from site:
1.9.3 (c0b9eafbfb53286ce31e75e2b6c976ee4d345473) 
Windows x64

**Additional context**
It's possible that the Qt tags are ill-formed since they are generated by QDoc (i.e. could be a bug with that), though personally (without knowing all of the ins-and-outs of Doxygen) I would also somewhat expect external tags to not be capable of producing documentation like this under any circumstances so I'd be a little surprised if that was the case.

Possibly a regression of #5258, but I don't believe these namespaces are nested under anything not shown in the above output.

albert-github:
I've just pushed a proposed patch, pull request #9245

oblivioncth:
Thanks for making the change so quickly.

Once it's merged I'll download the related artifact and confirm the fix.

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.4.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
[...]

PR #9245, opened 2022-04-01
issue #9241 Undocumented external namespaces from .tags files are shown in output Namespace index

albert-github:
Namespace references from tag files should not appear in the namespace list.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -1708,7 +1708,10 @@ static void writeNamespaceTree(const NamespaceLinkedRefMap &nsLinkedMap,FTVHelp 
 {
   for (const auto &nd : nsLinkedMap)
   {
-    writeNamespaceTreeElement(nd,ftv,rootOnly,addToIndex);
+    if (nd->isLinkableInProject())
+    {
+      writeNamespaceTreeElement(nd,ftv,rootOnly,addToIndex);
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/index.cpp
```diff
@@ -1708,7 +1708,10 @@ static void writeNamespaceTree(const NamespaceLinkedRefMap &nsLinkedMap,FTVHelp
 {
   for (const auto &nd : nsLinkedMap)
   {
-    writeNamespaceTreeElement(nd,ftv,rootOnly,addToIndex);
+    if (nd->isLinkableInProject())
```
doxygen, line 1711: This will also hide documented namespaces imported via tag files. Isn't the real problem that undocumented namespaces appeared in the tag file?

albert-github: I don't think that the problem is that undocumented namespaces appear in the tag file as undocumented classes appear also in the tag file and the documented and undocumented classes don't appear in the tree of the project that uses this tag file.
I created a small example to illustrate this: [example.tar.gz](https://github.com/doxygen/doxygen/files/8495000/example.tar.gz)

And in the output we see:

image

So no classes from project 1 but the namespaces from project 1 are shown.

I think logic of the proposed implementation is correct.

oblivioncth: > This will also hide documented namespaces imported via tag files. Isn't the real problem that undocumented namespaces appeared in the tag file?

I would like to clarify that in the original issue when I stated "undocumented external namespaces..." I meant external namespaces that were **not documented in my project**. The namespaces in my example picture are indeed documented in the original QDoc project that generated those tags.

I reported this as a bug as I was under the impression that the purpose of importing tags is so that a user's documentation can easily reference/link to external documentation (i.e. its dependencies), and not much more.

In my opinion, the primary listing pages (i.e. Namespaces, Classes, Files, etc.) should only list members that are part of the project in question, and not those that are from external projects, unless for whatever reason you explicitly added documentation for them in your own. They way the Namespaces page appears now gives the false impression that my library defines those extra namespaces, when in realty it doesn't define or add to them in any way.

As Albert pointed out, the Classes listing already behaves the "ideal" way (as does Files). This, along with the fact that the external namespaces that do appear have no brief descriptions, no pages associated with them, and aren't even linked to the external documentation that they are documented in, makes this seem like much more of a bug than intentional.

At least, that's how it seems to me.

Issue #9185, opened 2022-03-05
tag file exntry without filename


ChristianTackeGSI:
**Describe the bug**

external tag files with entries that do not have a `<filename>` element give links to non-existant files.

**Expected behavior**

If there is no filename in the tag file, maybe set the generated entry to `artificial = true`, so that it is visible in diagrams/etc, but not clickable? Like the `BUILTIN_STL_SUPPORT` things?

**To Reproduce**

This is a very minimal example. Not being really useful, but shows the problem.

`foo.h`:
```c++
#include <TObject.h>

class MyClass : TObject {};
```

`ext.tag`:
```xml
<?xml version='1.0' encoding='UTF-8' standalone='yes' ?>
<tagfile>
  <compound kind="class">
    <name>TObject</name>
  </compound>
</tagfile>
```

`Doxyfile`:
```
TAGFILES = ext.tag
```

This draws an inheritance / collaboration diagram for MyClass with TObject in it. And TObject is clickable, but sends you to a non-existant page.

**Version**
Debian-11, amd64, 1.9.1-1

albert-github:
There are 2 forms of `TAGFILES` references:
1. `TAGFILES = ext.tag=<reference to html documentation>`
    Here an explicit position of the relevant HTML documentation to be linked to is given.
2. `TAGFILES = ext.tag`
    This requires that the relevant HTML documentation to be linked to is in a directory that is a direct sub-directory of the place where the `ext.tag` file can be found.

See also the doxygen documentation https://www.doxygen.nl/manual/external.html, https://www.doxygen.nl/manual/config.html#cfg_generate_tagfile and https://www.doxygen.nl/manual/config.html#cfg_tagfiles

Personally I would always use the explicit version 1 for the references to the tag files

Some questions:
- how is the `ext.tag` generated?
- where can the documentation be found. In the resulting error message we see a reference to `.../html/classTObject.html` so it cannot find the relevant descriptions.

Small side note: the current doxygen version is 1.9.3 (most likely won't influence your problems).


As far as I can see the problem is usage.

ferdymercury:
If I guess correctly, the external tag file refers to the one from ROOT, that can be downloaded from https://root.cern/reference/

It was generated using this Doxyfile: https://github.com/root-project/root/blob/master/documentation/doxygen/Doxyfile#L2311

albert-github:
That might indeed be the case, but in the mentioned tag file I do see:
```
....
<compound kind="class">
[...]

PR #9247, opened 2022-04-03
issue #9185 tag file entry without filename

albert-github:
In case the the `filename` isn't mentioned in the tag file show the item but don't link to it (but still mention it as external at the appropriate places).
For this the original `filename` has te be "recorded" as later on the filename to be used might be overwritten.

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -777,7 +777,7 @@ ClassDefImpl::ClassDefImpl(
     bool isSymbol,bool isJavaEnum)
  : DefinitionMixin(defFileName,defLine,defColumn,removeRedundantWhiteSpace(nm),0,0,isSymbol)
 {
-  setReference(lref);
+  setReference(lref,fName);
   m_impl = new ClassDefImpl::IMPL;
   m_impl->compType = ct;
   m_impl->isJavaEnum = isJavaEnum;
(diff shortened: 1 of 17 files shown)


src/classdef.cpp
```diff
@@ -3324,7 +3324,7 @@ bool ClassDefImpl::isLinkable() const
   }
   else
   {
-    return isReference() || isLinkableInProject();
+    return (isReference() && !isInValidTagReference()) || isLinkableInProject();
```
doxygen, line 3327: try to avoid double negation, i.e. `isValidTagReference()` is easier to read and understand than `!isInvalidTagReference()` , is there any case where `isReference()` does not also imply `isValidTagReference()`? If not then better make it part of `isReference()`

albert-github: A quick search gave that there were 8 places where the new construct is used and 131 more constructs where only `isReference` is used.

Regarding the double negation: Yes it is most of the times hard to understand what is going on in that case though here I found it much clearer and easier to use the double negation.

doxygen: The followup question I have is why is the valid tag check needed in the 8 cases, but not in the other 131 cases? What is the criteria to include/not include the extra check? I ask from a maintenance point of view, i.e. if I add a new `isReference()` somewhere when do I also need the valid tag check?

albert-github: This is always hard to tell, the basic idea is to prevent to create a link that points to nowhere.
The `!isInValidReference` has to be added where a request is made to check / create a link, as this cannot be done due to the missing filename.

PR #9248, opened 2022-04-03
bug 334859 HTMLHELP: Underscore in chm file name prevents linking between chm files

albert-github:
The problem here looks like a bit of a quirk in the html help compiler. Created a note in the documentation.


albert-github:
@doxygen can we mention this just in the documentation or should we try to fix this in the `htmlhelp.cpp` or `tagreader.cpp`?

PR #9258, opened 2022-04-06
bug 783768 QCH files: classes & their constructors get conflicting keyword handling

albert-github:
The problem in not only present in the Qt help (qhp / qch) but also in the HTML Help (chm).
All items with the same name are put onto one heap. For the index this is corrected by adding the function signatures, analogous to the javascript search functionality of doxygen.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8428593/example.tar.gz)


ferdymercury:
Thanks for finding out! (I guess this supersedes https://github.com/doxygen/doxygen/pull/573 from @kossebau )

Issue #6524, opened 2018-09-20
Markdown formats missing in doxygen outputs.


Slyware:
Hi.

The ouputs files (html, rtf, ...) don't interpret ==highlight== and ~~strikethrough~~ format tags nor html tags as <u>underline</u>. I know that they may not be in the official Markdown standard but they are largely used by a many Markdown editors or viewers _**like Github does!**_.
Unlike **bold** or *italic* format tags, the above tags are displayed as raw source text in the outputs.
Is there a way to add these tags in the Doxygen parser or to customise the preprocessor?
Thanks.
SJ

Doxygen version: 1.8.14

albert-github:
In the current master version of doxygen some features should already be available. To see what you want can you please attach a, small, self contained example (source+config file in a tar or zip) that allows us to reproduce the problem?

Slyware:
Hi.
Here is an example of MD source edited in Typora (with option "Highlitghing" checked).
Ok, I saw that some of these formats are Markdown extensions but they are commonly used and so usefull.
[SupportDoxygen.zip](https://github.com/doxygen/doxygen/files/2409878/SupportDoxygen.zip)
Typora rendering is here: ![image](https://user-images.githubusercontent.com/43435546/45940086-39006080-bfd7-11e8-8884-6f74168dee45.png)
Thanks a lot.
SJ

albert-github:
Code has been integrated in master on github.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.8.15.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9267, opened 2022-04-10
bug_707283 Add super- and subscript to markdown support.

albert-github:
Add:
- superscript by surrounding the superscripted text by `^` characters
- subscript by surrounding the subscripted text by `~` characters

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8458796/example.tar.gz)

### DIFF
--- a/doc/markdown.doc
+++ b/doc/markdown.doc
@@ -522,6 +522,26 @@ For example:
 
     ![Doxygen Logo](https://www.doxygen.org/images/doxygen.png){html: width=50%, latex: width=5cm}
 
+\subsection md_super_sub_script Superscript and subscript
+
+Standard Markdown does not support superscript or subscript. In doxygen there are a number of
+possibilities to create superscript and subscript:
+- through formulas (see section \ref formulas "formulas")
+- by means of the `<SUP>` (see  \ref htmltag_SUP "\<SUP\>") and `<SUB>` (see  \ref htmltag_SUB "\<SUB\>") 
+  HTML commands.
+
+To allow for output format simple superscript and subscript texts it is possible to use:
+- superscript by surrounding the superscripted text by `^` characters
+- subscript by surrounding the subscripted text by `~` characters
+
+Example:
+```
+  H~2~O is a liquid.  2^10^ is 1024.
+```
+will result in:
+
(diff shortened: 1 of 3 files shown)


doxygen:
@albert-github Since this is not really standard markdown, I'm a bit worried about false positives/confusion, e.g. someone writing `r^2 = x^2 + y^2` and getting `r<sup>2 = x</sup>2 + y^2`

albert-github:
Indeed the proposed implementation would give some problems here.

From the referenced https://pandoc.org/MANUAL.html#superscripts-and-subscripts

> Superscripts may be written by surrounding the superscripted text by ^ characters; subscripts may be written by surrounding the subscripted text by ~ characters. Thus, for example,
>
> H~2~O is a liquid.  2^10^ is 1024.
>
> The text between `^...^` or `~...~` may not contain spaces or newlines. If the superscripted or subscripted text contains spaces, these spaces must be escaped with backslashes. (This is to prevent accidental superscripting and subscripting through the ordinary use of ~ and ^, and also bad interactions with footnotes.) Thus, if you want the letter P with ‘a cat’ in subscripts, use `P~a\ cat~`, not `P~a cat~`.

Maybe we should also impose the restriction:
> The text between `^...^` or `~...~` may not contain spaces or newlines.

doxygen:
Another problem seems that for Github `~strikethrough~` appears as ~strikethrough~, so there is also different interpretations for the same notation among Markdown dialects.

albert-github:
@doxygen It is correct that doxygen markdown handles strikethrough  differently from GitHub markdown, this is documented in the doxygen manual:

> 5.1.8 Strikethrough
> To strikethrough a text fragment you start and end the fragment with two tildes.
> Examples:
> `~~double tilde~~`
> See section Emphasis and strikethrough limits for more info how doxygen handles emphasis / strikethrough spans
slightly different than standard Markdown / GitHub Flavored Markdown.

This implementation is done based on issue #6524 / #6535

The underlying problem is of course that the strikethrough is not defined in the original markdown definition (https://daringfireball.net/projects/markdown/syntax) nor in the common marks specification (https://spec.commonmark.org/0.30/).

Issue #9269, opened 2022-04-11
Not able to compile with MSYS2/mingw (file too big)


MKHen:
**Describe the bug**
When compiling under Windows MSYS2 gcc/x86_64-w64-mingw32 I am getting the following error:
Assembler messages:
Fatal error: CMakeFiles/doxymain.dir/docparser.cpp.obj: file too big
...
Assembler messages:
Fatal error: CMakeFiles/doxymain.dir/context.cpp.obj: file too big

**Version**
I used the files from the master.

**Additional context**
When compiling with -Os all works well.
I've used set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj") and checked it in the outcomming makefiles.
-- The C compiler identification is GNU 11.2.0
-- The CXX compiler identification is GNU 11.2.0

albert-github:
Analogous to the Cygwin implementation in `CMakeLists.txt` you will probably need to add heer:
```
   set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Og")

   if (CMAKE_BUILD_TYPE STREQUAL  "")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O1")
   endif()
```

As you can see we had similar problems on Cygwin, but due to missing test environment for `MSYS2/mingw` we didn't yet add it.

Please report back when this solves the issue for you.

MKHen:
I was able to compile it with the CMAKE_CXX_FLAGS set to -O1 or even Os.

Sadly I cannot tell you the sha of the files. Just downloaded an renamed them.

As cmake command I just used the one from the homepage:
mkdir build
cd build
cmake -G "Unix Makefiles" ..
make (sometimes make -j12)

CMAKE_SYSTEM_NAME is Windows.

I've just downloaded and tried with 9258d25. Behaviour is the same as described.

**Additional:**
Ever thought of inserting
find_program(CCACHE_FOUND ccache)
if(CCACHE_FOUND)
    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
endif(CCACHE_FOUND)

albert-github:
An additional question (due to lack of knowledge of the MSYS2/MinGW possibilities):
- did you compile from a Windows command  window or a MSYS2/MinGW shell?

albert-github:
Thank you for the very quick feedback,
[...]

PR #9271, opened 2022-04-12
issue #9269 Not able to compile with MSYS2/mingw (file too big)

albert-github:
Adding compilation flags for MSYS2/MinGW environment analogous to Cygwin environment.
(As it is a different environment an extra block)


CMakeLists.txt
```diff
@@ -96,6 +96,20 @@ if (CMAKE_SYSTEM_NAME MATCHES "CYGWIN")
    endif()
 endif()
 
+if (CMAKE_SYSTEM_NAME MATCHES "Windows")
+   if ((CMAKE_GENERATOR MATCHES "MinGW Makefiles") OR
+       (CMAKE_GENERATOR MATCHES "MSYS Makefiles") OR
+       (CMAKE_GENERATOR MATCHES "Unix Makefiles"))
+
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
```
MKHen, line 104: Not sure if needed. Makes no difference in my case.

MKHen (approved):

CMakeLists.txt
albert-github: It might be that the `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")` is not needed anymore in combination with the `-O1` or `-Og`, but in the past you needed it and for Cygwin we also needed it in the past so I left it for compatibility.
I don't think it hurts, but I will have to do some tests ...

albert-github: I just finished a number of tests and it looks like the `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")` is not necessary anymore.
The description in the assembler man page says:
```
       -mbig-obj
           On PE/COFF target this option forces the use of big object file format, which allows more than 32768 sections.
```
I don't think it hurts and might prevent some problems in the future, however an extra flag set is always a bit tricky.

@doxygen Should we leave the flag or should we remove it?

doxygen: @albert-github I think we should remove it if not needed.

albert-github: OK, I;ve removed it (but we should keep it in mind for the future).

Issue #9276, opened 2022-04-13
File-Line is omitted when suggesting possible candidates depending on noMatchCount


ferdymercury:
**Describe the bug**
When suggesting "possible candidates", the file and line is omitted if there are no "uniquely" matching ones.

**Expected behavior**
Possible candidates printed include file-line.

**Screenshots**

```
/opt/root_src/core/metacling/src/TClingClassInfo.cxx:507: warning: no matching class member found for 
  TClingMethodInfo TClingClassInfo::GetMethodWithArgs(const char *fname, const char *arglist, bool objectIsConst, Longptr_t *poffset, EFunctionMatchMode, EInheritanceMode) const
Possible candidates:
  'TClingMethodInfo TClingClassInfo::GetMethodWithArgs(const char *fname, const char *arglist, Longptr_t *poffset, ROOT::EFunctionMatchMode mode=ROOT::kConversionMatch, EInheritanceMode imode=kWithInheritance) const'
  'TClingMethodInfo TClingClassInfo::GetMethodWithArgs(const char *fname, const char *arglist, bool objectIsConst, Longptr_t *poffset, ROOT::EFunctionMatchMode mode=ROOT::kConversionMatch, EInheritanceMode imode=kWithInheritance) const'
```

vs

```
/opt/root_src/core/metacling/src/TClingClassInfo.cxx:752: warning: no uniquely matching class member found for 
  void TClingClassInfo::Init(const Decl *decl)
Possible candidates:
  'void TClingClassInfo::Init(const char *name)' /opt/root_src/core/metacling/src/TClingClassInfo.h:165
  'void TClingClassInfo::Init(const clang::Decl *)' /opt/root_src/core/metacling/src/TClingClassInfo.h:166
  'void TClingClassInfo::Init(int tagnum)' /opt/root_src/core/metacling/src/TClingClassInfo.h:167
  'void TClingClassInfo::Init(const clang::Type &)' /opt/root_src/core/metacling/src/TClingClassInfo.h:168
```

**To Reproduce**
[candidates.zip](https://github.com/doxygen/doxygen/files/8484112/candidates.zip)

**Version**
1.9.4 (e449f7367f5b7f6efadba28747dba591e1120bc5*)

**Additional context**
https://github.com/doxygen/doxygen/blob/master/src/doxygen.cpp#L6039
https://github.com/doxygen/doxygen/issues/9143#issuecomment-1098031112

[...]

PR #9277, opened 2022-04-14
issue #9276 File-Line is omitted when suggesting possible candidates depending on noMatchCount

albert-github:
There is no reason why to make the File-Line depending on the `noMatchCount`.


ferdymercury:
Thanks for this! Works great now.

Issue #9266, opened 2022-04-09
doxygen runtime went from 10minutes to >14hours (lots of `using namespace`)


martinus:
**Describe the bug**
Newest doxygen version is impracticably slow when there are lots of `using namespace` in the code. I stopped generating after 14 hours. Older versions take 10 minutes for the same codebase.

**Expected behavior**
Don't take forever

**To Reproduce**
Unfortunately I cannot share our code. It might be reproducible with several files and lots of namespaces, I can try to do that if you think that helps

**Version**
I checked out the latest version, and did a `git bisect` to find out when the slowdown was introduced. It first happened in this commit: 8a93f4e00b6922526935909312dabbd9a566d5cc. Before this commit a stripped down codebase takes 0.5 seconds to generate, after that commit I stopped it after several minutes. I did all my tests in Linux, with clang++

**Stack trace**
I used [hotspot](https://github.com/KDAB/hotspot) to figure out where doxygen was hanging (for this I built `master`), and got this flamegraph:
![image](https://user-images.githubusercontent.com/14386/162568380-fe6c0ad5-3be1-4572-bed3-5fffb5d93956.png)

**Additional context**
Looking at the changes in 8a93f4e00b6922526935909312dabbd9a566d5cc I suspect the issue is now too often a fresh `AccessStack` or `VisitedNamespaces` or `StringUnorderedSet` is used when previously a non-empty container was used

albert-github:
You pointed to a probable cause of the problem (thanks), though you also wrote:
> Newest doxygen version is impracticably slow when there are lots of using namespace in the code. I stopped generating after 14 hours. Older versions take 10 minutes for the same codebase.

- Which is the "Newest doxygen version" version you tried (output of `doxygen -v`)
- Which is the most recent "Older versions" you tried?

especially the "Newest doxygen version" is of important.

The source code you used:
- which language is it written in?
- is it public domain or propriety?

martinus:
> Which is the "Newest doxygen version" version you tried (output of doxygen -v)

I've built doxygen myself, this is the output of `doxygen -v` of the newest version that I've tried:

1.9.4 (de4a15d48dff81693c60043af52974cec7926c79)

> Which is the most recent "Older versions" you tried?

The most recent old version (last good version) I tried:

1.9.0 (02cdddb491739a18413025ddd587879bd8f5305f)

Then, first bad version is this:

1.9.0 (8a93f4e00b6922526935909312dabbd9a566d5cc)

[...]

PR #9283, opened 2022-04-20
Code generation optimizations

martinus:
After #9266 I did some profiling of doxygen generation of our C++ codebase. Based on that I have done a few optimizations that bring a 9x speedup, from ~49 minutes down to just over 5 minutes.

The most important change was the first commit, which using `const& for LinkedRefMap` instead of copies.

Note that the commit "Used hash data structures for SymbolMap" seems to cause the regression test `027_extends.c` to fail, not sure if this is an issue?

### DIFF
--- a/src/dotrunner.cpp
+++ b/src/dotrunner.cpp
@@ -160,7 +160,7 @@ void DotRunner::addJob(const QCString &format, const QCString &output,
     return;
   }
   auto args = QCString("-T") + format + " -o \"" + output + "\"";
-  m_jobs.emplace_back(format.str(), output, args, srcFile, srcLine);
+  m_jobs.emplace_back(format, output, args, srcFile, srcLine);
 }
 
 QCString getBaseNameOfOutput(const QCString &output)
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -7721,7 +7721,7 @@ static void addToIndices()
   {
     if (pd->isLinkableInProject())
     {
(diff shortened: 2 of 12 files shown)


albert-github:
The:
> Note that the commit "Used hash data structures for SymbolMap" seems to cause the regression test 027_extends.c to fail, not sure if this is an issue?

is an issue, either the test results so far were incorrect (means update of test result required) or are now incorrect (code has to be corrected).
To visually check the test results you can run `make tests TEST_FLAGS="--keep --xml --xhtml --id 27"` and look in the `testing/test_output_027` subdirectory of your build environment.

I did the "approve and run" on the GitHub Actions workflow to see what kind of errors are given by the `027_extends.c` test, without looking into your code.

martinus:
Thanks @doxygen! Yes, I used [hotspot](https://github.com/KDAB/hotspot) which shows a nice flamegraph of the hotspots. It's awesome, I use it a lot to find bottlenecks. I saw lots of copies happen to data that's only read, so I tried to use `const&` when possible and it seems to have worked very well :slightly_smiling_face: 

I'm building doxygen with RelWithDebInfo like so:

```sh
CC="ccache clang" CXX="ccache clang++" CMAKE_BUILD_TYPE=RelWithDebInfo cmake -G Ninja ../src
```

Then hotspot is able to show nice call stacks, and lots of copies of `std::unordered_map` happen which are really costly.

I've rebased and removed the problematic change, now the test work for me

PR #9290, opened 2022-04-24
Support visible attribute for all elements in the layout file

albert-github:
Currently not all elements have the possibility to set the visible attribute in the layout file, so the only way was to remove the element from the layout file which also could lead to conflicts with the default loaded file.


doc/customize.doc
```diff
@@ -380,14 +380,20 @@ its name prefixed with a dollar sign, e.g.
 This was mainly added for backward compatibility.
 Note that the \c visible attribute is just a hint for doxygen.
 If no relevant information is available for a certain piece it is
-omitted even if it is set to \c yes (i.e. no empty sections are generated).
+omitted even if it is set to \c yes (i.e. no empty sections are generated).<br>
+Note also that not all elements have a \c visible attribute shown in the layout file.
+This attribute can be used to customize the output anyway but should be used used with
+caution as unexpected results (e.g.incorrect links) might appear.
```
doxygen, line 386: If I read this as a user, then what I am I supposed to do with this information? I would say that we either offer a feature that always works (and if it does not despite our best efforts it can be regarded as a bug that needs to be fixed) or we don't offer it. If we really need to add this disclaimer, I prefer not to provide it.

albert-github: I don't like disclaimers either and I agree that in case it is not working we have to see this as a bug.

I added the disclaimer as in the original version (see the 1.9.3 manual) there was already the disclaimer:
> Warning
>      at the moment you should not remove elements from the layout file as a way to hide information. Doing so can
cause broken links in the generated output!

(and as removing the line has the same effect as `visible="no"` I added the disclaimer.

I'll remove the disclaimer and reformulate about the `visible=` in case it is not shown in the layout file (or should we add everywhere in the layout file the `visible="yes"` to make it more explicit?).

doxygen: I would keep it implicit for the cases where we do not expect `visible="no"` is likely to be used, so probably for the cases where it is now not present. Just not to clutter the file too much.

src/layout.cpp
```diff
@@ -765,6 +736,14 @@ static ElementCallbacks::StartCallback startCb(void (LayoutParser::*fn)(const XM
...
 }
 
+static ElementCallbacks::StartCallback startCb(void (LayoutParser::*fn)(const XMLHandlers::Attributes &,LayoutDocManager::LayoutPart,QCString,LayoutNavEntry::Kind),
+                                        LayoutDocManager::LayoutPart part,
+                                        QCString scope,
+                                        LayoutNavEntry::Kind nav
+                                       )
+{
+  return [fn,part,scope,nav](LayoutParser &parser,const XMLHandlers::Attributes &attr) { (parser.*fn)(attr,part,scope,nav); };
+}
```
ferdnyc, line 746: I cannot express how much I hate this source layout. I know, I know: It mirrors the other functions above it exactly, and code changes should always follow the prevailing style... I didn't say it was _wrong_, I said I hate it. I hate the way **all** of them are formatted.

Would a PR that reformatted them to something along these lines be considered for merging? Or is it not worth bothering, because the devs are fine with the current ultra-widescreen abomination? (After all, everyone's entitled to their opinion, however wrong it may be. :wink:)

```c++
static ElementCallbacks::StartCallback
startCb(
    void (LayoutParser::*fn)(
        const XMLHandlers::Attributes&,
        LayoutDocManager::LayoutPart,
        QCString,
        LayoutNavEntry::Kind),
    LayoutDocManager::LayoutPart part,
    QCString scope,
    LayoutNavEntry::Kind nav)
{
  return [fn,part,scope,nav](
      LayoutParser &parser,
      const XMLHandlers::Attributes &attr)
      {
          (parser.*fn)(attr,part,scope,nav);
      };
}
```

Or perhaps a compromise, like...

```c++
static ElementCallbacks::StartCallback
startCb(
  void (LayoutParser::*fn)(const XMLHandlers::Attributes&,LayoutDocManager::LayoutPart,QCString,LayoutNavEntry::Kind),
  LayoutDocManager::LayoutPart part,
  QCString scope,
  LayoutNavEntry::Kind nav)
{
  return [fn,part,scope,nav](LayoutParser &parser,const XMLHandlers::Attributes &attr) {
    (parser.*fn)(attr,part,scope,nav);
  };
}
```

albert-github: Consistency in the layout a a very good thing. The way how the layout of the code should look like is always a source for a good dispute between developers and all layouts have their pros and cons.
Also the length of a line is a problem, in the past it was restricted to 72 / 80 characters later 132 characters, there are only a few language standards that limit the line length.

Looking at:
```
static ElementCallbacks::StartCallback startCb(void (LayoutParser::*fn)(const XMLHandlers::Attributes &,LayoutDocManager::LayoutPart,QCString,LayoutNavEntry::Kind),
                                        LayoutDocManager::LayoutPart part,
                                        QCString scope,
                                        LayoutNavEntry::Kind nav
                                       )
```
I would even prefer:
```
static ElementCallbacks::StartCallback startCb(void (LayoutParser::*fn)(const XMLHandlers::Attributes &,LayoutDocManager::LayoutPart,QCString,LayoutNavEntry::Kind),
                                               LayoutDocManager::LayoutPart part,
                                               QCString scope,
                                               LayoutNavEntry::Kind nav
                                              )
```
or 
```
static ElementCallbacks::StartCallback startCb(void (LayoutParser::*fn)(const XMLHandlers::Attributes &,
                                                                        LayoutDocManager::LayoutPart,QCString,
                                                                        LayoutNavEntry::Kind
                                                                       ),
                                               LayoutDocManager::LayoutPart part,
                                               QCString scope,
                                               LayoutNavEntry::Kind nav
                                              )
```
or
```
static ElementCallbacks::StartCallback startCb(void (LayoutParser::*fn)(const XMLHandlers::Attributes &,
                                                                        LayoutDocManager::LayoutPart,QCString,
                                                                        LayoutNavEntry::Kind),
                                               LayoutDocManager::LayoutPart part,
                                               QCString scope,
                                               LayoutNavEntry::Kind nav)
```
and to be honest I definitely don't like a style like:
```
static ElementCallbacks::StartCallback
startCb(
```
where the `static` and return type are on a different line than the name of the function.

PR #9292, opened 2022-04-25
add missing newline in docu

ferdymercury:


ferdymercury:
@albert-github not sure if and how doxygen.1 should be updated, too

albert-github:
Would be nice to update it too, as it now looks like:
image

(goto the doc directory and run here `man ./doxygen.1`, don't forget the `./` as otherwise you will get the man page from the installation).

I did some searching (https://linux.die.net/man/7/man) and found the following solution:
```
   Use doxygen to compare the used configuration file with the template configuration file
.RS 0
   without replacing the environment variables
.RE
```
which gives:
image


probably this should also be done for the text under point 4 and after point 8)

ferdymercury:
> Would be nice to update it too, as it now looks like: image
> 
> (goto the doc directory and run here `man ./doxygen.1`, don't forget the `./` as otherwise you will get the man page from the installation).
> 
> I did some searching (https://linux.die.net/man/7/man) and found the following solution:
> 
> ```
>    Use doxygen to compare the used configuration file with the template configuration file
> .RS 0
>    without replacing the environment variables
> .RE
> ```
> 
> which gives: image
> 
> probably this should also be done for the text under point 4 and after point 8)

Done as suggested, thanks.

Issue #9148, opened 2022-02-22
No uniquely matching class member, but no candidates are printed.


ferdymercury:
**Describe the bug**
We get 100+ warnings of the type:
```
/opt/root_src/io/sql/src/TSQLFile.cxx:2668: warning: no uniquely matching class member found for 
  void TSQLFile::Streamer(TBuffer &b)
```
but then no "candidates" are proposed.

**Expected behavior**
Either the adjective "uniquely" should be removed, or the "potential candidates", if any, should be listed.

**Screenshots**
Log is attached: [consoleText.txt.zip](https://github.com/doxygen/doxygen/files/8118260/consoleText.txt.zip)

**To Reproduce**
I couldn't yet achieve a "small" reproducer, somehow it only appears in the "big build".

**Version**
1.9.4 (a85850bf5b32843d6cde6b9a25b147c9a40bbc8a*)

albert-github:
You didn't provide the used Doxyfile but from the other questions I see a repeating pattern, so maybe a bit of a wild guess, have a look at:
```
IGNORE_PREFIX          = T
EXCLUDE_PATTERNS    = ....
```

albert-github:
I assume that the normal declaration is from `core/base/inc/Rtypes.h`.
What are the settings for `virtual_keyword` and for `overrd`?

Might be useful yo set `INPUT` to `INPUT=io/sql/src/TSQLFile.cxx` and try to run just the doxygen paryt as `doxygen -d preprocessor`.

ferdymercury:
Thanks for the hint. Attached is the preprocessor output with just the TSQLFile.cxx and .h as input.
[preproc.txt.zip](https://github.com/doxygen/doxygen/files/8119323/preproc.txt.zip)

overrd = override, and virtual_keyword is empty, because:
`ClassDefOverride(TSQLFile, 1)` calls (via Rtypes.h) the following:
`_ClassDefOutline_(name,id,,override)`
which has one empty argument (the third, i.e. the virtual_keyword), maybe that's confusing for doxygen?

This one then declares the Streamer function as:
`virtual_keyword void Streamer(TBuffer&) overrd;`
=>
`void Streamer(TBuffer&) override;`

We do see issues with anonymous namespaces in the full log, but not in this particular class, as far as I can evaluate.

Thanks again!

ferdymercury:
Ok, I think I have found the culprit:
`#include TFile.h: not found! skipping...`

which means I should add this file to the include_path, too.

[...]

PR #9293, opened 2022-04-25
issue #9148 No uniquely matching class member, but no candidates are printed.

ferdymercury:


albert-github:
- Looks like you incorporated here also the fix from #9292
- As title of the proposed pull request a bit more descriptive title should be used like:
   `issue #9148 No uniquely matching class member, but no candidates are printed.`

ferdymercury:
> * Looks like you incorporated here also the fix from [add missing newline in docu #9292](https://github.com/doxygen/doxygen/pull/9292)
> 
>     * As title of the proposed pull request a bit more descriptive title should be used like:
>       `issue #9148 No uniquely matching class member, but no candidates are printed.`

good catch. amended now.

Issue #9302, opened 2022-04-29
incomplete doxygen help message


ferdymercury:
**Describe the bug**
A common situation in GitHub is that users are requested to run the debug / preprocessor to see what might be going wrong. A common answer by users is: how do I run that?

**Expected behavior**
`doxygen --help` or `man doxygen` contain information about all usage flags.

For example:
`doxygen -d qhp` or `-d preprocessor` --> Runs the debug option of ...

Ideally also the website is updated: https://www.doxygen.nl/manual/doxygen_usage.html

**To Reproduce**
`doxygen --help`
`man doxygen`

**Version**
1.9.4 (b81ef4d433445de75404689f3d8c06853bb39b5f*)

ferdymercury:
Proposed patch: https://github.com/doxygen/doxygen/pull/9304

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.4.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9304, opened 2022-04-30
issue #9302 incomplete doxygen help message

ferdymercury:
* Man and --help were out of sync
* Added info on how to print debugging flags on both
* Update web message wrt troubleshooting

### DIFF
--- a/doc/doxygen.1
+++ b/doc/doxygen.1
@@ -21,7 +21,7 @@ doxygen [configName]
 .TP
 4) Use doxygen to generate a template file controlling the layout of the generated documentation:
 .IP
-doxygen -l [layoutFileName]
+doxygen \fB\-l\fR [layoutFileName]
 .IP
 .RS 0
    In case layoutFileName is omitted DoxygenLayout.xml will be used as filename.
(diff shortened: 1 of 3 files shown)


albert-github:
I'm a bit confused by the line (in `doxygen --help` and in the man page):
```
1) Use doxygen to generate a template configuration file*:
```
is it meant here as wild card or as a footnote (I know it is a footnote here but still..., I don't have a good solution for it though as real footnotes are not possible in ASCII text)

ferdymercury:
> I'm a bit confused by the line (in `doxygen --help` and in the man page):
> 
> ```
> 1) Use doxygen to generate a template configuration file*:
> ```
> 
> is it meant here as wild card or as a footnote (I know it is a footnote here but still..., I don't have a good solution for it though as real footnotes are not possible in ASCII text)

Maybe we could use instead `(*)`?

albert-github:
It is hard to get a good solution for this in all types of terminal windows as superscripts are not supported everywhere and where they are supported the probably are supported in different ways.
The suggestion of `(*)` will also have some problems (think about Windows where there can exists already the directory `Program Files (x86)` and about what further wild names users will come up with)

Issue #9301, opened 2022-04-29
outdated year


ferdymercury:
**Describe the bug**
When you print `doxygen --help`, it returns year 2021 instead of 2022.

**Expected behavior**
Year is updated.

**Screenshots**
This needs to be changed:
`src/doxygen.cpp:  msg("Doxygen version %s\nCopyright Dimitri van Heesch 1997-2021\n\n",qPrint(versionString));`

Other places needing update:
```
doc/markdown.doc: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/msc.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/fileinfo.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/dir.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/constexp.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/latexgen.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/scanner.l: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/dir.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/cmdmapper.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/regex.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/membergroup.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/docsets.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/textstream.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/constexp_p.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/fileinfo.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/lexscanner.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/dia.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/conceptdef.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/utf8.cpp: * Copyright (C) 1997-2021 by Dimitri van Heesch.
src/doxygen_lex.h: * Copyright (C) 2021 by Dimitri van Heesch.
src/htmldocvisitor.h: * Copyright (C) 1997-2021 by Dimitri van Heesch.
[...]

PR #9305, opened 2022-04-30
issue #9301 outdated year

ferdymercury:
Old years are bumped


doxygen:
@ferdymercury The way I understand copyright law is that the year only needs to be updated when a (significant) change is made to a file. So just updating the year in a separate commit without further changes is not really useful. See also https://opensource.stackexchange.com/questions/6389/how-do-the-years-specified-in-a-copyright-statement-work

ferdymercury:
> @ferdymercury The way I understand copyright law is that the year only needs to be updated when a (significant) change is made to a file. So just updating the year in a separate commit without further changes is not really useful. See also https://opensource.stackexchange.com/questions/6389/how-do-the-years-specified-in-a-copyright-statement-work

ohh, I understand. Thanks for clarification, let's close this then.

PR #9307, opened 2022-05-02
Correcting README

FarzamDorostkar:
The current description of Doxygen implies that VHDL is a "programming language", which is not correct. VHDL is a hardware description language (HDL), and HDLs should never be thought of as programming languages since they are totally different concepts. You might be interested to correct the same mistake on the project website as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@ Doxygen
 Doxygen is the de facto standard tool for generating documentation from
 annotated C++ sources, but it also supports other popular programming
 languages such as C, Objective-C, C#, PHP, Java, Python, IDL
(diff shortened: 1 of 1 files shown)


albert-github:
Probably also the formulation in the manual should be adjusted accordingly, the source is in doc/index.doc

ferdymercury:
and also https://github.com/doxygen/doxygen/blob/master/cmake/packaging.cmake#L21

FarzamDorostkar:
@albert-github @ferdymercury In [my second commit](https://github.com/doxygen/doxygen/pull/9307/commits/3d3441e1e1a958a22de15473f24eb13bf7d5ac74) I have adjusted the files you mentioned plus doc/starting.doc accordingly.

PR #9318, opened 2022-05-06
fold long template params

karthikeyann:
Fold template params in separate lines if number of template params is greater than 2.
For example: Class with more than 2 template parameters that have parameters in separate lines have better readability. https://nvlabs.github.io/cub/classcub_1_1_arg_index_input_iterator.html

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -1316,7 +1316,8 @@ void ClassDefImpl::writeTemplateSpec(OutputList &ol,const Definition *d,
       {
         Argument a = *it;
         if (al.size()>2) {
-          ol.docify("\n    ");
+          ol.lineBreak();
+          ol.writeNonBreakableSpace(4);
         }
         linkifyText(TextGeneratorOLImpl(ol), // out
           d,                       // scope
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -1315,6 +1315,9 @@ void ClassDefImpl::writeTemplateSpec(OutputList &ol,const Definition *d,
       while (it!=al.end())
       {
         Argument a = *it;
+        if (al.size()>2) {
+          ol.docify("\n    ");
+        }
         linkifyText(TextGeneratorOLImpl(ol), // out
           d,                       // scope
           getFileDef(),            // fileScope
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -2179,8 +2179,7 @@ void HtmlGenerator::addIndexItem(const QCString &,const QCString &)
 
 void HtmlGenerator::writeNonBreakableSpace(int n)
 {
-  int i;
-  for (i=0;i<n;i++)
+  for (int i=0;i<n;i++)
   {
     m_t << "&#160;";
   }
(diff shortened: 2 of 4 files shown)


albert-github:
The basic idea looks OK, though I tried to reproduce it wit the CUB 1.16.0 release and specifically the `arg_index_input_iterator.cuh` file (I renamed it to `arg_index_input_iterator.h`) so I have the example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8644839/example.tar.gz)

In the original version I see:
image
and in the updated version this is still the case.

Looking at the, non persistent, link you gave it looks like:
image

but it looks like this has been generated with doxygen 1.8.4 (May 19, 2013) together with some local templates (CSS), the later I cannot check as they are not in the mentioned release and on https://github.com/NVIDIA/cub I don't see them either).
I used the 1.9.4 version.

When I use the 1.8.4 version I get:
image


- did you try it against the current master version?
- what are the results before and after the cahange?
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that you used to show the improvement? Please don't add external links as they might not be persistent.

karthikeyann:
Thank you for quick review! 
[NVlabs/cub](https://github.com/NVlabs/cub/blob/gh-pages/docs/extra_stylesheet.css) has HTML_EXTRA_STYLESHEET with  `div.compoundTemplParams {  white-space: pre-wrap;  }`
white-space `pre-wrap` property allows to show the printed newline.
- Attached updated self-contained [example.tar.gz](https://github.com/doxygen/doxygen/files/8646619/example.tar.gz)

generated with 1.9.4 with this PR change, looks like
image

generated with 1.8.4 with this PR change, looks like
image

- Without `pre-wrap`, generated with 1.9.4 with this PR change, looks like
image

Options I could see are either leave it to the user to use `pre-wrap` in their `white-space` css property or use `<br />&#160;`.

albert-github:
I'm not really in favor of using things like `<br />&#160;` and the requirement of the `white-space: pre-wrap;` as
- this is only for HTML
- does it work with other formats as well (LaTeX, docbook, rtf, ...)
- this is now applied for the `<h3>` part as well
- will influence other usages of `div.textblock` and `div.compoundTemplParams` as well.

I think more appropriate might be to use a table here or placing it in an own class (though it has to be checked how this will work on other output formats).

karthikeyann:
`<h3>` css part was added for 1.8.4 version, but not required from 1.9.4 version and master. Ignore this since it's not relevant for master.

newline and space are replaced with `lineBreak()`, `writeNonBreakableSpace()` call.
Also, updated `writeNonBreakableSpace()` to work for 
- Latex (verified 4 space), 
- RTF (verified but shows 8 space) (old code uses extra space in the end  `"\\~ "`)
- HTML(verified 4 space).  
man(verified 4 space) doesn't need update. 
docbook spacing does work or `src/docbookgen.cpp` is not used.

albert-github:
Regarding the remark about docbook in the comment https://github.com/doxygen/doxygen/pull/9318#issuecomment-1121311035, I've just pushed a proposed patch, pull request #9326

karthikeyann:
@doxygen

Issue #9319, opened 2022-05-07
Doc build fails with cairo 1.17.6


antonio-rojas:
**Describe the bug**
Building docs with cairo 1.17.6 fails with this error
```
cd /home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples && ../bin/doxygen diagrams.cfg
error: Failed to extract bounding box from generated diagram file /home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__coll__graph.pdf
/home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__coll__graph.dot:1: error: Problems running dot: exit code=0, command='dot', arguments='"/home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__coll__graph.dot" -Tpdf -o "/home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__coll__graph.pdf"'

error: Failed to extract bounding box from generated diagram file /home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__inherit__graph.pdf
/home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__inherit__graph.dot:1: error: Problems running dot: exit code=0, command='dot', arguments='"/home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__inherit__graph.dot" -Tpdf -o "/home/antonio/Software/Arch/packages/doxygen/trunk/src/doxygen-Release_1_9_4/build/examples/../latex/examples/diagrams/latex/class_a__inherit__graph.pdf"'
[...]
```
Error goes away after downgrading cairo to 1.17.4. Graphviz is at 3.0.0

albert-github:
- Do you have the `class_a__coll__graph.pdf` as created with cairo 1.17.4 and 1.17.6?

albert-github:
[...]

PR #9320, opened 2022-05-07
issue #9319 Doc build fails with cairo 1.17.6

albert-github:
The `\MediaBox` field is written as:
```
"   /MediaBox [ 0 0 %f %f ]
```
bur read with
```
sscanf(p+bblen,"%d %d %d %d",&x,&y,width,height)
```
this has been corrected.


albert-github:
By chance, in the llvm project,  I found an example there is a number with a dot:
```
/MediaBox [0 0 500 126.58]
```
this was parsed OK till now as it concerned the last field, but would would give problems in case it was in the third field.

PR #9329, opened 2022-05-11
refactoring: split DotNode::writeLabel from DotNode::writeBox

makelinux:
to allow further generic development.


doxygen:
@makelinux Please check Github actions, your change seems to break generating doxygen's own documentation.

PR #9341, opened 2022-05-16
EXPAND_AS_DEFINED not expanded recursively

albert-github:
When having:
```
#define DOC_NODES DN(DocWord)           DN_SEP DN(DocLinkedWord)    DN_SEP

#define DN(x) constexpr const char *docNodeName(const x &n) { return #x; }
#define DN_SEP
DOC_NODES
#undef DN
#undef DN_SEP
```
with the settings:
```
QUIET                  = YES
EXTRACT_ALL            = YES
ENABLE_PREPROCESSING   = YES
MACRO_EXPANSION        = YES
EXPAND_ONLY_PREDEF     = YES
ALWAYS_DETAILED_SEC    = YES
PREDEFINED             =
EXPAND_AS_DEFINED     += DN
EXPAND_AS_DEFINED     += DN_SEP
EXPAND_AS_DEFINED     += DOC_NODES
```
The result is that `DOC_NODES` is expanded, but not the `DN` and `DN_SEP` used inside.
So the functions:
```
constexpr const char *  docNodeName (const DocWord &n)
constexpr const char *  docNodeName (const DocLinkedWord &n)
```
are not shown.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8699888/example.tar.gz)

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -277,7 +277,7 @@ INCLUDE_PATH           = libmd5 \
                          libmscgen
 INCLUDE_FILE_PATTERNS  =
 PREDEFINED             =
-EXPAND_AS_DEFINED      =
+EXPAND_AS_DEFINED      = DOC_NODES DN DN_SEP
 SKIP_FUNCTION_MACROS   = YES
 #---------------------------------------------------------------------------
 # Configuration options related to external references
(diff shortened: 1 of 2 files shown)


albert-github:
Has been obsoleted by means of commit https://github.com/doxygen/doxygen/commit/387f508b2e41179a9f70b2914132c2bf12d07b43

PR #9353, opened 2022-05-25
deduplicate writeGraphHeader

makelinux:
deduplicate writeGraphHeader
    
No need to redefine node shape.

### DIFF
--- a/src/dotgroupcollaboration.cpp
+++ b/src/dotgroupcollaboration.cpp
@@ -316,25 +316,6 @@ bool DotGroupCollaboration::isTrivial() const
 
 void DotGroupCollaboration::writeGraphHeader(TextStream &t,const QCString &title) const
 {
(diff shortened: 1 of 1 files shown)


src/dotgraph.h
```diff
@@ -36,6 +36,7 @@ class DotGraph
                  m_zoomable(TRUE), m_urlOnly(FALSE) {}
     virtual ~DotGraph() = default;
 
+    static void writeGraphHeader(TextStream& t, const QCString& title = QCString());
```
doxygen, line 39: Does this need to become public? I would expect keeping it protected would work fine as DotGroupCollaboration publicly inherits from DotGraph.

makelinux: Right. Fixed.

makelinux: @doxygen , would you like to review?

PR #9361, opened 2022-05-30
Specify input encoding based on a file pattern

albert-github:
Sometimes a few files in a project are encoded according to another encoding then the default `INPUT_ENCODING`.
With the setting `INPUT_FILE_ENCODING` at pattern and encoding can be given to overcome this problem.

Found by Fossies for the project: symfony
See also: https://github.com/symfony/symfony/discussions/46477

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/8799196/example.tar.gz)

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -11407,6 +11407,10 @@ void adjustConfiguration()
         "Check the 'INPUT_ENCODING' setting in the config file!\n",
         qPrint(Config_getString(INPUT_ENCODING)),qPrint("UTF-8"),strerror(errno));
   }
+  else
+  {
+    portable_iconv_close(cd);
+  }
 
   // check and split INPUT_FILE_ENCODING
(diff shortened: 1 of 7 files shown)


src/doxygen.cpp
```diff
@@ -11396,6 +11397,47 @@ void adjustConfiguration()
...
+    }
+    else
+    {
+      QCString pattern = mapStr.left(i).stripWhiteSpace().lower();
+      QCString encoding = mapStr.mid(i+1).stripWhiteSpace().lower();
+      if (pattern.isEmpty() || encoding.isEmpty())
+      {
+        continue;
+      }
+      void *cd = portable_iconv_open("UTF-8",encoding.data());
```
doxygen, line 11430: I'm missing the corresponding portable_iconv_close() or was there a reason to omit it?

albert-github: It was just forgotten, I've corrected it.

PR #9363, opened 2022-05-31
Changed dot tool launch method

udaken:
On Windows platforms, prevented conhost from starting when starting dot tool.

### DIFF
--- a/src/dotrunner.cpp
+++ b/src/dotrunner.cpp
@@ -197,7 +197,7 @@ bool DotRunner::run()
       srcFile = m_jobs.front().srcFile;
       srcLine = m_jobs.front().srcLine;
     }
-    if ((exitCode=Portable::system(m_dotExe,dotArgs,FALSE))!=0) goto error;
+    if ((exitCode=Portable::system(m_dotExe,dotArgs,true))!=0) goto error;
   }
   else
   {
@@ -206,7 +206,7 @@ bool DotRunner::run()
       srcFile = s.srcFile;
       srcLine = s.srcLine;
       dotArgs=QCString("\"")+m_file+"\" "+s.args;
-      if ((exitCode=Portable::system(m_dotExe,dotArgs,FALSE))!=0) goto error;
+      if ((exitCode=Portable::system(m_dotExe,dotArgs,true))!=0) goto error;
     }
   }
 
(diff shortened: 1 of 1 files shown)


udaken:
> * What is the disadvantage of the fact that this extra process is started (might give slight increase in memory usage and time usage though)

The main reason is the time used for processing.
It takes time to start processing (probably due to anti-virus software).

In the project I used, the clean usage time generated from an empty directory was reduced.(debug build)

- 8a9c41ee3dad073cf20090732de06e6e4fca05ae :  199s
- 2cf0a5ea4526870b648985d7f10fe19d75d267e3: 160s

>     * which version of dot are you using?

I am using 2.44.1(64bit version).

```
> dot -V
dot - graphviz version 2.44.1 (20200629.0846)
```
>     * which version of Windows are you using?

Windows 10 64bit Version 21H2 (build 19044)

>     * which shell / command prompt are you using?

I ran it with cmd and powershell on the Windows Terminal.

>     * Which compiler are you using?

Visual Studio 2017( cmake with `-G "Visual Studio 15 2017" `)

> On my Windows 10 system I don't see differences between the versions with and without the patch, it looks at my side always like:

Sorry, I forgot to add one commit.

It works well in my environment as shown below.
image


Check 2cf0a5ea4526870b648985d7f10fe19d75d267e3

doxygen:
@udaken I agree with Albert that the proposed logic is strange, and what about the other uses of `Portable::system()` that have `commandHasConsole` (implicitly) set to `true`? Should these also be run with `SEE_MASK_NO_CONSOLE`?

Gabrielcarvfer:
@doxygen sorry for sidetracking here, but I believe his end goal is the same so it shouldn't be that bad.

I was testing a project build yesterday and noticed the Dot part is the slowest part on Windows (82 minutes with 16 threads for 36k dot files vs 6 minutes taken by doxygen). 

Passing all .dot files to a single Dot process at once took 36 minutes, more than twice as fast just for not launching as many processes (which are super expensive on Windows, especially when under load). `measure-command{dot -Tpng -O *.dot}`

Then I tried using CREATE_SUBDIRS and passing each subdirectory to a different Dot process. It ended up taking about 9 minutes.
`measure-command{Get-ChildItem -Recurse -Directory | Foreach-Object -Parallel {cd $_.FullName; dot -Tpng -O *.dot}}`

PR #9375, opened 2022-06-06
Enabling include files for lex files

albert-github:
The motivation for this is that make the rules and code as used in the different lexers more consistent as now some definitions are used in multiple lexers (e.g. for white space) and they are not always consistent.
This patch provides the infrastructure so that late on it is easy to add the include files.

The (f)lex program is missing an include facility and the suggestion (https://stackoverflow.com/questions/66197941/defining-lex-include-files-for-flexers) was to write an own pre-processor.
The code presented here functions as a pre-processor and consists of a number of states:
- pre processing the `.l` file to include files and remember where these lines are included
- creating the so called scan states include file, this was already present but now uses the result of the previous step
- translate the pre-processed `.l` file to an `.cpp` file
- post-process the file created `.cpp` file from  the previous step so that line numbers and used files refer to the original file(s) and the now adjusted `cpp` file.

### DIFF
--- a/addon/doxywizard/CMakeLists.txt
+++ b/addon/doxywizard/CMakeLists.txt
@@ -80,15 +80,33 @@ add_custom_command(
 set_source_files_properties(${GENERATED_SRC_WIZARD}/configdoc.cpp PROPERTIES GENERATED 1)
 
 set(LEX_FILES config_doxyw)
+
+# Currently the complete list of lex include files for the dependency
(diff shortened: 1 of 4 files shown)


doxygen:
@albert-github I did a bit of tweaking in [this commit](https://github.com/doxygen/doxygen/commit/cc353e3507aa3a9fc934dd882e06ad8e1b42a613). Please check if I didn't break anything by accident.

PR #9383, opened 2022-06-09
refactor DotLegendGraph::computeTheGraph: consolidate common edge attributes

makelinux:
and remove multiple duplicated definition

Excerpt from doc documentation:

If a default attribute is defined using a `node`, `edge`, or `graph` statement,
or by an attribute assignment not attached to a node or edge,
any object of the appropriate type defined afterwards will
inherit this attribute value.

Reference:
 - https://graphviz.org/doc/info/lang.html#lexical-and-semantic-notes


src/dotgroupcollaboration.cpp
```diff
@@ -316,25 +316,7 @@ bool DotGroupCollaboration::isTrivial() const
...
-  t << "{\n";
-  if (Config_getBool(DOT_TRANSPARENT))
-  {
-    t << "  bgcolor=\"transparent\";\n";
-  }
-  t << "  edge [fontname=\"" << fontName << "\",fontsize=\"" << fontSize << "\","
-    "labelfontname=\"" << fontName << "\",labelfontsize=\"" << fontSize << "\"];\n";
-  t << "  node [fontname=\"" << fontName << "\",fontsize=\"" << fontSize << "\",shape=box];\n";
+  DotGraph::writeGraphHeader(t, title);
+  t << "  node [shape=box];\n";
```
albert-github, line 320: I thought the `shape="box"` was not necessary anymore for nodes.

makelinux: see #9353  for writeGraphHeader

PR #9390, opened 2022-06-10
issue 9386 Problems with the directory dependency graph

albert-github:
Seen
- the discussion with the mentioned issue
- the fact that a `record` is not really suited / intended for the us doxygen has for it
- the fact that the "problem" already exists since `dot` version 2.40 (as far as we can tell)

I think it is better to use `shape="box"` globally instead of `shape=record` even when GraphViz will improve `dot` to suit our needs as the problem remains in the older versions of `dot` (which wil probably be used for some time).

### DIFF
--- a/src/dotgraph.cpp
+++ b/src/dotgraph.cpp
@@ -301,7 +301,7 @@ void DotGraph::writeGraphHeader(TextStream &t,const QCString &title)
          "labelfontname=\"" << fontName << "\","
          "labelfontsize=\"" << fontSize << "\"];\n";
   t << "  node [fontname=\"" << fontName << "\","
-         "fontsize=\"" << fontSize << "\",shape=record];\n";
+         "fontsize=\"" << fontSize << "\",shape=\"box\"];\n";
 }
 
 void DotGraph::writeGraphFooter(TextStream &t)


src/dotgraph.cpp
```diff
@@ -301,7 +301,7 @@ void DotGraph::writeGraphHeader(TextStream &t,const QCString &title)
          "labelfontname=\"" << fontName << "\","
          "labelfontsize=\"" << fontSize << "\"];\n";
   t << "  node [fontname=\"" << fontName << "\","
-         "fontsize=\"" << fontSize << "\",shape=record];\n";
+         "fontsize=\"" << fontSize << "\",shape=\"box\"];\n";
```
makelinux, line 304: Comas and quotes are optional for graphviz and are confusing in embedded code. IMHO better just `shape=box`

makelinux: BTW I have fix to make attributes configurable.

albert-github: I think it would be better to have the `"` present as it is more conform with the other code like `fontsize`, `fontname` etc., though the word `box` is just one word and `fontsize` and `fontame` could have e.g a space in it which  makes here the quotes mandatory.

I don't get the point of "are confusing in embedded code", cannot think of any confusion.

albert-github: Regarding https://github.com/doxygen/doxygen/pull/9390#discussion_r894450582 (configurable), depends to which level it is configurable but I'm not sure about it. Maybe you should make as proposed pull request out of what you have.

makelinux: I am proposing #7839 for already two years and one day. Now I spit it into small chunks. The next chunk is #9353.

PR #9392, opened 2022-06-11
Enabling dependencies for include files in lex files

albert-github:
In the pull request #9375 the possibility was created to have lex include files for common parts.
The missing part was the dependencies of between the `.l` file and the included files.
By means of the stack exchange question:
- https://stackoverflow.com/questions/62713117/dynamic-dependency-within-custom-source-files
- https://stackoverflow.com/questions/72518401/cmake-dependencies-on-included-files

a solution was found for the dependency problem by:
- using the `DEPFILE` directive
- creating the dependency file (essential is that the dependencies i.e. the included file all are on one line).

### DIFF
--- a/addon/doxywizard/CMakeLists.txt
+++ b/addon/doxywizard/CMakeLists.txt
@@ -81,16 +81,17 @@ set_source_files_properties(${GENERATED_SRC_WIZARD}/configdoc.cpp PROPERTIES GEN
 
 set(LEX_FILES config_doxyw)
 
-# Currently the complete list of lex include files for the dependency
-# The dependency should be a more dynamic list based on the content of the lex file
-set(LEX_INC_FILES)
-
 foreach(lex_file ${LEX_FILES})
     add_custom_command(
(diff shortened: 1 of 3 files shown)


lamm45:
DEPFILE support for Makefiles was added in CMake 3.20:
https://cmake.org/cmake/help/latest/command/add_custom_command.html#generating-files

However, cmake_minimum_required is currently set to 3.3, but the instructions explicitly specify Makefiles as the generator:
https://doxygen.nl/download.html

For example, CMake 3.16.3 gives the following error (repeated multiple times):
```
CMake Error at src/CMakeLists.txt:120 (add_custom_command):
  add_custom_command Option DEPFILE not supported by Unix Makefiles
```

albert-github:
Correct.
The `DEPFILE` support was added on different times for the different engines.

- can you update to a newer version of cmake (preferred)
- otherwise as a workaround just remove locally the `DEPFILE` from the `src/CMakeLists.txt`  and the addon/doxywizard/CMakeLists.txt`


@doxygen which strategy to follow
- increase the minimal required cmake version to 3.21 (my preference)
- make some workaround code

doxygen:
@albert-github Ubuntu 20.04 ships with version 3.16 (which I also use on Windows), so I think we should stick with that as the minimum version for now and make a workaround.

albert-github:
I've just pushed a proposed workaround patch, pull request #9442

albert-github:
The workaround code for the older CMake versions has been integrated in master on GitHub.

PR #9387, opened 2022-06-10
delete TAGFILE for build error:

Angie-JA-Dou:
error: Tag file 'example.tag' does not exist or is not a file.

PR #9395, opened 2022-06-13
Dependency between examples.

albert-github:
When building the `tag` example a prerequisite is that the `example` example has been build (as the `tag` example needs the `example.tag` file, so we have first generate the `example` example before building the `tag` example).
An explicit dependency structure for relations between examples has been defined.

(based on the problem as reported in the proposed patch:  delete TAGFILE for build error: #9387 )

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -17,6 +17,9 @@ foreach (f  ${EXAMPLE_FILES})
     list(APPEND OUT_EXAMPLE_FILES "${PROJECT_BINARY_DIR}/examples/${f}")
 endforeach()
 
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I saw some duplication so I restructured things a bit. See https://github.com/doxygen/doxygen/commit/09a6b3e8ddca2328d08ae4715f42df1d07d55700 Please check if it still works as you intended.

albert-github:
Due to the chosen order (i.e. the `example` example is generated before the `tag` example) but by moving the `EXAMPLE_tag` to the front of the list it looks like it is OK.

Issue #9096, opened 2022-02-04
Call Graph corrupt and/or incomplete


DUChost:
The call graph is incomplete or/and wrong at least at the marked positions.

I have played around with the different parameters of DoxyGen and also rearranged the C file, unfortunately all without success.

The whole project is in the ZIP file. In the subdirectory "Files" is also the used configuration for DoxyGen.

Best regards UweC

[DOXY_BUG_REPORT.zip](https://github.com/doxygen/doxygen/files/8004782/DOXY_BUG_REPORT.zip)
![call_graph_bug](https://user-images.githubusercontent.com/55077974/152568990-8405f672-2081-4c01-8867-b54a7ee251b4.png)
![ApplicationC](https://user-images.githubusercontent.com/55077974/152569609-a8bb2e98-ef7b-4195-85b9-31743ec2b65e.png)

albert-github:
It looks like the problem is caused by the variable ` usSysTick`
In SRC/Application.h it is defined as a local variable:
```
  UInt16_t usSysTick;
```

and in SRC/Systick_WDT.c as:
```
static volatile UInt16_t usSysTick = 0;
```

In principle these 2 values should not influence each other in the documentation (like they also don't in the compiler).

As a workaround you could redefine local variable in SRC/Application.h to e.g.  `usSysTickLow` or use the setting `EXTRACT_STATIC=YES`

DUChost:
Dear Albert,

yes, using EXTRACT_STATIC works in my case: Thanks!
Do you have any other suggestions for my files and by the way, how did you solve that problem?

--> Did you recognize that "Process_RealTime" is not called by "main" but via callback from "SysTick_WDT"?!

Again, Thanks a lot!

albert-github:
> how did you solve that problem?

Trial and error, i.e. reducing the number of files and after that reducing the remaining files, a bit of luck and some educational guessing / experience.

Note: that when using `EXTRACT_STATIC=YES` and using `SOURCE_BROWSER=YES` in the function `Process` the local variable `usSysTick` is "linked" to the `static volatile` variable! So better to use the`usSysTickLow` workaround.

> --> Did you recognize that "Process_RealTime" is not called by "main" but via callback from "SysTick_WDT"?!

The problem here is that in `Appication.c` in the function `main` we have:
```
SysTick_Init( Process_RealTime );
```
[...]

Issue #5958, opened 2018-07-02
References for one function can inherit References from subsequent non documented function (Origin: bugzilla #762670)


doxygen:
status RESOLVED severity *normal* in component *general* for *---*
Reported in version *1.8.11* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[DoxygenReferencesTest.zip](http://ftp.stack.nl/pub/users/dimitri/bugzilla/attachments//bug762670_322350_DoxygenReferencesTest.zip)_ (ID 322350)

On 2016-02-25 10:51:44 +0000, Tomasz MoÅ wrote:

> Created attachment 322350
> REFERENCES_RELATION test case showing "inheritance" from undocumented functions
> 
> When REFERENCES_RELATION is set to YES, function can "inherit" references from subsequent undocumented function.
> 
> The test case reproduces the bug in Doxygen 1.8.11.
> 
> Expected behaviour:
>   get_int() function documentation should include:
>   References ARRAY_ELEMENTS, E_ONE, E_TWO, iOne, iTwo.
> 
> Actual behaviour:
>   get_int() function documentation includes:
>   References ARRAY_ELEMENTS, E_ONE, E_TWO, iOne, iTwo, iUndocumentedFour, iUndocumentedOne, iUndocumentedThree, iUndocumentedTwo, and UndocumentedStruct::one.
> 
> Attached is complete working example.

On 2016-03-20 10:53:36 +0000, Dimitri van Heesch wrote:

> Confirmed. Should be fixed in the next GIT update.

On 2016-09-05 13:45:18 +0000, Dimitri van Heesch wrote:

> This bug was previously marked ASSIGNED, which means it should be fixed in
> doxygen version 1.8.12. Please verify if this is indeed the case. Reopen the
> bug if you think it is not fixed and please include any additional information 
> that you think can be relevant (preferably in the form of a self-contained example).

albert-github:
It looks like that for this example there are some new problems since 1.8.15, see also #8061

PR #9396, opened 2022-06-14
issue #9096 Call Graph corrupt and/or incomplete (2)

albert-github:
In case of the problem as reported by @septatrix, the problem is caused by the setting of the `currentMemberDef` to 0 in case the member "does not exist" / is not documented. When documenting the (here enum value) or setting `EXTRACT_ALL=YES` the problem does not occur.

The problem was introduced when fixing  References for one function can inherit References from subsequent non documented function (Origin: bugzilla 762670) #5958.
```
Commit: 9abcad810b8d41d338d501ff5b32524e1ced7f33 [9abcad8]
Date: Sunday, March 20, 2016 11:51:25 AM
```

Removing the setting of `currentMemberDef=0` fixes the current issue but also the original issue is still OK.

Note: that the the in the code mentioned `bug762760` is  not correct this has to be `bug762670`.

### DIFF
--- a/src/code.l
+++ b/src/code.l
@@ -2835,12 +2835,6 @@ static bool getLinkInScope(yyscan_t yyscanner,
         return TRUE;
       }
     }
-    else // found member, but it is not linkable, so make sure content inside is not assigned
-         // to the previous member, see bug762760
-    {
-      DBG_CTX((stderr,"unlinkable member %s\n",qPrint(md->name())));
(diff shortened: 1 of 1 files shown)


septatrix:
I just compiled locally with this patch applied and it works as expected, thanks!

Would this reintroduce the old bug or is that now fixed elsewhere?

Issue #9413, opened 2022-06-20
problem with \file attribute and soft-link


aotto1968:
**Describe the bug**
I get the following error:
`…/tcl_MpGrep.doc:2: warning: the name '…/MpGrep.doc' supplied as the argument in the \file statement is not an input file`

The problem is that `tcl_MpGrep.doc` is a soft-link from `MpGrep.doc` and `MpGrep.doc` has the header

```
/**
\file       theLink/docs/man/MpGrep.doc
\brief      \$Id: pages.doc 341 2020-01-22 21:32:38Z dev1usr $
…
```

> ls -al tcl_MpGrep.doc
lrwxrwxrwx 1 dev1usr users 15 26. Feb 2021  tcl_MpGrep.doc -> MpGrep.doc

the `MpGrep.doc` is used as template for different files (tcl,cc,…), so that I can't just rename it.

**Expected behavior**
I expect that that the `\file` command is able to deal with _soft-link_

**To Reproduce**
just do a simple _soft-link_ from a **.doc** file.

**Version**
1.9.3

albert-github:
Recently some improvements have been made in respect to links, which landed in the current master.
- Please have a try with the current master (1.9.5 (886349348ad5dc60152e0d5b92e6ffbaf5f4a921)
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that shows the problem that should be fixed with the proposed patch? Please don't add external links as they might not be persistent.

albert-github:
A quick revisit and attempt with the current master, and this looks OK to me.
The example setup is like:
```
.:
total 17
 4 drwxrwxr-x+ 1 Auser  Agrp  0 Jun 20 17:06 .
 1 -rw-rw-r--+ 1 Auser  Agrp 36 Jun 20 17:04 Doxyfile
 0 lrwxrwxrwx  1 Auser  Agrp 20 Jun 20 17:03 the_real_file.h -> orig/the_real_file.h
 0 drwxrwxr-x+ 1 Auser  Agrp  0 Jun 20 17:02 orig
12 drwxrwx---+ 1 Auser  Agrp  0 Jun 20 17:00 ..

./orig:
total 5
4 drwxrwxr-x+ 1 Auser  Agrp  0 Jun 20 17:06 ..
0 drwxrwxr-x+ 1 Auser  Agrp  0 Jun 20 17:02 .
[...]

PR #9423, opened 2022-06-27
Prohibit in-place doxygen builds

albert-github:
Based on some build problems in issue #9413,  it is checked that there is no attempt to build doxygen in-place (i.e. that the main source directory and the build directory have to be different directories).
(The original problem shows errors like:
```
cycle in dependency tree for target 'generated_src\configimpl.corr'
```
)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,8 +36,10 @@ set_property(CACHE force_qt PROPERTY STRINGS OFF Qt6 Qt5)
 
 SET(enlarge_lex_buffers "262144" CACHE INTERNAL "Sets the lex input and read buffers to the specified size")
 
-if ("${PROJECT_BINARY_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}")
-  message(FATAL_ERROR "Doxygen cannot be generated in-place, the build directory (${PROJECT_BINARY_DIR}) has to differ from the doxygen main directory (${PROJECT_SOURCE_DIR})\nPlease don't forget to remove the already file created file 'CMakeCache.txt' and the directory 'CMakeFiles'!")
+if(enable_coverage)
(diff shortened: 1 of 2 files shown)


albert-github:
In-place build is indeed not recommended and a big advantage of cmake is that it makes it easy to do an out of place build (very useful when building multiple versions like debug / release and / or platforms, in my case Windows and Cygwin).

I agree that when 1.9.3 did build in-place we have to look whether it is possible to build the current master in place as well.
I'll have a look into the matter.

albert-github:
I just ran the in-place build for the 1.9.3 release with nmake and got 
```
NMAKE : fatal error U1071: cycle in dependency tree for target 'src\configimpl.l'
Stop.
NMAKE : fatal error U1077: '"D:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX86\x86\nmake.exe"' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"D:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX86\x86\nmake.exe"' : return code '0x2'
```

So looks like the 1.9.3 version did not build in-place either at least not with `nmake`

albert-github:
Under Cygwin I get:
```
make[2]: Circular src/configimpl.l <- src/configimpl.l dependency dropped.
```
but the compilation continues.

I found the problematic part, it is about the enabling of the coverage possibility.
I think that it is best to protect this possibility with the conditional `if(enable_coverage)` and make it when it is enabled a `FATAL _ERROR`. As the coverage currently only applies to gcc it is possible to make a warning out of it.
I'll update this proposed pull request.

PR #9435, opened 2022-07-04
bug 539818 C Preprocessor enum printing trick not handled properly by Doxygen

albert-github:
The determination of the end quote was not handled properly, so when inside a string we had, at the end, also skip the end quote.
Special attention had to be paid to  escaped quotes.

(Extended) example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9038609/example.tar.gz)

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -889,7 +889,7 @@ void linkifyText(const TextGeneratorIntf &out, const Definition *scope,
   size_t strLen = txtStr.length();
   if (strLen==0) return;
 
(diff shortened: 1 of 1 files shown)


albert-github:
In the current situation we see:

image

In the patched situation we see:

image

Removing the line `if (insideString) index++` gives:

image


So it looks like the line is not needed anymore.

My first test was with a version where there were no backslashes in the strings and at that moment it was necessary. Now that the backslash tests are build in it looks like it is not necessary anymore.

I'll update the commit.

Issue #9440, opened 2022-07-09
[Regression] Search field focus clearing broken and odd ".html" label


Krinkle:
**To Reproduce**

After focussing search field and clicking outside of it, the label is no longer "Search" but is now `.html`.

Also, the previoue behaviour where the word "Search" is removed when clicked, is no longer working. Now the text "Search" stays behind and has to be manually removed.

**Expected behavior**

It says the same.

**Screenshots**

<img width="249" alt="Screenshot 2022-07-09 at 16 36 45" src="https://user-images.githubusercontent.com/156867/178125921-bd27a476-3aee-473d-98a0-92c33a81e4a1.png">

**Version**

Doxygen 1.9.4, from Homebrew, on macOS.

**Additional context**

This is a regression, as it does not affect Doxygen 1.9.1.

The code was recently changed in https://github.com/doxygen/doxygen/commit/be61d64e75e1b9df5ad463130bbea69a8657db9a, which might have introduced the bug. I don't see an obvious mistake though. It removed a `false` argument in the `new SearchBox` call. That appears correct as there are now only 4 instead of 5 arguments, passed, and those appear to be passed in the correct places with the correct meaning. Yet, upon debugging in Firefox devtools, the `.html` extension value ends up as `this.searchLabel`.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.5.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9441, opened 2022-07-10
issue #9440: Fix search label regression

Krinkle:
### Background:

The extension value (`.html`) was injected instead of the search
label.

This meant that upon focussing the field it was no longer cleared,
so the "Search" text remains in the input field and thus had
to be removed manually.

It also meant that after de-focussing the field (click anywhere),
the field value was assigned to `.html` which made it show up in the
user interface even.

### Change:

Use the HTML5 placeholder attribute instead, which is a browser
built-in that does this automatically. It displays the value
while the field is empty, and makes it dissappear when the empty
field is focussed.

This has the added benefit of no longer destroying the user input
if you type something and then click or tab elsewhere before
submitting the form.

Also:

* Remove the now-unused code for managing this value.
* Synchronise the signature of the two SearchBox js classes to avoid
  mistakes like this in the future. The `inFrame` argument is not
  used in extsearch.js. Replace it with an equally-unused `extension`
  arg for consistency with search.js.

Fixes https://github.com/doxygen/doxygen/issues/9440.

### DIFF
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -1172,7 +1172,7 @@ void HtmlGenerator::startFile(const QCString &name,const QCString &,
     m_t << "<script type=\"text/javascript\">\n";
     m_t << "/* @license magnet:?xt=urn:btih:d3d9a9a6595521f9666a5e94cc830dab83b65699&amp;dn=expat.txt MIT */\n";
     m_t << "var searchBox = new SearchBox(\"searchBox\", \""
-        << m_relPath<< "search\",'" << theTranslator->trSearch() << "','" << Doxygen::htmlFileExtension << "');\n";
+        << m_relPath<< "search\",'" << Doxygen::htmlFileExtension << "');\n";
     m_t << "/* @license-end */\n";
     m_t << "</script>\n";
   }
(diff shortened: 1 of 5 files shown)


Krinkle:
I tested it locally on macOS. I'm quite inexperienced with C and C++. I found my way through the cmake/flex/bison deps, and ran `cd build/ && cmake .. && make && make install` and then repeated the last two portions after making changes.

I noticed that it was it wasn't picking up the searches to the JS files. I did try `make all` as something semi-random I found that perhaps would pick it up or force it, but no dice. I also tried touching `htmlgen` and `resourcemgr` source files directly but while they did recompile that also didn't seem to make it pick up the JS changes. I'm guessing these are embedded and chached in a way that the Makefile doesn't know about.

```
 ~/Development/doxygen/build (master+)
$ ack extsearch
> generated_src/resources.cpp
> 6272:static const unsigned char extsearch_js_data[] = 
```

I `rm -rf`'ed the generated_src directory and after that the change got picked up. If there was a different way I could have done this, for next time, I'd love to know!

Thanks!

albert-github:
Regarding the remark about the build process in https://github.com/doxygen/doxygen/pull/9441#issuecomment-1179633184, this looks OK.
I tested it (on Windows, I don't have a system with macOS) and didn't see a problem, only problem I can think of is i relation to the code in `src/CMakeLists.txt`:
```
# all resource files
if (${CMAKE_VERSION} VERSION_EQUAL "3.11.0" OR ${CMAKE_VERSION} VERSION_GREATER "3.11.0")
    file(GLOB RESOURCES CONFIGURE_DEPENDS ${PROJECT_SOURCE_DIR}/templates/*/*)
else()
    file(GLOB RESOURCES ${PROJECT_SOURCE_DIR}/templates/*/*)
endif()
```
hence
- which version of cmake are you using?
- which version of macOS are you using?

Krinkle:
>  * which version of cmake are you using?
>  * which version of macOS are you using?

cmake 3.23.2 ([Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/be935ab1190dc21d109ac7f34fc29f1b33d0d94e/Formula/cmake.rb)), macOS 10.15.

PR #9442, opened 2022-07-10
Workaround for DEPFILE

albert-github:
The `DEPFILE` possibility in CMake is only available in the newer versions of CMake and also the availability is  platform dependent.
The latest addition is for Windows / XCode and this is in version 3.21, so we restrict us to this version for determining whether or not the `DEPFILE` possibility exists.

### DIFF
--- a/addon/doxywizard/CMakeLists.txt
+++ b/addon/doxywizard/CMakeLists.txt
@@ -81,7 +81,30 @@ set_source_files_properties(${GENERATED_SRC_WIZARD}/configdoc.cpp PROPERTIES GEN
 
 set(LEX_FILES config_doxyw)
 
+if (NOT depfile_supported)
+    # In case the DEPFILE possibility is not supported the complete list of lex include files for the dependency has to be used
+    set(LEX_INC_FILES)
+endif()
+
 foreach(lex_file ${LEX_FILES})
+    if (depfile_supported)
+        add_custom_command(
+            COMMAND ${PYTHON_EXECUTABLE} ${PROJECT_SOURCE_DIR}/src/pre_lex.py ${PROJECT_SOURCE_DIR}/addon/doxywizard/${lex_file}.l  ${GENERATED_SRC_WIZARD}/${lex_file}.l ${GENERATED_SRC_WIZARD}/${lex_file}.corr ${GENERATED_SRC_WIZARD}/${lex_file}.d ${PROJECT_SOURCE_DIR}/src
+            DEPENDS ${PROJECT_SOURCE_DIR}/src/pre_lex.py ${PROJECT_SOURCE_DIR}/addon/doxywizard/${lex_file}.l
+            DEPFILE ${GENERATED_SRC_WIZARD}/${lex_file}.d
+            OUTPUT  ${GENERATED_SRC_WIZARD}/${lex_file}.l ${GENERATED_SRC_WIZARD}/${lex_file}.corr ${GENERATED_SRC_WIZARD}/${lex_file}.d
+        )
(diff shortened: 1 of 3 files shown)


addon/doxywizard/CMakeLists.txt
```diff
@@ -81,7 +81,30 @@ set_source_files_properties(${GENERATED_SRC_WIZARD}/configdoc.cpp PROPERTIES GEN
 
 set(LEX_FILES config_doxyw)
 
+if (NOT depfile_supported)
+    # In case the DEPFILE possibility is not supported the complete list of lex include files for the dependency has to be used
+    set(LEX_INC_FILES)
+endif()
+
 foreach(lex_file ${LEX_FILES})
```
doxygen, line 89: @albert-github instead of not having dependencies when `DEPFILE` is not supported, would it also be an idea to always have all flex files just depend on all flex include files. Then the rule is simple and fixed. Of course, it will result in too much being rebuilt, but I expect the number of flex include files to be limited, their change frequency to be low, and the overhead of regenerating and compiling some extra flex files is also not that big.

albert-github: I don't think it is a good idea to have this fixed rule.
- the number of include files will be limited and the frequency of changes will be low, but when changing e.g. something for Fortran it would be strange and time consuming that the scanners for e.g. C, Python, vhdl would be regenerated  as well.
- it is error-prone (easy to forget to add a file), the `DEPFILE` handles this automatically
- it is a pity that CMake added the `DEPFILE` possibility in such a late state, but I think we should make use of this better and modern possibility when available.

doxygen: ok fair enough. I only now see you already have the full list of dependencies in case the option is not supported. I was under the impression it would not have any dependencies then.

addon/doxywizard/CMakeLists.txt
```diff
@@ -81,7 +81,30 @@ set_source_files_properties(${GENERATED_SRC_WIZARD}/configdoc.cpp PROPERTIES GEN
...
+    if (depfile_supported)
+        add_custom_command(
+            COMMAND ${PYTHON_EXECUTABLE} ${PROJECT_SOURCE_DIR}/src/pre_lex.py ${PROJECT_SOURCE_DIR}/addon/doxywizard/${lex_file}.l  ${GENERATED_SRC_WIZARD}/${lex_file}.l ${GENERATED_SRC_WIZARD}/${lex_file}.corr ${GENERATED_SRC_WIZARD}/${lex_file}.d ${PROJECT_SOURCE_DIR}/src
+            DEPENDS ${PROJECT_SOURCE_DIR}/src/pre_lex.py ${PROJECT_SOURCE_DIR}/addon/doxywizard/${lex_file}.l
+            DEPFILE ${GENERATED_SRC_WIZARD}/${lex_file}.d
+            OUTPUT  ${GENERATED_SRC_WIZARD}/${lex_file}.l ${GENERATED_SRC_WIZARD}/${lex_file}.corr ${GENERATED_SRC_WIZARD}/${lex_file}.d
+        )
+        set_source_files_properties(${GENERATED_SRC_WIZARD}/${lex_file}.l PROPERTIES GENERATED 1)
```
doxygen, line 97: are the `set_source_file_properies()` not also needed in the `else()` case?

albert-github: Indeed the `set_source_file_properies()` should be present in all cases.
Also in the workaround case the `.....d` files are missing as arguments / `OUTPUT`s

PR #9443, opened 2022-07-12
bug_665439 FILE_VERSION_FILTER incorrectly run for file references from a tagfile with filename "<tagfile>:/my/path"

albert-github:
The `FILE_VERSION_FILTER` should not be applied on tag files

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9092838/example.tar.gz)

### DIFF
--- a/src/filedef.cpp
+++ b/src/filedef.cpp
@@ -1526,6 +1526,18 @@ void FileDefImpl::acquireFileVersion()
   if (!vercmd.isEmpty() && !m_filePath.isEmpty() &&
       m_filePath!="generated" && m_filePath!="graph_legend")
   {
+    const StringVector &tagFileList = Config_getList(TAGFILES);
+    for (const auto &s : tagFileList)
+    {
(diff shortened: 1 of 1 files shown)


src/filedef.cpp
```diff
@@ -1526,6 +1526,18 @@ void FileDefImpl::acquireFileVersion()
   if (!vercmd.isEmpty() && !m_filePath.isEmpty() &&
       m_filePath!="generated" && m_filePath!="graph_legend")
   {
+    const StringVector &tagFileList = Config_getList(TAGFILES);
```
doxygen, line 1529: I'd expect this loop can be simplified by just checking for `isReference()`, or do you see a case where that would not work?

albert-github: I'm not sure whether or not the the reference (i.e. just above it by means of `setReference(lref)`) is set or not. This should be checked by means of the example.

A quick test with the example shows that the `p1.h` in is set as reference, so should be OK.

Issue #9047, opened 2022-01-17
Support @param-taint in PHP (permit dashes in command aliases?)


Krinkle:
**Describe the bug**

Unable to remove lines like `@param-taint $foo none` from the HTML output.

Currently this results in:

```
LinkRenderer.php:266: warning: argument '$target' from the argument list of
  LinkRenderer\makeBrokenLink has multiple @param documentation sections
```
<img width="604" alt="Screenshot 2022-01-17 at 23 29 39" src="https://user-images.githubusercontent.com/156867/149847431-7a73634e-8861-411e-b81e-260aa6782423.png">

**Expected behavior**

```php
/**
 * Class that generates HTML anchor link elements for pages.
 *
 * @since 1.28
 */
class LinkRenderer {
	/**
	 * @param LinkTarget|PageReference $target
	 * @param-taint $target none
	 * @param string|HtmlArmor|null $text
	 * @param array $extraAttribs
	 * @param array $query
	 * @return string
	 */
	public function makeBrokenLink(
		$target, $text = null, array $extraAttribs = [], array $query = []
	) {
	}
}
```
```
ALIASES                = "codeCoverageIgnore=\noop" \
                         "codingStandardsIgnoreEnd=\noop" \
                         "codingStandardsIgnoreStart=\noop" \
                         "phan=\noop" \
                         "param-taint=\noop" \
                         "suppress=\noop"
```

**Version**
doxygen 1.9.2 and doxygen 1.9.3.

**Additional context**

```
error: Illegal ALIASES format 'param-taint=\noop'. Use "name=value" or "name{n}=value", where n is the number of arguments
```

I suspect there is (or once was) syntax ambiguity where a command that takes no argument needs to be distinct from other content. E.g. the full-stop in `Hello @foo.`, which seems reasonable to allow as separator. I guess that means we currently also permit the dash to be used in this way, e.g. `Hello @foo- How are you?`. I'm hoping that maybe this is okay to change by considering the dash in this case (if there isn't a space after it) to be considered part of the command.

Or alternatively, we could perhaps do something clever where we first give it a chance to match a command including the adjecent dash, and if not known as command, fallback to rendering separately.

albert-github:
The `-` is indeed the character that is not possible in an ID (in this case a command name), the question that arises is why do you need a `-` in the ID, can't you use an `_`.?


[...]

PR #9451, opened 2022-07-16
issue #9047 Support @param-taint in PHP (permit dashes in command aliases)?

albert-github:
Add the possibility to have a dash in an ALIASES.


albert-github:
Looks like that the generation of the documentation now fails with
```
doc/commands.doc:3461: error: Found unknown command '\LaTeX'
```
looking into the problem ...

Issue #7946, opened 2020-08-05
Add filename and line number support to tags, ala __FILE__ and __LINE__ macros


iwolf81:
Hello,

I'd like to tag a line of code such that the HTML output identifies the filename and line number in a manner similar to what the C/C++ preprocessor does with the \_\_FILE\_\_ and \_\_LINE\_\_ macros. 

For example, given the following code segment:

foo.c:
line 965:  bar();  /// @test{file,lineno} Suite A : TC 1 : BP 2

I'd like to see Doxygen generate the following for the Test List output:

Global bar (void)
foo.c:864 Suite A : TC 1: BP 2

By identifying the filename and line number in this manner, I could create a tool that automatically updates the filenames and line number for breakpoints in debugger scripts when the source files are modified. The same capability could be used to create and automatically update a Software Design Document to Source Code traceability matrix, as suggested by @ls6777 in [Doxygen Create Link to specific line of code](https://stackoverflow.com/questions/59327898/doxygen-create-link-to-specific-line-of-code).

Thank you,
Ira

thom-sch:
I'd also like to ask for an enhancement of Doxygen like above.

I'm  working on documenting Delphi source codes using a set of ALIASES.

One of the main goals was to have included the Delphi source code snippets
as well as the C++ *.hpp snippets (generated by Delphi for C++ builder).
So most of the ALIASES have a parameter which refers to the
currently parsed source file, as an (shortened) example ALIAS DClass:

  \DClass{2}="\class \1 ^^ ... \dontinclude \2.pas ^^ \snippet ...source code snippet... "
  with
  - \1 = class name to document
  - \2 = current source file's name (without extension .pas)

There are some more ALIASES like this to document other entities (types, functions etc.),
and each of them gets the source file's name as parameter.

Things would be much easier if there would be tags as mentioned above like:
  - __FILENAME__: file name only, with extension
  - __FILE__: file name only, without extension
  - __FILEDIR__: file's directory
  - __FILEPATH__: complete file path

In this case parameter \2 of ALIAS \DClass (and all other ALIASES) could be
omitted resulting in better readability and avoiding to repeat the source
file's name in every other ALIAS, in this case:

  \DClass{1}="\class \1 ^^ ... \dontinclude __FILE__.pas ^^ \snippet ...source code snippet... "

Regards from Meerbusch
Thomas Schneider

albert-github:
I've just pushed a proposed patch, pull request #9461

[...]

PR #9461, opened 2022-07-19
issue #7946 Add filename and line number support to tags, ala __FILE__ and __LINE__ macros

albert-github:
Implementation of the basic commands `\filename` and `\fileline` and for the `\filename` command some parts of it by means of options.


doxygen:
Looks like this should have referenced #7946

PR #9463, opened 2022-07-19
Configure dot graph, nodes and arrows attributes

makelinux:
Configure dot graph, nodes and arrows attributes
    
    For example, default:
    DOT_NODE_ATTR = shape=box
    ┌────────┐     ┌────────┐
    │ caller │────►│ callee │
    └────────┘     └────────┘
    
    new options:
    DOT_NODE_ATTR = shape=plain
    
        caller────►callee
    
    DOT_NODE_ATTR = shape=plaintext
    
       caller  ────►  callee
    
    Features of only nodes without boxes:
    - a root node is in bold font
    - truncated nodes are in italic
    
    Changes and features:
    - writeBox renamed to generic writeNode and reworked
    - arrows and nodes can be configured
    - DOT_FONTNAME, DOT_FONTSIZE are replaced with DOT_GRAPH_ATTR,
    DOT_COMMON_ATTR, DOT_EDGE_ATTR, DOT_NODE_ATTR
    - removed duplicated code and used DotGraph::writeGraphHeader

Before:
image

With
DOT_NODE_ATTR = shape=plaintext height=0.2 width=0.4 :
graph_legend


src/dotdirdeps.cpp
```diff
@@ -155,6 +155,7 @@ static void drawDirectory(TextStream &t, const DirDef *const directory, const Do
       "fillcolor=\"" << getDirectoryBackgroundColor(directory->level()-startLevel)   << "\", "
       "color=\""     << getDirectoryBorderColor(property)                            << "\", "
       "URL=\""       << addHtmlExtensionIfMissing(directory->getOutputFileBase())    << "\""
+      "tooltip=\""   << directory->briefDescriptionAsTooltip() << "\""
       "];\n";
```
albert-github, line 159: I assume this change has to do with #9467 and it looks a bit out of order for this proposed pull request.
I think it would be better to create a separate pull request for this change.

makelinux: Sure. But please review the rest of PR. I try to merge it for two years!

testing/obsolete
```diff
@@ -0,0 +1,9 @@
+# Usage:
```
doxygen, line 1: is this file needed for something?

doxygen:
@makelinux If you can look at the merge conflict and the remark I put at the `testing/obsolete` file, then it looks good to be merged.

testing/obsolete
makelinux: It is manual test for new code in function `Config::updateObsolete`. It is not connected to test suite. It could be connected latter.

PR #9470, opened 2022-07-22
dotdirdeps: briefDescriptionAsTooltip

### LINKED ISSUE #9467: feature request: directory tooltip contains brief message
For example include dependency graph tooltips already contains text of brief message.

makelinux:
implements  #9467

### DIFF
--- a/src/dotdirdeps.cpp
+++ b/src/dotdirdeps.cpp
@@ -138,6 +138,14 @@ static std::string getDirectoryBorderStyle(const DotDirProperty &property)
   return style;
 }
 
+static TextStream &common_attributes(TextStream &t, const DirDef *const dir, const DotDirProperty &prop)
+{
+  return t <<
+    "style=\""   << getDirectoryBorderStyle(prop) << "\", "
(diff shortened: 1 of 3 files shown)


src/dotdirdeps.cpp
```diff
@@ -155,6 +155,7 @@ static void drawDirectory(TextStream &t, const DirDef *const directory, const Do
       "fillcolor=\"" << getDirectoryBackgroundColor(directory->level()-startLevel)   << "\", "
       "color=\""     << getDirectoryBorderColor(property)                            << "\", "
       "URL=\""       << addHtmlExtensionIfMissing(directory->getOutputFileBase())    << "\""
+      "tooltip=\""   << directory->briefDescriptionAsTooltip() << "\""
```
doxygen, line 158: I would expect a similar operation as is done by`escapeTooltip()` for `DotNode` is needed before writing the result to avoid quotes in the description to cause invalid output.

makelinux: Done

PR #9472, opened 2022-07-23
Placing of the version number in the head template

albert-github:
In the old pull request #8215 it was mentioned as a side note:
> The <!DOCTYPE html attribute should really come first in the document. Perhaps the autogenerated doxygen version line 

comment could come at the end? or second line?
In the "HTML standard" it is written:
> A DOCTYPE is a required preamble.

It is strange that comment cannot be at the beginning of the file though.

The choice is to put the version comment directly inside the `<head>` tag.

### DIFF
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -1132,13 +1132,17 @@ void HtmlGenerator::writeStyleSheetFile(TextStream &t)
 
 void HtmlGenerator::writeHeaderFile(TextStream &t, const QCString & /*cssname*/)
 {
-  t << "<!-- HTML header for doxygen " << getDoxygenVersion() << "-->\n";
-  t << ResourceMgr::instance().getAsString("header.html");
+  QCString head =  ResourceMgr::instance().getAsString("header.html");
+  int endPlace = head.find("<head>") + 7;
+
+  t << head.left(endPlace);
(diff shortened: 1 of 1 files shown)


src/htmlgen.cpp
```diff
@@ -1132,13 +1132,17 @@ void HtmlGenerator::writeStyleSheetFile(TextStream &t)
 
 void HtmlGenerator::writeHeaderFile(TextStream &t, const QCString & /*cssname*/)
 {
-  t << "<!-- HTML header for doxygen " << getDoxygenVersion() << "-->\n";
-  t << ResourceMgr::instance().getAsString("header.html");
+  QCString head =  ResourceMgr::instance().getAsString("header.html");
+  int endPlace = head.find("<head>") + 7;
```
doxygen, line 1136: This look a bit hackish to me. It depends on header.html containing `<head>` and where is the magic number 7 coming from when `<head>` is only 6 characters? To skip over an assumed newline?

Maybe better to just add the comment after the header?

albert-github: I thought about the comment after the header but it would be far down and land after the comment `<!-- end header part -->`.
The header.html is the default doxygen header and this would, I think always, contain the `<head>`  tag on a single line. Hence also the magic number 7 i.e. the `<head>` plus the `\n`.

I can however quite easily change the pull request so the line is a the end, please advice.

Issue #9482, opened 2022-07-31
[regression] incorrect setting of replacement for DOT_FONTNAME and DOT_FONTSIZE


albert-github:
When having, with the current master (1.9.5 (1e5aa121826c53e50b2e290efaa3f8bcd19bfb7c)) the simple `Doxyfile`:
```
DOT_FONTNAME           = Helvetica
DOT_FONTSIZE           = 20
HAVE_DOT = YES
CALL_GRAPH=YES
CALLER_GRAPH=YES
```
we get with the command `doxygen -x`: 
```
warning: Tag 'DOT_FONTNAME' at line 1 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'DOT_FONTSIZE' at line 2 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
# Difference with default Doxyfile 1.9.5 (1e5aa121826c53e50b2e290efaa3f8bcd19bfb7c)
HAVE_DOT               = YES
DOT_COMMON_ATTR        = "fontname=Helvetica fontsize=10 fontname=Helvetica fontsize=20"
DOT_EDGE_ATTR          = "labelfontname=Helvetica labelfontsize=10 labelfontname=Helvetica labelfontsize=20"
CALL_GRAPH             = YES
CALLER_GRAPH           = YES
```
The both warnings are correct, but the duplication is not as the wrong, in this case size, is chosen.
This is nicely illustrated with the simple source file:
```
/// \file

/// \brief first fie
/// \details first fie
int fie_1()
{
  fie_2();
}

/// \brief second fie
/// \details second fie
int fie_2()
{
}
```

With the doxygen 1.9.4 version we see:
![image](https://user-images.githubusercontent.com/5223533/182019291-9c302475-475f-4e49-a12d-1509a7cccb6d.png)

and with the mentioned master version we get:
![image](https://user-images.githubusercontent.com/5223533/182019306-817a9542-34bb-4fe3-97a6-d635364a851d.png)

Note with the 1.9.4. version the size is 20 and with the current master 10.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9228553/example.tar.gz)

albert-github:
@makelinux regression due to #9463

doxygen:
[...]

PR #9483, opened 2022-07-31
remove fontsize from default DOT_NODE_ATTR

makelinux:
because it is already defined in DOT_COMMON_ATTR

(fixes) #9482

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3645,7 +3645,7 @@ to be found in the default search path.
       </docs>
     </option>
     <option type='string' id='DOT_NODE_ATTR' format='string' depends='HAVE_DOT'
-      defval='fontsize=10 shape=box height=0.2 width=0.4'>
+      defval='shape=box height=0.2 width=0.4'>
       <docs>
 <![CDATA[
  \c DOT_NODE_ATTR is concatenated with \ref cfg_dot_common_attr "DOT_COMMON_ATTR".


albert-github:
The usage of the double `fontsize=10` in the definitions should indeed be prevented.
The problem seems to be solved, though I don't like the fact that that when using `doxn -x` I still see:
```
DOT_COMMON_ATTR        = "fontname=Helvetica fontsize=10 fontname=Helvetica fontsize=20"
DOT_EDGE_ATTR          = "labelfontname=Helvetica labelfontsize=10 labelfontname=Helvetica labelfontsize=20"
```
in the result (same problem with `doxygen -u`).

makelinux:
OK. I have draft of the solution.

makelinux:
> The usage of the double `fontsize=10` in the definitions should indeed be prevented. The problem seems to be solved, though I don't like the fact that that when using `doxn -x` I still see:
> 
> ```
> DOT_COMMON_ATTR        = "fontname=Helvetica fontsize=10 fontname=Helvetica fontsize=20"
> DOT_EDGE_ATTR          = "labelfontname=Helvetica labelfontsize=10 labelfontname=Helvetica labelfontsize=20"
> ```
> 
> in the result (same problem with `doxygen -u`).

@albert-github , please see #9484

doxygen:
@makelinux This fix is already included in commit https://github.com/doxygen/doxygen/commit/a44f6a7e5caf1fbf5fe271c99046a51b315d5f56

PR #9484, opened 2022-07-31
DotAttributes based on map

makelinux:
See commits

### DIFF
--- a/src/configimpl.l
+++ b/src/configimpl.l
@@ -43,6 +43,7 @@
 #include "fileinfo.h"
 #include "dir.h"
 #include "textstream.h"
+#include "dotgraph.h"
 
 #define YY_NO_INPUT 1
 #define YY_NO_UNISTD_H 1
@@ -2064,13 +2065,11 @@ void Config::checkAndCorrect(bool quiet, const bool check)
 
 }
 
-static void appendAttribute(QCString& a, const char * name, ConfigObsolete* value)
+static void appendAttribute(DotAttributes& attr, QCString name, ConfigObsolete* value)
 {
   if (!value || !value->valueStringRef()->length())
     return;
-  if (a.length())
-    a += ' ';
-  a += name + QCString("=") + value->valueStringRef()->quoted();
+  attr[name] = *value->valueStringRef();
 }
 
 void Config::updateObsolete()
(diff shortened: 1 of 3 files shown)


src/configimpl.l
```diff
@@ -2064,13 +2065,11 @@ void Config::checkAndCorrect(bool quiet, const bool check)
 
 }
 
-static void appendAttribute(QCString& a, const char * name, ConfigObsolete* value)
+static void appendAttribute(DotAttributes& attr, QCString name, ConfigObsolete* value)
 {
   if (!value || !value->valueStringRef()->length())
     return;
```
albert-github, line 2071: Probably better to have:
```
if (!value || value->valueStringRef()->isEmpty()) return;
```
so that when accidentally there is just  just white space in the string it will be caught as well.
(I just noticed this now).

src/dotgraph.cpp
```diff
@@ -103,6 +104,38 @@ static bool insertMapFile(TextStream &out,const QCString &mapFile,
 
 //--------------------------------------------------------------------
 
+DotAttributes::DotAttributes(QCString input)
+{
+  std::string tag = R"(\s*(\w+)\s*)";
+  std::string unquoted = R"(\s*([^\s,"]+)\s*)";
+  std::string quoted = R"(\s*\"([^"]*)\"\s*)";
```
albert-github, line 111: - How will a string like `"text1\"text2"` be handled?
- Is the special character `_` possible in the input string?

makelinux: unmatched text will be assigned to key `_unmatched_`

makelinux: underscore is possible.

makelinux: strings with quotes inside are not supported

doxygen:
@makelinux @albert-github According to https://graphviz.org/doc/info/lang.html escapes in quotes strings are supported. I reworked the code a bit, please check if it is ok like this.

makelinux:
> @makelinux @albert-github According to https://graphviz.org/doc/info/lang.html escapes in quotes strings are supported. I reworked the code a bit, please check if it is ok like this.

regex for quoted string with escapes: `string quoted = R"(\s*\"((?:[^"\\]|\\.)*)\"\s*)";`

According specs attribute separator is not just comma.

Conventional C/C++ spacing after comas and around operators would be nice.

`std::regex_replace` can be used to change an attribute.

doxygen:
@makelinux I'd like to avoid `std::regex` after finding out the hard way it really slowed down doxygen significantly (not really a problem for your case, but better keep things consistent). You are right about the separators; I've fixed that with this commit https://github.com/doxygen/doxygen/commit/6f020734520f2945f8f2821deb78ee0beb7c0c62. Not sure what you mean with "conventional C/C++ spacing".

makelinux:
@doxygen , have you any presences of `clang-format` configuration or another source code formatting tool?

doxygen:
@makelinux I looked at `clang-format` but I cannot configure it in such a way it does not introduce a lot of spaces in expressions. Seems no way to control this behavior (see this question https://stackoverflow.com/questions/67245250/can-i-tell-clang-format-to-not-add-space-around-arithmetic-operators).

makelinux:
> @makelinux I looked at `clang-format` but I cannot configure it in such a way it does not introduce a lot of spaces in expressions. Seems no way to control this behavior (see this question https://stackoverflow.com/questions/67245250/can-i-tell-clang-format-to-not-add-space-around-arithmetic-operators).

Actually spaces around arithmetic operations and after commas are common standard practices for readability.

PR #9488, opened 2022-08-01
[regression] bug 786576 regression with ALIASES or image latex interpretation

albert-github:
With the CGAL Program we got:
```
doc/Arrangement_on_surface_2/Arrangement_on_surface_2.txt:2779: warning: Unexpected token TK_HTMLTAG found as part of a image
```
as the code (shortened) was like:
```
<table border=0>
<tr>
<td>\image html sphere.png</td>
<tr align="center">
<td>(a) Sphere</td>
</tr>
</table>
```
the problem is the `</td>` as this is now not seen properly (in the past it was ignored!).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9234259/example.tar.gz)

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -965,6 +965,7 @@ void DocParser::defaultHandleTitleAndSize(const int cmd, DocNodeVariant *parent,
     }
     else if (tok==TK_HTMLTAG)
     {
+      tokenizer.unputString(context.token->name);
       break;
     }
     if (!defaultHandleToken(parent,tok,children))
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -963,6 +963,10 @@ void DocParser::defaultHandleTitleAndSize(const int cmd, DocNodeVariant *parent,
       // special case: no title, but we do have a size indicator
       break;
     }
+    else if (tok==TK_HTMLTAG)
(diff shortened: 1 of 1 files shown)


albert-github:
This is a regression on #9477

src/docparser.cpp
```diff
@@ -963,6 +963,10 @@ void DocParser::defaultHandleTitleAndSize(const int cmd, DocNodeVariant *parent,
       // special case: no title, but we do have a size indicator
       break;
     }
+    else if (tok==TK_HTMLTAG)
+    {
+      break;
```
doxygen, line 968: should we no unput the string like is done below? For a closing tag is it probably ok to ignore it, but an opening tag should not be ignored, like in the following example:

```html
<table border=0>
<tr>
<td>\image html sphere.png<tr align="center">
<td>(a) Sphere</td>
</tr>
</table>
```

albert-github: Correct, none of the tags should be ignored. I looked like it was working after the warning was gone but this was luck.

Code was inspired by the "height=/width=" above, but that is a different situation where later on the part after the part after the `=` is processed and there there is no unput necessary. 

The pull request has been updated.

PR #9493, opened 2022-08-06
bug 564252 HTML output for pure virtual function with "throws()" hint is wrong

albert-github:
The parts after the closing round bracket of a function definition are stored in multiple places, so in case of
```
virtual myclass() const trow() = 0;
```
throw part is stored in a separate string and the `cons t` and `=0` part are stored in another string but concatenated, during output this was not handled properly, now the `=0` part is stored separately.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9275155/example.tar.gz)


doxygen:
@albert-github I've come up with a more local solution, see https://github.com/doxygen/doxygen/commit/09edf55a01f8889698b56fda264744f3172b7d81. Let me know if I missed anything.

albert-github:
At first glance looks OK

Issue #9502, opened 2022-08-11
Fortran containing a single apostrophe (') in comments interferes with Doxygen parsing.


brtnfld:
**Describe the bug**
Using an apostrophe (') in the comments causes the #ifdef to be treated as comments.

**Expected behavior**
#ifdef should not be treated as a comment.

**To Reproduce**

TEST.F90

```
!> @ingroup TEST
!!
!! @brief This user's problem. 
!!
MODULE TEST

#ifdef DOXYGEN_FORTRAN
  INTERFACE DEF
    MODULE PROCEDURE DEFI
    MODULE PROCEDURE DEFR
  END INTERFACE
#else
  INTERFACE NODEF
    MODULE PROCEDURE NODEFI
    MODULE PROCEDURE NODEFR
  END INTERFACE
#endif

CONTAINS

END MODULE TEST
```
This code will result in both NODEF and DEF being processed, regardless of whether DOXYGEN_FORTRAN is defined.

**Version**
I used 1.8.14, but I tried 1.9.5, and it has the same issue.

**Additional context**
The workaround is to use ```&apos; ``` instead of (').

albert-github:
Besides the mentioned case there are also the cases:
- comment after a statement
- fixed format code comment

I've just pushed a proposed patch, pull request #9503

albert-github:
I've just pushed a proposed patch, pull request #12280 in favor of the proposed PR #9503 (to get a proper merge in git)

albert-github:
- Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in Doxygen version 1.18.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed, and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9503, opened 2022-08-12
issue #9502 Fortran containing a single apostrophe (') in comments interferes with Doxygen parsing.

albert-github:
In the preprocessor we have to ignore Fortran comment not only at the beginning of the line but also after a statement like `INTEGER :: I !< The user's comment`.
Furthermore in Fortran fixed format code also lines that have `C`, `c` or `*` at the first position should be skipped (see also util.cpp and commentcnv.l).


albert-github:
Obsoleted by #12280  (to get a proper merge).

PR #9510, opened 2022-08-15
Fortran private/public statements

melven:
Improvements for handling `public` and `private` in Fortran.

### DIFF
--- a/src/fortranscanner.l
+++ b/src/fortranscanner.l
@@ -291,7 +291,7 @@ ATTR_SPEC (EXTERNAL|ALLOCATABLE|DIMENSION{ARGS}|{INTENT_SPEC}|INTRINSIC|OPTIONAL
 ACCESS_SPEC (PRIVATE|PUBLIC)
 LANGUAGE_BIND_SPEC BIND{BS}"("{BS}C{BS}((,{BS}NAME{BS}"="{BS}"\""(.*)"\""{BS})|(,{BS}NAME{BS}"="{BS}"'"(.*)"'"{BS}))?")"
 /* Assume that attribute statements are almost the same as attributes. */
-ATTR_STMT {ATTR_SPEC}|DIMENSION|{ACCESS_SPEC}
+ATTR_STMT {ATTR_SPEC}|DIMENSION
 EXTERNAL_STMT (EXTERNAL)
 
 CONTAINS  CONTAINS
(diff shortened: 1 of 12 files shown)


melven:
Hi, this is one pull request for a subset of changes I made in 2016 to improve Fortran support.
Took me some time to rebase this on the current master.

Please review - I have more changes for future pull requests and would really like to see better Fortran support in doxygen ;)

src/fortranscanner.l
```diff
@@ -899,8 +914,23 @@ private                                 {
...
+                                          yy_pop_state(yyscanner);
+                                          int outerState = YY_START;
+                                          yy_push_state(currentState,yyscanner);
+                                          if( outerState == Start || outerState == ModuleBody )
+                                          {
+                                            if ((yyextra->current_root) && 
+                                                (yyextra->current_root->section == Entry::CLASS_SEC
+                                                 || yyextra->current_root->section == Entry::NAMESPACE_SEC))
+                                            {
+                                              name = yyextra->current_root->name + "::" + name;
```
albert-github, line 929: Is it guaranteed that `yyextra->current_root->name` is lowercase otherwise it is might be better to "lower()" it as well.

src/fortranscanner.l
```diff
@@ -672,11 +675,9 @@ extends{ARGS}                           {
                                         }
 public                                  {
                                           yyextra->current->protection = Public;
-                                          yyextra->typeProtection = Public;
                                         }
```
albert-github, line 678: Why is the type protection removed here?

src/fortranscanner.l
```diff
@@ -672,11 +675,9 @@ extends{ARGS}                           {
...
 private                                 {
                                           yyextra->current->protection = Private;
-                                          yyextra->typeProtection = Private;
```
albert-github, line 679: Why is the type protection removed here?

src/fortranscanner.l
```diff
@@ -654,8 +657,8 @@ SCOPENAME ({ID}{BS}"::"{BS})*
                                           }
 
                                           yy_push_state(Typedef,yyscanner);
-                                          yyextra->current->protection = yyextra->defaultProtection;
-                                          yyextra->typeProtection = yyextra->defaultProtection;
+                                          yyextra->current->protection = Package; // invalid in Fortran, replaced below
+                                          yyextra->typeProtection = Public;
```
albert-github, line 661: On which Fortran standard based rule the protection is set to "Public", e.g. module variables are by default Private

albert-github (commented):
It is hard to tell whether all paths for the new constructs (i.e. for types) are run through correctly.

The routine `addModule` can probably do with a `lower()` on the `name` as here the `yytext` (see lines 630 and 634).

src/fortranscanner.l
melven: Default in Fortran is public, from the [F2003 handbook]((https://doi.org/10.1007/978-1-84628-746-6):
> If the accessibility of a particular module entity is not explicitly specified, the moduleʹs default accessibility applies to that > entity. The moduleʹs default accessibility is
> specified by the accessibility statement without an access-id list; if there is no such
> statement, the default accessibility is PUBLIC

albert-github: From head I thought that I remembered that the default accessibility in Fortran was PRIVATE though when I look in the Fortran 2018 standard  I see that it is PUBLIC>
(Been working lately to much with C++ where the default, for a class, is private.)

src/fortranscanner.l
melven: I put a `name.lower().str()` in the line below - this seems more consistent with the rest of the code.

src/fortranscanner.l
melven: This only refers to the visibility of the type itself (thats the line `type, public :: T1`) and does not change the default for the member variables of the type... I've added a case in the test to check that.

melven: (but I'll have a look at the standard again)

src/fortranscanner.l
melven: Same as above, this only refers to the type itself, not its contained members...

src/fortranscanner.l
albert-github: Is a bit against my feeling.
gfortran can have a bug but it is a good indication.
- Which version of gfortran did you use to test?

melven: Now that you asked: same behavior for gfortran 9.4, 12.1 and current ifort.

albert-github: Thanks, gives some more confidence.

PR #9511, opened 2022-08-15
Test Fortran inheritance over modules

melven:
Remark: there are no changes in C++ or Fortran behavior - just two unit
tests with similar code in C++ and Fortran.


melven:
Hi,

I had some old code from 2016 that improved inheritance in Fortran. This now already works correctly with the current master.

Perhaps, the test cases might still be useful - what do you think?

PR #9514, opened 2022-08-15
Fix build on GNU/Hurd

sthibaul:
There is no path length limitation there, even via pathconf.  But glibc
provides a getcwd function that allocates the buffer dynamically so we can
just leverage that.

### DIFF
--- a/filesystem/filesystem.hpp
+++ b/filesystem/filesystem.hpp
@@ -4081,6 +4081,13 @@ GHC_INLINE path current_path(std::error_code& ec)
         return path();
     }
     return path(std::wstring(buffer.get()), path::native_format);
+#elif defined(__GLIBC__)
+    std::unique_ptr<char, decltype(&std::free)> buffer { ::getcwd(NULL, 0), std::free };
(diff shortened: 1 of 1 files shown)


albert-github:
Thanks for pointing this out, though the file  `filesystem.hpp` is an imported function from the https://github.com/gulrak/filesystem project (i.e. the file: https://github.com/gulrak/filesystem/blob/master/include/ghc/filesystem.hpp), so the problem should be reported upstream at the filseystem project as well.

sthibaul:
Ok, submitted on https://github.com/gulrak/filesystem/pull/154

albert-github:
I see that the fix https://github.com/gulrak/filesystem/pull/154 has been integrated in the master version on https://github.com/gulrak/
I just created an issue about the plans for a new release https://github.com/gulrak/filesystem/issues/164

albert-github:
I see that the new release of filesystem is out for some time and that it has been integrated in the doxygen master as well.

doxygen:
which means it should be fixed in doxygen version 1.9.7.

PR #9532, opened 2022-08-22
Extend `\showdate` with time possibilities

albert-github:
Extend the `\showdate` command with the possibility to specify the time

### DIFF
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -1426,10 +1426,10 @@ contains \c TEST, or \c DEV
   for an example.
 
 <hr>
-\section cmdshowdate \\showdate "<format>" [ <date> ]
+\section cmdshowdate \\showdate "<format>" [ <date_time> ]
 
   \addindex \\showdate
-  Shows the date based on the given \<format\> and \<date\>. Where the \<format\> is a string in which the following tokens have a special meaning:
+  Shows the date and time based on the given \<format\> and \<date_time\>. Where the \<format\> is a string in which the following tokens have a special meaning:
   | Code | Description |
   | :--- | :---------- |
   | \%y | The last two decimal digits of the year. If the result is a single digit it is prefixed by 0.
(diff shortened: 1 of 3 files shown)


doc/commands.doc
```diff
@@ -1452,34 +1452,58 @@ contains \c TEST, or \c DEV
...
   - \showdate "%a %d-%m-%y" 14-3-2015
   - \showdate "%m.%d%y" 14-3-2015
+  - \showdate "%lA %0d-%0m-%Y %H:%M:%S" 14-3-2015 03:04:15
+  - \showdate "%lA %0d-%0m-%Y %H:%M:%S" 14-3-2015 03:04
   \endcode
   In case `OUTPUT_LANGUAGE=english` this results in:
   - Saturday 14-03-2015
   - Sat 14-3-15
   - 3.1415
+  - Saturday 14-03-15 3:4:15
```
doxygen, line 1498: This is a strange representation of 24h time, I would expect `03:04:15` (so just like the input format).

doc/commands.doc
```diff
@@ -1452,34 +1452,58 @@ contains \c TEST, or \c DEV
...
+  - Saturday 14-03-15 3:4:0
```
doxygen, line 1499: why the 0 seconds? this is not according to the time format given.

doc/commands.doc
albert-github: It is but the choice of the user is `%H:%M:%S` but might have been better to use `%H:%0M:%S`

doc/commands.doc
albert-github: This is the default value used when not specified and the user specified `%H:%M:%S` as format so including seconds.

doc/commands.doc
doxygen: I assume we want to follow e.g. strftime (https://man7.org/linux/man-pages/man3/strftime.3.html) so `%M` has range `00` to `59`, so with leading 0. There is no need for a version without leading 0 in my opinion.

doc/commands.doc
doxygen: ok, I see. 

Still a bit inconsistent as this seems to be the only value that when omitted in the format is replaced by a 0. Maybe it should just be an error, i.e. either omit the `:%S` or specify the seconds.

doc/commands.doc
albert-github: I tried to stay close to the definition as used e.g. for the month number in the `\showdate` command
>  | \%m | The month as a decimal number.
>  | \%0m | The month as a decimal number. If the result is a single digit, it is prefixed with 0.

The definition used in the strftime for month and hour are:

>        │%m  │                                             │
>        │    │        The month number, formatted with two │
>        │    │        digits (from `01' to `12'). [tm_mon] │

and

>        │%H  │                                             │
>        │    │        The hour (on a 24-hour clock),       │
>        │    │        formatted with two digits (from `00' │
>        │    │        to `23'). [tm_hour]

and with the date command and on https://cplusplus.com/reference/ctime/strftime/
> %m     month (01..12)
and
>  %H     hour (00..23)

So the initial implementation is also not following this definition.
With the current `\showdate` definitions it is possible have (when necessary) just e.g. a minutes or month value between 0 and 9 but also the values between 00 and 09.

doc/commands.doc
albert-github: It is the choice of the user in this case to have an output format the specifies to output the seconds but the user did not specify this value in its input.

doxygen: We can also conclude that what the user has specified is incomplete/wrong and give an error? 

The put it another way, if the user doesn't specify the minutes value should we not also fill in 0 there? Same for hour? i.e. why is the seconds value treated in a special way? Is there a use case for it?

doc/commands.doc
doxygen: I think we should try to follow standards where possible. Maybe this one https://strftime.org/ that is used for Python can be followed. It has `%m` (standard) and the `%-m` extension for the non-zero padded variant.

doc/commands.doc
albert-github: > We can also conclude that what the user has specified is incomplete/wrong and give an error?

could be concluded and can be implemented but will be a bit of work.

> The put it another way, if the user doesn't specify the minutes value should we not also fill in 0 there? Same for hour? i.e. why is the seconds value treated in a special way? Is there a use case for it?

It is more a gut feeling that the user is not interested in the seconds value / doesn't can / will specify the seconds but will specify the hour and minute.

doc/commands.doc
albert-github: I think it would be possible to follow this "standard", I can adjust it in that way.
- Should I change it?

doxygen: I can also do it as part of the merge. Just let me know what you prefer

albert-github: Please do it as part of a merge.

PR #9542, opened 2022-08-27
Incorrect version number in $(VERSION) and thus documentation

albert-github:
When the doxygen VERSION number is changed there are no changes to other configuration entities and one builds the doxygen documentation locally the old version number is still present.

Reference regarding the solution:
- https://stackoverflow.com/questions/73511087/getting-environment-variable-during-runtime-from-file-in-cmake
- https://stackoverflow.com/questions/24246037/how-to-make-cmake-reconfiguration-depend-on-custom-file

(happened in my cloned repositories after a release as the rest was already up to date)


albert-github:
Looks like still a small problem, investigating...

albert-github:
Looks like Murphy was hitting quite well.

There is a more modern approach, so this pull request is superseded by #9543

PR #9551, opened 2022-08-29
Dot uml class diagram fix

petyovsky:
# Bugfix of malformed UML class diagram generated thru graphviz.

Recently was introduced new doxygen's config key: `DOT_NODE_ATTR` with default value: `shape=box, ...`.

Due this Doxygen generate malformed UML inheritance and collaboration diagrams,
because shape type `box` isn't Graphviz's `Record-based node`.
Only Graphviz's shape types: `record` and `Mrecord` supports `record-based label`. See [dot User's Manual](https://www.graphviz.org/pdf/dotguide.pdf), Ch. 21, p. 6.
Record-based shape represent recursive lists of fields, which are drawn as alternating horizontal and vertical rows of boxes.
Special characters are: Literal braces, vertical bars and angle brackets.

Patch set shape type to `shape=record` which is the record-based but only for UML inheritance and collaboration diagrams.

In the future, it might be a good idea to introduce a new doxygen configuration key: `DOT_UML_NODE_ATTR` with default value: `shape=record, ...`,
which will also allow the user to choose the `Mrecord` shape for the UML diagrams of inheritance and collaboration.


petyovsky:
doxygen_dot_class_inher_coll_diagram_bug

PR #9553, opened 2022-08-29
Vhdl process flowchart bugfix

petyovsky:
# Bugfix of malformed VHDL process flowchart generated thru graphviz.

Patch fixed bug in generating graphviz's source describing the VHDL process flowchart.
The bug manifested itself as the generation of empty nodes in the VHDL process flowchart.

Solution is to ignore the empty section of the VHDL variable definition in the method `FlowChart::addFlowChart()`.
This is section between `process` and `begin` keywords, where any source text is missing, probably a bug in the VHDL source parser.


petyovsky:
Before:
doxygen_dot_vhdl_flowchart_before

After:
doxygen_dot_vhdl_flowchart_after

PR #9556, opened 2022-08-30
Extending internal documentation with some version information

albert-github:
Enhancing the doxygen internal documentation with overviews of:
- release dates
- versions where special commands entered doxygen
- versions where  settings have entered / become obsolete in doxygen

Adding a small description of the doxygen internal commands

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -111,7 +111,12 @@ WARN_LOGFILE           =
 #---------------------------------------------------------------------------
 # Configuration options related to the input files
 #---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


doxygen:
@albert-github Nice work! Did you create a script for this?

albert-github:
In the past I created some very bad looking scripts for this and I store the intermediate results from all versions on my disk.
The scrips also changed in time as the information needed to be obtained from other places (e.g. the settings commands now are all nicely located in the config.xml file, but in the past I had to get it from config.l)

PR #9555, opened 2022-08-30
Better VHDL translation support

petyovsky:
Added translation specialization for VHDL in the `trMemberFunctionDocumentation()` method, (see method body with `OPTIMIZE_OUTPUT_VHDL` condition in the `translator_en.h` file).

This VHDL specialization string `Member Function/Procedure/Process Documentation` is based on the translation string in the `trFunctionAndProc()` method.

Added the possibility to translate the `flowchart: ` string in `htmldocvisitor.cpp` and todo remark describing this here was removed.
Appended translation section 1.9.6 to the `translator_en.h` file.
Flowchart string added as new method `trFlowchart()` to this section and string was changed to: `Flowchart: `.

Added `class TranslatorAdapter_1_9_5` to the file `translator_adapter.h`.
Changed dependency of up-to-date translations: `br`, `cz`, `nl`, `pt` to `TranslatorAdapter_1_9_5`.

I would like to ask the maintainers to check that I have made all the changes to the files correctly and have not messed anything up. Thank you.

PR #9563, opened 2022-09-01
Check for unknown encoding of source file

albert-github:
Based on the discussion with PR #9555 checking for the file encoding

PR #9564, opened 2022-09-01
Check for unknown encoding of source file

albert-github:
Based on the discussion with PR #9555 checking for the file encoding

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -66,6 +66,18 @@ jobs:
     - name: Checkout doxygen
       uses: actions/checkout@v3
       
+    - name: Check file types
+      shell: bash
+      run: |
+        export result=`find . | xargs file | grep -v "  ASCII text" | grep -v "  BibTeX text file, ASCII text" | grep -v "  C source, ASCII text" | grep -v "  C source, UTF-8 Unicode text" | grep -v "  C++ source, ASCII text" | grep -v "  C++ source, UTF-8 Unicode text" | grep -v "  FIG image text, version 3.2, ASCII text" | grep -v "  Git index, version 2" | grep -v "  Git pack index, version 2" | grep -v "Git pack, version 2" | grep -v "  HTML document, ASCII text" | grep -v "  HTML document, UTF-8 Unicode text" | grep -v "  LaTeX 2e document, ASCII text" | grep -v "  LaTeX document, ASCII text" | grep -v "  LaTeX document, UTF-8 Unicode text" | grep -v "  LaTeX table of contents, ASCII text" | grep -v "  MS Windows icon resource" | grep -v "  PDF document, version 1.5" | grep -v "  PHP script, ASCII text" | grep -v "  PNG image data" | grep -v "  POSIX shell script, ASCII text executable" | grep -v "  Perl script text executable" | grep -v "  PostScript document text conforming DSC level" | grep -v "  Python script, ASCII text executable" | grep -v "  Qt Resource Collection file" | grep -v "  ReStructuredText file, ASCII text" | grep -v "  SVG Scalable Vector Graphics image" | grep -v "  UTF-8 Unicode text" | grep -v "  UTF-8 Unicode text, with very long lines" | grep -v "  XML 1.0 document text" | grep -v " XML 1.0 document, ASCII text" | grep -v "  XML 1.0 document, UTF-8 Unicode text" | grep -v "  current ar archive" | grep -v "  directory" | grep -v "  empty" | grep -v "  exported SGML document, ASCII text" | grep -v "  makefile script, ASCII text" | grep -v "  troff or preprocessor input, ASCII text" | wc -l`
+        if [ $result -ne 0 ]; then
+          echo "Found files with unknown encoding!"
+          echo "=================================="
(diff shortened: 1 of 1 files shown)


.github/workflows/build_cmake.yml
```diff
@@ -66,6 +66,18 @@ jobs:
...
... HTML document, ASCII text" | grep -v "  HTML document, UTF-8 Unicode text" | grep -v "  LaTeX 2e document, ASCII text" | grep -v "  LaTeX document, ASCII text" | grep -v "  LaTeX document, UTF-8 Unicode text" | grep -v "  LaTeX table of contents, ASCII text" | grep -v "  MS Windows icon resource" | grep -v "  PDF document, version 1.5" | grep -v "  PHP script, ASCII text" | grep -v "  PNG image data" | grep -v "  POSIX shell script, ASCII text executable" | grep -v "  Perl script text executable" | grep -v "  PostScript document text conforming DSC level" | grep -v "  Python script, ASCII text executable" | grep -v "  Qt Resource Collection file" | grep -v "  ReStructuredText file, ASCII text" | grep -v "  SVG Scalable Vector Graphics image" | grep -v "  UTF-8 Unicode text" | grep -v "  UTF-8 Unicode text, with very long lines" | grep -v "  XML 1.0 document text" | grep -v " XML 1.0 document, ASCII text" | grep -v "  XML 1.0 document, UTF-8 Unicode text" | grep -v "  current ar archive" | grep -v "  directory" | grep -v "  empty" | grep -v "  exported SGML document, ASCII text" | grep -v "  makefile script, ASCII text" | grep -v "  troff or preprocessor input, ASCII text" | wc -l`
```
doxygen, line 72: there must a less convoluted way to check this. I don't think this is very maintainable/portable.

albert-github: I agree that is is a bit of a brute force attack and might be a bit heavy on maintenance though I didn't see any way to properly filter out the files and be prepared for future file types (that might have to be accessed and added).
I don't expect that it often will happen that new file types are added.
For this reason too I, for the time being, only added it to just one test build.

doxygen: @albert-github but problem was the unexpected BOM marker, right? Isn't there a more direct way to check for it, i.e. like shown here: https://stackoverflow.com/a/65841914/784672

albert-github: The original problem was the BOM marker but there are other encodings that I think that can cause problems as well. During my tests I set some files to other encodings (set the encoding with Notepad++) as well:
```
./src/translator_en.h:                                                  C++ source, UTF-8 Unicode (with BOM) text
./src/translator_cz.h:                                                  C++ source, UTF-8 Unicode (with BOM) text
./src/translator_br.h:                                                  C++ source, Little-endian UTF-16 Unicode text
./src/translator_pt.h:                                                  C++ source, Big-endian UTF-16 Unicode text
```
I didn't test whether or not, in this case e.g. the Brazilian and Portuguese translation gave problems as well.
I didn't test the `UTF-32 LE` and `UTF-32 BE` as it looks like, my version of, Notepad++ doesn't offer these encodings directly.

Therefore I decided to use the current implementation.

doxygen: @albert-github Using the same approach as you did, but with a slightly less explict grep that combines all patterns, I come to this:

```sh
find . -type f | xargs file | grep -E -v "(ASCII text|UTF-8 Unicode text|PNG|SVG|PDF|XML|ar archive|PostScript|empty|Qt Resource|MS Windows)"
```
which seems to pass the test on the 1.9.5 release and should fail for any of the files above.

Would this work?

albert-github: I did a quick test in the original PR #9563 (see the github action: https://github.com/doxygen/doxygen/runs/8177752269?check_suite_focus=true#step:3:30) and got the warnings:
```
./.git/objects/pack/pack-812c70f1aa167a15c50526183d6ae4a07f5fc3bc.pack: Git pack, version 2, 1197 objects
./.git/objects/pack/pack-812c70f1aa167a15c50526183d6ae4a07f5fc3bc.idx:  Git pack index, version 2
./.git/hooks/fsmonitor-watchman.sample:                                 Perl script text executable
./.git/index:                                                           Git index, version 2, 1068 entries
./src/translator_en.h:                                                  C++ source, UTF-8 Unicode (with BOM) text
./src/translator_cz.h:                                                  C++ source, UTF-8 Unicode (with BOM) text
./src/translator_br.h:                                                  C++ source, Little-endian UTF-16 Unicode text
./src/translator_pt.h:                                                  C++ source, Big-endian UTF-16 Unicode text
```
the last 4 are correct, the first 3 aren't and should also have been removed.

For the current set of files (besides the 3 files mentioned above) it looks like to work but I'm not sure that it will work for other, now unknown, file types. One can never know but in the original constellation I think there is a better chance due to the longer strings.

doxygen: The hidden directories should be excluded of course:

```sh
find . -not -path '*/.*' -type f | xargs file | grep -E -v "(ASCII text|UTF-8 Unicode text|PNG|SVG|PDF|XML|ar archive|PostScript|empty|Qt Resource|MS Windows)"
```

Longer strings do require more maintenance, so I still prefer the compact version.

albert-github: Looks a good test.
The difference is that I think that strings line "ASCII text" and "UTF-8 Unicode text" might be a bit general texts and lead to overseeing some problems so I chose for the longer versions.

PR #9565, opened 2022-09-02
Warning during compilation on Windows systems

albert-github:
Fixing a (first set) of warnings on Windows systems as found by means of setting `/Wall`

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -503,17 +503,13 @@ class LayoutParser
           QCString()
         }
       };
-      LayoutNavEntry::Kind kind = static_cast<LayoutNavEntry::Kind>(0);
       // find type in the table
       int i=0;
       QCString type = XMLHandlers::value(attrib,"type");
       while (mapping[i].typeStr)
       {
         if (mapping[i].typeStr==type)
-        {
-          kind = mapping[i].kind;
           break;
-        }
         i++;
       }
       if (mapping[i].typeStr==0)
@@ -530,6 +526,7 @@ class LayoutParser
         m_invalidEntry=TRUE;
         return;
       }
+      LayoutNavEntry::Kind kind = mapping[i].kind;
       QCString baseFile = mapping[i].baseFile;
       QCString title = XMLHandlers::value(attrib,"title");
       bool isVisible = elemIsVisible(attrib) && parentIsVisible(m_rootNav);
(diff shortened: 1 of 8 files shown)


src/xmldocvisitor.cpp
```diff
@@ -926,11 +926,11 @@ void XmlDocVisitor::operator()(const DocImage &img)
                 altValue, img.isInlineImage());
 
   // copy the image to the output dir
-  FileDef *fd;
+  FileDef *fd = 0;
   bool ambig;
   if (url.isEmpty() && (fd=findFileDef(Doxygen::imageNameLinkedMap,img.name(),ambig)))
   {
-    copyFile(fd->absFilePath(),Config_getString(XML_OUTPUT)+"/"+baseName);
+    if (fd) copyFile(fd->absFilePath(),Config_getString(XML_OUTPUT)+"/"+baseName);
```
doxygen, line 933: this extra `if (fd)` should not be needed, `fd` is already checked in the expression on the previous line.

src/layout.cpp
```diff
@@ -503,7 +503,7 @@ class LayoutParser
           QCString()
         }
       };
-      LayoutNavEntry::Kind kind;
+      LayoutNavEntry::Kind kind = static_cast<LayoutNavEntry::Kind>(0);
```
doxygen, line 506: this is not an improvement in my opinion. I think it is better to keep the enum uninitialized and then add a warning plus early return with an error on line 518 if line 514 was never executed.

src/xmldocvisitor.cpp
albert-github: Indeed apparently `File *fd = 0` already removed the warning.

src/layout.cpp
albert-github: This is a bit a complex situation.
- the warning is already present from the block starting in line 519
- in line 499 we see `static_cast<LayoutNavEntry::Kind>(0),` 
- in line 506 the variable was not initialized, so what would have been the best guess? I chose 0 as above in line 499 (actually any value would do).
- we could introduce a `LayoutNavEntry::Unknown` but this might cause problems / warnings in other places (like defaults in switch statements). I'm not afraid that `LayoutNavEntry::Unknown`` would be used as this would have been caught by the test in line 519.

albert-github: Looks like I found a relatively easy solution.

doxygen: 👍

PR #9577, opened 2022-09-05
Update translator_fr.h

tititiou36:
Fix some HTML syntax:
   - one closing </p>
   - a <p> .. </p>
This has apparently no impact on the generated .html files, but aligns the French doc on the reference one.


albert-github:
The issue description looks like a bit truncated, can you update this here on github?

The changes have indeed no effect on the output as doxygen will add these (in this case) automatically.
We see that in the English translation they are present as well, but in the Dutch translation they are not present either.

tititiou36:
My understanding of https://www.doxygen.nl/manual/langhowto.html is that the English version is the reference.
My goal was to reduce to the minimum the diff output between fr and en.
A the result seems unchanged, feel free to ignore the patch.
I just noticed it, so I proposed a fix, nothing more, nothing less.

PR #9581, opened 2022-09-06
bug 609517  wrong output with optional names for bitfields in C structs

albert-github:
Specifically handle the bit-field case where the name field contains `int`, `char`, `long`, `short` as they would be types.


albert-github:
Looks like that this issue is still present (revisited based on #9905)

PR #9590, opened 2022-09-11
Portuguese translators updated to 1.9.6.

fjtc:
Portuguese translators updated to 1.9.6.


albert-github:
From the differences, on Github, it looks like the files is saved in DOS mode. Can you please check and correct when necessary.

fjtc:
The problem is not related to line endings, it is a bit worse than that. I used VS Code to edit the files this time. It auto formatted the code on save and I didn't notice it. Please reject this pull request because I'll send a new one with the format back to the original one.

PR #9598, opened 2022-09-13
Warning during compilation on Windows systems

albert-github:
This is based on the warning type: 5219 implicit conversion from 'type-1' to 'type-2', possible loss of data

### DIFF
--- a/src/diagram.cpp
+++ b/src/diagram.cpp
@@ -628,14 +628,14 @@ void TreeDiagram::drawBoxes(TextStream &t,Image *image,
           }
           else
           {
-            xf = di->xfPos()/static_cast<float>(gridWidth);
+            xf = di->xfPos()/gridWidth;
             if (doBase)
             {
-              yf = di->yfPos()/static_cast<float>(gridHeight)+static_cast<float>(superRows)-1;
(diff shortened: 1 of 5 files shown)


src/diagram.cpp
```diff
@@ -626,14 +628,14 @@ void TreeDiagram::drawBoxes(TextStream &t,Image *image,
           }
           else
           {
-            xf = di->xPos()/static_cast<float>(gridWidth);
+            xf = di->xfPos()/static_cast<float>(gridWidth);
```
doxygen, line 631: by using xfPos() the float cast of gridWidth could be removed, or does that then still give a compiler warning?

albert-github: Interesting thought.
I did a quick test and the warning does not re-appear, makes code easier to read.
I'm gone work on it.

albert-github: I corrected the mentioned conversion.

- Should for the `static_cast<float>(superRows)` a separate variable be created like:
   ```
   float superRowsF = static_cast<float>(superRows)
   ```
   so that the construct
   ```
   yf = static_cast<float>(superRows)-1-di->yfPos()/gridHeight;
   ```
   can be rewritten as:
   ```
   yf = superRowsF-1-di->yfPos()/gridHeight;
   ```

PR #9605, opened 2022-09-15
Unknown entities in the preprocessor conditionals

albert-github:
With e.g. the bash package we get warnings like:
```
y.tab.c:728: warning: preprocessing issue while doing constant expression evaluation: syntax error: input=' (! 0L  &&  0L          && 0L<=  (( 0L) (! (! (( 0L) 0 < ( 0L) -1))   ? ( 0L) -1   : (((( 0L) 1 << ((0L *  8 ) - 2)) - 1) * 2 + 1))) )'
```
note e.g. the `(0L)0` investigation lead to the fact that this is due to a construct with `(int)` i.e. a type cast.

Further investigation lead to the possibility of `sizeof(int)` that was not recognized.
- type casts of basic types are ignored
- sizeof basic types are inserted into the condition

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9576872/example.tar.gz)
(found by Fossies for the bash package)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -2752,7 +2752,7 @@ static const char *processUntilMatchingTerminator(const char *inputStr,QCString 
  */
 static QCString removeIdsAndMarkers(const QCString &s)
 {
-  //printf("removeIdsAndMarkers(%s)\n",s);
+  //printf("removeIdsAndMarkers(%s)\n",qPrint(s));
   if (s.isEmpty()) return s;
   const char *p=s.data();
   char c;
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I've changed the implementation a bit to avoid code duplication and elaborate expressions, see https://github.com/doxygen/doxygen/commit/70473b456cc35349316b61cb642277230d6bfaec. Let me know if you see issues with it.

albert-github:
I found some issues:
- compilation on github, OK on my 32-bit Version 19.16.27034 for x86 (2017) error on 64-bit 19.28.29914 for x64 (2019)
- a missing tests, were also not in my original code
- shadowed variable

I've just pushed a proposed patch, pull request #9614

albert-github:
Code for the problems as mentioned in https://github.com/doxygen/doxygen/pull/9605#issuecomment-1259203066 has been integrated in master on GitHub (PR #9614)

PR #9608, opened 2022-09-21
Add building of the doxygen internal documentation to CMake

albert-github:
The doxygen internal documentation is part of the GitHub actions and is published to http://doxygen.github.io/doxygen, so it is logical create the possibility to build the internal documentation with the aid of CMake and deploy those results.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -393,9 +393,15 @@ jobs:
       if: matrix.config.name == 'Ubuntu Latest GCC Release'
       
     - name: Generate Internal documentation
-      shell: bash
+      shell: cmake -P {0}
       run: |
-        build/bin/doxygen Doxyfile
+        execute_process(
+          COMMAND cmake --build build --target docs_internal
+          RESULT_VARIABLE result
+        )
+        if (NOT result EQUAL 0)
+          message(FATAL_ERROR "Building internal documentation failed")
+        endif()
       if: matrix.config.name == 'Ubuntu Latest GCC Release'
 
     - name: Publish Internal documentation to Github pages
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,7 @@ if (build_doc)
     add_subdirectory(doc)
 endif ()
 
+add_subdirectory(doc_internal)
 add_subdirectory(addon)
 
 enable_testing()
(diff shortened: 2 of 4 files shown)


ferdymercury:
idea/suggestion, you could also use the specific CMake command doxygen_add_docs.

See one example here:
https://github.com/AliceO2Group/Monitoring/blob/dev/doc/CMakeLists.txt#L22

albert-github:
Interesting thought, I was not aware of the `doxygen_add_docs` though I have some, big, reservations against the usage of it:
- we have already a `Doxyfile`, so we would need to import this `Doxyfile` into the `CMakeLists.txt` (prepended with `DOXYGEN_`) i.e. at least the ones different from the default settings. In case of new settings (like we recently had e.g. `DOT_EDGE_ATTR`) one has to be very careful with spaces as the handling in not yet known by `doxygen_add_docs`
- it sets under the hood some defaults different to the doxygen defaults, so when one wants to have these one has to explicitly set these again (or incorporate the whole
- the `doxygen_add_docs` still contains settings and tests that have become obsolete, so should be made dependent on the doxygen version (most notably is `MSCGEN_PATH` which became obsolete in doxygen version 1.8.16 as the code was integrated into doxygen).

ferdymercury:
> 
> * we have already a `Doxyfile`, so we would need to import this `Doxyfile` into the `CMakeLists.txt` (prepended with `DOXYGEN_`) i.e. at least the ones different from the default settings.

I guess this could help in the future https://gitlab.kitware.com/cmake/cmake/-/issues/23107

> In case of new settings (like we recently had e.g. `DOT_EDGE_ATTR`) one has to be very careful with spaces as the handling in not yet known by `doxygen_add_docs`
> * it sets under the hood some defaults different to the doxygen defaults, so when one wants to have these one has to explicitly set these again (or incorporate the whole
> * the `doxygen_add_docs` still contains settings and tests that have become obsolete, so should be made dependent on the doxygen version (most notably is `MSCGEN_PATH` which became obsolete in doxygen version 1.8.16 as the code was integrated into doxygen).

Oh I see. I guess CMake devs are not maintaining too much this interface.

albert-github:
- The mentioned issue would indeed be a good improvement.
- I don't know who maintains the doxygen part in CMake or whether it has been just a one time contribution to CMake from a user. The problem would always be that only the newest version of CMake could catch up with doxygen in respect to used settings (but it is always possible to use older versions of doxygen in CMake, as one can query the DOXYGEN_VERSION)

ferdymercury:
The function was introduced by @zaufi, but the latest maintenance was done by @nnemkin 1 year ago: https://gitlab.kitware.com/cmake/cmake/-/blame/master/Modules/FindDoxygen.cmake

PR #9610, opened 2022-09-24
GHC_NO_DIRENT_D_TYPE should be defined on Haiku

Begasus:
This also fixes detection for Haiku.

### DIFF
--- a/filesystem/filesystem.hpp
+++ b/filesystem/filesystem.hpp
@@ -80,6 +80,9 @@
 #elif defined(__QNX__)
 #define GHC_OS_QNX
 #define GHC_NO_DIRENT_D_TYPE
+#elif defined(__HAIKU__)
+#define GHC_OS_HAIKU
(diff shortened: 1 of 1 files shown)


albert-github:
Did you also report this bug upstream i.e. https://github.com/gulrak/filesystem as otherwise it will be lost when filesystem.hpp is updated?

Begasus:
> 

Done now. :)

Begasus:
Checked with a build for Grantlee, documents created successfully. :)

Doxygen-Grantlee

gulrak:
I change the code with [filesystem #142](https://github.com/gulrak/filesystem/issues/142) (published with v1.5.12) to auto-detect the presence of a `d_type` field, that removed the need to define a macro for any new target system that doesn't have it.

Nevertheless thank you for the upstream notice!

Begasus:
All that's needed is to update the file here to reflect the new one upstream? (can't find a sollution to set it to use a system version when packaging filesystem)

PR #9623, opened 2022-10-03
Making the arrows with the `<details>` tag  look like those used in the treeview

albert-github:
Making the arrows with the `<details>` tag  look like those used in the treeview, resulting in adding the settings in doxygen.css

There were some problems in the output introduced in the comment scanner
- multiple `<summary>` tags were just left there whilst only the first one is relevant for the `<details>` tag, the other `<summary>` tags are just new paragraphs
- the `<summary>` tag should only remain in case we are in inside a `<details>` section, otherwise it is handled as a brief description (here the `<summary>` was already ignored, but now it is not even added).
- the order of the resulting `<p>` tags in the HTML output was not correct for XHTML (the `<summary>` tag  was inside a `<p>` tag, this has been corrected

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -419,7 +419,6 @@ struct commentscanYY_state
 
   QCString         guardExpr;
   int              roundCount = 0;
-  int              HTMLDetails = 0;
   std::vector<int> HTMLDetailsSummary;
 
   bool             insideParBlock = FALSE;
@@ -638,15 +637,13 @@ STopt  [^\n@\\]*
                                           REJECT;
                                         }
 <Comment>"<"{DETAILS}{ATTR}">"          { // start of a HTML style details description
-                                          yyextra->HTMLDetails++;
                                           yyextra->HTMLDetailsSummary.push_back(0);
                                           setOutput(yyscanner,OutputDoc);
                                           addOutput(yyscanner,yytext);
                                         }
 <Comment>"</"{DETAILS}">"               { // end of a HTML style details description
-                                          if (yyextra->HTMLDetails)
+                                          if (yyextra->HTMLDetailsSummary.size())
                                           {
-                                            yyextra->HTMLDetails--;
                                             yyextra->HTMLDetailsSummary.pop_back();
                                           }
                                           addOutput(yyscanner,yytext);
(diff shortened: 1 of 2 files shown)


src/commentscan.l
```diff
@@ -420,6 +420,7 @@ struct commentscanYY_state
   QCString         guardExpr;
   int              roundCount = 0;
   int              HTMLDetails = 0;
+  std::vector<int> HTMLDetailsSummary;
```
doxygen, line 423: Looks like there is an invariant `HTMLDetails==HTMLDetailsSummary.size()`. 
If so `HTMLDetails` can be replaced by `HTMLDetailsSummary.size()`

albert-github: Indeed this is the case, adjusted code.

src/commentscan.l
```diff
@@ -706,15 +712,35 @@ STopt  [^\n@\\]*
...
                                           setOutput(yyscanner,OutputDoc);
                                           addOutput(yyscanner,yytext);
                                         }
 <Comment>"</"{SUMMARY}">"               { // start of a .NET XML style detailed description
-                                          addOutput(yyscanner,yytext);
-                                          if (!yyextra->HTMLDetails) setOutput(yyscanner,OutputDoc);
+                                          if (yyextra->HTMLDetails)
+                                          {
+                                            int id = yyextra->HTMLDetailsSummary[yyextra->HTMLDetails-1];
+                                            if (id == 1)
```
doxygen, line 733: what is special about `id==1`? Shouldn't this be `id>0`? I can imagine that 0 is the special in being the first, but 1?

src/commentscan.l
```diff
@@ -706,15 +712,35 @@ STopt  [^\n@\\]*
                                         }
 <Comment>"<"{SUMMARY}">"                { // start of a .NET XML style brief description
                                           if (!yyextra->HTMLDetails) setOutput(yyscanner,OutputBrief);
-                                          addOutput(yyscanner,yytext);
+                                          else
+                                          {
+                                            int id = yyextra->HTMLDetailsSummary[yyextra->HTMLDetails-1];
+                                            if (id == 0)
+                                            {
+                                              addOutput(yyscanner,"\\htmlonly </p> \\endhtmlonly");
```
doxygen, line 717: What about other output formats? 

I think this should preferably be handled in HtmlDocVisitor. Do you have an example of the problem, so I can have a look?

src/commentscan.l
albert-github: I have not been able to create an example where `id == 1` gives problems.
small nested example
```
/** the fie
 <details>
 <summary> The summary 1</summary>
 The details 1


 some
 <summary> The summary 1</summary>
 The details 1
 <details>
 <summary> The summary 1 nested</summary>
 The details 1 nested


 some
 <summary> The summary 1 nested</summary>
 The details 1 nested
 <details>
 <summary> The summary 1 nested 2</summary>
 The details 1 nested 2


 some
 <summary> The summary 1 nested 2</summary>
 The details 1 nested 2
 </details>
 </details>
 </details>
*/
void fie3();
```
works fine with `id == 1` but fails with `id >= 1`.

src/commentscan.l
albert-github: I didn't get problems with other formats, though I think the entire construction with `\htmlonly` can be omitted when it is possible to move the content of the first `<summary.` style change to the beginning of the handling of the `<details>` in the function `int DocHtmlDetails::parse(DocNodeVariant *thisVariant)` of `docnode.cpp `, but I don't yet know how to do this properly.

PR #9626, opened 2022-10-05
Fix a number of compiler warnings

albert-github:
```
.../markdown.cpp(3060,7): warning: variable 'sp' set but not used [-Wunused-but-set-variable]
  int sp=0;

doxygen.cpp(1348,7): warning: variable 'iteration' set but not used [-Wunused-but-set-variable]
  int iteration=0;

dotfilepatcher.cpp(338,7): warning: variable 'lineNr' set but not used
      [-Wunused-but-set-variable]
  int lineNr=1;

scanner.l(6825,47): warning: use of bitwise '|' with boolean operands
      [-Wbitwise-instead-of-logical]
                                          if ((yyextra->docBlockName=="verbatim") | (yyextra->docBlockName=="code") || (yyex...
                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                                  ||
scanner.l(6825,47): note: cast one or both operands to int to silence this warning

mscgen_lexer.l(47,20): warning: unused function 'stateToString' [-Wunused-function]
static const char *stateToString(int state);
```

- corrected syntax problem
- removed unused variables
- doxygen.cpp commented as `iteration` was still present in commented `printf` statements

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -3055,15 +3055,6 @@ QCString Markdown::processBlocks(const QCString &s,const int indent)
   int i=0,end=0,pi=-1,ref,level;
   QCString id,link,title;
 
-  // get indent for the first line
-  end = i+1;
-  int sp=0;
-  while (end<=size && data[end-1]!='\n')
-  {
-    if (data[end-1]==' ') sp++;
-    end++;
-  }
-
 #if 0 // commented m_out, since starting with a comment block is probably a usage error
       // see also http://stackoverflow.com/q/20478611/784672
 
--- a/libmscgen/mscgen_lexer.l
+++ b/libmscgen/mscgen_lexer.l
@@ -44,7 +44,6 @@ static Boolean        lex_utf8 = FALSE;
 /* Local function prototypes */
 static void newline(const char *text, unsigned int n);
 static char *trimQstring(char *s);
-static const char *stateToString(int state);
 
 #if USE_STATE2STRING
 static const char *stateToString(int state);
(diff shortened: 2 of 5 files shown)


src/markdown.cpp
```diff
@@ -3055,15 +3055,6 @@ QCString Markdown::processBlocks(const QCString &s,const int indent)
   int i=0,end=0,pi=-1,ref,level;
   QCString id,link,title;
 
-  // get indent for the first line
```
doxygen, line 3058: why is this whole section removed? `end` is still used at line 3078? maybe you meant to remove only `sp` usage?

albert-github: You meant the usage in the lines:
```
  while (i<size)
  {
    findEndOfLine(data,size,pi,i,end);
    // line is now found at [i..end)

    int lineIndent=0;
    while (lineIndent<end && data[i+lineIndent]==' ') lineIndent++;
...
```
yes it is used here but when we look in `findEndOfLine` the first statements here are:
```
{
  TRACE(data);
  // find end of the line
  int nb=0;
  end=i+1;
```
so it `end` is reset.
Actually the declaration of end could take place inside the while loop.
(I first removed the `sp` and had at that moment also a look at `end` and saw what is written above, so I removed the complete loop).

doxygen: Ah you're right, I see it now.

Issue #9628, opened 2022-10-09
Warning: "Unexpected token TK_HTMLTAG found as part of a title section"


Albrecht-S:
**Describe the bug**
When generating the documentation of the [FLTK project](https://www.fltk.org) a bogus warning "`Unexpected token TK_HTMLTAG found as part of a title section`" appears in current Doxygen `master`. This is a regression WRT Doxygen 1.9.4 and earlier (1.9.5 not tested).

**Expected behavior**
These warnings should not appear. The token "`TK_HTMLTAG`" does not exist in the FLTK source code but I found it in the Doxygen code.

**Screenshots**
No screenshots. Warnings appearing with current FLTK documentation:
```
.../preface.dox:74: warning: Unexpected token TK_HTMLTAG found as part of a title section
.../preface.dox:74: warning: Unexpected token TK_HTMLTAG found as part of a title section
.../preface.dox:80: warning: Unexpected token TK_HTMLTAG found as part of a title section
.../preface.dox:80: warning: Unexpected token TK_HTMLTAG found as part of a title section
.../preface.dox:80: warning: Unexpected token TK_HTMLTAG found as part of a title section
.../preface.dox:80: warning: Unexpected token TK_HTMLTAG found as part of a title section
```
Link to original source file [preface.dox](https://github.com/fltk/fltk/blob/4d17cf310dc953703bf3a60763b681eb217bc3e9/documentation/src/preface.dox#L74). Note that the preceding lines of those mentioned in the warnings contain HTML `<tt>` statements. I'll follow-up with a partial workaround (removing these `<tt>` tags).

**To Reproduce**
I'm currently not able to create a small reproducer since the warnings are not related to our FLTK source code. However, it's easy to reproduce the warnings with current FLTK code:
```
git clone https://github.com/fltk/fltk.git
cd fltk
# optionally check out `master` as of today:
git checkout 4d17cf310dc953703bf3a60763b681eb217bc3e9
mkdir build
cd build
# Make sure that `doxygen` can be found in your path so cmake finds it.
# I assume you know how to do this, but anyway:
# example: export PATH=/path/to/doxygen/build/bin:$PATH
cmake ..
make html
```
Please ignore warnings like "Tag 'CLASS_DIAGRAMS' ... has become obsolete. These warnings are due to an older Doxyfile (as you certainly know).

**Version**
$ doxygen --version
1.9.6 (59aa82c1385054dadfb636375e7e45be6b95672c)
(self built)

**Additional context**
[...]

PR #9630, opened 2022-10-10
issue #9628 Warning: "Unexpected token TK_HTMLTAG found as part of a title section"

albert-github:
Since 1.9.5 a warning is given when a HTML tag is used inside the title of a `\par` command and the HTML tag is ignored. A number of HTML tags can be used inside a '\par` statement without any problems.

### DIFF
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -2698,7 +2698,157 @@ void DocTitle::parse(DocNodeVariant *thisVariant)
   int tok;
   while ((tok=parser()->tokenizer.lex()))
   {
-    if (!parser()->defaultHandleToken(thisVariant,tok,children()))
+    if (tok == TK_HTMLTAG)
+    {
+      QCString tagName = parser()->context.token->name;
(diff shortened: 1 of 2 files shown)


src/docnode.cpp
```diff
@@ -2698,7 +2698,157 @@ void DocTitle::parse(DocNodeVariant *thisVariant)
   int tok;
   while ((tok=parser()->tokenizer.lex()))
   {
-    if (!parser()->defaultHandleToken(thisVariant,tok,children()))
+    if (tok == TK_HTMLTAG)
```
doxygen, line 2701: It looks like there is now quite some overlap/duplication with the `defaultHandleToken()` used in the else, which also handles `tok==TK_HTMLTAG` for a selected set of supported tags. I'd prefer to avoid duplication. Which commands are now handled differently than `defaultHandleToken()`? And are things like `<div>` and `<pre>` really supported as title for all output formats?

Albrecht-S:
Note that the underlying issue #9628 has obviously been fixed in doxygen 1.9.6. Thanks

Issue #9615, opened 2022-09-27
doxygen C# regression


Labouh:
**Describe the bug**
With release 1.9.4 and 1.9.5 we got a lot of parameter and return type not documeted (in fact it seems for all methods!) on C# files that does not exists with previous version 1.9.1.
For instance:
The following simple C# declaration:
``` c#
        /// <summary>
        /// RestorePLCCOnfig functionality
        /// </summary>
        /// <returns>Asynchronous result with error information</returns>
        private async Task<FluentResults.Result> RestorePLCCOnfig()

```
Return an invalid:
```
warning: return type of member MvtMain.Services.TiaPortalConfigManager.TiaPortalConfigManager.RestorePLCCOnfig is not documented
```
**Expected behavior**
The previous C# code should be seen as documented by doxygen and not return warning (it was the case with version 1.9.1).

**To Reproduce**
Any C# code methods is seen as undocumented.

**Version**
1.9.4 and 1.9.5, it worked with 1.9.1

albert-github:
With just the code snippet I'm not able to reproduce the problem  (probably missing some surrounding like class etc.).

- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

Labouh:
[Doxyfile.zip](https://github.com/doxygen/doxygen/files/9655944/Doxyfile.zip)

Here an simple example...

With 1.9.1 there is no warning about parameters missing... (only about unsupportedtag)

With 1.9.5, there is one:
Main.cs(21): warning: parameters of member CSharpTutorials.Program.Main are not documented

Note: it is the case with any C# file

Labouh:
The first warning is normal (as I set Doxyfile only to get warning), so I get it with all version.

The real issue is the second warning that is wrong as the method *is* documented, and 1.9.1 does not output such a false warning 

```
        /// <summary>
        /// Main
        /// </summary>
        /// <param name="args">args</param>
        static void Main(string[] args)
```

Labouh:
But yes you are right, if I set an output format, there is no more the warning...

I use this to find warning without generated the documentation... so it is quite usefull.

albert-github:
I've just pushed a proposed patch, pull request #9633

PR #9633, opened 2022-10-11
issue #9615 doxygen C# regression

albert-github:
The problem is that in case non of the output generators HTML, LaTeX, RTF, docbook or Man are used not all the settings are done before the tests are started. We see this problem also when we just set `GENERATE_XML`, we get the warning like"
```
Main.cs(21): warning: parameters of member CSharpTutorials.Program.Main are not documented
```

by introducing a new output generator that will only be enabled in case non of the generators HTML, LaTeX, RTF, docbook or Man is enabled, the settings are done, so the tests can be executed properly.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -58,6 +58,7 @@
 #include "docbookgen.h"
 #include "defgen.h"
 #include "perlmodgen.h"
+#include "validatinggen.h"
 #include "reflist.h"
 #include "pagedef.h"
 #include "bufstr.h"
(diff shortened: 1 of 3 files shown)


src/doxygen.cpp
```diff
@@ -12462,6 +12463,12 @@ void generateOutput()
     g_outputList->add<RTFGenerator>();
     RTFGenerator::init();
   }
+  if (!(generateHtml && generateLatex && generateDocbook && generateMan && generateRtf))
+  {
+    g_outputList->add<ValidatingGenerator>();
+    ValidatingGenerator::init();
+  }
+
```
doxygen, line 12471: Would it be easier to do this instead?

```c++
  if (g_outputList->size()==0)
  {
    Doxygen::suppressDocWarnings=TRUE;
  } 
```

albert-github: No I don't think so.
Doxygen consists, unfortunately, over 2 types of output generators (HTML / RTF / LaTeX / Docbook / Man / ... versus XML / SQLite / ...) and a number of tests / updates are only done in the first type of generators and hence when none of these generators we need a dummy generator that sees to it that the tests / updates are executed.

It might be that the test could be reduced to `g_outputList->size()==0` (or even `!g_outputList->size()`) though as this would be safer for the future.

PR #9636, opened 2022-10-12
Fix unreadable file warnings for broken symlinks with EXCLUDE_SYMLINKS is set

pinchartl:
When walking directories to search for source files, Doxygen prints a warning for each file that isn't readable. This can be caused by broken symlinks. When EXCLUDE_SYMLINKS is set to YES, symlinks are skipped anyways, so there is little point in printing a warning. Refactor the readDir() and readFileOrDirectory() functions to avoid the warning in that case.


pinchartl:
I'm not familiar with the Doxygen coding style, so feel free to rearrange the code if the empty body of `if` statements is a style problem.

Issue #7421, opened 2019-11-28
PlantUML anchor hyperlink within page


eraytufan:
Hello,

I'm using plantUML and created a hyperlink to one of the anchors (subsection) of the page. However I cannot seem to direct the link to the anchor as I have wanted. Is there any way I can use "\ref" inside the "@startUML" or how can I create an anchor in the UML .svg file.

```
@startuml
start
:Some [[#myanchor link]];
end
@enduml
```

this code goes to directly: 
/inline_umlgraph_1.svg#myanchor 

instead of:
/myfunction_8c.html#myanchor

albert-github:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.
- Please also specify the doxygen version used.

eraytufan:
[plantUmlDoxygen.zip](https://github.com/doxygen/doxygen/files/3901574/plantUmlDoxygen.zip)

Okay I have attached an example file. When you click the link, it opens it in a new page while I try to go to anchor.

albert-github:
I think this issue is related to #6539

PR #9644, opened 2022-10-19
bug 751096 Possibility for svg images with \image command in LaTeX

albert-github:
When using a svg image with the \image command in LaTeX we get the message:
```
! LaTeX Error: Unknown graphics extension: .svg.

See the LaTeX manual or LaTeX Companion for explanation.
```
when generating the refman.pdf.

Enabling can be done by:
-    using the svg package
      note:
      - LATEX_CMD should not be quoted in the `make.bat` / `Makefile`
      - properly handle underscores in svg images (see also: https://tex.stackexchange.com/a/525613/44119)
-    installing the external package Inkscape.
-    enabling the shell escape with the LaTeX processor (normally the option -enable-write18).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9820633/example.tar.gz)

Supersedes #7421

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -294,8 +294,8 @@ static void writeLatexMakefile()
   }
   TextStream t(&f);
   // inserted by KONNO Akihisa <konno@researchers.jp> 2002-03-05
-  QCString latex_command = theTranslator->latexCommandName().quoted();
-  QCString mkidx_command = Config_getString(MAKEINDEX_CMD_NAME).quoted();
+  QCString latex_command = theTranslator->latexCommandName();
+  QCString mkidx_command = Config_getString(MAKEINDEX_CMD_NAME);
   QCString bibtex_command = "bibtex";
   QCString manual_file = "refman";
   const int latex_count = 8;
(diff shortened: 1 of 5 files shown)


src/latexgen.cpp
```diff
@@ -294,8 +294,8 @@ static void writeLatexMakefile()
   }
   TextStream t(&f);
   // inserted by KONNO Akihisa <konno@researchers.jp> 2002-03-05
-  QCString latex_command = theTranslator->latexCommandName().quoted();
-  QCString mkidx_command = Config_getString(MAKEINDEX_CMD_NAME).quoted();
+  QCString latex_command = theTranslator->latexCommandName();
```
doxygen, line 297: I think this change will break things if an existing Doxyfile contains something like this:

```
LATEX_CMD=C:\Program Files\MiKTeX\miktex\bin\x64\latex.exe
```

doxygen: If adding `-enable-write18` is always required anyway when the image format SVG is used, then doxygen could also detect this and only then add the option itself, rather than relying on the user to do this.

albert-github: In principle yes but due the security risks I think it would be better to do it more explicitly and not implicitly, although for other parts doxygen also automatically starts sub programs, like `dot` of filters etc., so it should be possible to add it implicitly (and place a warning in the manual).

jessegv:
It seems like this has stalled and is not being merged.  Is there any plan to merge this or an updated version of this?

PR #9645, opened 2022-10-20
Removing hierarchy / namespace from layout file results in incorrect RTF / LaTeX document

albert-github:
When removing from the `DoxygenLayout.xml` the parts
```
    <tab type="namespaces" visible="no" title="">
      <tab type="namespacelist" visible="yes" title="" intro=""/>
      <tab type="namespacemembers" visible="yes" title="" intro=""/>
    </tab>
```
and
```
      <tab type="hierarchy" visible="no" title="" intro=""/>
```
this leads to errors due to missing files `namespaces` and `hierarchy` when building the PDF from the LaTeX files and when combining the RTF files.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9832021/example.tar.gz)

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -4553,9 +4553,13 @@ static void writeIndex(OutputList &ol)
     }
     if (Config_getBool(SHOW_NAMESPACES) && (documentedNamespaces>0))
     {
-      ol.startIndexSection(isNamespaceIndex);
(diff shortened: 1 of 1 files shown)


albert-github:
Issue was found when building CGAL documentation in LaTeX (/ RTF) mode.
Problem might be in other places as well but didn't show up with CGAL (so this proposed pull request must also be seen as an example how to handle other problems, when they arise).

PR #9647, opened 2022-10-23
Incorrect interactive svg file for dot

albert-github:
When using a `\dot` command and having `svg` as format and `INTERACTIVE_SVG` set, the svg file generated by doxygen is incorrect. The message we get is like:
```
XML Parsing Error: mismatched tag. Expected: </svg>.
Location: file:///.../html/dot_inline_dotgraph_1.svg
Line Number 332, Column 3:
````
and xmllint gives:
```
html\dot_inline_dotgraph_1.svg:332: parser error : Opening and ending tag mismatch: svg line 53 and g
</g>
    ^
html\dot_inline_dotgraph_1.svg:334: parser error : Extra content at the end of the document
        <g id="navigator" transform="translate(0 0)" fill="#404254">
        ^
```
Doxygen relies on a `<title>` tag inside the originally generated svg file, but in the newer versions of dot this is not present anymore (still present in 2.38 but absent already in 2.50). The problem can be solved by explicitly defining a 'title' with the digraph in the dot setup.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/9846735/example.tar.gz)

### DIFF
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -4011,7 +4011,56 @@ int DocPara::handleCommand(DocNodeVariant *thisVariant,const QCString &cmdName, 
         parser()->defaultHandleTitleAndSize(CMD_DOT,vDocVerbatim,dv->children(),width,height);
         parser()->tokenizer.setStateDot();
         retval = parser()->tokenizer.lex();
-        dv->setText(parser()->context.token->verb);
+        // with interactive SVG it is important that a digraph has a title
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I've created a simpler solution by changing the dotfileparser, so it can work without title tag.

albert-github:
As graphviz had chosen to remove the title tag in case no title was given I also went that way, but in principle every tag would do.

PR #9670, opened 2022-11-09
Add option to have warning on undocumented enumeration member

Blabla51:
I have seen this feature requested several times, so here it is 😄 

The configuration parameter `WARN_IF_ENUM_MEMBER_UNDOCUMENTED` is now available to detect undocumented enumeration members.

Warnings are displayed in the following way, here using the `examples/enum.cfg` file:
`/home/doxygen/examples/enum.h:4: error: Documentation for enum member 'Enum_Test::TEnum::Val2' is missing.`

~I also added `/html/*` and `/latex/*` folders in the `.gitignore` file to avoid having the documentation generated by the examples to be in the list of changes on git. If it is used for something else (like the internal documentation), I can remove this change.~

~I also fixed a line return that was missing on a warning message in `src/doxygen.cpp`.~


src/doxygen.cpp
```diff
@@ -11797,7 +11797,7 @@ void searchInputFiles()
   }
   if (Doxygen::inputNameLinkedMap->empty())
   {
-    warn_uncond("No files to be processed, please check your settings, in particular INPUT, FILE_PATTERNS, and RECURSIVE");
+    warn_uncond("No files to be processed, please check your settings, in particular INPUT, FILE_PATTERNS, and RECURSIVE\n");
```
albert-github, line 11800: Good catch though it would be better to have such a change not hidden in another PR but in a separate PR.

albert-github:
Thanks fro the PR.

> I have seen this feature requested several times

do you have an example of the issue(s) ?

> I also added /html/* and /latex/* folders in the .gitignore file to avoid having the documentation generated by the examples to be in the list of changes on git. If it is used for something else (like the internal documentation), I can remove this change.

looks like a local issue at your place there is no html / latex file generated at the root level of doxygen, must be from a run you did on a wrong place,  it doesn't hurt to have html and latex excluded. Though it would be better to have such a change not hidden in another PR but in a separate PR.

> I also fixed a line return that was missing on a warning message in src/doxygen.cpp.

Good catch though it would be better to have such a change not hidden in another PR but in a separate PR.

src/doxygen.cpp
Blabla51: Do you want me to create another merge request for that ?

.gitignore
```diff
@@ -20,3 +20,6 @@
 tags
 
 .idea
+
+/html/*
+/latex/*
```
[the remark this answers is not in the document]
Blabla51: I'm not sure. I was looking at `/examples/baseexample.cfg`, and the output directory is `OUTPUT_DIRECTORY   = ..`, so I would expect the output to the root of the repo `/`

Also, if you want to keep this change, do you want me to create another PR for that?

src/doxygen.cpp
albert-github: I think this would be better and cleaner.

.gitignore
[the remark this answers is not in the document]
albert-github: I now understand how you got the `html` and `latex`, the examples are not intended for running them standalone, but as part of the documentation generation in the build process (`make docs` and `make docs_chm`), so other people might indeed also run into this problem.

I think this would be better and cleaner to have a separate PR for this.

Blabla51:
I do not see issues created, but there are some stackoverflow questions about it:
https://stackoverflow.com/questions/55415652/missing-doxygen-warning-for-undocumented-enum-class-constants
https://stackoverflow.com/questions/38452152/forcing-doxygen-c-to-warn-if-macros-or-enums-are-undocumented

Let me know if you want to have different PR for the different improvements in the code. Thanks !

albert-github:
Problem with the stack overflow questions is that they are not recorded with doxygen as problems and might be lost, sometimes I advise to create an issue (but I see I didn't do it in this case) or create a PR for it straight away.

Regarding the 
>Let me know if you want to have different PR for the different improvements in the code.

see my remarks with the review, please separate PRs.

Blabla51:
Alright, I'll do that by the end of the day tomorrow, thanks for the feedback :)

Blabla51:
Done, I'll create other PR for the reverted changes.

Let me know if there are other changes needed.

Blabla51:
@albert-github do you need to approve something so that the workflow can be run ?

albert-github:
The problem with the button is that it says "Approve and Run" in other words 2 actions. The "Run" part would be OK, though I didn't think about all the, subtle, consequences of the setting when set (I just remarked about the problems I saw regarding mixing topics). I'll just do the "Run".

Blabla51:
@albert-github Sorry to bother again, but how does it work for the pull requests to be accepted ? Is there something else I should do ? :)

albert-github:
I don't decide whether or not a proposed pull request is accepted so I cannot tell you.

doxygen:
I change/shortened the option name to `WARN_IF_UNDOC_ENUM_VAL` because of Doxyfile layout restrictions.

PR #9676, opened 2022-11-11
Added latex and html folders in gitignore

Blabla51:
The configuration in the `examples` folder is using the root of the project as an output base directory.

When generating the examples individually, the documentation is generated in the `/html/` and `/latex/` folders.

I might not be the only one trying to generate the examples individually while adding some features to the code, so it is maybe worth to add those folders in the `.gitignore` file, if those folders are not used for any other purpose.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@
 tags
 
 .idea
+
+/examples/html/*
+/examples/latex/*


doxygen:
@Blabla51 what about the folders `templates/html/` and `templates/latex/`. These should not be gitignored. I think it is better to make sure the generated documentation for the examples ends up in the build directory. Or are you building in the source folder?

Blabla51:
@doxygen I only added the `/html/` and `/latex/` folders at the root directory. It should not have an impact on the `/template/` folder.

To explain the background maybe better, I having been working on https://github.com/doxygen/doxygen/pull/9670 and I just wanted to test my changes on the `/examples/enum.h` file, which is related to the `/examples/enum.cfg` configuration. So I went inside the `/examples/` and ran `doxygen ./enum.cfg`. It then needs to have the `/html/` and `/latex/` folders created so that it can generate the documentation. Is it not intended to work in this way?

albert-github:
As I wrote in: https://github.com/doxygen/doxygen/pull/9670#discussion_r1018878096:
> I now understand how you got the html and latex, the examples are not intended for running them standalone, but as part of the documentation generation in the build process (make docs and make docs_chm), so other people might indeed also run into this problem.

The problem is cased by the `OUTPUT_DIRECTORY   = ..` in the `baseexample.cfg`

albert-github:
I've just pushed a proposed patch, pull request #9677 which addresses the path problem.

I did not adjust / address the `.gitignore` change this now could (probably) be `/examples/html` and `/examples/latex`

Note that there is still a problem regarding the creation of  directories in the standalone case, though this is addressed in the proposed pull request #8444

Blabla51:
Alright, I'll wait for https://github.com/doxygen/doxygen/pull/9677 to be merged, and I'll update this PR afterward

albert-github:
Code of #9677 has been integrated in master on GitHub

Blabla51:
I made the changes 

I had to force-push because I messed up some things while merging from master, should be good now

Issue #1166, opened 2018-07-01
Read error when multi-byte characters in file path name (Origin: bugzilla #145186)


doxygen:
status NEW severity *major* in component *general* for *---*
Reported in version *1.5.3-SVN* on platform *Other*
Assigned to: Dimitri van Heesch

On 2004-06-30 11:02:26 +0000, Leo Liang wrote:

    Doxygen will report "Error: source D:/vxml_ips/ÃÃ¯ÃÃ´Â»Â¥ÃÂªÃÃ¸ is not a
    readable
    file or directory", when the path contains multi-byte characters, e.g.
    chinese.
    
    The platform is Windows 2000, locale is P.R.China, using GB2312 (simplifed 
    chinese) as local charset.
    
    I'v traced into the source code, found that the promblem is in QT codes.
    QFile::encodeName() at last calls QString::local8Bit() to transform unicode 
    string to non-unicode string, on Windows platform, it calls
    qt_winQString2MB(),
    qt_winQString2MB() use WideCharToMultiByte to convert string into
    CP_ACP(ANSI)
    code page, but not the code page specified in locale setting, so the result
    is
    wrong.
    
    Bug 145066 may be the same promblem.

On 2004-06-30 13:14:49 +0000, Leo Liang wrote:

    After checked more codes, I found this program is not i18n friendly:(
    Thought it uses unicode internal, but only do ANSI/latin1 transform, if the 
    local code page is other than ANSI, error occurs.

On 2004-06-30 17:50:24 +0000, Dimitri van Heesch wrote:

    I do not have the i18n knowledge to fix this I'm afraid. So if you
    could provide a patch I'm happy to include it, otherwise you should
    restrict directories to ansi character only.

On 2004-07-01 00:57:40 +0000, Leo Liang wrote:

    I had tried to fix it, but seems it's not so easy. The problems are in QT's 
    code. Maybe we had to wait for QT to resolve it?

On 2004-07-03 13:28:45 +0000, Dimitri van Heesch wrote:

    *** Bug 145066 has been marked as a duplicate of this bug. ***

On 2004-07-05 13:29:06 +0000, xujin2000@sina.com wrote:

    I have tried the following modification:
    
    1.Add #include <qtextcodec.h> in doxywizard.cpp and config.cpp;
    
    2.In function static bool loadConfig( QString loadFile ),
    I change   if (!Config::instance()->parse(loadFile)) to
      QTextCodec *pCodec = QTextCodec::codecForLocale();
[...]

PR #9681, opened 2022-11-17
fix error file path with non-ansi code

Jiangshan00001:
fix this error:

https://github.com/doxygen/doxygen/issues/1166

### DIFF
--- a/src/diagram.cpp
+++ b/src/diagram.cpp
@@ -30,6 +30,8 @@
 #include "classlist.h"
 #include "textstream.h"
 
+#include "filesystem.hpp"
+
 //-----------------------------------------------------------------------------
 
 class TreeDiagram;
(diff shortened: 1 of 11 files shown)


doxygen:
@Jiangshan00001 can you squash all your commits into a single one? Makes it easier to review and all commits belong to the same issue anyway.

doxygen:
@Jiangshan00001 I've decide to implement this in a bit different/cleaner way. Since you didn't squash your commits, I couldn't really use this pull request, and decided to push my own [commit](https://github.com/doxygen/doxygen/commit/e002e293d9dc956a0634b3a4bcc8d93e655582d5).

Jiangshan00001:
@doxygen fine.
i have not figured out how to squash the commits...

doxygen:
which means it should be fixed in doxygen version 1.9.6.

Issue #9701, opened 2022-11-27
Relative Links in Markdown to Anchors do not get translated correctly in HTML output


RobertClassen:
**Steps to reproduce:**
1. Download and unpack example: [DoxygenExample.zip](https://github.com/doxygen/doxygen/files/10098544/DoxygenExample.zip)
2. Open contained `Documentation~/index.html` file (static shortcut to generated `Documentation~/html/index.html`) and unfold hierarchy on the left, result should look like this: 
![DoxygenExample](https://user-images.githubusercontent.com/61211020/204141839-19339e72-6d9f-4c84-b461-27c63b46a229.png)
3. Mouse over and compare links for "SomeCategory" and "SomeOtherCategory" in left hierarchy vs right table → same links, works as expected
4. Compare links to individual methods → links in hierarchy work, links in table lead to "Page not found" notification in browser
5. (Open top-level `README.md` file in text editor and compare links at bottom)

**Expected result:** 
Appending the same relative link with `#` followed by the anchor (header) name should translate the generated link to the link in the hierarchy.
(Clicking a method name in the table should result in jumping/scrolling to and briefly highlighting the method name title the same way clicking the method name in the hierarchy does.)

**Use case:** 
The top-level `README.md` is used to give an overview of the contents on a GitHub repository page where the links in the table work.
I included `../README.md` in `INPUT` in the `Documentation~/Doxyfile` to show the same overview after opening the above `index.html` instead of showing an empty page and to allow for easier navigation via the custom table instead of having to either unfold and dig through the hierarchy or use the search function (which would require already knowing some class/method names to search for).

I tried searching but the closest issue I found was [this](https://stackoverflow.com/questions/59685012/doxygen-markdown-links-to-main-page-do-not-work), leading to https://github.com/doxygen/doxygen/pull/7511.

**Doxygen Version:** 
Tested with 1.9.1 and 1.9.5

**P.S.:** Possibly the same or a related issue exists with the "Contents" link under the "MyExample" title which links to the wrong `README.md` file, although this is just added as an extra example for relative links in the same `README.md` file while links to other `README.md` files are my main concern.

albert-github:
Looking at 2 links represented in the table:
```
[SomeOtherCategory]: Runtime/Scripts/SomeOtherCategory/README.md
[SomeOtherCategory.MyOtherMethod0]: Runtime/Scripts/SomeOtherCategory/README.md#MyOtherMethod0
```
[...]

PR #9703, opened 2022-11-28
issue #9701 Relative Links in Markdown to Anchors do not get translated correctly in HTML output

albert-github:
Determining the language from the filename should be done without the anchor. Also other operations should be done without the anchor (like absolute path), but we have to append the anchor later on again.

### DIFF
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -699,9 +699,19 @@ DocRef::DocRef(DocParser *parser,DocNodeVariant *parent,const QCString &target,c
   QCString     anchor;
   //printf("DocRef::DocRef(target=%s,context=%s)\n",qPrint(target),qPrint(context));
   ASSERT(!target.isEmpty());
-  SrcLangExt lang = getLanguageFromFileName(target);
+  QCString locTarget = target;
+  QCString locAnchor;
+  int lp = -1;
+  lp = locTarget.findRev('#');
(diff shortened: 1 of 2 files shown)


doxygen:
@albert-github When I try this pull request on the example provided with issue #9701, I get more warnings than without it.
New onces include:
```
DoxygenExample/README.md:2: warning: unexpected token TK_EOF as the argument of ref
```
Just an observation, didn't investigate in detail.

albert-github:
Strange I only get:
- warning about the configuration file, logical as the Doxyfile is from version 1.9.1
- warning about the non existing ../Editor which is logical as the directory doesn't exist

I just unpacked the zip file and ran doxygen in the `Documentation~` directory.

RobertClassen:
>DoxygenExample/README.md:2: warning: unexpected token TK_EOF as the argument of ref

That would be `Lorem Ipsum [Contents]`
which GitHub recognizes as a short form for `Lorem Ipsum [Contents][Contents]`
and resolves it to line 16: `[Contents]: ../README.md#Contents`

Interestingly, as mentioned in the issue, Doxygen resolves this to the incorrect `README.md` file - the one in the `Documentation~` folder which I put there to give users instructions on how to install and run Doxygen to get the local documentation - instead of the main `README.md` file at the top-level. Seems like it's ignoring the `../` for some reason?

Issue #7764, opened 2020-05-08
[Request] about python `_` and `__` generate different permissions.


625781186:
## 1.
I want `_` to generate protect and `__` to generate private ,

But now _ also generate private .

I think it's best to add a configuration option to implement it.

Actually, `_` in python class can be subclassed and rewritten, but `__` can't.

![image](https://user-images.githubusercontent.com/30521429/81369993-9868a100-9126-11ea-9fb3-45c8e0fd09d2.png)

## 2.
`_` has a higher priority than`@protected` , 
I don't think it's reasonable.

![image](https://user-images.githubusercontent.com/30521429/81370357-86d3c900-9127-11ea-92b1-fdcacc1cdbd5.png)

albert-github:
See also discussion with issue #7759 

- Please add references (when available) to relevant PEPs
- Please add a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem shows the requests?  Please don't add external links as they might not be persistent.

625781186:
> See also discussion with issue #7759 and proposed pull request #6555
> 
> * Please add references (when available) to relevant PEPs
> * Please add a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem shows the requests?  Please don't add external links as they might not be persistent.

1. PEPs.
https://www.python.org/dev/peps/pep-0008/
[Descriptive: Naming Styles](https://www.python.org/dev/peps/pep-0008/#id36)
[Method Names and Instance Variables](https://www.python.org/dev/peps/pep-0008/#id47)

![image](https://user-images.githubusercontent.com/30521429/81405063-85c38b80-9169-11ea-865e-78fbd2d8f4e2.png)

2. demo .
[doxy_error.zip](https://github.com/doxygen/doxygen/files/4598895/doxy_error.zip)

Issue #7759, opened 2020-05-07
[Question] Some comments on python . 


625781186:
## 1. 
`@property` as a member variable rather than a function.
Can it be done now?

![image](https://user-images.githubusercontent.com/30521429/81253620-91289100-905b-11ea-9cf9-6600e4e6be56.png)

## 2. 
I want `_` to generate `protect` and `__` to generate `private` , 

But now `_` also generate `private`  .

Is there any way to do that ?

![image](https://user-images.githubusercontent.com/30521429/81253687-b0272300-905b-11ea-9129-b58c6764f058.png)

## 3. 
How to Ignore some comments : 

### 3.1
I have some code like this for the IDE to fold, but now that it's recognized as an instruction, how do I ignore it?

![image](https://user-images.githubusercontent.com/30521429/81253776-f2e8fb00-905b-11ea-9a53-5965ef300e1f.png)

![image](https://user-images.githubusercontent.com/30521429/81253782-f7151880-905b-11ea-840f-7d6a9d069657.png)

### 3.2. The same, 

![image](https://user-images.githubusercontent.com/30521429/81253910-465b4900-905c-11ea-9eee-3859a3ae8f83.png)

![image](https://user-images.githubusercontent.com/30521429/81253932-54a96500-905c-11ea-8bde-90bc9adc176a.png)

## 4 . 
This comment generates `two arrows`. 

Are these details mentioned anywhere in the document?

I can't seem to find them. 

Can you tell me where it is? Thanks!

![image](https://user-images.githubusercontent.com/30521429/81253866-2b88d480-905c-11ea-9d2d-cb2fb874789d.png)

albert-github:
1. @property as a member variable rather than a function.
    Please show an example where this goes wrong:
    - attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.
    - Please also specify the doxygen version used.
2. I want _ to generate protect and __ to generate private
   Python has no knowledge about private it is a loose ad-hoc rule (followed by a number of people), so it is hard for doxygen to follow this rule. Doxygen has commands like `private` probably this wil bring something.
3. `##` is handled as (continuation) comment in doxygen, so here is not realy anything that can be done, only thing might be might be
    - instruct the IDE to use a single `#`
    - use a `FILTER_PATTERNS` in doxygen that can filter out these lines
4. This comment generates `two arrows`.
[...]

PR #9706, opened 2022-12-02
Adding protection protected for Python

TreeN0de:
Adding conditions to differ between private and protected in python.

### DIFF
--- a/src/pyscanner.l
+++ b/src/pyscanner.l
@@ -496,9 +496,16 @@ STARTDOCSYMS      "##"
                         yyextra->current->startLine = yyextra->yyLineNr;
                         yyextra->current->bodyLine  = yyextra->yyLineNr;
                         yyextra->current->type.resize(0);
-                        if (yyextra->current->name.at(0)=='_') // mark as private
+                        if (yyextra->current->name.at(0)=='_') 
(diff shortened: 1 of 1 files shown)


albert-github:
Relevant in relation to this proposed pull request might be:  #7764, #6555 and #7759.

TreeN0de:
Could someone please start the checks for me? This was my first merge request and I do not have the necessary permissions to start it myself.

albert-github:
The failing checks are caused by the fact that MiKTeX doesn't work properly, in the mean time we replaced it with tlmgr.

albert-github:
I was looking at the proposed pull request #6555 ad saw there that the functions with names starting and ending with a double underscore were classified as `public`, now they are `private`, An example is `__init__`, so is `__init__` a `public` or a `private` member of a class?

TreeN0de:
Functions in python starting with `__` and ending  with `__` are special functions of every object in python. This functions can be accessed indirect. For example `__init__` is called when an new instance of an object is created. As I don‘t known of an ohter way to define a constructor in Python, I would declare this functions as public.
I thought, that I didnt' touched the spezific section (line in 1568  pyscanner.l)

Issue #7320, opened 2019-10-14
Python class attributes typing support


Aranaevens:
Hello !

It seems like Doxygen doesn't support typing inside a function in Python. [#6462](https://github.com/doxygen/doxygen/issues/6462) fixed it for functions declaration with types but not inside the function it seems.

If I write :
```
## Constructor.
# @param stuff: stuff
def __init__(self):
    ## a comment
    self.an_int: int = 2
    ## another comment
    self.another_int = 3
```

The former doesn't appear in the Doxygen html page while the latter appears with "another comment" as the documentation of the attribute.

It works similarly for static attributes, found it while writing QT something like:
```
class MyClass(QWidget):
    ## Doc for default
    DEFAULT_COLOR: QColor = QColor(0, 0, 0, 255)
    ## Doc for event
    EVENT_COLOR: QColor = QColor(255, 255, 255, 255)
```
On the html output by Doxygen I had a single entry in public static attributes named "QColor" with both documentations.

albert-github:
So actually the type hints can be used on other places than the function headings as well.

For completeness (and no cut / paste necessary):
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.
- Please also specify the doxygen version used.

Aranaevens:
[doxygen_type_python.zip](https://github.com/doxygen/doxygen/files/3724478/doxygen_type_python.zip)

Here is the zip with the above code, I added an example with a standalone typed string that just doesn't show up at all. I included the Doxyfile I used for generating it if that may help.

Doxygen version used (output of --version):
1.8.16 (cfd73d5c4d1a66c620a3b7c08b72a3f3c3f94255*)

str55:
For reference, what we'd like to see supported is [PEP 526](https://www.python.org/dev/peps/pep-0526/) which is similar to the well-supported [PEP 484](https://www.python.org/dev/peps/pep-0484/)

lwpiotr:
Hello, I second this issue, especially since it does not seem to have been fixed for nearly 2 years. In my case:

`
@dataclass
class DataTree():
    ## File handle
    _file: ROOT.TFile = None
    ## File name
    _file_name: str = None

`

results in 
![Screenshot-20220727102932-472x304](https://user-images.githubusercontent.com/54011073/181200270-59fb2627-74b2-415a-9662-476590fbd817.png)

which makes documenting dataclasses quite tricky...

[...]

PR #9709, opened 2022-12-04
Adding Python PEP 526 typing support

TreeN0de:
Adding Python [PEP 526](https://www.python.org/dev/peps/pep-0526/) typing support for variables.
Maybe relevant for: #7320

### DIFF
--- a/src/pyscanner.l
+++ b/src/pyscanner.l
@@ -309,7 +309,8 @@ STARTDOCSYMS      "##"
                         yyextra->packageCommentAllowed = FALSE;
                         BEGIN(VariableDec);
                       }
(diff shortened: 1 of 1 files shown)


TreeN0de:
I' ve looked into the failures. The reasons are some timeout in webrequest to download software. My change doesn't affect this part. But since I don't have the option to run the checks again, I don't know what to do.

albert-github:
Your changes don't influence the downloads of the testing software, but sometimes there are some timeouts (not influenceableby doxgen) that cause the problems with the tests. I restarted the failing tests and they now ran without problems.

Issue #9715, opened 2022-12-06
segmentation fault (recursive classes?)


Barokoli:
I am getting a segmentation fault when running doxygen.
Tested versions 1.9.2, 1.9.5
Tested OS: Mac 12.6 (M1) and also on gitlab runner alpine docker image

Running on mostly c# code (Unity Project)

I tried to get a core dump but failed due to multiple reasons. Maybe you can help me with that.

I tried to build with XCode but it fails to generate with the error:
`
CMake Error in src/CMakeLists.txt:
  The custom command generating

    /Users/sebastian/Coding/Tools/doxygen/build/generated_src/configvalues.h

  is attached to multiple targets:

    generate_configvalues_header
    doxycfg
    doxymain

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".`

I then build with "Unix Makefiles" (I prefer)
With ("-DCMAKE_BUILD_TYPE=Debug") it didn't produce a coredump or similar just segfaulted.

I then ran with lldb and was able to track down the issue. I am suspecting it is related to recursive class parsing.

Here is the info i got out of lldb:
Backtrace:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=2, address=0x16f603fe0)
  * frame #0: 0x000000010011ca84 doxygen`std::__1::__wrap_iter<BaseClassDef const*>::__wrap_iter(this=0x000000016f603ff8, __x=0x0000600001757640) at wrap_iter.h:158 [opt]
    frame #1: 0x000000010011ca74 doxygen`std::__1::vector<BaseClassDef, std::__1::allocator<BaseClassDef> >::__make_iter(this=<unavailable>, __p=<unavailable>) const at vector:1530:12 [opt]
    frame #2: 0x0000000100102fcc doxygen`std::__1::vector<BaseClassDef, std::__1::allocator<BaseClassDef> >::begin(this=<unavailable>) const at vector:1547:12 [opt]
    frame #3: 0x0000000100102e20 doxygen`ClassDefImpl::isBaseClass(this=0x0000600000c6a1f0, bcd=0x0000600000e71440, followInstances=true) const at classdef.cpp:3378:27 [opt]
    frame #4: 0x0000000100102ea4 doxygen`ClassDefImpl::isBaseClass(this=0x0000600000c6a1f0, bcd=0x0000600000e71440, followInstances=true) const at classdef.cpp:3389:20 [opt]
[...]

PR #9716, opened 2022-12-06
Issue #9715 prevent recursion

Barokoli:
Prevent recursion for abstract c# classes that inherit from themselves.
Moved the recursion check in front of the function call in isBaseClass in classdef.cpp since the function call itself could produce an infinite recursion.
Added an equality check to prevent an infinite loop in diagram.cpp


albert-github:
Seen the comments with #9715 this problem is fixed and thus this proposed pull request is obsolete.

Issue #9719, opened 2022-12-12
\ref command does not insert title of referenced page (works on 1.8.14, broken in 1.9.5)


KasperHesse:
**Describe the bug**
Under Doxygen 1.8.14, using the `\ref` command to reference a named page, the inserted link would take on the title of the page in question. Under Doxygen 1.9.5, this no longer works, and the link is named after the page's `<name>` and not the title.

**Expected behavior**
The link should reflect the title of the page.

**Screenshots**
Expected result (from Doxygen 1.8.14)
![image](https://user-images.githubusercontent.com/47347446/207039609-10a88aa2-d0b2-4392-b73d-4e4a4392ef5c.png)

Actual result (from Doxygen 1.9.5)
![image](https://user-images.githubusercontent.com/47347446/207039643-0e124e57-96b9-4ea9-bfb7-58d906c9671f.png)

**To Reproduce**
```bash
mkdir doxy && cd doxy
echo "@mainpage The main page\nThis is the main page referencing \\ref pPage1 and @ref pPage2" >> main.md
echo "@page pPage1 Page number one\nThis is page number one" >> p1.md
echo "\\page pPage2 The second page\nThis is the second page" >> p2.md
doxygen .
```

**Version**
1.9.5 vs 1.8.14. 
lsb_release:
``` 
LSB Version:    :core-4.1-amd64:core-4.1-noarch:cxx-4.1-amd64:cxx-4.1-noarch:desktop-4.1-amd64:desktop-4.1-noarch:languages-4.1-amd64:languages-4.1-noarch:printing-4.1-amd64:printing-4.1-noarch
Distributor ID: Rocky
Description:    Rocky Linux release 8.7 (Green Obsidian)
Release:        8.7
Codename:       GreenObsidian
```

albert-github:
We see that the translation of the `@page is done as follows:
```
======== Markdown =========
---- input -------
@page pPage1 Page number one
This is page number one

---- output -----
@page md_p1 Page number one\ilinebr @anchor pPage1
This is page number one

=========
```

The extra anchor command was necessary in doxygen but probably we should add an extra option to the `@anchor` containing the label like:
```
@anchor{<label>}
```
which would in the given case give as resulting markdown output:
```
@page md_p1 Page number one\ilinebr @anchor{Page number one} pPage1
 ```

WilliamHeuser:
[...]

PR #9728, opened 2022-12-16
issue #9719 \ref command does not insert title of referenced page

albert-github:
Adding the `title` as optional text as option to the `\anchor` command.

### DIFF
--- a/doc/commands.doc
+++ b/doc/commands.doc
@@ -2191,11 +2191,14 @@ Commands to create links
   This command adds (text) to the \LaTeX , DocBook and RTF index.
 
 <hr>
-\section cmdanchor \\anchor <word>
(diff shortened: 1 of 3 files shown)


albert-github:
Investigating ...

albert-github:
Found solution.

doxygen:
@albert-github In this case, I found it better to introduce an internal `@ianchor` command, as the title is only really needed internally.

albert-github:
I didn't choose for the `\anchor` version as:
- I didn't see any risk in adding the option
- I think it might have advantages to have an anchor with a default text (so one can refer to it just with the anchor id and one doesn't have to provide the text every time one uses the anchor. Might be useful in case of multiple usages of parts that are no sections etc. but subparts of sections etc. that one wants to address).

PR #9729, opened 2022-12-18
Consistency of end of program

albert-github:
In case `-d time` is used the "finished...` was not printed, made consistent

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -12693,6 +12693,14 @@ void generateOutput()
          Portable::getSysElapsedTime()
         );
     g_s.print();
+    bool restore=FALSE;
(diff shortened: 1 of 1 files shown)


src/doxygen.cpp
```diff
@@ -12693,6 +12693,14 @@ void generateOutput()
          Portable::getSysElapsedTime()
         );
     g_s.print();
+    bool restore=FALSE;
+    if (Debug::isFlagSet(Debug::Time))
```
doxygen, line 12697: the same check is already done at line 12689 it seems, so I would expect this guard is always true and the code can be simplified.

albert-github: Indeed, fixed.

PR #9753, opened 2022-12-29
Automatic generation history information for commands / tags

albert-github:
Automatic generation of history information in internal documentation for commands and tags from version 1.9.0 onward

### DIFF
--- a/doc_internal/cmds_tags.py
+++ b/doc_internal/cmds_tags.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+import argparse, glob, itertools, re, shutil, os, sys
+import subprocess
+import shlex
+
+
+cmds_reg = re.compile(r'.refitem *cmd[a-z]* *')
+tags_obsolete_reg = re.compile(r"^.*option *type=.obsolete.*id='([^']*)'.*")
+tags_active_reg = re.compile(r"^.*option .*id='([^']*)'.*")
+
+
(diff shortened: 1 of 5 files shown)


doc_internal/cmds_tags.py
```diff
@@ -0,0 +1,182 @@
...
+            proc = subprocess.Popen(shlex.split(cmd),stdout=subprocess.PIPE,stderr=subprocess.PIPE,encoding=encoding) # Python 3 with encoding
+            return proc.stdout.read()
+
+
+def get_versions():
+    raw_versions = xpopen('git tag')
+    versions = []
+    strip_versions = []
+    versions.append("Release_1_8_20")
+    pre = "Release_1_9_"
```
doxygen, line 48: will this continue to work for future version tags like e.g. `Release_1_10_0` or `Release_2_0_0`?

albert-github: No at that moment we have to one of the following
1. add another loop over `pre` with the new prefixes 
2. process the the first part after Release so we get the major and minor version number
3. move the "1.9" releases into the md file and change `Release_1_9_` in  `Release_1_10_` / `Release_2_0_` (and adjust `versions.append("Release_1_8_20")` as well

For this version I didn't want to make it too complicated, probably the second option would be the best

albert-github: Added functionality for versions like 1.10.0 and 2.0.0

Issue #9718, opened 2022-12-10
class diagram's width is not correct and class member appears weird


DamonsJ:
I installed doxygen from url: https://www.doxygen.nl/files/Doxygen-1.9.5.dmg in my macbook
I tried to generate class diagram from my test code(with config file in zip file):
[doxygen-test.zip](https://github.com/doxygen/doxygen/files/10199913/doxygen-test.zip)

macos: 12.4
install graphviz using : brew install graphviz
dot version is :
dot - graphviz version 7.0.4 (20221203.1631)

here is my problem:

 member function's name is too long to show in one line , so I tried to change the width to 0.8 value in "shape=record,height=0.2,width=0.2", and regenerate with "shape=record,height=0.2,width=0.8".
but it has no effect, it appears like :
![image](https://user-images.githubusercontent.com/28893028/206851583-ef7403b8-2c4b-4280-8636-a429a7104698.png)

so how can I make long function name show in one line?

by the way, the class member is the attached image start with '#', it is weird , right?
it should start with '+' or '-' ?

thanks!

albert-github:
The items listed with `#` are in my opinion correct as they are protected members. See your input file `d1.h` and e.g. https://en.wikipedia.org/wiki/Class_diagram

albert-github:
The multi column approach is not as straight forward as it looked like at first glance.

I see that you have some long function names that are cut into multiple parts, before (in an other issue #9717) I wrote:
> I didn't test it but as far as i know this is outside the scope of doxygen as the line break is done by the dot program.

this is not true, doxygen does some splitting which is based on the setting  `DOT_WRAP_THRESHOLD`, setting this to a larger value than the default might help already.

DamonsJ:
@albert-github  
thanks very much!
it works if I change the setting DOT_WRAP_THRESHOLD!

thanks!

albert-github:
I had a look at the "alignment problem and used the following code (original code from doxygen, node2 and image on the left, alternative HTML-like code, node3 image on the right):
```
digraph "Class" {
  bgcolor="transparent";

  edge [fontname=Helvetica,fontsize=10,labelfontname=Helvetica,labelfontsize=10];
  node [fontname=Helvetica,fontsize=10,shape=record,height=0.2,width=0.8];

[...]

PR #9758, opened 2022-12-31
issue #9718 class diagram's width is not correct and class member appears weird

albert-github:
Updating the UML look to use the HTML like labels of dot so that the alignment in the images is better.


albert-github:
investigating ...

albert-github:
fixed

PR #9773, opened 2023-01-05
Escape `-` sign in fontname settings substring as well

albert-github:
Not only substrings with white space have to be quoted but also when containing `-`

(Based on the comments https://github.com/doxygen/doxygen/issues/9319#issuecomment-1371783416 and https://github.com/doxygen/doxygen/issues/9319#issuecomment-1372056305)


albert-github:
Added extra proposed pull request #9775

PR #9785, opened 2023-01-11
Consistency  in HTML output with multiple runs

albert-github:
When running a project twice on the same source (leaving the HTML output in place) still the resulting HTML output differs slightly between both run. (Necessary is that interactive SVG is enabled)

It is just a cosmetic change in the output files, but helps when looking for differences between the 2 runs (with e.g. diff).
(The problem was found by Fossies)


doxygen:
@albert-github I decided to fix the problem at the source (i.e. remove the newline placed after the marker) instead of fixing it afterwards.

PR #9792, opened 2023-01-15
Collect warnings and output them at the end

albert-github:
Introduced a new setting for `WARN_AS_ERROR` i.e. `FAIL_ON_WARNINGS_PRINT`. Especially on CI systems when running large projects it can be hard to obtain all the warnings as either they are written to a file and the file is not accessible (anymore) or they are written between the progress messages. With this new option the warnings are given always at the end of the run.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10420367/example.tar.gz)


src/message.cpp
```diff
@@ -30,33 +30,41 @@ static const char *    g_warningStr = "warning: ";
...
-  if (!logFile.isEmpty())
+  if (!g_warnlogFile.isEmpty())
   {
-    if (logFile == "-")
+    if (g_warnlogFile == "-")
     {
       g_warnFile = stdout;
     }
-    else if (!(g_warnFile = Portable::fopen(logFile,"w")))
+    else if (!(g_warnFile = Portable::fopen(g_warnlogFile,"w")))
```
Ryanf55, line 57: Hey there, if this log file has a foldername in it, but the folder does not exist, this call is not creating the folder. Then, the log error message appears, and then my CI instance looking for the log output in that folder does not find it.

PR #9795, opened 2023-01-16
Handling of verbatim type sections in preprocessor

albert-github:
A number of problems were identified in `pre.l`
- missing verbatim type commands i.e. `startuml` and `msc`
- conditions for formulas (`\f$` etc.) were not defied properly and also the end conditions were not defined properly
- verbatim sections were not handled properly inside non used `cond` parts, they still should be handled properly so also a terminating `\endcond` is handled here OK


doxygen:
@albert-github I tried to reduce the amount of code duplication (see https://github.com/doxygen/doxygen/commit/f2c31cb36e8ff17db8992a9283421a79ad9e0bcd). Please check if I might have overlooked something.

albert-github:
@doxygen Unfortunately there was a small regression regarding the formula ends, see #9803

Issue #9781, opened 2023-01-10
Doxygen regression when creating PDF output 1.9.6 -> master of 10.01.2023


StefanBattmer:
**Describe the bug**
The PDF generation is no longer generating an output but stops with an error message now:

```
! Undefined control sequence.
l.3 ...of the manual, you will get an \doxysectref
                                                  {introduction}{p.}{mvBF_pa...

? 
! Emergency stop.
l.3 ...of the manual, you will get an \doxysectref
                                                  {introduction}{p.}{mvBF_pa...

!  ==> Fatal error occurred, no output PDF file produced!
```

**Expected behavior**
A correctly generated PDF document

**Screenshots**
none

**To Reproduce**
Could try to come up with something when the attached snippets don't ring a bell but I am afraid that this might turn out difficult this time...

**Version**
1.9.6 is still working fine, The master from today (10.01.2023) is not

**Stack trace**
Have attached the 2 relevant sections from the output instead for diffing. Hope that helps.

[DoxygenIssue9781.zip](https://github.com/doxygen/doxygen/files/10381631/DoxygenIssue9781.zip)

albert-github:
From the 2 text files we see that in the 1.9.6 version you already got a warning:
```
LaTeX Warning: Reference `mvBF_page_introduction' on page 1 undefined on input
line 3.
```
and apparently this gives a fatal error in the current master 1.9.7 (19420b7ae5fb764654df9d949d62429ae1432df2) due to the fix #9761 so looks like the problem was already present but never noticed.

It is hard to tell what caused the warning in the 1.9.6 version in the first place (did doxygen give a warning regarding `mvBF_page_introduction` ?
There is probably a source file that refers to `mvBF_page_introduction` (`\ref`, \link\`, ...) but it is not possible to tell what happened.

To be able to find the problem:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

(Note when referring to a version please give the information from `doxygen -v` and not just a date (might be multiple versions created on one day.)

StefanBattmer:
Thanks @albert-github for pointing me to #9761! We will first dig deeper on our side then!

albert-github:
Well I think this problem has 2 sides:
- your side fixing the problem
- doxygen side shouldn't crash, so when you found the problem on your side it might be "easily" possible to create a small example so we can see what we can do on the doxygen side as well to prevent the crash.

[...]

PR #9805, opened 2023-01-21
Prevent list items in section titles

albert-github:
When having the, a bit unusual, construct:
```
\section ThePage_subsection 1. The First Sub-Section
```
and we use the reference like:
```
\ref ThePage_subsection
```
we get in the LaTeX generation the error:
```
! LaTeX Error: Something's wrong--perhaps a missing \item.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...

l.4 \end{DoxyEnumerate}}}
```
due to the construct:
```
Check it here \mbox{\hyperlink{_the_page_ThePage_subsection}{
\begin{DoxyEnumerate}
\item The First Sub-\/\+Section
\end{DoxyEnumerate}}}
```

This is all because of the `1.` i.e. a list item in the section title.

(all based on the discussion in:  Doxygen regression when creating PDF output 1.9.6 -> master of 10.01.2023 #9781 )

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10472228/example.tar.gz)

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -1599,7 +1599,7 @@ int DocParser::internalValidatingParseDoc(DocNodeVariant *parent,DocNodeList &ch
 
   if (doc.isEmpty()) return retval;
 
-  tokenizer.init(doc.data(),context.fileName,context.markdownSupport);
+  tokenizer.init(doc.data(),context.fileName,context.markdownSupport,context.insideHtmlLink);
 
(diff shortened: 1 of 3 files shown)


StefanBattmer:
@albert-github: I have tested your patch! Works like a charm! Thanks!

PR #9806, opened 2023-01-21
Layout of copyright notice in manual

albert-github:
- the copyright year and the word "by" were visibly not separated
- the `"as is"` could be broken because of the used space, this is prevented.

### DIFF
--- a/doc/index.doc
+++ b/doc/index.doc
@@ -122,12 +122,12 @@ The third part provides information for developers:
 \addindex license
 \addindex GPL
 
-Copyright &copy; 1997-\showdate "%Y" by <a href="mailto:doxygen@gmail.com">Dimitri van Heesch</a>.<p>
+Copyright &copy; 1997-\showdate "%Y" &nbsp;by <a href="mailto:doxygen@gmail.com">Dimitri van Heesch</a>.<p>
 
 Permission to use, copy, modify, and distribute this software and its
 documentation under the terms of the GNU General Public License is hereby
 granted. No representations are made about the suitability of this software
-for any purpose. It is provided "as is" without express or implied warranty.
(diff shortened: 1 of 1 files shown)


doc/index.doc
```diff
@@ -122,12 +122,12 @@ The third part provides information for developers:
 \addindex license
 \addindex GPL
 
-Copyright &copy; 1997-\showdate "%Y" by <a href="mailto:doxygen@gmail.com">Dimitri van Heesch</a>.<p>
+Copyright &copy; 1997-\showdate "%Y" &nbsp;by <a href="mailto:doxygen@gmail.com">Dimitri van Heesch</a>.<p>
```
doxygen, line 125: I understand this will force a space, but I think it would be more intuitive if the space after `"\showdate %y"` is preserved. What do you think?

albert-github: That would have been nice but I think that again would limit probably other possibilities where it would be possible to join things directly with the (in this case year), like the result of other commands as at that moment the string would not a proper forma string. In a bit an artificial example:
```
\showdate "%Y" \showdate "%Y"
```
would result now in `20232023` but when the space is retained it would be `2023 2023`.
when using
```
\showdate "%Y"\showdate "%Y"
```
we get:
```
warning: invalid <date_time> argument for command '\showdate'
```

(I know that in this special case I could use "%Y%Y")

doxygen:
@albert-github I think I found a way that keeps the space if there is one, while allowing the optional <date_time> parameter if the pattern matches, and if not then exclude the remaining characters from being part of the \showdate command. See https://github.com/doxygen/doxygen/commit/07d9f9fa03e44683500f3142c48b744755d428b0 for the fix and some examples.

Issue #9802, opened 2023-01-20
doxmlparser can't properly parse enumerated types


dhowland:
**Describe the bug**
The generated code that implements doxmlparser apparently has a bug.

When parsing enumerated value types, the `name` property is always `None`.  In other words, you can't read the enumerated values.

I have fixed the bug in the following commit to my fork, here: 19d77e6ce50ccbf3877af9fc892b0f33c373569c

That commit log contains further context.  Given that this is generated code, I can't send a pull request.  Honestly I'm not sure if this is something that is fixable, but I discovered the issue and fixed it, so I'm making a public record.

**Expected behavior**
The `name` property should be populated.

**To Reproduce**
See attached demonstration.
[bugdemo.zip](https://github.com/doxygen/doxygen/files/10469658/bugdemo.zip)

**Version**
Doxygen 1.9.5
Python 3.10

albert-github:
I'm not sure about this, looking at the generated code we see:
```
    def buildChildren(self, child_, node, nodeName_, fromsubclass_=False, gds_collector_=None):
        if nodeName_ == 'name' and child_.text is not None:
            valuestr_ = child_.text
            valuestr_ = self.gds_parse_string(valuestr_, node, 'name')
            valuestr_ = self.gds_validate_string(valuestr_, node, 'name')
            obj_ = self.mixedclass_(MixedContainer.CategorySimple,
                MixedContainer.TypeString, 'name', valuestr_)
            self.content_.append(obj_)
            self.name_nsprefix_ = child_.prefix
```

and in the `class MixedContainer`:
```
    def __init__(self, category, content_type, name, value):
        self.category = category
        self.content_type = content_type
        self.name = name
        self.value = value
    def getCategory(self):
        return self.category
    def getContenttype(self, content_type):
        return self.content_type
    def getValue(self):
        return self.value
    def getName(self):
        return self.name
```
so wouldn't it be possible  to obtain the "value" of the "name" in this way?

dhowland:
Well there is no public getter method that produces `content_`.  However the implementation of `bugdemo.py` would look like this:

'''
for compounddef in root.get_compounddef():
    for sectiondef in compounddef.get_sectiondef():
        if sectiondef.get_kind() == DoxSectionKind.ENUM:
            for memberdef in sectiondef.get_memberdef():
                for enumvaluedef in memberdef.get_enumvalue():
                    for item in enumvaluedef.content_:
                        if item.getName() == 'name':
[...]

PR #9809, opened 2023-01-22
Issue #9802 enumvalueType should not be a mixed type

dhowland:
This will fix a quirk with doxmlparser.py that makes it difficult to parse Enums.


dhowland:
My original commits were intended for discussion, not for merging.  I suggest you squash this if you intend to merge.

I apologize that I did not regenerate compound.py myself, I don't have the necessary toolchain on my Windows machine.

PR #9819, opened 2023-01-26
Autotrim snippet left whitespace

devel0:
This PR address problem asked [here](https://stackoverflow.com/questions/55955554/remove-leading-indentation-white-space-in-doxygen-snippet) by trimming automatically left whitespace for a length max possible common to all lines of given snippet.

The data of related unit test changed accordingly, following a shot of what changes about:

image


albert-github:
It is not good to refer to an external example as this might not be persistent, so add it always to the PR / issue (when reasonably possible).

Extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10518372/example.tar.gz)


Some remarks:
- in some languages the starting spacing is important, e.g. fixed format Fortran code
- what to do with `<tabs>` as they are not handled now (`<tabs>` are not a good idea anyway in my opinion, but this a matter of taste).
- what will happen when mixing `<spaces>` and `<tabs>` (see also the setting `TAB_SIZE`).

devel0:
I agree with your assessment, an option could be to add a config var like this:

**Doxygen config**

`SNIPPET_TRIMLEFT = NO`
Default behavior ( no trim at all )

`SNIPPET_TRIMLEFT = YES`
As following ( note: TTTT represents tab space conversion respect to TAB_SIZE )

image

[trim-left-tab-handling.docx](https://github.com/doxygen/doxygen/files/10519784/trim-left-tab-handling.docx)

Further more a third option could

 `SNIPPET_TRIMLEFT = TABCONVERT`

image

[trim-left-tab-convert-handling.docx](https://github.com/doxygen/doxygen/files/10519807/trim-left-tab-convert-handling.docx)

devel0:
I reverted changes to the unit test 049 having that is honored within default SNIPPET_TRIMLEFT = NO option.

Added SNIPPET_TRIMLEFT enum config option ( NO, YES ) and implemented behavior described for the YES option.

Following html view of case1:
html1

case2:
html2

note aside, the html output I see that trims left whitespaces so that my case2 results without beginning space, it's not a problem btw this is the result of the function about the case:
html2_res

src/config.xml
```diff
@@ -565,6 +565,17 @@ Go to the <a href="commands.html">next</a> section or return to the
...
       </docs>
     </option>
+    <option type='enum' id='SNIPPET_TRIMLEFT' defval='NO'>
+      <docs>
+<![CDATA[
+ Inserted snippet will not trimmed by default.
+ 
+ Setting \c SNIPPET_TRIM_LEFT to YES will remove lines common starting spaces/tab.
+]]>
+      </docs>
```
albert-github, line 575: - ... will not trimmed .... -> ... will not be trimmed ....
- to YES -> to \c YES
- there is an inconsistency in the name `SNIPPET_TRIMLEF` and later `SNIPPET_TRIM_LEFT`
- the line starting with "Setting" could be made a bit more explicit and also refer to the `TAB_SIZE` setting, maybe something like:
  ```
  Setting \c SNIPPET_TRIMLEFT to \c YES will remove from the lines in the snippet the starting spacing that is common to all the lines (for `<tabs>` also the setting of  \ref cfg_tab_size "TAB_SIZE" is taken into consideration.
  ```
- why is an `enum` used and not a `bool` as type?

src/util.cpp
```diff
@@ -6139,39 +6139,82 @@ std::vector<std::string> splitString(const std::string& str)
  * such that it does not remove non-spacers
  */ 
 QCString stripLeft(const QCString& text)
-{
+{
```
albert-github, line 6142: See to it that there is no empty spacing at the end of a lien.

src/config.xml
albert-github: regarding the last remark, regarding `enum` <-> `bool` this is probably because of the comment that it still in the code:
>   // to be used in case of implement SNIPPET_TRIMLEFT = TABCONVERT
>   // auto tabSize = Config_getInt(TAB_SIZE); 

so looks like some work in progress.

src/util.cpp
```diff
@@ -6139,39 +6139,82 @@ std::vector<std::string> splitString(const std::string& str)
  * such that it does not remove non-spacers
  */ 
 QCString stripLeft(const QCString& text)
-{
+{  
   auto lines = splitString(text.str());
+  
+  // to be used in case of implement SNIPPET_TRIMLEFT = TABCONVERT
+  // auto tabSize = Config_getInt(TAB_SIZE);
```
albert-github, line 6146: Why not always do the `detabbing`?

In `markdown.cpp` there is a method `QCString Markdown::detab(const QCString &s,int &refIndent)` so it might be useful to reuse this routine (as it calculates the minimum start spacing and translates the `tabs` to spaces).
So do something like first detab the original text and on the result perform the trimming.

devel0: thanks you for your double checks:
- enum is used for future extension and specialization of behavior as you noted
- replaced `(for `<tabs>` also the setting of  \ref cfg_tab_size "TAB_SIZE" is taken into consideration.` because effectively the "YES" option would not consider TAB_SIZE actually it will be used eventually for a mixed mode TABCONVERT
- I take a look to the `QCString Markdown::detab(const QCString &s,int &refIndent)`

devel0: Tried `detab` but doesn't work for the need, it doesn't remove spaces ( [see here](https://github.com/doxygen/doxygen/blob/aed5276eb131252a6d4eaf4b00ee145127541f1f/src/markdown.cpp#L3227) )

image

albert-github: You return the `detab` text but this is just changing tabs into spaces and determines the `refIndent`, this should be fed into `stripLeft` and here the "superfluous" spaces at the left could be removed.

devel0: You right, but effectively seems to me a forcing of modularization, actually proposed PR do the thing producing expected result, of course if you feel that this is necessary you could change as your need. (note: detab is a private method and the environment which planned to be used is for markdown text, future changes to its behavior could brake desired effect of SNIPPER_TRIMLEFT behavior ).
One thing that missed is the unit test, when I have some space I will add if you consider the PR could ok.

albert-github:
I've been doing some thinking about the underlying issue here.

**Usage of tabs**

The usage of tabs always is in for some surprises especially when (horribly) spaces and tabs are mixed.
Say we have a small example like:
```
  void fie()
  {
    printf("Hello world\n");
  }
```
and
```
  void fie()
  {
    printf("Hello world\n");
  }
```
they both look ok but when we  look in the code they are different as in the second case there are 2 spaces followed by a tab (tab setting is 5).
In the above `S/T` notation this looks like:
```
SSvoid fie()
SS{
SSSSprintf("Hello world\n");
SS}
```
and
```
SSvoid fie()
SS{
SSTprintf("Hello world\n");
SS}
```
after trim left this would lead to (in `S/T` notation again:
```
void fie()
{
SSprintf("Hello world\n");
}
```
and
```
void fie()
{
Tprintf("Hello world\n");
}
```

combining this in a mockup would lead to:

image

and here we see that in the tab indented trim left version the printf is to far to the right as the tab is expanded after the trim left and thus results in 4 spaces again whilst normally it would also take into account the 2 initial spaces (a tab is not a fixed number of spaces but jumps to the next tab position).

The function `detab` in markdown is only used in the `markdow.cpp` but it doesn't use / convert any markdown facilities, so it would be easy to make it a publicly accessible  or a global function.

So I think the detabbing is always necessary.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10529465/example.tar.gz)


**Usage of the setting SNIPPET_TRIMLEF**

This means an extra setting variable and it will also apply to all (code) snippets.
I think it might be better to have an option with the `\snippet` command with the name `trimleft` so one can write things like:
```
\snippet{trimleft} examples.cpp before::fixup
```

(the `\snippet` command does already some option handling).

devel0:
Thanks for detailed explanation, I agree with all your points; actually I haven't space to allocate now but in future I get back to this topic and if not yet solved by someone else I could work on it further in directions you pointed here.
My intent was to forward a little bit the topic about this issue in doxygen and I think the target was in that sense reached:
- we know doxygen actually lacks this feature and could useful to have it implemented
- we have some rationalized specs to work on it

albert-github:
I've just pushed an alternative approach, i.e. proposed patch, pull request #9827

albert-github:
Code of #9827 has been integrated in master on GitHub

Issue #6381, opened 2018-07-03
search.js causes Content Security Policy violation with evaluated javascript


ArktekniK:
When hosting generated HTML documentation any on page that implements the content security protocol header with (even quite permissive values), the search functionality is broken. Specifically, this affects us when we are hosting API documentation on Jenkins CI, using the publishHTML api.

This is due to the fact that search.js attempts to use an unsafe eval when updating the search box with results. Although this instance of eval is implicit, the following page clarifies some of the security concerns. [See MDN here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval#Do_not_ever_use_eval!)

Please see chrome error from console:
```
Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of 
script in the following Content Security Policy directive: "script-src 'self' 'unsafe-inline'".

SearchBox.OnSearchFieldChange @ search.js:204
```
You can see there that we have even relaxed the CSP header values, but it's not enough. To solve this issue, a server would need to send the `script-src unsafe-eval` value to the browser. Obviously this is a security concern, so I would like to propose a fix.

The offending code is here:
```js
if (searchValue != "") // non-empty search
{
    // set timer for search update
    this.keyTimeout = setTimeout(this.name + '.Search()', // implicit eval
                                 this.keyTimeoutLength);
}
```
I presume that the `name` is used to solve the issue whereby the function passed to `setTimeout` is executed on the `window` context, rather than the appropriate `this`. It is not necessary and there are better ways to solve that problem.

The following alternative will work. See [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind#Browser_compatibility)
```
// set timer for search update
this.keyTimeout = setTimeout(this.Search.bind(this),  // Bind the correct context to the Search function
                              this.keyTimeoutLength);
```

Thanks!

albert-github:
Robert,

Please create an issue for this in the issue tracker on github (
https://github.com/doxygen/doxygen/issues/new  doxygen recently moved bug
tracer from Bugzilla to Github!).


Albert

ArktekniK:
Hi Albert,

Not sure I understand... I did file the bug on GitHub!

Rob

albert-github:
Hi Rob,

Was a mistake from my side I thought it was coming from a doxygen forum, but a little bit later I saw that it was already from github (perfect), so I removed the comment again on github.
[...]

PR #9830, opened 2023-02-01
avoid csp insecure eval on search box.

sbiscigl:
From the issue [issues/6381](https://github.com/doxygen/doxygen/issues/6381) there was a suggestion on an update to the seach.js template to avoid csp violations for `unsafe-eval` which are currently seeing in our published docs. 

`Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'unsafe-inline'".`

This adds the suggested change from the issue, combing through the comments on the issue, i don't see any blocking comments, rather that other places need to be updated unrelated to search. This is to fix the search issue we are seeing. We have created and are using a fork.

### DIFF
--- a/templates/html/search.js
+++ b/templates/html/search.js
@@ -156,7 +156,7 @@ function SearchBox(name, resultsPath, extension)
 
   this.OnSearchSelectHide = function()
   {
-    this.hideTimeout = setTimeout(this.name +".CloseSelectionWindow()",
+    this.hideTimeout = setTimeout(this.CloseSelectionWindow.bind(this),
                                   this.closeSelectionTimeout);
   }
 
@@ -211,7 +211,7 @@ function SearchBox(name, resultsPath, extension)
       if (searchValue != "") // non-empty search
       {
         // set timer for search update
-        this.keyTimeout = setTimeout(this.name + '.Search()',
+        this.keyTimeout = setTimeout(this.Search.bind(this),
                                      this.keyTimeoutLength);
       }
       else // empty search field


templates/html/search.js
```diff
@@ -156,7 +156,7 @@ function SearchBox(name, resultsPath, extension)
 
   this.OnSearchSelectHide = function()
   {
-    this.hideTimeout = setTimeout(this.name +".CloseSelectionWindow()",
+    this.hideTimeout = setTimeout(this.Search.bind(this),
                                   this.closeSelectionTimeout);
```
albert-github, line 160: I have my doubts that this is correct as when on selects the magnifying glass in the search box we get a drop-down box with e.g "All" / "Classes" / "Namespaces" / "Functions" / ... and this drop-down box remains when it looses focus whilst in the current version the drop-down box vanishes.

sbiscigl: yeah you're right, let me investigate a bit

sbiscigl: alright updated the PR and tested for that test case, that drop down box should now disappear when not in focus.

sbiscigl:
Hey sorry to bother @albert-github jus wondering if you could take a peek again, i think i addressed the concerns, give a shout if im missing something though

albert-github:
I had already a look at it and the result was now OK, though I cannot say / don't know anything about the underlying csp problem

sbiscigl:
from testing, at least for us, ti fixed the search/csp issue. apologies as i am a first time contributor here, but curious if you need anything else from me in the review process.

albert-github:
Maybe @doxygen has some review comments, I don't have (but like I said I know anything about the underlying csp problem)

doxygen:
Looks ok, so I've merged it.

sbiscigl:
Thanks!

Issue #9831, opened 2023-02-02
@exception description paragraph does not accept @ref tag


MarkShackman:
Both @param and @exception say that their description "paragraph has no special internal structure. All visual enhancement commands may be used inside the paragraph."

However, @exception descriptions don't accept @ref commands, whereas @param descriptions do:

```
## @file throwsBug

##
# Example class
class Example():

  def function(p):
    """ Example function
    """
    pass

  ## Documentation using doxygen style comment
  # @param p @ref throwsBug.Example.function() "function()"
  # @throws @ref throwsBug.Example.function() "function()"
  def useDoxygenStyle(p):
    pass
```
Using doxygen 1.9.6 with an unchanged Doxyfile.

albert-github:
The documentation of `\throws` gives:

> 24.103 \throws <exception-object> { exception description }


and as \`throws` is synonymous with `\exception`:

> 24.79 \exception <exception-object> { exception description }
>
> Starts an exception description for an exception object with name <exception-object>. Followed by a description
of the exception. The existence of the exception object is not checked. The text of the paragraph has no special
internal structure. All visual enhancement commands may be used inside the paragraph. Multiple adjacent
\exception commands will be joined into a single paragraph. Each exception description will start on a new
line. The \exception description ends when a blank line or some other sectioning command is encountered.
See section \fn for an example.

I think the problem is that you forgot to mention the `<exception-object>` with the `\param` we can see that you used `p` as `<parameter-name>`.

Adding the `<exception-object>` will solve the problem.

I think the problem is usage.

MarkShackman:
Ah, true.
Perhaps refactoring has tightened up on syntax, as the following is working for me under doxygen 1.8.18:
```
   @throws [@ref exceptions.SLException.SLException "SLException"] if unable to communicate
```
The 1.8.18 build uses an edited doxyfile with doxypy.py as a filter, so I'll have to investigate further....

albert-github:
When I use your syntax from https://github.com/doxygen/doxygen/issues/9831#issuecomment-1414106515 in 1.8.18 I get (and a warning):

```
throwsBug.py:14: warning: unexpected token in comment block while parsing the argument of command throws
```

![image](https://user-images.githubusercontent.com/5223533/216399865-c64cedd1-75f4-43af-a500-b0b325f89bf3.png)


so likely that filter does something with the `@exception`

MarkShackman:
[...]

PR #9833, opened 2023-02-03
issue #9831 @exception description paragraph does not accept @ref tag

albert-github:
Handle the `"""` in the comment converter properly in case `PYTHON_DOCSTRING=NO` so e.g. `ALIASES` are converted as well.

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -1233,12 +1233,7 @@ void convertCppComments(const BufStr &inBuf,BufStr &outBuf,const QCString &fileN
   clearCommentStack(yyscanner);
   yyextra->vhdl = FALSE;
 
-  if (Debug::isFlagSet(Debug::CommentCnv))
-  {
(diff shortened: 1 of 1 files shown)


src/commentcnv.l
```diff
@@ -1214,7 +1233,12 @@ void convertCppComments(const BufStr &inBuf,BufStr &outBuf,const QCString &fileN
   clearCommentStack(yyscanner);
   yyextra->vhdl = FALSE;
 
-  DebugLex debugLex(Debug::Lex_commentcnv,__FILE__, qPrint(fileName));
+  if (Debug::isFlagSet(Debug::CommentCnv))
```
doxygen, line 1236: did you mean to remove the DebugLex line here?

albert-github: No I didn't have the intention to remove that line, during my debugging I wanted to see the input of commentcnv and forgot to revert this.

I reverted it now.

PR #9841, opened 2023-02-09
docs: Ensure examples' "click here for output"  HTML links point to the most useful place

ferdnyc:
Since output examples are supposed to help the user understand code by way of practical demonstration, ensuring that each of the manual's "Click here for the corresponding HTML documentation..." links points to the most useful demonstration of the associated code is a worthwhile goal.

While _most_ of those links already are pointed at optimal locations within the sample output, a few were not, pointing at the (often totally blank) main index page, or to a particular class documentation page that leaves the reader with no easy means of finding the _rest_ of the sample content (unless they resort to exploring the navigation bar at the top of the page).

This PR is a result of me going through the _entire_ manual, finding (hopefully) every example with an associated output link, and making sure that the page it points to is the best possible demonstration of the concept being illustrated by that particular chunk of source code. (I _think_ I looked at all of them, although I can't promise I didn't miss any.)

For each one that I felt could do a better job of demonstrating the associated Doxygen feature, I updated the link to target what I felt was the best possible representation of the code:output relationship.

In choosing targets, I focused on delivering the reader to a page from which they could reach **all** of the output using nothing more than the links right on that page. IOW, without forcing them to use the navbar. So, some links might point to the class or module overview page, rather than the documentation of any one class. Even though the target page has much **less** content that way, it's the one that directly links to **ALL** of the content, in a way that a child class page typically doesn't. (At least, not directly.)

This PR contains one commit per file updated, which is nearly a one-commit-per-link granularity. But I'd fully encourage squashing it into a single commit for merging, or if it's preferred I'll happily do that myself and update the branch.


albert-github:
Good that a number of places have been corrected. Hopefully they are now in accordance with the LaTeX output.

You already wrote:
> This PR is a result of me going through the entire manual, finding (hopefully) every example with an associated output link, and making sure that the page it points to is the best possible demonstration of the concept being illustrated by that particular chunk of source code. (I think I looked at all of them, although I can't promise I didn't miss any.)

though in the `commands.doc`  I think there are 3 missed, with the following commands:
- `\extends`
- `\implements`
- `\memberof`

ferdnyc:
> though in the `commands.doc` I think there are 3 missed, with the following commands:
> 
> * `\extends`

Ah! I missed that one because there was no source code block preceding it. My method was basically to scan for blocks of `<pre>`-formatted source and look for a link underneath, which clearly was not the best approach.

> * `\implements`
> * `\memberof`

And, ditto and ditto, on the other two.

I've updated all three to link to the documentation for `manual.c` itself. Thanks!

ferdnyc:
Hmm. Looking at `\example` again... I originally left it alone, because it was already linking to https://www.doxygen.nl/manual/examples/example/html/examples.html.

But that page is really kind of spare. Perhaps the link should go directly to the `Example_Test` class reference, instead. That page links to the `example_test.cpp` doc, so everything is accessible from there. ...It links there in two different places, in fact.

WDYT?

albert-github:
I think that the best way would be to have it link to the place that is best suited and is consistent with the LaTeX output.
Probably (for the `\extends` command) would be to link to struct_car.html

ferdnyc:
> I think that the best way would be to have it link to the place that is best suited and is consistent with the LaTeX output.
> Probably (for the `\extends` command) would be to link to struct_car.html

Mmmm, I see what you mean. Yeah, that probably is better.

Or how about `struct_vehicle.html`? That way they have links to both parent and child classes, and the multiple subclasses are visible. (Linking to either Car or Truck effectively hides the other one from view, and makes the inheritance look like just a straight line.)

ferdnyc:
Eh, I changed them all again so they link to `struct_car.html`, as you said. I don't know why I'm being so difficult, force of habit I guess.

albert-github:
Might also be a good candidate I have no real preference, but we also have to se to it that the LaTeX output will link to the same place.

ferdnyc:
> Might also be a good candidate I have no real preference, but we also have to se to it that the LaTeX output will link to the same place.

\*nod\* I just went through the HTML docs and the PDF, side-by-side, and tweaked several of them... IMHO every HTML link is now the "moral equivalent" of its LaTeX counterpart.

(A couple aren't _precisely_ equivalent, because for instance in the PDF the example section for [Modules](file:///home/ferd/rpmbuild/REPOS/doxygen/_build_lexfiles/html/grouping.html#Modules) starts with the first group. But doing that in HTML would give you no easy way to see the rest of the groups, so instead I link to the module list page, something that doesn't exist in the PDF.)

ferdnyc:
And I fixed my VHDL link, which was missing the HTML extension. Duh.

PR #9842, opened 2023-02-09
Fix variable name collision with clang mingw

Biswa96:
This changes the variable name from 'stat' to 'isStatic'.
    Otherwise the 'stat' name come into conflict with 'stat'
    macro in mingw-w64 headers. This fixes the following error:

    commentscan.l:3074:21: error: no member named '_stat64' in 'Entry'
      yyextra->current->stat = TRUE;
      ~~~~~~~~~~~~~~~~  ^

    The 'stat' macro is defined in sys/stat.h file in mingw-w64
    like this '#define stat _stat64'

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -3070,7 +3070,7 @@ static bool handleInternal(yyscan_t yyscanner,const QCString &, const StringVect
 static bool handleStatic(yyscan_t yyscanner,const QCString &, const StringVector &)
 {
   struct yyguts_t *yyg = (struct yyguts_t*)yyscanner;
-  yyextra->current->stat = TRUE;
+  yyextra->current->isStatic = TRUE;
   return FALSE;
 }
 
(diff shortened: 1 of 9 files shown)


Biswa96:
I can try add CI build with mingw gcc and clang if required.

albert-github:
Regarding:
> I can try add CI build with mingw gcc and clang if required. Only 012_cite test fails in mingw.

Most likely problem is that you didn't install perl.
Maybe the `--noredir` option in the `TEST_FLAGS` with `make tests` gives some more information.

Biswa96:
perl (v5.32.1) is already installed in the mingw environment. There are some known issues about path handling with mingw perl. So, I did not focused on that part. By the way, this pull request is not related to that test failure.

albert-github:
I know the PR is not related to the test environment.

This might indeed be a path handling issue.
More information can be obtained when using `-d extcmd` and `-d cite` with doxygen. The later leaving some intermediate files on the disk for further inspection but this is less convenient in a CI environment.

maybe you should create a small script that calls doxygen with some options like
```
doxygen -d extcmd $*
```
and give the name of the script through the `TEST_FLAGS`  (option `--doxygen`) so you get the information about the external commands called.

PR #9844, opened 2023-02-09
Build: Re-enable LEX_FILES_{H,CPP}, streamline

ferdnyc:
This PR re-enables use of the CMake `${LEX_FILES_H}` and `${LEX_FILES_CPP}` variables, in a way that's compatible with CMake versions much older than 3.5, even -- this code should work with even CMake 3.0.

Support for older CMake versions is actually a moot point, since `deps/spdlog/CMakeLists.txt` contains this line:

https://github.com/doxygen/doxygen/blob/8887ee5387624fafa7a46dafa6a4f8183d45d1c6/deps/spdlog/CMakeLists.txt#L3

So, since the addition of spdlog the effective minimum CMake version for doxygen's own builds has been CMake 3.10. But I promise you, if that restriction were lifted the code in this PR would still work in much, _much_ older CMake releases.

#### Why?

The problem with older CMake versions wasn't support for listing source files in variables, it was how the variables were being constructed. Examining the `CMakeCache.txt` of a configured build tree made the issue apparent. Configuring a build tree from the current `master` branch produces:

```cmake
LEX_FILES_H:INTERNAL= ;/tmp/_build/generated_src/scanner.l.h; ;/tmp/_build/generated_src/code.l.h; ;/tmp/_build/generated_src/pyscanner.l.h; ;/tmp/_build/generated_src/pycode.l.h; ;/tmp/_build/generated_src/fortranscanner.l.h; ;/tmp/_build/generated_src/fortrancode.l.h; ;/tmp/_build/generated_src/vhdlcode.l.h; ;/tmp/_build/generated_src/pre.l.h; ;/tmp/_build/generated_src/declinfo.l.h; ;/tmp/_build/generated_src/defargs.l.h; ;/tmp/_build/generated_src/doctokenizer.l.h; ;/tmp/_build/generated_src/commentcnv.l.h; ;/tmp/_build/generated_src/commentscan.l.h; ;/tmp/_build/generated_src/constexp.l.h; ;/tmp/_build/generated_src/xmlcode.l.h; ;/tmp/_build/generated_src/sqlcode.l.h; ;/tmp/_build/generated_src/lexcode.l.h; ;/tmp/_build/generated_src/lexscanner.l.h; ;/tmp/_build/generated_src/configimpl.l.h
```

The variables were being constructed as a list of values, but each item of that list had an additional, separate list item in front of it that contained only a single space character.

CMake understands how to explode lists-in-variables into members, when listing arguments to a command, It can even expand a list **of arguments**, when calling a command. (I make use of that in my logic to handle the conditional `DEPFILE` arguments to add_custom_command()`.) But older CMake versions weren't smart enough to handle **empty** list arguments without breaking.

With the code in this PR, `LEX_FILES_H` is generated cleanly:
```cmake
LEX_FILES_H:INTERNAL=/tmp/_build2/generated_src/code.l.h;/tmp/_build2/generated_src/commentcnv.l.h;/tmp/_build2/generated_src/commentscan.l.h;/tmp/_build2/generated_src/constexp.l.h;/tmp/_build2/generated_src/declinfo.l.h;/tmp/_build2/generated_src/defargs.l.h;/tmp/_build2/generated_src/doctokenizer.l.h;/tmp/_build2/generated_src/fortrancode.l.h;/tmp/_build2/generated_src/fortranscanner.l.h;/tmp/_build2/generated_src/lexcode.l.h;/tmp/_build2/generated_src/lexscanner.l.h;/tmp/_build2/generated_src/pre.l.h;/tmp/_build2/generated_src/pycode.l.h;/tmp/_build2/generated_src/pyscanner.l.h;/tmp/_build2/generated_src/scanner.l.h;/tmp/_build2/generated_src/sqlcode.l.h;/tmp/_build2/generated_src/vhdlcode.l.h;/tmp/_build2/generated_src/xmlcode.l.h
```

CMake understands how to process **that** list. So it works everywhere, even ancient CMake 3.0/3.1/3.2.

#### What else?
- I sorted the `LEX_FILES` list out of necessity, so that I could compare it to the list of items I was removing from the `add_target()` call and ensure I removed only the items from the variable(s).
- After discovering that the `configimpl` files weren't on the list (they're added to the static target, instead), added code to keep them out of `LEX_FILES_H` and `LEX_FILES_CPP`.
- Line-wrapped the `add_custom_target()` command I edited, to use more vertical space, and less horizontal. (I can undo that on request. it admittedly amounts to nothing more than a personal preference.)
- With the addition of a couple of temporary CMake variables, I managed to streamline and DRY out the `add_custom_command()` call that processes lexer files (eliminating the nearly-identical duplicate branched on `depfiles_supported`).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -94,53 +94,65 @@ add_custom_command(
 )
 set_source_files_properties(${GENERATED_SRC}/resources.cpp PROPERTIES GENERATED 1)
 
-set(LEX_FILES scanner 
+set(LEX_FILES
     code
-    pyscanner
-    pycode
-    fortranscanner
-    fortrancode
-    vhdlcode
-    pre
+    commentcnv
+    commentscan
+    configimpl
+    constexp
     declinfo
     defargs
     doctokenizer
-    commentcnv
-    commentscan
-    constexp
-    xmlcode
-    sqlcode
+    fortrancode
+    fortranscanner
     lexcode
     lexscanner
-    configimpl)
+    pre
+    pycode
+    pyscanner
+    scanner
+    sqlcode
+    vhdlcode
+    xmlcode
+)
 
 if (NOT depfile_supported)
     # In case the DEPFILE possibility is not supported the complete list of lex include files for the dependency has to be used
     set(LEX_INC_FILES)
 endif()
 
-# unfortunately ${LEX_FILES_H} and ${LEX_FILES_CPP} don't work in older versions of CMake (like 3.6.2) for add_library
+set(LEX_FILES_H)
+set(LEX_FILES_CPP)
+set(_depfile_args)
+
 foreach(lex_file ${LEX_FILES})
-    set(LEX_FILES_H ${LEX_FILES_H} " " ${GENERATED_SRC}/${lex_file}.l.h CACHE INTERNAL "Stores generated files")
-    set(LEX_FILES_CPP ${LEX_FILES_CPP} " " ${GENERATED_SRC}/${lex_file}.cpp CACHE INTERNAL "Stores generated files")
+    # configimpl is handled specially (it's in the static lib)
+    if (NOT lex_file STREQUAL "configimpl")
+        list(APPEND LEX_FILES_H ${GENERATED_SRC}/${lex_file}.l.h)
+        list(APPEND LEX_FILES_CPP ${GENERATED_SRC}/${lex_file}.cpp)
+    endif()
+
+    set(_generated_files
+        ${GENERATED_SRC}/${lex_file}.l
+        ${GENERATED_SRC}/${lex_file}.corr
+        ${GENERATED_SRC}/${lex_file}.d)
 
     if (depfile_supported)
-        add_custom_command(
-            COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/pre_lex.py ${CMAKE_CURRENT_LIST_DIR}/${lex_file}.l  ${GENERATED_SRC}/${lex_file}.l ${GENERATED_SRC}/${lex_file}.corr ${GENERATED_SRC}/${lex_file}.d ${CMAKE_CURRENT_LIST_DIR}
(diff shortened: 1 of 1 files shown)


doxygen:
Thanks. Looks a lot cleaner indees. The spdlog dependency could be made optional. Support is completely compiled out for a release build.

src/CMakeLists.txt
```diff
@@ -94,53 +94,65 @@ add_custom_command(
...
+            ${CMAKE_CURRENT_LIST_DIR}/pre_lex.py
+            ${CMAKE_CURRENT_LIST_DIR}/${lex_file}.l
+            ${_generated_files}
+            ${CMAKE_CURRENT_LIST_DIR}
+        DEPENDS
+            ${CMAKE_CURRENT_LIST_DIR}/pre_lex.py
+            ${CMAKE_CURRENT_LIST_DIR}/${lex_file}.l
+        ${_depfile_args}
+    )
+    set_source_files_properties(${_generated_files} PROPERTIES GENERATED 1)
```
ferdnyc, line 155: I left this is because it was being done before, but it _shouldn't_ be necessary. The `OUTPUT` files of `add_custom_command` will [automatically be marked `GENERATED`](https://cmake.org/cmake/help/latest/command/add_custom_command.html#generating-files:~:text=Each%20output%20file%20will%20be%20marked%20with%20the%20GENERATED%20source%20file%20property%20automatically.) by CMake itself. That's been documented behavior since at least CMake 3.0, if not earlier.

The only reason I can think of for re-applying the `GENERATED` property manually would be to ensure that it's visible in another scope. (And that's a relatively-recent CMake 3.18+ feature.) Like, this would ensure that the files are seen as `GENERATED` from the root build directory:
```cmake
if (CMAKE_VERSION VERSION_GREATER "3.18")
  set_source_files_properties(${_generated_files} 
    DIRECTORY ${CMAKE_BINARY_DIR}
    PROPERTIES GENERATED 1)
endif()
```

...But since CMake 3.20, [`GENERATED` is **always** globally-visible](https://cmake.org/cmake/help/latest/prop_sf/GENERATED.html#prop_sf:GENERATED), so even that's not really necessary.

Issue #9839, opened 2023-02-07
Server side search is returning wrong results


splint33:
Hello, I have recently upgraded my Doxygen installation from 1.9.3 to 1.9.6.
I am using the server side search, in order to search into function brief & details in my Python project.

For example, I currently have the keyword "TWS" in 23 brief functions.

On Doxygen 1.9.3, when searching for "TWS", all 23 functions are showing up on the result page.
On Doxygen 1.9.6, when searching for "TWS", a completly unlinked function is showing up in the result page, with "Matches: tws(23)".

Searching "TWS" on Doxygen 1.9.3:
![image](https://user-images.githubusercontent.com/4028855/217312041-26d64cfe-b09f-4f54-99d4-6c97d2bdb59e.png)

Searching "TWS" on Doxygen 1.9.6:
![image](https://user-images.githubusercontent.com/4028855/217311138-9c4daed1-474e-4717-8846-318d449f5360.png)

splint33:
I have crafted a little Stub project with four4 functions, three of them with the keyword "TWS" in their brief tag, and another one, completely unrelated, without TWS in the brief tag.

The doxyfile is located in the configuration folder, and I ran "doxygen at_ms.doxyfile" on the configuration folder in order to generate the documentation.

The documentation is then displayed using a web server.
You will need to change the OUTPUT_DIRECTORY, as it is pointing to a local network share.

As mentioned before, the search is working perfectly on Doxygen 1.9.3, and seems to be broken on 1.9.6.

When crafting the Stub project, I have found that the wrongly displayed results change based on its name.
I hope you can find & fix the issue, if you need more information, please let me know.

[StubProject.zip](https://github.com/doxygen/doxygen/files/10683437/StubProject.zip)

albert-github:
Thanks for the small project.
The essential settings here are (I think)::
```
SEARCHENGINE           = YES
SERVER_BASED_SEARCH    = YES
EXTERNAL_SEARCH        = NO
```

I see that there are some changes, between the 1.9.3 version and the 1.9.6 version, though I cannot test it as I have no search server running.

@doxygen do you have an idea?

splint33:
I have done more tests, it seems that the problem appeared in the version 1.9.4, and was not present in the version 1.9.3.

albert-github:
@splint33 Interesting observation.

I did a diff between the results of the 1.9.3 and 1.9.4 versions and see:
```
[...]

PR #9849, opened 2023-02-10
issue #9839 Server side search is returning wrong results

albert-github:
In case of a non external server search do not add the items to the search index file.


albert-github:
Obsoleted by means of an alternative / better implementation in https://github.com/doxygen/doxygen/commit/134e0e351c9d0009ac411a0fa195644f771fe673

Issue #9760, opened 2022-12-31
Disable a specific warning for the duration of processing a specified piece of code


iljamaki:
Please see [this](https://stackoverflow.com/questions/74969475/how-to-suppress-a-doxygen-warning) for a description of the issue and an example case where it is a problem, and the comments there for reasoning why ability to temporarily prevent a particular warning from being issued would likely be the sustainable solution. I'm happy to answer any further questions you may have.

Doxygen version 1.9.6 for Windows 10 Pro 64bit.

albert-github:
Please don't refer to external sources as they might not be persistent (and might make the issue non-understandable), but include de text in the issue.

iljamaki:
Ok, here's the main points of the question posted there. I'd like to emphasize though that the intent would be more to have a general solution i.e. to actually be able to disable the warning for certain code block than to find workarounds for this particular example case. Another issue more related to this particular example code is then to find out why Doxygen fails to find the declarations from the #include'd header file. So here goes:

Is there a way to suppress the `documented symbol 'foo' was not declared or defined` warning for certain parts of the code?

So first of all, I need to have `WARN_AS_ERROR=FAIL_ON_WARNINGS` so that Doxygen exit code will be non-zero when there are (valid) warnings so that the pull request checker can detect issues correctly. In this case, however, Doxygen incorrectly claims that some documented symbols are not declared because it's unable to parse their declarations due to the way their declarations come from external library code (see below). So that particular warning would need to be suppressed for the definitions of those particular functions. The documentation for those functions should still be processed and warnings about any other issues in that processing should be issued.

This is one possible way to make Doxygen issue an unjustified warning:

in file c.h:

```
//! blabla
namespace A
{
  //! blabla
  namespace B
  {
    //! blabla
    class C : public D
    {
#include "lib.h" // This is where the function declarations are but Doxygen does not seem to figure it
    };
  }
}
```
in file c.cpp:

```
using namespace A::B;

/*! The declaration of this member function was in lib.h.
 *  I may document something additional here but Doxygen collects this function without that also, which is desired.
 *  There are hundreds of these functions declared in lib.h and #include'd into several classes.
 *  Missing documentation for param b should issue a warning but as soon as documentation for param b is added,
 * no warnings should be issued and Doxygen exit code should be zero.
 *  \param a blahblah
*/
void C::foo( int a, int b )
{
}
```

albert-github:
Some small questions:
- how does `lib.h` look like?
[...]

Issue #9436, opened 2022-07-06
fortran - INCLUDE file command not recognised


pak80:
When writing fortran, I typically use the INCLUDE command to break up long files into several more-manageable smaller ones, e.g. in modules with many different subroutines. However, doxygen (1.9.3) does not recognise that the content/definitions in the included file are part of the module itself; which make the generated documentation less coherent. Of course, the lack of recognition of  INCLUDE can have other effects on documentation as well; but it is very clear in this module case.

Generally I find INCLUDE an important tool if I need to re-use snippets of code in several different positions in the code (perhaps complicated sets of definitions; notably - once upon a time - common blocks), or if some routines are machine-generated just before (re)compilation and need to be incorporated without continually re-altering the main source code.

Essentially, if doxgen sees a line

INCLUDE 'file.f'

inside a file, it should include its contents there (and then not use it again). I think from other discussion here this functionality is supported in other contexts (e.g. #include), so I might hope that it would not be too difficult to add for fortran also. I currently use .f as a suffix for all my fortran - even includes - so it would be nice if doxygen then ignored any independent matches, but me changing the suffix or other naming convention for include files to avoid double-appearances would not be too annoying.

I attach code for a test case. It contains two simple modules, one using an INCLUDE which then doesn't get documented properly; and another (not using an INCLUDE) which *is* documented as expected, so you can see the difference.

[doxytest.zip](https://github.com/doxygen/doxygen/files/9055767/doxytest.zip)

PR #9851, opened 2023-02-11
bug 549438 VERBATIM_HEADERS - only works with header files with file type

albert-github:
Also check whether the file is included so also files with unknown and without extensions are caught as include files (besides files with a  known header extension).

Extended example: [example_extended.tar.gz](https://github.com/doxygen/doxygen/files/10713915/example_extended.tar.gz)

### DIFF
--- a/src/filedef.cpp
+++ b/src/filedef.cpp
@@ -1426,7 +1426,7 @@ bool FileDefImpl::generateSourceFile() const
   bool verbatimHeaders = Config_getBool(VERBATIM_HEADERS);
   return !isReference() &&
          (sourceBrowser ||
-           (verbatimHeaders && guessSection(name())==Entry::HEADER_SEC)
+           (verbatimHeaders && (guessSection(name())==Entry::HEADER_SEC || !m_includedByMap.empty()))
          ) &&
          !isDocumentationFile();
 }


src/filedef.cpp
```diff
@@ -1426,7 +1426,7 @@ bool FileDefImpl::generateSourceFile() const
   bool verbatimHeaders = Config_getBool(VERBATIM_HEADERS);
   return !isReference() &&
          (sourceBrowser ||
-           (verbatimHeaders && guessSection(name())==Entry::HEADER_SEC)
+           (verbatimHeaders && (guessSection(name())==Entry::HEADER_SEC || !m_includedByMap.empty()))
```
ferdnyc, line 1429: This will work, it looks like... but it'll mean that every file, of any type, that's `#include`d by another file is treated as a header.

That may seem correct, or even tautological, but I'm not sure it entirely is. The problem with `#include` is that it's just a preprocessor directive that pulls in the contents of other files. It doesn't actually say anything about the _nature_ of that file.

There are some coding styles where `.cpp` files get `#include`d into others, for various reasons. (Qt's `moc` sets up generated files that way, as one example.) Even if it's used in a `#include`, I'm not sure it makes sense to treat a `.cpp` file as a header.

It might be better to give `guessSection()` the ability to _optionally_ return `Entry::HEADER_SEC` for filenames with no extension, or for extensionless names found only in certain designated directories.

It's a more complex fix, admittedly, but it might have fewer potentially-unwanted side effects.

albert-github: I think that the problem can be present but that the effects are normally limited by means of the setting of `INCLUDE_FILE_PATTERNS`

ferdnyc: @albert-github That's the thing, this fix bypasses `FILE_PATTERNS` _and_ `INCLUDE_FILE_PATTERNS`. (necessarily, since `guessSection()` ignores both of those configurations as well. Those configs determine which files Doxygen processes in various ways. _This_ affects whether a file is included as a source file when `VERBATIM_HEADERS` is _on_ but `SOURCE_BROWSER` is off. (In those cases, only headers should have browsable sources.)

`guessSection()` returns `Entry::SOURCE_SEC` or `Entry::HEADER_SEC` (or `0`, if there's no extension or an unknown extension) based _solely_ on a hardcoded set of extension mappings, so it will ignore all of the configuration.

albert-github: I did a quick test which the extended example by adding 2 files (contents analogous to bb) and the lines (in main.,cpp):
```
#include "abs.cba"
#include <xyz.zxy>
```
and the result is:

image

(I made the links red for better visibility).
So these "rubbish" names are not linked and I didn't find any further reference either, so I think the problem does not exist or do you have a, small, example showing the problem?

ferdnyc: @albert-github 
Is that with this patch, without? If it's _without_, the issue isn't the links, it's the inclusion of the source. What you'll find, if you click those links, is that the "aa.h" link leads to the source of the header file, but the "bb" link leads to the documentation for that file, because the source isn't included. (By using `VERBATIM_HEADERS`, it should be. That's the need for this patch, or some other fix.)

If it's _with_ this patch, the "rubbish files" wouldn't be included because those files still aren't being processed by doxygen at all, which is good. But if I were to have a `#include "somefile.cpp"` in a file, where both `somefile.cpp` and the file that includes it are being processed by Doxygen, then this patch would cause `somefile.cpp` to be treated as a header (so its source would be included in the docs, if `VERBATIM_HEADERS` is enabled but not `SOURCE_BROWSER`). Which is the concern I was raising.

ferdnyc: (Duh, it's your patch, of course you know the reasons for it. Sorry.)

But here's an example of a project that will be affected by this patch:

[doxy_qt.tar.gz](https://github.com/doxygen/doxygen/files/10752358/doxy_qt.tar.gz)

That's a completely-valid Qt project, it will compile and run. But it's "weird", because (for some reason) the user got it in their head to `#include "MainWindow.cpp" from inside `main.cpp`, instead of using it as a separate source file. (Maybe they wanted to write the class inside `main.cpp`, but it's under a different license. Maybe someone told them it will speed up their builds. Who knows?)

Without this patch, `MainWindow.cpp` is treated as a source file; the `#include "MainWindow.cpp"` line in `main.cpp` is linked to the _documentation_ page, but its source isn't included. With this patch, `MainWindow.cpp` becomes a header file, and its source ends up in the docs.

ferdnyc: Holy carp! And if you're curious how common it is to `#include "something.cpp"`, [this GitHub code search](https://github.com/search?q=include+%5C.cpp+extension%3Acpp+extension%3Acxx+extension%3Acc+extension%3AC&type=Code&ref=advsearch&l=&l=) returns almost _**40 MILLION**_ code results that broadly fit that pattern. 

(There are surely some false positives, because GitHub code search is "fuzzy" — which is just confounding, in a _code_ search — but regardless, it's clearly _much_ more common than I ever imagined.)

ferdnyc: Ugh,ignore those numbers. The search is _so_ fuzzy, that one matched every file that contains both "include" and "cpp" _ANYWHERE_ in the text. (And quite possibly, files that contain only "include" _or_ "cpp" as well.)

[This one](https://github.com/search?l=&p=3&q=%22include%2A.cpp%22+extension%3Acpp+extension%3Acxx+extension%3Acc+extension%3AC&ref=advsearch&type=Code) is more realistic, and returns 187,000 matches, which seems much more plausible. And still, plenty common. (Just browsing the first few pages of results, it's clear there are a lot of valid matches.)

albert-github:
Regarding the include files there are actually a number of types of include files
- what is considered normally as an include file (like `stdio.h` or `QMainWindow`)
- include files containing code (like `MainWindow.cpp` in your example, indeed a strange but valid construct. I think a "better" example of this type is given in the comment https://github.com/doxygen/doxygen/issues/9760#issuecomment-136826275, the "lib.h" and also in #9436)


I was already thinking about the "code include" as they should be verbatim in the code, so I was thinking about having during the read step (i.e. before the doxygen preprocessor, pre.l kicks in) a possibility to include verbatim some files when they are mentioned in a `#include` (or for Fortran in an `INCLUDE` statement).
- create a setting `VERBATIM_INCLUDE` with (wildcard and recursive) filepaths / filenames to be verbatim included in the code
- execute the enhanced read step analogous to the current read function, replacing the "verbatim includes"

In this way the "code include" files will not be visible anymore in the include graphs and also not in the `m_includedByMap`, though maybe it is still desired to have the names of the verbatim included files in the include graph as it are still files that are included (has to be checked how to implement, but shouldn't be to problematic as we have a list of verbatim included files).

Nice advantage is that in this way the code of this proposed pull request is still correct / valid.

@doxygen  @ferdnyc what do you think of this general idea?

PR #9757, opened 2022-12-31
Fix conversion of ghc::filesystem::path to std::string

Biswa96:
This fixes the following compiler error with clang

```
portable.cpp:652:10: error: no matching constructor for initialization of 'std::ofstream' (aka 'basic_ofstream<char>')
  return std::ofstream(fs::path(fileName.str()), mode);
         ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This happens because std::ifstream does not have any constructor with ghc::filesystem::path, instead with filesystem::path.

PR #9854, opened 2023-02-13
Fix compiler error with mingw clang

Biswa96:


Biswa96:
@HazardyKnusperkeks Would you like to review these changes? I have tested it with gcc and clang and the unicode test passes with both compilers.

albert-github:
I assume that this is based on the comment: https://github.com/doxygen/doxygen/pull/9757#issuecomment-1422640043 of the initial fix (#9757) and revering it in most cases except for the specific clang / mingw case:
```
Commit: 289e76aa67646b4fc59cb2a8e713b03893917afd [289e76a]
Date: Wednesday, January 4, 2023 8:38:14 PM
Commit Date: Wednesday, January 4, 2023 8:38:19 PM
Fix conversion of ghc::filesystem::path to std::string

This fixes the following compiler error with clang

portable.cpp:652:10: error: no matching constructor for initialization of 'std::ofstream' (aka 'basic_ofstream<char>')
```

doxygen:
I expect this will break things again for other Windows compilers like Visual Studio. Did you test it with that?

Biswa96:
I have tested in mingw environment only. The Visual Studio environment is being checked in CI already.

Issue #9855, opened 2023-02-13
HTML: TOC: Smooth scrolling


tim-gromeyer:
**Describe the error**.
When clicking on a section/link in the table of contents, it does not scroll smoothly. When clicking in the navigation bar, it scrolls smoothly.

**Expected behavior**
Smooth scrolling as when clicking a section in the navigation bar.

**To Reproduce**
[reproduce.zip](https://github.com/doxygen/doxygen/files/10726034/reproduce.zip)

**Version**
Doxygen version: `1.9.6 (4586b5cfaa3d46d51f6a51882951d15644c49edf)` (from release)
System:
```
Distributor ID:	Linuxmint
Description:	Linux Mint 21.1
Release:	21.1
Codename:	vera
```

albert-github:
When in the clicking in the navigation tree:
![image](https://user-images.githubusercontent.com/5223533/218707972-d6c22a01-0fc3-4524-9bd1-2bbb2f99f364.png)
on "Another section" it "smoothly" scrolls to the section.
When clicking in the "Table of Contents":
![image](https://user-images.githubusercontent.com/5223533/218708193-bebe6695-cdfe-483c-b05d-354179c95918.png)
on "Another section" it jumps to the section. The later also occurs when clicking on "Section 2" in the running text:
![image](https://user-images.githubusercontent.com/5223533/218708438-1119e93f-df51-492e-8465-e7d42631765e.png)

This is not an easy thing to solve (as already remarked in https://github.com/jothepro/doxygen-awesome-css/issues/107) and the question is which parts should be affected:
- just the "Table of contents"
- all hyperlinks
- ...

@doxygen any idea how / where to accomplish this?

tim-gromeyer:
> ... and the question is which parts should be affected:
> 
> * just the "Table of contents"
> * all hyperlinks
> * ...

I think it would be good to rework the whole section linking. It is complicated to link to a section because it uses the label `autotoc_md` + the section number. So if you add another section before the section you link to, you have to update all the links because the section you want to link to now has a different number.

In summary, I would say that reworking the table of contents alone is not enough.

Ideas other than a reworking are welcome.

albert-github:
The `autotoc_md` are not for general use, just for internal use. When one wants to use a label one either has to use commands like `\section` and `\anchor` or in the markdown case use the explicit label as described in https://www.doxygen.nl/manual/markdown.html#md_header_id

So I don't think reworking labels is necessary.

[...]

Issue #8378, opened 2021-02-11
Handling markdown page references without label


arwedus:
**Describe the bug**

Doxygen handles Markdown pages quite well when you have a label for the first L1 heading in a .md file, like this:

`# Data Handling                             {#p_data_handling}`

Then, referencing in Markdown style `[data handling](doc/data_handling.md)` and in Doxygen commands (`\subpage p_data_handling`) jumps to the right page.

The problem: PHP Markdown extra is not really a popular dialect, and this extension is not CommonMark and not [github flavored markdown](https://github.github.com/gfm/), is not understood by VS code plug-in or Bitbucket. Thus, the L1 heading shows with a big fat label anchor everywhere else.

Instead, we created the following construct:

```
\page p_input_handling Input Handling

# Input Handling
```

With that, references in Markdown style `[input handling](doc/input_handling.md)` jump to an empty page in Doxygen. Doxygen commands references (`\subpage p_data_handling`) jump to the right page.

If I just use the L1 heading and, references in Markdown style `[output handling](doc/output_handling.md)` jump to the right page in Doxygen. Doxygen commands references (`\subpage doc/output_handling.md`) jump to the right page, but the command (subpage, in this case) does not make the output_handling Page a subpage in the page tree.

**Expected behavior**

1. Doxygen should never create empty markdown pages for unresolved page links.
2. Doxygen should offer some auto-label for Markdown pages that start with a L1 heading (possibly after a meta section or ToC).
3. Doxygen should support Markdown-style references in the \subpage command consistently.

Recommendation for page auto-label:

a. Simply the page name, all special characters replaced by underscore (e.g. "input handling.md" -> input_handling)
b. As a), but "input_handling_md" or "md_input_handling")
c. The first L1 heading, reformatted: e.g. "Input Handling" becomes "input-handling", as [on Microsoft Docs](https://docs.microsoft.com/en-us/azure/devops/project/wiki/markdown-guidance?view=azure-devops#anchor-links)

**To Reproduce**

See attached example.

**Version**

Tested on Doxygen 1.9.1 win x64, but present in all older versions too.

arwedus:
[doxygen_md_ref_example.zip](https://github.com/doxygen/doxygen/files/5965316/doxygen_md_ref_example.zip)

arwedus:
[...]

PR #9858, opened 2023-02-14
Markdown: Heading auto id

tim-gromeyer:
As I mentioned in #9855, I don't like the default title ID. I also didn't want to use commands or give the title an ID as described [here](https://www.doxygen.nl/manual/markdown.html#md_header_id). That's why I created this PR to change the default title id. I have already tested it with HTML and PDF. It worked.

### DIFF
--- a/src/definition.cpp
+++ b/src/definition.cpp
@@ -36,6 +36,7 @@
 #include "utf8.h"
 #include "indexlist.h"
 #include "fileinfo.h"
+#include "markdown.h"
 
 //-----------------------------------------------------------------------------------------
 
@@ -400,7 +401,7 @@ void DefinitionImpl::writeDocAnchorsToTagFile(TextStream &tagFile) const
     //printf("%s: writeDocAnchorsToTagFile(%d)\n",qPrint(name()),m_impl->sectionRef.size());
     for (const SectionInfo *si : m_impl->sectionRefs)
     {
-      if (!si->generated() && si->ref().isEmpty() && !si->label().startsWith("autotoc_md"))
+      if (!si->generated() && si->ref().isEmpty() && !set_contains(Markdown::ids(), si->label()))
       {
         //printf("write an entry!\n");
         if (m_impl->def->definitionType()==Definition::TypeMember) tagFile << "  ";
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -33,26 +33,24 @@
 
 #include <stdio.h>
 
-#include <unordered_map>
-#include <functional>
+#include <algorithm>
 #include <atomic>
+#include <functional>
+#include <unordered_map>
 
-#include "markdown.h"
-#include "growbuf.h"
-#include "debug.h"
-#include "util.h"
-#include "doxygen.h"
 #include "commentscan.h"
-#include "entry.h"
-#include "commentcnv.h"
 #include "config.h"
-#include "section.h"
+#include "debug.h"
+#include "doxygen.h"
+#include "entry.h"
+#include "fileinfo.h"
+#include "growbuf.h"
+#include "markdown.h"
 #include "message.h"
 #include "portable.h"
 #include "regex.h"
-#include "fileinfo.h"
-#include "utf8.h"
 #include "trace.h"
+#include "util.h"
 
 enum class ExplicitPageResult
 {
@@ -1845,7 +1843,52 @@ static bool isHRuler(const char *data,int size)
   return n>=3; // at least 3 characters needed for a hruler
 }
 
-static QCString extractTitleId(QCString &title, int level)
+QCString doxygenHtmlIdentifier()
+{
+  static int count = 0;
+  QCString result = "autotoc_md";
+  result.append(std::to_string(count));
+  ++count;
+  return result;
+}
+
+std::unordered_map<std::string, int> id_count;
+QCString githubHtmlIdentifier(const std::string &str)
+{
+  std::string result;
+
+  for (auto c : str) {
+    if (std::isspace(c))
+      result.push_back('-');
+    else if (std::ispunct(c))
+      continue;
+    else
+      result.push_back(c);
+  }
+
+  result = convertUTF8ToLower(result);
+
+  if (result.empty())
+    return doxygenHtmlIdentifier();
+
+  // Adding end digits if an identical header already exists
+  int &count = id_count[result];
+  if (count > 0) {
+    result += "-" + std::to_string(count);
+  }
+  count++;
+
+  return result;
+}
+
+// Use set because it has better lookup times and every item is unique
+std::set<QCString> m_ids;
+const std::set<QCString> &Markdown::ids()
+{
+  return m_ids;
+}
+
+QCString Markdown::extractTitleId(QCString &title, int level)
 {
   AUTO_TRACE("title={} level={}",Trace::trunc(title),level);
   // match e.g. '{#id-b11} ' and capture 'id-b11'
@@ -1856,15 +1899,34 @@ static QCString extractTitleId(QCString &title, int level)
   {
     std::string id = match[1].str();
     title = title.left(match.position());
+    int &count = id_count[id];
+    if (count > 0) {
+      warn(m_fileName,
+           m_lineNr,
+           "An automatically generated id already has the name '%s'!",
+           id.c_str());
+    }
     //printf("found match id='%s' title=%s\n",id.c_str(),qPrint(title));
     AUTO_TRACE_EXIT("id={}",id);
-    return QCString(id);
+    return id;
   }
-  if ((level > 0) && (level <= Config_getInt(TOC_INCLUDE_HEADINGS)))
-  {
-    static AtomicInt autoId { 0 };
+  if ((level > 0) && (level <= Config_getInt(TOC_INCLUDE_HEADINGS))) {
+    static HEADING_AUTO_IDENTIFIER_t behavior = Config_getEnum(HEADING_AUTO_IDENTIFIER);
     QCString id;
-    id.sprintf("autotoc_md%d",autoId++);
+
+    switch (behavior)
+    {
+      case HEADING_AUTO_IDENTIFIER_t::DOXYGEN:
+        id = doxygenHtmlIdentifier();
+        break;
+      case HEADING_AUTO_IDENTIFIER_t::GITHUB:
+        id = githubHtmlIdentifier(title.str());
+        break;
+      default:
+        id = doxygenHtmlIdentifier();
+    }
+
+    m_ids.insert(id);
     //printf("auto-generated id='%s' title='%s'\n",qPrint(id),qPrint(title));
     AUTO_TRACE_EXIT("id={}",id);
     return id;
@@ -3144,7 +3206,7 @@ static ExplicitPageResult isExplicitPage(const QCString &docs)
   return ExplicitPageResult::notExplicit;
 }
 
-QCString Markdown::extractPageTitle(QCString &docs,QCString &id, int &prepend)
+QCString Markdown::extractPageTitle(QCString &docs, QCString &id, int &prepend, bool *extractedId)
 {
   AUTO_TRACE("docs={} id={} prepend={}",Trace::trunc(docs),id,prepend);
   // first first non-empty line
@@ -3175,6 +3237,8 @@ QCString Markdown::extractPageTitle(QCString &docs,QCString &id, int &prepend)
       convertStringFragment(title,data+i,end1-i-1);
       docs+="\n\n"+docs_org.mid(end2);
       id = extractTitleId(title, 0);
+      if (extractedId != nullptr)
+        *extractedId = true;
       //printf("extractPageTitle(title='%s' docs='%s' id='%s')\n",title.data(),docs.data(),id.data());
       AUTO_TRACE_EXIT("result={}",Trace::trunc(title));
       return title;
@@ -3189,6 +3253,8 @@ QCString Markdown::extractPageTitle(QCString &docs,QCString &id, int &prepend)
   {
     docs=docs_org;
     id = extractTitleId(title, 0);
+    if (extractedId != nullptr)
+      *extractedId = true;
   }
   AUTO_TRACE_EXIT("result={}",Trace::trunc(title));
   return title;
@@ -3298,9 +3364,11 @@ void MarkdownOutlineParser::parseInput(const QCString &fileName,
   QCString docs = fileBuf;
   Debug::print(Debug::Markdown,0,"======== Markdown =========\n---- input ------- \n%s\n",qPrint(fileBuf));
   QCString id;
+  bool extractedId = false;
   Markdown markdown(fileName,1,0);
-  QCString title=markdown.extractPageTitle(docs,id,prepend).stripWhiteSpace();
-  if (id.startsWith("autotoc_md")) id = "";
+  QCString title = markdown.extractPageTitle(docs, id, prepend, &extractedId).stripWhiteSpace();
+  if (extractedId)
+    id = "";
   int indentLevel=title.isEmpty() ? 0 : -1;
   markdown.setIndentLevel(indentLevel);
   QCString fn      = FileInfo(fileName.str()).fileName();
--- a/src/tagreader.cpp
+++ b/src/tagreader.cpp
@@ -40,6 +40,7 @@
 #include "section.h"
 #include "containers.h"
 #include "debug.h"
+#include "config.h"
 
 // ----------------- private part -----------------------------------------------
 
@@ -457,7 +458,10 @@ class TagFileParser
         case InMember:
         case InPackage:
         case InDir:
-          if (m_curString.endsWith("autotoc_md")) return;
+          static HEADING_AUTO_IDENTIFIER_t behavior = Config_getEnum(HEADING_AUTO_IDENTIFIER);
+          if (behavior == HEADING_AUTO_IDENTIFIER_t::DOXYGEN)
+            if (m_curString.startsWith("autotoc_md"))
+              return;
           break;
         default:
           warn("Unexpected tag 'docanchor' found");
--- a/src/config.xml
+++ b/src/config.xml
@@ -704,6 +704,16 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+    <option type='enum' id='HEADING_AUTO_IDENTIFIER' defval='DOXYGEN' depends='MARKDOWN_SUPPORT'>
+      <docs>
+<![CDATA[
+ The \c HEADING_AUTO_IDENTIFIER tag can be used to specify the behavior used to generate ids for the Markdown headings.
+ Note: Every id is unique.
+]]>
+      </docs>
+      <value name="DOXYGEN" desc="Use autotoc_md + a number starting at 0."/>
+      <value name="GITHUB" desc="Uppercase to lowercase, replace all sort of spaces with '-', remove all punctations."/>
+    </option>
     <option type='bool' id='AUTOLINK_SUPPORT' defval='1'>
       <docs>
 <![CDATA[
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -5367,7 +5367,7 @@ endparagraph:
 
 int DocSection::parse(DocNodeVariant *thisVariant)
 {
-  AUTO_TRACE("start {} level={}",parser()->context.token->sectionId,m_level);
+  AUTO_TRACE("start {} level={}", parser()->context.token->sectionId, m_level);
   int retval=RetVal_OK;
   auto ns = AutoNodeStack(parser(),thisVariant);
 
@@ -5442,8 +5442,12 @@ int DocSection::parse(DocNodeVariant *thisVariant)
     }
     else if (retval==RetVal_Subsubsection && m_level<=Doxygen::subpageNestingLevel+2)
     {
-      if ((m_level<=1+Doxygen::subpageNestingLevel) && !parser()->context.token->sectionId.startsWith("autotoc_md"))
-          warn_doc_error(parser()->context.fileName,parser()->tokenizer.getLineNr(),"Unexpected subsubsection command found inside %s!",g_sectionLevelToName[m_level]);
+      if ((m_level <= 1 + Doxygen::subpageNestingLevel)
+          && set_contains(Markdown::ids(), parser()->context.token->sectionId))
+        warn_doc_error(parser()->context.fileName,
+                       parser()->tokenizer.getLineNr(),
+                       "Unexpected subsubsection command found inside %s!",
+                       g_sectionLevelToName[m_level]);
       // then parse any number of nested sections
       while (retval==RetVal_Subsubsection) // more sections follow
       {
@@ -5452,11 +5456,13 @@ int DocSection::parse(DocNodeVariant *thisVariant)
                                 parser()->context.token->sectionId);
         retval = children().get_last<DocSection>()->parse(vDocSection);
       }
-      if (!(m_level<Doxygen::subpageNestingLevel+2 && retval == RetVal_Subsection)) break;
+      if (!(m_level < Doxygen::subpageNestingLevel + 2 && retval == RetVal_Subsection))
+        break;
     }
     else if (retval==RetVal_Paragraph && m_level<=std::min(5,Doxygen::subpageNestingLevel+3))
     {
-      if ((m_level<=2+Doxygen::subpageNestingLevel) && !parser()->context.token->sectionId.startsWith("autotoc_md"))
+      if ((m_level <= 2 + Doxygen::subpageNestingLevel)
+          && set_contains(Markdown::ids(), parser()->context.token->sectionId))
         warn_doc_error(parser()->context.fileName,parser()->tokenizer.getLineNr(),"Unexpected paragraph command found inside %s!",g_sectionLevelToName[m_level]);
       // then parse any number of nested sections
       while (retval==RetVal_Paragraph) // more sections follow
@@ -5483,7 +5489,7 @@ int DocSection::parse(DocNodeVariant *thisVariant)
                   retval==RetVal_EndInternal
                  );
 
-  AUTO_TRACE_EXIT("retval={}",DocTokenizer::retvalToString(retval));
+  AUTO_TRACE_EXIT("retval={}", DocTokenizer::retvalToString(retval));
   return retval;
 }
 
@@ -5605,7 +5611,7 @@ void DocRoot::parse(DocNodeVariant *thisVariant)
 
   // first parse any number of paragraphs
   bool isFirst=TRUE;
-  DocPara *lastPar=0;
+  DocPara *lastPar = nullptr;
   do
   {
     {
@@ -5622,11 +5628,14 @@ void DocRoot::parse(DocNodeVariant *thisVariant)
         lastPar = par;
       }
     }
-    if (retval==RetVal_Paragraph)
+    if (retval == RetVal_Paragraph)
     {
-      if (!parser()->context.token->sectionId.startsWith("autotoc_md"))
+      if (!set_contains(Markdown::ids(), parser()->context.token->sectionId))
       {
-         warn_doc_error(parser()->context.fileName,parser()->tokenizer.getLineNr(),"found paragraph command (id: '%s') outside of subsubsection context!",qPrint(parser()->context.token->sectionId));
+        warn_doc_error(parser()->context.fileName,
+                       parser()->tokenizer.getLineNr(),
+                       "found paragraph command (id: '%s') outside of subsubsection context!",
+                       qPrint(parser()->context.token->sectionId));
       }
       while (retval==RetVal_Paragraph)
       {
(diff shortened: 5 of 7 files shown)


albert-github:
I don't think this will work as there are other places in doxygen that rely on the `autotoc_md`.

tim-gromeyer:
Ah, I fond these places:
- https://github.com/doxygen/doxygen/blob/master/src/definition.cpp#L403
- https://github.com/doxygen/doxygen/blob/master/src/docnode.cpp#L5445
- https://github.com/doxygen/doxygen/blob/master/src/docnode.cpp#L5459
- https://github.com/doxygen/doxygen/blob/master/src/docnode.cpp#L5627
- https://github.com/doxygen/doxygen/blob/master/src/docnode.cpp#L5658
- https://github.com/doxygen/doxygen/blob/master/src/docnode.cpp#L5687
- https://github.com/doxygen/doxygen/blob/master/src/markdown.cpp#L3303
- https://github.com/doxygen/doxygen/blob/master/src/tagreader.cpp#L460

I'll make it WIP as long as this is not fixed.

tim-gromeyer:
The tests have already specified Id's. Since the id now contains the title (instead of `autotoc_md` + the number) the ids in the tests are wrong.

see:
- https://github.com/doxygen/doxygen/blob/master/testing/024/indexpage.xml#L10

albert-github:
My opinion is don't change, the procedure has been well thought about and one should as user (nearly) never link to automatically generated anchors as they might change, so they should only be used internally by / in the program.

albert-github:
Did you try e.g. the example in the attached file: [example.tar.gz](https://github.com/doxygen/doxygen/files/10744027/example.tar.gz)
as far as I can tell this now throws the warnings:
```
.../intro.md:5: warning: found subsection command (id: 'introduction') outside of section context!
.../intro_2.md:5: warning: found subsection command (id: 'introduction-1') outside of section context!
```

tim-gromeyer:
Yeah, this is due to this line: <https://github.com/doxygen/doxygen/blob/master/src/docnode.cpp#L5687>.  
I'm still searching for a solution there.

tim-gromeyer:
Does anybody know why we check for `autotoc_md` [here](https://github.com/doxygen/doxygen/blob/master/src/definition.cpp#L403)

albert-github:
The repository gives:
```
Commit: 5b5db5372a769fc15b29981164014b42815f9ae6 [5b5db53]
Date: Sunday, December 15, 2019 11:33:40 AM
Prevent writing automatic anchors to the tag file
```

Still I don't understand what the use case would be for this change, like as said before in https://github.com/doxygen/doxygen/pull/9858#issuecomment-1430904540 the anchors should only be used internally.

tim-gromeyer:
> Still I don't understand what the use case would be for this change

For better compatibility with GitHub. In GitHub READMEs I often see that another section has been linked to. GitHib itself also generates ids for the headings. Doxygen also generates ids for headings, butt the name is different. You can assign an id by adding `{ #<your_id> }`, but that doesn't work with GitHub. Example: <https://github.com/jothepro/doxygen-awesome-css/blob/main/docs/tricks.md#diagrams-with-graphviz-tricks-graphviz>

>like as said before in [#9858 (comment)](https://github.com/doxygen/doxygen/pull/9858#issuecomment-1430904540) the anchors should only be used internally.

I understand that but the standard ids of Doxygen limit the compatibility of READMEs between Doxygen and GitHub.

albert-github:
I thought so, there are already a number of issues regarding the links in markdown files (see e.g. #8378 and the comment https://github.com/doxygen/doxygen/issues/8378#issuecomment-1303586122).
The major problem is the fact that links like `(...)[...]` are not standardized / the mechanism to be used is not described.

tim-gromeyer:
> The major problem is the fact that links like `(...)[...]` are not standardized / the mechanism to be used is not described.

I completely agree with you. We could implement the heading auto-identifier as GitHub has, and add an option to switch between the default (`autotoc_md`) and GitHub's. The option would prevent existing websites to break.

GitHub's and Pandoc's mechanisms are described here: mity/md4c#175

albert-github:
Thanks for the link to the (in-official) description, it is the first time I see something like it.
When implementing something it should probably be as close as possible to the GitHub / pandoc conventions though, GitHub won't probably not understand some "pandoc" ids as `.`  keeps it (what does GitHub do with it)?

It should also be taken into account that in doxygen link ids are global.

Note that with the tag reader it should bet taken into account that the read file might be from an older version of doxygen and thus can have `autotoc_md...` so that this needs to be filtered in the same way as before.

tim-gromeyer:
> When implementing something it should probably be as close as possible to the GitHub / pandoc conventions though, GitHub won't probably not understand some "pandoc" ids as `.` keeps it (what does GitHub do with it)?

GitHub removes the `.`. See here: <https://github.com/tim-gromeyer/stackoverflow/blob/tests/README.md>.
The implementation I made is pretty close to GitHub's. The only difference I know is that when the id is empty it falls back to `autotoc_md`.

> Note that with the tag reader it should bet taken into account that the read file might be from an older version of doxygen and thus can have `autotoc_md...` so that this needs to be filtered in the same way as before.

I would add an option like `HEADING_AUTO_IDENTIFIER` which defaults to `DOXYGEN` and can also be set to `GITHUB` and maybe `PANDOC`. This would prevent any compatibility problems.

tim-gromeyer:
@albert-github I added a option. Would you adjust the text or leave it like it is now.

tim-gromeyer:
If anyone wants to test the GitHub implementation: https://godbolt.org/z/qcvrx1cff

doxygen:
@tim-gromeyer when you are ready with your changes, can you squash your commits to make it easier to review. And please do not do unnecessary changes like sorting #include's

src/markdown.cpp
```diff
@@ -1846,18 +1843,22 @@ static bool isHRuler(const char *data,int size)
 }
 
 std::unordered_map<std::string, int> id_count;
-QCString githubHtmlIdentifier(std::string str)
+QCString githubHtmlIdentifier(const std::string &str)
 {
-  // Replace space with '-'
-  std::replace(str.begin(), str.end(), ' ', '-');
+  static std::string_view germanSharpS = "ß";
```
doxygen, line 1857: There are thousands of non-ascii characters, why is only this one getting a special treatment? You should assume the string has UTF-8 encoding, see utf8.h for helper functions.

tim-gromeyer: For the ß, I know GitHub leaves it in the id, see https://github.com/tim-gromeyer/stackoverflow/blob/tests/README.md

tim-gromeyer:
> And please do not do unnecessary changes like sorting #include's

I'm sorry for sorting your includes. My IDE does it automatically.

tim-gromeyer:
@doxygen I finally squashed it! You can review it if you want to. Just ignore the `pandocHtmlIdentifier`. I couldn't get the function working and the code in the function is a total mess.

tim-gromeyer:
Before anybody reviews, I'm currently working on an better approach using some helper functions from utf8.h

src/tagreader.cpp
```diff
@@ -457,7 +457,6 @@ class TagFileParser
         case InMember:
         case InPackage:
         case InDir:
-          if (m_curString.endsWith("autotoc_md")) return;
           break;
```
albert-github, line 460: Removing this line will give problems when a tag file is used from the 1.8.16 version of doxygen is used as here in the `p1.tag` file of the given example there is the line:
```
<docanchor file="md_intro" title="Description">autotoc_md0</docanchor>
```

src/definition.cpp
```diff
@@ -395,13 +395,13 @@ void DefinitionImpl::addSectionsToIndex()
 
 void DefinitionImpl::writeDocAnchorsToTagFile(TextStream &tagFile) const
 {
-  if (!m_impl->sectionRefs.empty())
-  {
+#define VECTOR_CONTAINS(x, y) (std::find(x.begin(), x.end(), y) != x.end())
+  if (!m_impl->sectionRefs.empty()) {
     //printf("%s: writeDocAnchorsToTagFile(%d)\n",qPrint(name()),m_impl->sectionRef.size());
```
albert-github, line 400: Please keep also consistent layout, in this case the `{` bracket not at the end of the line but on the next line.
(multiple places)

src/definition.cpp
```diff
@@ -395,13 +395,13 @@ void DefinitionImpl::addSectionsToIndex()
 
 void DefinitionImpl::writeDocAnchorsToTagFile(TextStream &tagFile) const
 {
-  if (!m_impl->sectionRefs.empty())
-  {
+#define VECTOR_CONTAINS(x, y) (std::find(x.begin(), x.end(), y) != x.end())
```
albert-github, line 398: As this define is used multiple times wouldn't it be better to have it as a global lambda or (template) function or in a global include file?

src/docnode.h
```diff
@@ -1072,9 +1072,9 @@ class DocPara : public DocCompoundNode
     const HtmlAttribList &attribs() const { return m_attribs; }
     void setAttribs(const HtmlAttribList &attribs) { m_attribs = attribs; }
 
-  private:
```
albert-github, line 1075: For the private don't remove the spaces (consistency).

src/markdown.cpp
albert-github: Please also add in comment a list of rules to show how characters are replaced.

src/markdown.cpp
```diff
@@ -1856,15 +1937,28 @@ static QCString extractTitleId(QCString &title, int level)
...
+    return id;
   }
-  if ((level > 0) && (level <= Config_getInt(TOC_INCLUDE_HEADINGS)))
-  {
-    static AtomicInt autoId { 0 };
+  if ((level > 0) && (level <= Config_getInt(TOC_INCLUDE_HEADINGS))) {
+    static QCString behaviour = Config_getEnumAsString(HEADING_AUTO_IDENTIFIER);
     QCString id;
-    id.sprintf("autotoc_md%d",autoId++);
+    if (behaviour == "DOXYGEN")
```
albert-github, line 1954: Don't test on the string value but on the enum value like is done e.g. for `WARN_AS_ERROR` in `message.cpp`.
Makes it also possible to use a `switch` statement with additional benefit that some compilers will warn when a value is not handled.

albert-github (commented):
Some quick remarks / comments / things to change / add

tim-gromeyer:
I will change it next time when I have time again!

src/tagreader.cpp
tim-gromeyer: I dont really understand this. As far as I can see it starts with autotoc_md, not ends with. It ends with 0 but thats not what we checked for.

albert-github: I think you are right, probably a bug no one noticed, the tag file indeed gives quite clearly `autotoc_md0` so it should be `startsWith` instead of `endsWith`.

tim-gromeyer:
I fixed 1 last bug I found, fixed the code formatting ans squashed it.

src/tagreader.cpp
tim-gromeyer: Yeah and because it went unnoticed for so long it thought it's unnessasary and removed it. If you want I can add some code and check if we are in the doxygen Mode and than check if it starts with `autotoc_md`.

albert-github: I think it is better to add the correct test.

tim-gromeyer: I'll do it today tonight(about when I commited yesterday)

src/markdown.cpp
```diff
@@ -33,26 +33,24 @@
 
 #include <stdio.h>
 
-#include <unordered_map>
-#include <functional>
+#include <algorithm>
```
tim-gromeyer, line 36: I added the `<algorithm>` include. Nothing else changed here.

tim-gromeyer:
@doxygen @albert-github I think I fixed all the bugs and formatting issues (except the sorted includes in `markdown.cpp`).  
I think it is now ready for a review if nobody has any objections/minds.

src/markdown.cpp
```diff
@@ -1856,15 +1899,33 @@ static QCString extractTitleId(QCString &title, int level)
   {
     std::string id = match[1].str();
     title = title.left(match.position());
+    int &count = id_count[id];
+    if (count > 0) {
+      warn(m_fileName,
+           m_lineNr,
+           "A automatically generated id already has the name '%s'!",
```
albert-github, line 1906: 'A' -> 'An'

src/markdown.cpp
```diff
@@ -1856,15 +1899,33 @@ static QCString extractTitleId(QCString &title, int level)
...
     //printf("found match id='%s' title=%s\n",id.c_str(),qPrint(title));
     AUTO_TRACE_EXIT("id={}",id);
-    return QCString(id);
+    return id;
   }
-  if ((level > 0) && (level <= Config_getInt(TOC_INCLUDE_HEADINGS)))
-  {
-    static AtomicInt autoId { 0 };
+  if ((level > 0) && (level <= Config_getInt(TOC_INCLUDE_HEADINGS))) {
+    static HEADING_AUTO_IDENTIFIER_t behaviour = Config_getEnum(HEADING_AUTO_IDENTIFIER);
```
albert-github, line 1914: 'behaviour' -> 'behavior'
(UK-English versus US-English).
Also in other files.

src/markdown.cpp
```diff
@@ -1856,15 +1899,33 @@ static QCString extractTitleId(QCString &title, int level)
...
+    switch (behaviour) {
+    case HEADING_AUTO_IDENTIFIER_t::DOXYGEN:
+      id = doxygenHtmlIdentifier();
+      break;
+    case HEADING_AUTO_IDENTIFIER_t::GITHUB:
+      id = githubHtmlIdentifier(title.str());
+      break;
+    default:
+      id = doxygenHtmlIdentifier();
+    }
```
albert-github, line 1926: - '{' should be on next line
- 'case' (and code inside 'case') should be indented wit 2 more spaces.

tim-gromeyer:
@doxygen Friendly ping: Is it possible to merge the changes before there are any conflicts with the base branch. Of course only if nobody has any objections!

doxygen:
@tim-gromeyer I made some additional changes, see https://github.com/doxygen/doxygen/commit/113c1b5373356b49e74bd575c3226c1092108a84. Please check if I didn't break anything in the process.

tim-gromeyer:
The changes look good! I will test them tonight. If it does not work, I will report it.

albert-github:
Looks like I found a small problem (based on the flutter package).

I thought that when having:
```
[How the DeviceLab runs tests](#how-the-devicelab-runs-tests)
```
and
```
## How the DeviceLab runs tests
```
there would be a link here as well, but this doesn't seem the case, I only see links appear for single word headers.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10813190/example.tar.gz)

tim-gromeyer:
@albert-github Is the problem with my original implementation (I haven't deleted the branch yet) or wirh the modifications made by doxygen

albert-github:
I just tested with the modifications done as this is the version as used in the master version of doxygen (I don't have a version with just your changes).

When running doxygen with the `-d markdown` option I see:

```
======== Markdown =========
---- input -------
# Flutter DeviceLab


## Table of Contents
* [How the DeviceLab runs tests](#how-the-devicelab-runs-tests)
* [Running tests locally](#running-tests-locally)
* [Prerequisites](#prerequisites)
* [Adding tests to continuous integration](#adding-tests-to-continuous-integration)
* [Adding tests to presubmit](#adding-tests-to-presubmit)


## How the DeviceLab runs tests


## Running tests locally


### Prerequisites

## Adding tests to continuous integration

## Adding tests to presubmit

---- output -----
@page flutterdevicelab Flutter DeviceLab\ilinebr @ianchor{Flutter DeviceLab} md_aa\ilinebr


@section tableofcontents Table of Contents
* @ref "how-the-devicelab-runs-tests" "How the DeviceLab runs tests"
* @ref "running-tests-locally" "Running tests locally"
* @ref "prerequisites" "Prerequisites"
* @ref "adding-tests-to-continuous-integration" "Adding tests to continuous integration"
* @ref "adding-tests-to-presubmit" "Adding tests to presubmit"


@section howthedevicelabrunstests How the DeviceLab runs tests


@section runningtestslocally Running tests locally


@subsection prerequisites Prerequisites

@section addingteststocontinuousintegration Adding tests to continuous integration

@section addingteststopresubmit Adding tests to presubmit

=========
```
so in the translation fro, `##` to `@section` the problem occurs.

doxygen:
Found and fixed the issue (see https://github.com/doxygen/doxygen/commit/2bc76492eb4702b99ae0b9f5a6f5a91f1012a548). My bad.

albert-github:
Looks better (just tested my example and also the flutter package).

tim-gromeyer:
I have now tested it too and it works!

albert-github:
I see in the HTML directory some strange filenames appear when in the `DOXYGEN` mode i.e. the file name: `autotoc_md0.html`
whilst in the old situation this was: `md_aa.html`

The used source file **aa.md**:
```
# Some page

Some text
```
the used **Doxyfile**:
```
QUIET = YES
```

I don't think this is nice and that it should be corrected (even though it is an internal name).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10816316/example.tar.gz)

tim-gromeyer:
Unfortunately I have no idea how to fix this, maybe I'll figure it out.

tim-gromeyer:
@doxygen I found a bug(also in my version).
I wrote a comment unter the commit(https://github.com/doxygen/doxygen/commit/113c1b5373356b49e74bd575c3226c1092108a84):
To reproduce:
**index.md**:
```markdown
[TOC]

# Test {#test}

## Test
```

**Doxyfile**:
```
MARKDOWN_ID_STYLE = GITHUB
```

doxygen:
@albert-github , @tim-gromeyer Thanks for the quick feedback. Commit https://github.com/doxygen/doxygen/commit/e1ba6ae34de33b2d962005dcab359918c439039b should address both issues.

tim-gromeyer:
The error I found seems fixed, I'm glad I could help!

albert-github:
Looks like I found 2 more problems. With the code:
```
# bun

Some text

## Table of Contents

- [bun](#bun)
  - [`Bun.spawn` – spawn a process](#bunspawn--spawn-a-process)


## `Bun.spawn` – spawn a process

`Bun.spawn` lets you quickly spawn a process. Available as of Bun v0.2.0.
```
I get the warnings:
```
README.md:7: warning: unable to resolve reference to 'bun' for \ref command
README.md:8: warning: unable to resolve reference to 'bunspawn--spawn-a-process' for \ref command
```

1. "bun" the first header is promoted to the page level (`@page` command) but this make the "bun" link not present. Probably an extra `\ianchor bun` is needed here
2. the "bunspawn--spawn-a-process", this is a bit a mean one as it looks like we have a minus sign here (octal 055) but in fact it is s special minus sign consisting of three bytes (octal 342 200 223). Looking at the resulting markdown code:
   ```
   @page md__r_e_a_d_m_e bun\ilinebr

   Some text

   @section table-of-contents Table of Contents

   - @ref "bun" "bun"
     - @ref "bunspawn--spawn-a-process" "`Bun.spawn` – spawn a process"


   @section bunspawn-–-spawn-a-process <tt>Bun.spawn</tt> – spawn a process

   <tt>Bun.spawn</tt> lets you quickly spawn a process. Available as of Bun v0.2.0.
   ```
   we see that the special minus is not filtered out of the section label.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10831328/example.tar.gz)

albert-github:
Another one, regarding numbers:
```
# Note

As of 3.0.0, the History.md file has been deprecated.
[Please refer to the full commit logs available on GitHub](https://github.com/chaijs/chai/commits/master).

---

2.3.0 / 2015-04-26
==================

some more text
```
give a lot of warnings like:
```
xx.md:1: warning: Unexpected character '3' while looking for section label or title
```
and one:
```
xx.md:8: warning: Invalid or missing section label
```

This is due to the fact that the translation of
```
2.3.0 / 2015-04-26
==================
```
is:
```
@section 230--2015-04-26 2.3.0 / 2015-04-26
```
where the label is:
```
230--2015-04-26
```
so starting with a number which is not (yet) supported.

Example: [example_2.tar.gz](https://github.com/doxygen/doxygen/files/10831390/example_2.tar.gz)

tim-gromeyer:
> Another one, regarding numbers:

I already know this one. That's why I didnt implement pandoc. There was just no way to test if it works.

The bug with the `EN -` is a tricky one indeed. I think by using the ASCII number(45 I think) instead if the dash we can avoid the issue.

doxygen:
@albert-github with [this commit](https://github.com/doxygen/doxygen/commit/ca204d56b644e22d3f3bf1dec64a43a87076c219) I've tried to address the 2 "bun" related issues. The last issue with headers starting with a digit, I'm not sure yet if we can or want to support that (not allowed in HTML4 and XHTML).

albert-github:
There are a number sides on the number issue:
- to be honest I don't know what e.g. GitHub does with the labels starting with a number (didn't yet see an example, I only saw the header so far) 
- the warning given by doxygen, this is a bit of an annoying side effect of the `MARKDOWN_ID_STYLE=GITHUB` (as already used by Fossies)

As a solution I can think of:
- in case a (potential) label starts with a number we do a fallback to a more specific doxygen label by prepending to the `MARKDOWN_ID_STYLE=GITHUB` label a specific term like, `autotoc_md<number>_<label>` where `<number>` is an increasing number so it is always unique and `<label>` is the `MARKDOWN_ID_STYLE=GITHUB` type label. {Probably also gives later on the easiest conversion when we know what GitHub does).

albert-github:
I just found an example on GitHub in respect to numbers and they violate HTML4 / XHTML as in the example I found at 
https://github.com/oven-sh/bun/blob/main/src/CHANGELOG.md the "label" `0.41.0` is translated into `#0410`.

tim-gromeyer:
> * to be honest I don't know what e.g. GitHub does with the labels starting with a number (didn't yet see an example, I only saw the header so far)

It includes all the numbers, for example:

```markdown
2.3.0 / 2015-04-26
==================
```

will be `#230--2015-04-26`.
<br>
`0.41.0` is translated to `#0410` because GitHub removes the dots because they are punctuation.
<br>
Maybe we can add an warning, that it might break HTML4 and XHTML compatibility

albert-github:
I don't think it is a good idea to break HTML4 / XHTML compatibility I think it is better that we translate the (as example `0.41.0`) like given before:
- the definition place
   ```
   0.41.0
   -------
   ```
   becomes
   ```
   @section autotoc_md1_0410 "0.41.0"
   ```
- usage place
   ```
   [the chapter 0.41.0](#0410)
   ```
   becomes
   ```
   \ref autotoc_md1_0410 "the chapter 0.41.0"
   ```

tim-gromeyer:
Sounds like a good idea, but would be harder to implement.
I would suggest removing the `<number>` from `autotoc_md<number>_<label>` as long as the reason is "so it is always unique". The GitHub label itself already has a number at the end if it occurs in multiple places.

albert-github:
I agree that the `<number>`  is not necessary as in case of multiple identical `<label>`s a number is appended. This brings me straight away with the next problem I encountered (also in the "bun" package). I tried to, quickly, reproduce the problem but got slightly different results based on the following example:
**file_1.md**:
```
The first file

# Upcoming breaking changes
```
**file_2.md**:
```
The second file

- [Upcoming breaking changes](#upcoming-breaking-changes)


# Upcoming breaking changes
```
Here we don't get any warnings but we don't jump to the right place but in the original example we get:
```
README.md:96: warning: unable to resolve reference to 'upcoming-breaking-changes' for \ref command
README.md:174: warning: unable to resolve reference to 'bundns---lookup-a-domain' for \ref command
README.md:219: warning: unable to resolve reference to 'prefer-latest' for \ref command
```
so again warnings (there are more warnings here but they are  not really relevant for the case here).

As far as I can tell the problem arises from the fact that in GitHub the labels just have to be unique inside a file (and subsequent numbering is done just inside the file) whilst for doxygen the labels have to be globally unique.

Also in doxygen it is possible that a label is referred outside the file where it is defined but with GitHub this is not the case.

This is a hard problem, but probably an approach where the file name + path is incorporated into the label has to be chosen so that the labels remain unique over files and can be numbered inside the file.
The "mangling" probably has to be done at the place where the label is defined and used.

Note that the internal used labels have to be unique inside doxygen and that the translation from GitHub mangled labels to doxygen labels has to be done but that the translation the other way around is not necessary.


Small example: [example_small.tar.gz](https://github.com/doxygen/doxygen/files/10834190/example_small.tar.gz)
Original example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10834192/example.tar.gz)

tim-gromeyer:
> Also in doxygen it is possible that a label is referred outside the file where it is defined but with GitHub this is not the case.

If I have understood you correctly, this is not quite right, see https://github.com/tim-gromeyer/stackoverflow/blob/tests/README.md

> This is a hard problem, but probably an approach where the file name + path is incorporated into the label has to be chosen so that the labels remain unique over files and can be numbered inside the file. The "mangling" probably has to be done at the place where the label is defined and used.

This is indeed a difficult problem. Too difficult for me to solve it, because of the complexity of doxygen. But I think you are right. With the combination of the relative path and the filename it should be possible.

> Note that the internal used labels have to be unique inside doxygen and that the translation from GitHub mangled labels to doxygen labels has to be done but that the translation the other way around is not necessary.

albert-github:
Regarding the number problem I now also found a problem regarding a minus sign at the beginning, in fact the line:
```
### * tools/matchcompiler.py
```
translates into:
```
@subsection -toolsmatchcompilerpy * tools/matchcompiler.py
```
which is a correct translation, but doxygen does not like the `-` at the beginning of a label which results in:
```
cppcheck-2.10.2/tools/readme.md:3: warning: Invalid or missing section label
```
(Found by Fossies in the cppcheck 2.10.2 package)

tim-gromeyer:
> which is a correct translation, but doxygen does not like the `-` at the beginning of a label which results in:

Why does doxygen not like the `-` at the beginning? Is it some kind of validation that only valid identifiers are included? If so, we could skip checking in GitHub mode, since GitHub obviously doesn't care about XHTML and HTML4.

albert-github:
The `-` is probably also not allowed in HTML4 and XHTML as the beginning of a label.
> we could skip checking in GitHub mode, since GitHub obviously doesn't care about XHTML and HTML4.

But doxygen does care about XHTML and HTML4 as doxygen documentation is normally not read on GitHub but in a browser.

tim-gromeyer:
> The `-` is probably also not allowed in HTML4 and XHTML as the beginning of a label.

This might be true but the GitHub id mangling is HTML5 only. This should definitely be noted in the documentation. 

> But doxygen does care about XHTML and HTML4 as doxygen documentation is normally not read on GitHub but in a browser.

Yes, that's true, but can you name me a browser that doesn't support HTML5. That was an issue 8 years or so ago, but not anymore.

doxygen:
I think it is a matter of defining the purpose of `MARKDOWN_ID_STYLE=GITHUB`. 
If it is to

1) allow hand written GitHub pages to be (also) processed as-is by doxygen
2) allow cross-references between hand written pages and doxygen generated pages.

Then doxygen should not alter the anchors and links, support leading `-`s and digits, and also treat each page as a unique namespaces (at least for HTML). The consequence is then to only support HTML5, and when HTML4 or XHTML compliance is needed the user should switch back to the default `MARKDOWN_ID_STYLE=DOXYGEN`.

tim-gromeyer:
> I think it is a matter of defining the purpose of `MARKDOWN_ID_STYLE=GITHUB`. If it is to
> 
> 1. allow hand written GitHub pages to be (also) processed as-is by doxygen
> 2. allow cross-references between hand written pages and doxygen generated pages.

I think both, at least partially.

> Then doxygen should not alter the anchors and links, support leading `-`s and digits, and also treat each page as a unique namespaces (at least for HTML).

That sounds like a solution, a **complex** one.

> The consequence is then to only support HTML5, and when HTML4 or XHTML compliance is needed the user should switch back to the default `MARKDOWN_ID_STYLE=DOXYGEN`.

That's exactly what I was trying to say!

albert-github:
I think the purpose would be to: allow hand written GitHub pages to be (also) processed by doxygen. I explicitly left out the words "as-is".
I think doxygen should:
- take the `#` page / section / ... lines and translate (as done now under `MARKDOWN_ID_STYLE=GITHUB`) to a label. In case the resulting label starts with a digit or a `-` sign the generated label should be prepended by a reserved word (like `autotoc_md`).
- take the lines like `[...](...)` (as this is how GitHub refers to sections etc. where the part between `(` and `)`is taken and when it starts with a `#` it is stripped away (might be more than one, I saw it once so far but I think this is actually a typo in the code but it works) and when the remaining string (i.e. the label) starts with a digit or a `-` sign the label should be prepended by the reserved word

I think in this way the labels will work in all of HTML4, HTML5 and XHTML. Not doing this will limit the usability quite a bit.

tim-gromeyer:
This sounds like a good idea, and if I can help implement it, I'll give it a try.

albert-github:
Based on the ideas as given in https://github.com/doxygen/doxygen/pull/9858#issuecomment-1447835301for github labels staring with a digit of `-` I implemented it in the  proposed patch, pull request #9888

PR #9862, opened 2023-02-16
Fix server based search for PHP 8.1

dwrobel:
Fixes the following errors when using server based search:
```
 PHP Fatal error:  Array and string offset access syntax with curly \
   braces is no longer supported in search_functions.php on line 65
 PHP Fatal error:  Array and string offset access syntax with curly \
   braces is no longer supported in search_functions.php on line 325
 Uncaught Error: Call to undefined function each() in search_functions.php:205
 #0 search_functions.php(339): filter_results()
 #1 search_functions.php(353): run_query()
 #2 search.php(58): main()
 #3 {main}
   thrown in search_functions.php on line 205
```
Tested with:
```
 $ rpm -qv php-fpm
 php-fpm-8.1.15-1.fc37.x86_64
```

### DIFF
--- a/templates/html/search_functions.php
+++ b/templates/html/search_functions.php
@@ -62,10 +62,10 @@ function computeIndex($word)
   // Simple hashing that allows for substring search
   if (strlen($word)<2) return -1;
   // high char of the index
-  $hi = ord($word{0});
+  $hi = ord($word[0]);
   if ($hi==0) return -1;
   // low char of the index
-  $lo = ord($word{1});
+  $lo = ord($word[1]);
   if ($lo==0) return -1;
   // return index
   return $hi*256+$lo;
@@ -202,7 +202,7 @@ function combine_results($results,&$docs)
 function filter_results($docs,&$requiredWords,&$forbiddenWords)
 {
   $filteredDocs=array();
-  while (list ($key, $val) = each ($docs)) 
+  foreach ($docs as $key => $val)
   {
     $words = &$docs[$key]["words"];
     $copy=1; // copy entry by default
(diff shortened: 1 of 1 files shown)


dwrobel:
>(as the title explicitly stated PHP 8.1)
I used explicitly 8.1 only because that was the version I have access to and I could test using it.

> Will this also work with older versions of PHP i.e. PHP 8.0

I did a quick test for `foreach` on https://phpcodechecker.com/:
```
$arr = array(1, 2, 3, 4);
foreach ($arr as &$value) {
    $value = $value * 2;
}:
```
and it worked fine on all `5.6`, `7.4` and `8.2` versions.

The same for works fine for testing `$word[0]`:
```
function computeIndex($word)
{
  // Simple hashing that allows for substring search
  if (strlen($word)<2) return -1;
  // high char of the index
  $hi = ord($word[0]);
  if ($hi==0) return -1;
  // low char of the index
  $lo = ord($word[1]);
  if ($lo==0) return -1;
  // return index
  return $hi*256+$lo;
}
```

If you change back `$word[0]` to `$word{0}` then the checker reports error on `8.2`.

albert-github:
Gives quite a bit of confidence, though for the foreach you tested:
```
foreach ($arr as &$value)
```
whilst the new code states:
```
foreach ($docs as $key => $val)
```

as not a php expert are these the same or ...

dwrobel:
>whilst the new code states:
>foreach ($docs as $key => $val)

The following example with associative table also works for all versions:
```
$capitals = [
	'Japan' => 'Tokyo',
	'France' => 'Paris',
	'Germany' => 'Berlin',
	'United Kingdom' => 'London',
	'United States' => 'Washington D.C.'
];

foreach ($capitals as $country => $capital) {
	echo "The capital city of {$country} is $capital" . '<br>';
}
```

Issue #9861, opened 2023-02-15
Links among markdown files don't work if upstream path element contains space


dschlaep:
**Describe the bug**
Doxygen does not resolve markdown links of the form `[here](MyMarkdownFile.md)`  for me if my project resides on a path where an upstream path element contains a space, e.g., "path with space/MyProject".

Warnings: (note I removed non-relevant parts of the path and inserted `...` instead)

> Generating docs for page md_doc_additional_doc_MyMarkdownFile...
> /.../path with space/MyProject/doc/additional_doc/MyMarkdownFile.md:7: warning: unable to resolve reference to '/.../path' for \ref command

and

> Generating index page...
> /.../path with space/MyProject/README.md:12: warning: unable to resolve reference to '.../path' for \ref command
> 

This results in the following html documentation
```
Additional information can be found /.../path with [space/MyProject/doc/additional_doc/MyMarkdownFile.md](file:///.../path%20with%20space/MyProject/doc/html/_my_markdown_file_8md.html) "here".
```

**Expected behavior**
When I copy the exact same project to a path without spaces, then the markdown links are correctly generated and doxygen does not issue warnings.

The documentation is working correctly; the corresponding line to the one cited above is
```
Additional information can be found [here](file:///.../path_without_space/MyProject/doc/html/md_doc_additional_doc__my_markdown_file.html).
```

**Screenshots**
NA

**To Reproduce**
Attached [debug_doxygen_paths.zip](https://github.com/doxygen/doxygen/files/10745210/debug_doxygen_paths.zip) contains a simple "Hello Word" program in C with some extra markdown files `README.md` and `MyMarkdownFile.md` which link to each other:

> - `demonstrate_path_issue.sh` 
> - MyProject/
>     - doc/
>         - `Doxyfile`
>         - additional_doc/
>             - `MyMarkdownFile.md` which contains the link `[markdown README](README.md)`
>         - `doc_open.sh`
>    - src/
>        - `HW.c`
>    - `README.md` which contains the link `[here](doc/additional_doc/MyMarkdownFile.md)`
>    - `makefile`
> 

The bash script `demonstrate_path_issue.sh`
- creates two folders `"path_without_space"` and `"path with space"`
- copies "`MyProject/"` into each created folder
- enters each copy, i.e., `"path_without_space/MyProject"` and `"path with space/MyProject"`
- runs doxygen from "MyProject" using "doc/Doxyfile" to create the documentation -- see warnings that I copied above
[...]

PR #9863, opened 2023-02-17
issue #9861 Links among markdown files don't work if upstream path element contains space

albert-github:
Doxygen replaces the given path / filename with the full path and due the the spaces in the path this leads to "problems" during further processing. The problem does not only occur due to the "upstream" part but also to local path / filenames with spaces. By embedding the link name always in double quotes (`"`) the problem can be overcome when the double quotes are handled.

Extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10766441/example.tar.gz)

### DIFF
--- a/doc_internal/commands_internal.md
+++ b/doc_internal/commands_internal.md
@@ -12,6 +12,7 @@ and the version in which they were introduced.
 \refitem cmdiline \\iline
 \refitem cmdilinebr \\ilinebr
 \refitem cmdiliteral \\iliteral
+\refitem cmdiref \\iref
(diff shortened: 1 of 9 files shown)


dschlaep:
Many thanks, this is awesome! I tested with the extended example attached to this PR and with my original example attached to the issue.

albert-github:
An alternative code update has been made in master on GitHub (https://github.com/doxygen/doxygen/commit/b8c00336322b2f285677b5ca3aff4a25be353341), so this PR is obsolete.

PR #9864, opened 2023-02-17
Double identifiers due to mapping to same id of file names

albert-github:
During the translation of filenames etc. to e.g. the `@page` command a number of characters are mapped to the same id by means of an underscore resulting in warnings like:
```
additional_doc/MyMarkdownFile.md:1: warning: multiple use of page label 'md_additional_doc_MyMarkdownFile' with different titles, (other occurrence: .../additional+doc/MyMarkdownFile.md, line: 1)
```
(here it is for the directory names `additional_doc` and `additional*doc`, by using a more sophisticated translation this can be prevented

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10767332/example.tar.gz)


testing/055_markdown.md
```diff
@@ -1,6 +1,6 @@
 <!--
 // objective: test markdown parsing
-// check: md_055_markdown.xml
+// check: md_055__markdown.xml
```
doxygen, line 3: I think the original reference file (without the double underscope) can be removed, right?

albert-github: Yes, I had some trouibles with git and this one was, unfortunately, left over.
I removed it.

PR #9874, opened 2023-02-23
Portuguese translators updated to 1.9.7

fjtc:
This commit updates both Portuguese translators to 1.9.7. I also merged the Portuguese translator with the Brazilian Portuguese translator in order to make future updates easier.

This was possible because both languages are almost identical in their written form with minor variations in vocabulary and phrase constructs. In fact, there are some European Portuguese constructs that are not commonly used by Brazilians but are grammatically correct in Brazilian Portuguese.  I don't know if the same happens the other way around. 

From now on, the Portuguese translator will be a subclass of the Brazilian translator with the methods that do differ between those languages overridden in the Portuguese translator.  I hope this change will reduce the workload required to maintain both translators.

### DIFF
--- a/src/translator_br.h
+++ b/src/translator_br.h
@@ -19,6 +19,8 @@
  *    Thanks to Jorge Ramos, Fernando Carijo and others for their contributions.
  *
  * History:
+ * 20230219:
+ *  - Updated to 1.9.7;
  * 20220911:
  *  - Updated to 1.9.6;
  * 20220525:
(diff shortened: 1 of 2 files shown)


albert-github:
I think a very good idea to merge the both languages.

You wrote:
> From now on, the Portuguese translator will be a subclass of the Brazilian translator with the methods that do differ between those languages overridden in the Portuguese translator. I hope this change will reduce the workload required to maintain both translators.

I think it would be better to have the (European) Portuguese as base class and the Brazilian Portuguese as subclass, I see the Brazilian Portuguese as a variety of the (European) Portuguese. I also compare this with the Spanish where the South-American Spanish languages are varieties of the (European) Spanish. With the German language I see the same where e.g. Austrian German and Swiss German are varieties of the official German.
(Maybe this is a bit old-fashioned thinking and it is not correct though)

There is also an issue with the documentation in the Internationalization chapter where for the Brazilian (looks OK)
image
but for the Portuguese:
image

but this is something that is probably out of your scope and we have to address.

I was also thinking about a construct analogous to the `TranslatorAdapterBase` where for none defined translations the English translation is used. I can think of a situation, in the future as at the moment this is no problem as both translators are in one hand, where the (European) Portuguese is and the Brazilian Portuguese is not but in that case both languages are said to be up to date.

But all in all I still think it is a good idea to merge these languages.

@doxygen which steps to take?

fjtc:
Hi @albert-github,

I completely forgot about the automated translator verification tool. Since it works by perform a simple check in the translator file, looking for the presence and/or absence of the required methods, I don't think it is worth to change it just to because of the merge of the Portuguese translators. I'd love to work on a way to let the verification tool work with this case but I don't have time free time to work on it right now. Given that, I think it is better to split the translators again.

I decided to merge both translators because they share most of the methods and the newly added methods required a better way to manage the gender of some terms. This gave me the incentive to experiment the translator merge.

I'll close this request and create a new pull request with both translators as distinct entities again. To avoid some of the code duplication I will experiment some alternative options to see which one will work without disturbing the translator verification tool as it is an essential tool to keep the translations up to date.

Thanks for your feedback!

Issue #9883, opened 2023-02-27
Performance issues when HAVE_DOT is YES to make `\dot` work, and `CLASS_GRAPH=GRAPH` is also used


krolli:
In our code-base, we have been relying on `\dot` command working despite setting `HAVE_DOT=NO`. This has become problem as soon as https://github.com/doxygen/doxygen/issues/7273 has been resolved, as all uses of `\dot` command are now ignored.

After switching to `HAVE_DOT=YES`, uses of `\dot` work, but we have noticed an order-of-magnitude slowdown in doxygen performance. We disabled all graphs that we don't use but were enabled by default, leaving us with only `CLASS_GRAPH=GRAPH`, which helped performance a little bit, but it is still about 6x slower compared to just setting `HAVE_DOT=NO`.

We also verified that `CLASS_GRAPH=TEXT` gives us the same generation time as `HAVE_DOT=NO` to make sure there is nothing else that is causing significant amount of graphs to be generated.

Ultimately, we prefer built-in class graph generator over dot-based one, so we would like to have option to use that even though `HAVE_DOT=YES` is set.

We will probably be using `CLASS_GRAPH=TEXT` while iterating on documentation to keep things fast, and use `CLASS_GRAPH=GRAPH` on CI servers producing documentation over night. Though slowdown is an issue on CI as well, it's not as much of a problem as it is for documentation authors.

I don't know what would be the best solution to this problem in regards to overall doxygen architecture and design goals. One thing that would work perfectly well for us would be adding another possible value for `CLASS_GRAPH`, let's call it `BUILTIN_GRAPH`, which would force use of built-in class graph generator even when `HAVE_DOT=YES` is set. It should also be backwards compatible and not affect any existing configs using one of the previous values.

albert-github:
- Which version of doxygen are you using?
- You mention a factor 6 but what are some absolute times? Do we speak about second, minutes, hours?
- What is the used setting of `DOT_NUM_THREADS`?
- What is the used setting of `DOT_IMAGE_FORMAT`?
- Do you clean the output directory between doxygen runs (not advised in your case as it will also remove the `.md5` files and `.png` / `.svg` resulting image files, they signal that an unchanged graph file doesn't need to be regenerated. In case of the internal images this might also be a big speedup as the internal files are always regenerated).

krolli:
> Which version of doxygen are you using?

[...]

PR #9884, opened 2023-02-27
Issue #9883 Performance issues when HAVE_DOT is YES to make \dot work, and CLASS_GRAPH=GRAPH is also used

albert-github:
Define `INTERNAL_GRAPH` for the setting `CLASS_GRAPH` which always generates the class graphs with the internal diagram generator irrespective of the setting of `HAVE_DOT`


doxygen:
@albert-github Decided to use `BUILTIN` as option value name instead of `INTERNAL_GRAPH`

PR #9888, opened 2023-03-01
issue 9855 Markdown: Heading auto id starting with digit of minus sign

albert-github:
Based on the comment https://github.com/doxygen/doxygen/pull/9858#issuecomment-1447835301:
```
    take the # page / section / ... lines and translate (as done now under MARKDOWN_ID_STYLE=GITHUB) to a label. In case the resulting label starts with a digit or a - sign the generated label should be prepended by a reserved word (like autotoc_md).

    take the lines like [...](...) (as this is how GitHub refers to sections etc. where the part between ( and )is taken and when it starts with a # it is stripped away (might be more than one, I saw it once so far but I think this is actually a typo in the code but it works) and when the remaining string (i.e. the label) starts with a digit or a - sign the label should be prepended by the reserved word
```

this feature has been implemented.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10859639/example.tar.gz)

### DIFF
--- a/src/anchor.cpp
+++ b/src/anchor.cpp
@@ -84,6 +84,7 @@ std::string AnchorGenerator::generate(const std::string &label)
     }
     else
     {
+      if ((result[0] == '-') || (result[0]>='1' && result[0]<='9')) result = "autotoc_md" + result;
       int &count = p->idCount[result];
       // Add end digits if an identical header already exists
       if (count>0)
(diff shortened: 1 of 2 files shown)


doxygen:
@albert-github I made some further simplifications and also allowed a 0 as start of an anchor, see https://github.com/doxygen/doxygen/commit/7717739ef333b89d5c33c4c4fc0ed46b707d4cc2

albert-github:
@doxygen The intention was to have the `0` also in the list and should have been in. I did a grep for `isDigit` and did a copy / paste of the only one in the repository that did the start check at `'1'` but apparently that didn't register properly with me at that moment.

PR #9890, opened 2023-03-02
Text on start line of fenced code block should not be ignored.

albert-github:
When having code like:
~~~
# Releasing PyTorch

  - [Cutting release branches](#cutting-release-branches)

``` python github_analyze.py --repo-path ~/local/pytorch --remote upstream  --branch release/1.11 --milestone-id 26 --missing-in-branch ```

## Cutting release branches

```bash
git clone git@github.com:pytorch/pytorch.git
```
~~~

and setting:
```
MARKDOWN_ID_STYLE = GITHUB
```

we get a warning like:
```
RELEASE.md:3: warning: unable to resolve reference to 'cutting-release-branches' for \ref command
```

This is due to the fact that the parts after the first ` ``` ` is ignored
- in case a space follows the ` ``` ` it should be considered as a fenced comment block without language specifier
- the rest of the line should not be ignored (as this would automatically mean the the closing ` ``` ` isn't seen.
- closing `}` should be skipped from output.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10872063/example.tar.gz)

(Found by Fossies for pytorch package)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -2094,7 +2094,9 @@ static bool isFencedCodeBlock(const char *data,int size,int refIndent,
   } // not enough tildes
   if (i<size && data[i]=='{') i++; // skip over optional {
   int startLang=i;
-  if (data[i] != ' ')
+  lang = "";
+  char c = data[i];
+  if (c == '{' || (c>='A' && c<='Z') || (c>='a' && c<='z'))
   {
     while (i<size && (data[i]!='\n' && data[i]!='}' && data[i]!=' ')) i++;
     convertStringFragment(lang,data+startLang,i-startLang);
@@ -2109,7 +2111,6 @@ static bool isFencedCodeBlock(const char *data,int size,int refIndent,
       int endTildes=0;
       while (i<size && data[i]==tildaChar) endTildes++,i++;
       while (i<size && data[i]==' ') i++;
-      if (i==size || data[i]=='\n')
       {
         if (endTildes==startTildes)
         {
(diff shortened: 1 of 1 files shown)


albert-github:
Similar problems as with the start of the fenced code block can occur with the end of the code block as illustrated with:
~~~
# Table of Contents

 * [Performance annotations](#performance-annotations)

```/full/path/to/envoy/bazel-bin/source/exe/envoy-static:_Znwm```. The part before

```
.LFE73:
```


# Performance annotations

```cpp
void doHeavyLifting() {
```
~~~

- handling the end of the code block which doesn't need be at the end of a line at the end of the file
- determination of the language: only a `{` or an alphabetic character can start the name of a language

Example: [example_1.tar.gz](https://github.com/doxygen/doxygen/files/10873682/example_1.tar.gz)

(found by Fossies for the envoy package)

doxygen:
@albert-github I've made the different cases more explicit, which should also make parsing it a bit more robust and the code more clear, see https://github.com/doxygen/doxygen/commit/49dc16a5217905957c06dfec1a8b8ec79ac663f4

albert-github:
I didn't test it but wouldn.t this block things like: ` ```{.c++} `
with `EXTENSION_MAPPING = c++=cpp`

albert-github:
Looks like it is even worse, the `{c++}` isn't and wasn't supported at all but a filename like `aa.c++` is supported.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/10888220/example.tar.gz)

PR #9903, opened 2023-03-09
Compile with -Werror=pedantic on one platform

lrineau:
A CI check with `-Werror=pedantic` will avoid PRs like #9902.


lrineau:
This PR detects the warnings: https://github.com/doxygen/doxygen/actions/runs/4375832191/jobs/7657088636#step:25:290

I will rebase, once #9902 is merged.

.github/workflows/build_cmake.yml
```diff
@@ -25,7 +25,7 @@ jobs:
             os: ubuntu-20.04,
             build_type: "Debug", cc: "gcc", cxx: "g++",
             build_gen: "Unix Makefiles",
-            cmake_extra_opts: "-Dbuild_search=YES -Dbuild_app=YES -Dbuild_parse=YES -Dbuild_xmlparser=YES -Duse_sqlite3=ON"
+            cmake_extra_opts: "-Dbuild_search=YES -Dbuild_app=YES -Dbuild_parse=YES -Dbuild_xmlparser=YES -Duse_sqlite3=ON -DCMAKE_CXX_FLAGS='-Werror=pedantic'"
```
albert-github, line 28: It is good to remove these types of warnings though I don't think it is a good idea to have this on the command line and to "promote" this to an error. Furthermore I don't think it is actually an error as it more or less signals an empty statement, but it doesn't look nice in the code.
With GCC 11.3.0 O cannot reproduce the warning, nor with wit the the Microsoft compiler (not even when setting `\Wall`).

We have already `-Wpedantic` in the in file CompilerWarnings.cmake and we see in e.g. the GitHub Action https://github.com/doxygen/doxygen/actions/runs/4368671534/jobs/7641547392 already the warning:
```
[ 77%] Building CXX object src/CMakeFiles/doxymain.dir/dotgroupcollaboration.cpp.o
/home/runner/work/doxygen/doxygen/src/dotgroupcollaboration.cpp:56:2: warning: extra ‘;’ [-Wpedantic]
   56 | };
      |  ^
```

PR #9914, opened 2023-03-14
A Lax is a Fish, and not a Cookie

ecxod:
Cookie “doxygen_width” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,50 @@
+name: Build Debian Package
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-package:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout source code
+      uses: actions/checkout@v2
+
+    - name: Install build dependencies
+      run: sudo apt-get build-dep doxygen
+
+    - name: Build Doxygen from source
+      run: |
+        cd ${{ github.workspace }}
+        ./configure
+        make
+
+    - name: Create package directory
+      run: mkdir -p ${{ github.workspace }}/my-doxygen-package/usr/bin/
+
+    - name: Copy binaries to package directory
+      run: |
+        cp ${{ github.workspace }}/bin/doxygen ${{ github.workspace }}/my-doxygen-package/usr/bin/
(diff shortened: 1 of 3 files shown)


albert-github:
- What has the second change to do with the subject?
   -  When not related create a separate PR for it
- Why can this not be part of the `build_cmake.yml`

albert-github:
Same questions with the [44f5ea6](https://github.com/doxygen/doxygen/pull/9914/commits/44f5ea6212bdcd403778f59d5ee5447e785c0386) update  as with the [13c9892](https://github.com/doxygen/doxygen/pull/9914/commits/13c9892ac6278e7b51df337ebf3af38c8c8ecab3) update.

doxygen:
@ecxod `SameSite=Lax` is listed in the page to refer to, so why you say it is invalid? Does `SameSite=Strict` work cross browser with the same origin policy when the `file://` protocol is used?

ecxod:
It's always the case that when you get to know people better, they disappoint you. Do what you want. i will compile a fork and work with my fork. To use a cookie for something you can do with CSS is already a hammer, But a question : is one frame in one domain (http://left.example.com) and the other in another domain (http:// right.example.com) ?

doxygen:
> It's always the case that when you get to know people better, they disappoint you. 

Not sure what warrants such harsh reaction on a simple question.

> To use a cookie for something you can do with CSS is already a hammer, 

Care to explain?

> But a question : is one frame in one domain (http://left.example.com/) and the other in another domain (http:// right.example.com) ?

No, but doxygen's html pages are often viewed locally without web server, and one frame can be `file://some/path/file1.html` and another `file://some/path/file2.html` and then some browsers (such as Chrome) have the "same origin policy" implemented such that this is treated as two different domains.

ecxod:
You don't need a cookie for this. If you don't like css, you can use localstorage or a session for such things. Cookies are only necesary if you want to comunicate between javascript and the server.

Issue #9894, opened 2023-03-05
config.xml  to Chinese


wb2712:
**Describe the bug**
config to Chinese  by google translate
The final compiled program cannot run
In the process, some problems were solved and shared here

I use this software to generate manuals for some large-scale open source software, which helps me quickly understand, but the first run usually freezes and can only be forced to kill, and the second time is much better

FindIconv.cmake   
Unable to compile and run correctly, I ended up shielding this problem by a tricky way, but did not solve this problem

The following two issues are not fixed in my environment
#9854 
#8335

**Expected behavior**
Use my translated configuration files in the next version as soon as possible
Or to solve my compilation problem

**Screenshots**
![image](https://user-images.githubusercontent.com/112671903/222959833-86d22734-d5f7-4624-8025-363e0d5a32ae.png)
![image](https://user-images.githubusercontent.com/112671903/222960078-8a2004b3-0623-4907-9cdc-b9575f397445.png)
![image](https://user-images.githubusercontent.com/112671903/222960092-7f56c4fc-1d0d-4486-90c4-59541df89cae.png)
![image](https://user-images.githubusercontent.com/112671903/222960360-8243c36a-caf7-4a46-9ce2-10d10d860d9e.png)
![image](https://user-images.githubusercontent.com/112671903/222960436-071e63d3-32fb-4493-8650-e10e13f7acec.png)
![image](https://user-images.githubusercontent.com/112671903/222960554-ab5d341b-a673-4bb9-8052-38cab4def598.png)
![image](https://user-images.githubusercontent.com/112671903/222960594-4bb341c4-7dcf-415b-8e5d-d38612742500.png)
![image](https://user-images.githubusercontent.com/112671903/222960684-034ab4cb-a1fa-4d39-86bb-9d84fd0fc1a4.png)
![image](https://user-images.githubusercontent.com/112671903/222961016-31321864-3af4-427e-9183-c672761ec095.png)
![image](https://user-images.githubusercontent.com/112671903/222961086-d2703737-4240-4179-a018-c26e0e8309c9.png)
![image](https://user-images.githubusercontent.com/112671903/222961246-251778b3-1ceb-4a8f-8b55-2534a557bdb9.png)

**Additional context**

[config-chinese.txt](https://github.com/doxygen/doxygen/files/10891801/config-chinese.txt)
[configgen.txt](https://github.com/doxygen/doxygen/files/10891809/configgen.txt)

albert-github:
[...]

PR #9917, opened 2023-03-15
Make reading configuration utf-8 compatible

albert-github:
Based on some comments in #9894 make the possibility to read real utf-8 files hand write it properly to the generated files

### DIFF
--- a/src/configgen.py
+++ b/src/configgen.py
@@ -17,6 +17,16 @@ import sys
 import re
 import textwrap
 from xml.dom import minidom, Node
+import io
+
+def xprint(str):
+    if sys.version_info[0] == 2:
+        str1 = unicode(str)
+        sys.stdout.write(str1)
+        sys.stdout.write(u"\n")
+    else:
+        print(str)
+
 
 def transformDocs(doc):
     # join lines, unless it is an empty line
@@ -117,7 +127,7 @@ def addValues(var, node):
         if (n.nodeName == "value"):
             if n.nodeType == Node.ELEMENT_NODE:
                 name = n.getAttribute('name')
-                print("  %s->addValue(\"%s\");" % (var, name))
+                xprint("  %s->addValue(\"%s\");" % (var, name))
 
 
 def parseHeader(node,objName):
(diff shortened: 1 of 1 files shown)


src/configgen.py
```diff
@@ -17,6 +17,16 @@
 import re
 import textwrap
 from xml.dom import minidom, Node
+import io
+
+def xprint(str):
+    if sys.version_info[0] == 2:
+        str1 = unicode(str)
```
doxygen, line 24: I think this will create just an array of code-point mulitbyte characters, not UTF-8. I think you need to add `.encode('utf-8')` See also https://stackoverflow.com/a/10288345/784672

albert-github: This did work and I didn't see problems, I'll gave the `.encode('utf-8')` a try (again) and it gives the error:
```
[ 15%] Generating ../generated_src/configvalues.h
Traceback (most recent call last):
  File ".../doxygen/src/configgen.py", line 974, in <module>
    main()
  File ".../doxygen/src/configgen.py", line 788, in main
    xprint("/* WARNING: This file is generated!")
  File ".../doxygen/src/configgen.py", line 25, in xprint
    sys.stdout.write(str1)
TypeError: write() argument 1 must be unicode, not str

```

src/configgen.py
```diff
@@ -717,26 +733,32 @@ def parseFooterDoc(node):
...
 
 
 def main():
     if len(sys.argv)<3 or (not sys.argv[1] in ['-doc','-cpp','-wiz','-maph','-maps']):
         sys.exit('Usage: %s -doc|-cpp|-wiz|-maph|-maps config.xml' % sys.argv[0])
     try:
-        doc = xml.dom.minidom.parse(sys.argv[2])
+        if sys.version_info[0] == 2:
+            fh = open(sys.argv[2],'r')
+            sys.stdout = io.open(sys.stdout.fileno(), 'w')
```
doxygen, line 745: I think also python 2.7 supports encoding when using `io.open`, i.e. this seems to work fine:

```python
import io
import sys
fh = io.open('file.txt','r',encoding='utf-8')
sys.stdout = io.open(sys.stdout.fileno(), 'w', encoding='utf8')
sys.stdout.write(fh.read())
```

albert-github: This looks like to work for Python 2 and Python 3, do you want me to patch this or will you do it?

albert-github: The first test was with the default config.xml, I now ttried with the Chinese version and this gives problems:
```
[ 15%] Generating ../generated_src/configoptions.cpp

'ascii' codec can't encode characters in position 8373-8391: ordinal not in range(128)
```


So I think my original patch is OK.
```

doxygen:
@albert-github Ok, I still changed things a bit so that the normal `print()` can be used, see https://github.com/doxygen/doxygen/commit/5b3cd1de01880b1bc28c284d9f13135cbb14dede

albert-github:
Just checked it looks like it is functioning.

PR #9919, opened 2023-03-15
Limit index fields in hhk file of chm file.

albert-github:
Based on some comments in #9894 where a warning was shown like
```
Warning: Keyword string:
...
is too long.  The maximum size is 488 characters.
```
tit was found that this was due to an extremely long `value` field in the `index.hhk` (index file), limiting this field and placing an ellipse when necessary

### DIFF
--- a/src/htmlhelp.cpp
+++ b/src/htmlhelp.cpp
@@ -212,6 +212,14 @@ static QCString field2URL(const IndexField *f,bool checkReversed)
  */
 void HtmlHelpIndex::writeFields(std::ostream &t)
 {
+  /* to prevent 
+   *  Warning: Keyword string:
+   *    ...
+   *  is too long.  The maximum size is 488 characters.
+   */
+  size_t maxLen = 400;
+  QCString contStr = "...";
+
   std::sort(std::begin(m_map),
             std::end(m_map),
             [](const auto &e1,const auto &e2) { return e1->name < e2->name; }
@@ -265,7 +273,7 @@ void HtmlHelpIndex::writeFields(std::ostream &t)
         t << "  <LI><OBJECT type=\"text/sitemap\">";
         t << "<param name=\"Local\" value=\"" << field2URL(f.get(),FALSE);
         t << "\">";
-        t << "<param name=\"Name\" value=\"" << convertToHtml(m_recoder.recode(level1),TRUE) << "\">"
+        t << "<param name=\"Name\" value=\"" << convertToHtml(m_recoder.recode(level1),TRUE,maxLen,contStr) << "\">"
            "</OBJECT>\n";
       }
       else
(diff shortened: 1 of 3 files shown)


doxygen:
@albert-github Kept the changes a bit more local, see https://github.com/doxygen/doxygen/commit/b8db19dab69d66755ecb48dca7278580bfdd459d

albert-github:
@doxygen,

I've been thinking about this as well but it might lead to problems in case `s.left(400)` contains characters in the form that run into the code that transform it into `&lt;`, `&gt`m `&amp;` etc, so that the `s.left(400)` might leave halve a code.

albert-github:
Previous comment has been reformulated.

doxygen:
@albert-github Good point. I made this fix https://github.com/doxygen/doxygen/commit/0b5c6f68d297a06aa666dcbde83c656cffebb230, which assumes this is a very rare case so it does not have to be very efficient.

albert-github:
@doxygen I think this is better not it is not good.
Say that the length of `s` is 400 or less then we will directly run `return convertToHtml(s,true);`, but the `convertToHtml` can result in a string that is larger than 488 characters, this will be a rare case but still it can happen.

doxygen:
@albert-github correct, so when it now happens, the `do..while()` loop will start reducing the input size with 10 characters at a time, until the result after conversion is less than 450 characters.

albert-github:
@doxygen, I don't think your remark "correct" is correct as we will never get into the `do ... while()` loop but directly jump to the else part.

doxygen:
@albert-github Ah now I see what you mean. Indeed, that's not ok. We should also use the length after conversion. Second try https://github.com/doxygen/doxygen/commit/f482317986aa9e8b6cb062085682d124e1b5630d.

albert-github:
@doxygen I think this will work. For a small efficiency improvement I think it would even be better to change the lines:
```
      result = convertToHtml(s.left(maxLen));
      maxLen-=20;
```
into
```
      maxLen-=20;
      result = convertToHtml(s.left(maxLen));
```
as otherwise the initial conversation will be done twice in case of an initial conversation that is to long or the `while loop` should have the condition at the end (so a `do ... while()` loop.

doxygen:
@albert-github Doing the `maxLen-=20` earlier would be the same as starting with `maxLen=380`. The idea is that by truncating with `maxLen` and comparing the resulting length after conversion against `maxExpandedLen` that after the first iteration of the while loop the result is very likely already below `maxExandedLen` (so the expansion adds less than 50 extra characters). Only if this is not the case, the gap between `maxLen` and `maxExpandedLen` is increased in steps of 20 bytes until the expansion no longer exceeds the gap. The initial gap should be such that it is likely already enough the first time. What the right value is to cover say >95% of the cases, I don't know, so I assumed 50.

albert-github:
I see indeed the first conversion takes the full string and on failure the first 400 bytes are taken (back in my head was, incorrectly, that the first conversion was already with the first 400 bytes).

The limit of 400 characters was an arbitrary number (its should be a number less than 488), so even the 450 could be expanded to a larger number and the 400 could also at a higher number but I think this is not worth the effort as
- the estimate of 95% is probably even an underestimation I think that it is more like 99%
- when having such a long string probably it is a generated string and nobody will read / study it 

so the current choices are OK for me.

PR #9929, opened 2023-03-19
Unknown generated markdown id

albert-github:
When having an example like:
```
Pipenv: Python Development Workflow for Humans
==============================================

- @ref "pipenv-python-development-workflow-for-humans" "Pipenv"
```
and the settings:
```
QUIET = YES
MARKDOWN_ID_STYLE = GITHUB
```
we get a warning like:
```
README.md:4: warning: unable to resolve reference to 'pipenv-python-development-workflow-for-humans' for \ref command
```
this is due to the fact that a markdown `===` type of header doesn't get an automatic header when promoted to a page

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/11010927/example.tar.gz)

(Found by Fossies for the pipenv package)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -3365,6 +3365,11 @@ void MarkdownOutlineParser::parseInput(const QCString &fileName,
         {
           docs.prepend("@ianchor " +  generatedId + "\\ilinebr ");
         }
+        else
+        {
+          QCString autoId = AnchorGenerator::instance().generate(title.str());
+          docs.prepend("@ianchor " +  autoId + "\\ilinebr ");
+        }
         docs.prepend("@page "+id+" "+title+"\\ilinebr ");
       }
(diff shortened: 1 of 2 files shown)


src/markdown.cpp
```diff
@@ -3365,6 +3365,11 @@ void MarkdownOutlineParser::parseInput(const QCString &fileName,
         {
           docs.prepend("@ianchor " +  generatedId + "\\ilinebr ");
         }
+        else
+        {
+          QCString autoId = AnchorGenerator::instance().generate(title.str());
+          docs.prepend("@ianchor " +  autoId + "\\ilinebr ");
```
doxygen, line 3371: this misses the `{...}` title argument, and then will not be matched in commentscan.l line 1341.

albert-github:
I had some problems with the repository due to the merge of #9930 so I created a new proposed pull request #9937

Issue #842, opened 2018-07-01
Heirarchy of subgroups in LaTex, RTF is wrong (Origin: bugzilla #123963)


doxygen:
status NEW severity *major* in component *general* for *---*
Reported in version *1.5.3-SVN* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[hide-indices-small-toc-1.4.3.patch](https://sourceforge.net/projects/doxygen/files/attachments/bug123963_49223_hide-indices-small-toc-1.4.3.patch/download/download)_ (ID 49223)

On 2003-10-06 16:39:44 +0000, Nara Won wrote:

    I made subgroups and HTML output made nice tree for the heirarchy of 
    subgroups.
    
    But LaTex and RTF output have wrong heirarchy of subgroups or flattened 
    the subgroups so the outputs have wrong chapters.

On 2005-07-15 11:37:15 +0000, Michael McTernan wrote:

    Created attachment 49223
    Patch to prevent flattened index in LaTeX contents

On 2005-07-15 11:41:05 +0000, Michael McTernan wrote:

    I found something that sounds to be the same and posted the following with a
    patch to doxygen-develop on 7/7/2005.  I've also added the patch to this
    bug.
    
    Original message to doxygen-develop:
    
    I've been generating some beautiful LaTeX documentation but have found that
    the
    table of contents that LaTeX adds is sometimes a little big and flat!  
    
    For example, if LATEX_HIDE_INDICES=NO, the generated table of contents
    duplicates the section indices, and additionally flattens out module
    documentation.  The following is test input that illustrates this:
    
    /** \defgroup Group0 This is group 0.
     @{ */
    /** \defgroup Group1 This is group 1.
    @{ */
    /** \defgroup Group2 This is group 2.
    @{ */
    /** \defgroup Group3 This is group 3.
    @{ */
    
    int main();
    
    /** @} */
    /** @} */
    /** @} */
    /** @} */
    
    With this, the modules index (page 5 if this is the only input file) looks
    like
    the following:
    
    Chapter 1
    Module Index
    1.1 Modules
    Here is a list of all modules:
    This is group 0. . . . . . . . . . . . . . . . . . . 3
       This is group 1 . . . . . . . . . . . . . . . . . 4
          This is group 2. . . . . . . . . . . . . . . . 5
             This is group 3 . . . . . . . . . . . . . . 6
    
    Perfect!  But the main index (page 3) appears as the following:
    
    Contents
[...]

PR #9936, opened 2023-03-20
Support for basic hierarchical grouping in LaTeX output

mtc-mlx:
This addresses the LaTeX part of #842.

What is missing:
- ~RTF~
- ~Doing the same thing for pages and subpages~
- Doing the same thing for classes/namespaces. (Are these normally organized hierarchically? I haven't used Doxygen for C++, only for C.)
- More thorough testing.

**Note: the LaTeX TOC will still be controlled by `tocdepth`.** Maybe the default should be increased? However, the PDF bookmarks will contain all the links regardless of `tocdepth`:

image


mtc-mlx:
RTF and DocBook should now be fixed. All tests pass locally (except the ones that try to run latex/bibtex, which I don't have installed inside of Msys2). But in my normal usage, the LaTeX output compiles fine.

Unfortunately I do not have the time right now to also apply this to classes/namespaces. These will remain flat for now. Hopefully that is OK...

mtc-mlx:
Sorry for the retries; everything should be OK now. I overlooked the need for `etocsetlevel` since we don't use `etoc` in our custom stylesheet.

Due to the way the page headers are generated now, some of the `\addindex` commands in the Doxygen user manual should no longer be required (but they are harmless).

The first chapter of the Doxygen user manual is now called "Doxygen" instead of "Introduction"—again, due to the way page headers are generated. This could be overridden in the `\mainpage` command if needed, but perhaps it's an improvement? An introduction to the introduction didn't make much sense...

**Before:**
image

**After:**
image

albert-github:
Small remark:
The double "Introduction" does indeed not make a lot of sense but the change of the first "Introduction" into "Doxygen" make, in my opinion, even less sense.

It is easy to remove the second "Introduction" in the LaTeX output by moving the `<h2>Introduction</h2>` into the `\htmlonly` part but that would make an inconsistent look as this would remove the sub-header "Introduction" but later on there would be other sub-headers like "Doxygen license" and "User examples".
Maybe the sub-header should be renamed to e.g. "General introduction" (@doxygen what is your opinion?)

mtc-mlx:
Another option would be to check `mainPageHasOwnTitle()`  specifically for LaTeX, RTF, and DocBook output, and use `trMainPage` even if PROJECT_NAME is defined.

Although "Main Page" doesn't really make sense in a PDF, so I guess we would need a new translator string for "Introduction".

albert-github:
Actually the doxygen manual is created with a dedicated startup file `doxygen_manual.tex` that is not generated but provided and here the first "Introduction" is seti (, so for me it is strange that you get here "Doxygen".
How did you generate the manual? The normal way is by means of `make docs`.

I don't think it is is a good idea to make for  LaTeX, RTF, and DocBook output a specific `mainPageHasOwnTitle()`.

mtc-mlx:
I changed PageDefImpl::writeDocumentation to make it consistent with module/group pages, where the title is written inside the file page file instead of in the parent. So I removed the title override in `doxygen_manual.tex` in 4fc0d89. I guess I should have split that into a separate commit…

mtc-mlx:
I split the HTML and LaTeX generating steps into two, so we can have a version with `\mainpage Doxygen` and a version with `\mainpage Introduction`. This way at least the result is the same as before.

Does this seem like a suitable approach?

albert-github:
This doesn't look like a suitable approach to me, according to my knowledge the first word  "Introduction" in the manual comes from, the chapter title as specified in the dedicated startup file `doxygen_manual.tex`.
- What happens when you change in your setup (without your last changes) in `doxygen_manual.tex` the word "Introduction" into something else?

mtc-mlx:
> the first word "Introduction" in the manual comes from, the chapter title as specified in the dedicated startup file doxygen_manual.tex.

This is not true anymore in this PR, as I mentioned on 29 March. The `\chapter{<Page Title>}` is written to the same file as the page contents, for consistency with module pages. Therefore the "hack" of using a different page title in `doxygen_manual.tex` will no longer work; it would result in duplicate headings.

> What happens when you change in your setup (without your 2 last changes) in `doxygen_manual.tex` the word "Introduction" into something else?

I removed this completely from doxygen_manual.tex as I explained. See https://github.com/doxygen/doxygen/pull/9936/files#diff-dddc37ac29fbb733a31120d80d4227bd676628e627eb06ac068e7a26b209ccb3L142

mtc-mlx:
@doxygen Please let me know if there is anything I can do to help with this.

mtc-mlx:
Thanks! I appreciate all the work you guys do on Doxygen!

albert-github:
- I just generated the doxygen documentation.
  In the LaTeX / PDF part 3 new warnings:
  ```
  Package etoc Warning: unexpected value `7' in \etocsetlevel.
  Should be -2,-1, 0, 1, 2, 3, 4, 5, or 6. Set to 6 (=ignored) on input line 128.

  Package etoc Warning: unexpected value `8' in \etocsetlevel.
  Should be -2,-1, 0, 1, 2, 3, 4, 5, or 6. Set to 6 (=ignored) on input line 129.

  Package etoc Warning: unexpected value `9' in \etocsetlevel.
  Should be -2,-1, 0, 1, 2, 3, 4, 5, or 6. Set to 6 (=ignored) on input line 130.
  ```
  (line numbers refer to `doxygen_manual.tex`)

  These warnings should not be present (would it be possible to set them to 6?)

- In the "Contents" at the totp of the PDF document we see in the new version:
  image

  whilst in the old version we see:
  image

  so the following items are gone:
  - Putting documentation after members
  - Examples
  - Documentation at other places

  This also is the case for the table of contents in the chapter (here Chapter 4).
  - why does this happen here for the non numbered parts (also in e.g. the  Searching chapter, but not in the appendices)
  - does it have to do something with the maximum depth supported?

- There is also the minor cosmetic issue of the `Introduction` `Interoduction` as mentioned in https://github.com/doxygen/doxygen/pull/9936#issuecomment-1486448686

mtc-mlx:
Oh, this isn’t good. Let me try to fix this.

mtc-mlx:
I see the issue now: etoc doesn't accept levels higher than 6, but named paragraphs and subparagraphs like these now have levels 8 and 9, respectively, so etoc assigns them level six → hidden. :(

mtc-mlx:
@albert-github I edited `etoc.sty` to allow TOC levels up to 9 and raised the `tocdepth` counter to 8; this resolved the issue. Which would you prefer:

1. Try to upstream the patch to etoc (perhaps unlikely to be accepted), or
2. Add the necessary overrides to `header.tex` and `doxygen_manual.tex` and don't try to upstream the change to etoc

mtc-mlx:
1. I'll try to implement it as a disabled-by-default option.
2. The trouble is paragraph and subparagraph would be hierarchically above `subsubsubsection` and `subsusbsubsubsection`. So there's not an easy way to deal with this without redefining a few macros from etoc.

mtc-mlx:
I will send this etoc patch once I can download an email client that doesn't word-wrap at 80 characters...

~~~diff
--- etoc.dtx	2023-04-24 12:27:52.942064100 -0400
+++ etoc.dtx	2023-04-24 18:10:13.405620600 -0400
@@ -3052,8 +3052,8 @@
 been done by the package itself in its definition of its own line styles).
  
 The accepted levels run from |-2| to |6| inclusive.  Anything else is
-mapped to |6|, which is a dummy level, never displayed.  The package
-does:
+mapped to |6|, which is a dummy level, never displayed.  This can be increased
+to 12 by passing the deeplevels option.  The package does:
 \begin{verbatim}
 \etocsetlevel{book}{-2}
 \etocsetlevel{part}{-1}
@@ -5144,7 +5144,11 @@
 \newcounter{visibletoc}
 \renewcommand{\etocaftertitlehook}
    {\stepcounter{visibletoc}\etoctoccontentsline{visibletoc}{\thevisibletoc}}
-\etocsetlevel{visibletoc}{6}
+\ifEtoc@deeplevels
+  \etocsetlevel{visibletoc}{12}
+\else
+  \etocsetlevel{visibletoc}{6}
+\fi
 \begin{document}
    < document body >
 \subsection{Surprising uses of etoc}
@@ -8040,6 +8044,7 @@
 \DeclareBoolOption[false]{locallottotoc}
 }
 \DeclareBoolOption[true]{ouroboros}
+\DeclareBoolOption[false]{deeplevels}
 \DeclareDefaultOption{\PackageWarning{etoc}{Option `\CurrentOption' is unknown.}}
 \ProcessKeyvalOptions*
 \DisableKeyvalOption[action=error,package=etoc]{etoc}{lof}
@@ -8147,10 +8152,24 @@
 \expandafter\chardef\csname Etoc@4@@\endcsname  4
 \expandafter\chardef\csname Etoc@5@@\endcsname  5
 \expandafter\chardef\csname Etoc@6@@\endcsname  6
-\expandafter\let\expandafter\Etoc@@six@@\csname Etoc@6@@\endcsname
+\ifEtoc@deeplevels
+  \expandafter\chardef\csname Etoc@7@@\endcsname  7
+  \expandafter\chardef\csname Etoc@8@@\endcsname  8
+  \expandafter\chardef\csname Etoc@9@@\endcsname  9
+  \expandafter\chardef\csname Etoc@10@@\endcsname  10
+  \expandafter\chardef\csname Etoc@11@@\endcsname  11
+  \expandafter\chardef\csname Etoc@12@@\endcsname  12
+  \expandafter\let\expandafter\Etoc@@maxdepth@@\csname Etoc@12@@\endcsname
+\else
+  \expandafter\let\expandafter\Etoc@@maxdepth@@\csname Etoc@6@@\endcsname
+\fi
 \@ifclassloaded{memoir}{\def\Etoc@minf{-\thr@@}}{\def\Etoc@minf{-\tw@}}
 \let\Etoc@none@@  \Etoc@minf
-\expandafter\let\expandafter\Etoc@all@@ \csname Etoc@5@@\endcsname
+\ifEtoc@deeplevels
+  \expandafter\let\expandafter\Etoc@all@@ \csname Etoc@11@@\endcsname
+\else
+  \expandafter\let\expandafter\Etoc@all@@ \csname Etoc@5@@\endcsname
+\fi
 \let\Etoc@dolevels\@empty
 \def\Etoc@newlevel #1{\expandafter\def\expandafter\Etoc@dolevels\expandafter
         {\Etoc@dolevels\Etoc@do{#1}}}
@@ -8181,8 +8200,8 @@
 \def\etocsetlevel#1#2{\expanded{\noexpand\etoc@setlevel{#1}{#2}}}%
 \def\etoc@setlevel#1#2{%
    \edef\Etoc@tmp{\the\numexpr#2}%
-   \if1\ifnum\Etoc@tmp>\Etoc@@six@@0\fi\ifnum\Etoc@tmp<-\tw@0\fi1%
-      \in@{.#1,}{.none,.all,.figure,.table,.-3,.-2,.-1,.0,.1,.2,.3,.4,.5,.6,}%
+   \if1\ifnum\Etoc@tmp>\Etoc@@maxdepth@@0\fi\ifnum\Etoc@tmp<-\tw@0\fi1%
+      \in@{.#1,}{.none,.all,.figure,.table,.-3,.-2,.-1,.0,.1,.2,.3,.4,.5,.6,.7,.8,.9,.10,.11,.12,}%
 %    \end{macrocode}
 % Letter or other agnostic test.  First time I ever use \csa{@car} as I never
 % paid attention to its existence and in more macro-coding intensive context
@@ -8229,13 +8248,19 @@
       \PackageWarning{etoc}
          {Argument `\detokenize{#2}' of \string\etocsetlevel\space should
           represent\MessageBreak
-          -2, -1, 0, 1, 2, 3, 4, 5, or 6 but evaluates to \Etoc@tmp.\MessageBreak
-          The level of `#1' will be set to 6.\MessageBreak
+          -2, -1, 0, 1, 2, 3, 4, 5,
+            \ifEtoc@deeplevels
+            6, 7, 8, 9, 10, 11, or 12
+            \else
+            or 6
+            \fi
+          but evaluates to \Etoc@tmp.\MessageBreak
+          The level of `#1' will be set to \Etoc@@maxdepth@@.\MessageBreak
           Tables of contents will ignore `#1' as long\MessageBreak
-          as its level is 6.\MessageBreak
+          as its level is \Etoc@@maxdepth@@.\MessageBreak
           Reported}%
       \etocifunknownlevelTF{#1}{\Etoc@newlevel{#1}}{}%
-      \expandafter\let\csname Etoc@#1@@\endcsname\Etoc@@six@@
+      \expandafter\let\csname Etoc@#1@@\endcsname\Etoc@@maxdepth@@
    \fi
 }
 %    \end{macrocode}
@@ -8271,8 +8296,8 @@
 % We do not do a formal \csbc{etocsetlevel}|{figure}{6}| and anyhow the user
 % level interface forbids |figure| and |table| as first argument.
 %    \begin{macrocode}
-\let\Etoc@figure@@\Etoc@@six@@
-\let\Etoc@table@@\Etoc@@six@@
+\let\Etoc@figure@@\Etoc@@maxdepth@@
+\let\Etoc@table@@\Etoc@@maxdepth@@
 %    \end{macrocode}
 % \etocrelease{1.09g} adapts to \ctanpkg{hyperref} depending on whether the
 % latter is at |7.00u| or earlier.  Indeed internal changes to
@@ -8426,7 +8451,7 @@
 %    \begin{macrocode}
 \def\Etoc@etoccontentsline #1{%
   \let\Etoc@next\@gobblefour
-  \ifnum\csname Etoc@#1@@\endcsname=\Etoc@@six@@
+  \ifnum\csname Etoc@#1@@\endcsname=\Etoc@@maxdepth@@
   \else
    \Etoc@skipthisonefalse
 %    \end{macrocode}
@@ -9130,7 +9155,7 @@
 %    \begin{macrocode}
           \def\Etoc@do##1{%
               \ifnum\etoclevel{##1}>\etoclocaltop
-                     \expandafter\let\csname Etoc@##1@@\endcsname\Etoc@@six@@
+                     \expandafter\let\csname Etoc@##1@@\endcsname\Etoc@@maxdepth@@
               \fi}%
           \Etoc@dolevels
 }
@@ -9236,7 +9261,7 @@
       \def\Etoc@@startlocaltochook{\Etoc@@startlocallistof@setlevels{figure}}%
       \def\Etoc@listofhook{%
         \def\Etoc@do####1{%
-          \expandafter\let\csname Etoc@@####1@@\endcsname\Etoc@@six@@
+          \expandafter\let\csname Etoc@@####1@@\endcsname\Etoc@@maxdepth@@
         }%
         \Etoc@dolevels
       }%
@@ -9262,7 +9287,7 @@
       \def\Etoc@@startlocaltochook{\Etoc@@startlocallistof@setlevels{table}}%
       \def\Etoc@listofhook{%
         \def\Etoc@do####1{%
-          \expandafter\let\csname Etoc@@####1@@\endcsname\Etoc@@six@@
+          \expandafter\let\csname Etoc@@####1@@\endcsname\Etoc@@maxdepth@@
         }%
         \Etoc@dolevels
       }%
@@ -9341,7 +9366,7 @@
 \def\Etoc@testingcontentslinelocal #1{%
   \ifEtoc@stoptoc
   \else
-   \ifnum\csname Etoc@#1@@\endcsname=\Etoc@@six@@
+   \ifnum\csname Etoc@#1@@\endcsname=\Etoc@@maxdepth@@
    \else
     \global\expandafter\let\expandafter\Etoc@level\csname Etoc@#1@@\endcsname
     \if @\@car#1\@nil\else\global\let\Etoc@virtualtop\Etoc@level\fi
@@ -9383,7 +9408,7 @@
 \def\Etoc@testingcontentsline #1{%
   \ifEtoc@stoptoc
   \else
-   \ifnum\csname Etoc@#1@@\endcsname=\Etoc@@six@@
+   \ifnum\csname Etoc@#1@@\endcsname=\Etoc@@maxdepth@@
    \else
     \unless\ifnum\csname Etoc@#1@@\endcsname>\c@tocdepth
      \global\Etoc@isemptytocfalse
@@ -9878,7 +9903,7 @@
 }
 \def\Etoc@setstyle@a #1{%
     \edef\Etoc@tmp{\the\numexpr\csname Etoc@#1@@\endcsname}%
-    \if1\unless\ifnum\Etoc@tmp<\Etoc@@six@@ 0\fi\unless\ifnum\Etoc@tmp>\Etoc@minf 0\fi1%
+    \if1\unless\ifnum\Etoc@tmp<\Etoc@@maxdepth@@ 0\fi\unless\ifnum\Etoc@tmp>\Etoc@minf 0\fi1%
       \Etoc@standardlinesfalse
       \expandafter\Etoc@setstyle@b\expandafter\Etoc@tmp
     \else
@@ -11260,7 +11285,7 @@
 % consistency with internal macros, but too late it is already in user |.toc| files.
 %    \begin{macrocode}
 \def\etoc@setlocaltop #1{%
-  \ifnum#1=\Etoc@@six@@
+  \ifnum#1=\Etoc@@maxdepth@@
    \Etoc@skipthisonetrue
   \else
    \Etoc@skipthisonefalse
~~~

mtc-mlx:
It is sent now. Including the patch inline makes it easier for the recipient to comment and make suggestions.

jfbu:
Hi @mtc-mlx the patch has been received.  ok for upstream merge (after some adaptations).

doc/doxygen_manual.tex
```diff
@@ -123,6 +123,12 @@
 \tolerance=750
 \usepackage{etoc}
 \etocsettocstyle{\doxytocparskip}{\doxynormalparskip}
+\etocsetlevel{subsubsubsection}{4}
+\etocsetlevel{subsubsubsubsection}{5}
+\etocsetlevel{subsubsubsubsubsection}{6}
+\etocsetlevel{subsubsubsubsubsubsection}{7}
+\etocsetlevel{paragraph}{8}
+\etocsetlevel{subparagraph}{9}
```
jfbu, line 131: Actually this is not needed as etoc is left in "compatibility mode for line styles" i.e. there is no use of `\etocsetstyle`.  On the other hand, it does not hurt either.

jfbu: (and as you commented already, for this to work, upstream etoc currently at 1.2 needs an update)

templates/latex/header.tex
```diff
@@ -200,6 +200,12 @@
   % in page table of contents
   \usepackage{etoc}
   \etocsettocstyle{\doxytocparskip}{\doxynormalparskip}
+  \etocsetlevel{subsubsubsection}{4}
+  \etocsetlevel{subsubsubsubsection}{5}
+  \etocsetlevel{subsubsubsubsubsection}{6}
+  \etocsetlevel{subsubsubsubsubsubsection}{7}
+  \etocsetlevel{paragraph}{8}
+  \etocsetlevel{subparagraph}{9}
```
jfbu, line 208: same remark: these things would be needed only if one wanted to use etoc to modify the actual rendering inside the TOC, but currently Doxygen makes no use of `\etocsetstyle` so these  things are not strictly speaking needed.

jfbu:
Much to my embarrassment I initially posted here a wrong comment on etoc functioning.  I had to delete, so awful was my mistake! Sorry...

mtc-mlx:
Hi @jfbu, thank you for your quick response and cleaning up the patch. To respond to your comments via email:

- The reason I implemented it with a fixed number of levels instead of variable was mostly for simplicity and readability. Creating the levels dynamically would of course be possible, but probably not necessary.
- For the changelog message you mentioned via email, I don't really care—my patch is too trivial to qualify for copyright anyway 😉 . "see Doxygen PR#9936" would be OK with me.

@albert-github Is it OK to add an etoc version check and only pass the new `deeplevels` parameter if the installed version of etoc accepts it? This would mean accepting this minor regression until the update makes its way into the TeX distros.

jfbu:
@mtc-mlx @albert-github 

- it is now acted that next etoc, release `1.2a`, will have deeplevels option, based upon adapted patch gently provided by @mtc-mlx 

- although `1.2a` will be pushed to CTAN in a matter of days, for Doxygen to benefit from this it has to ship this etoc.sty, as some Linux distros lag years behind TeXLive

- problem: very old LaTeX installation will have problems with etoc `1.2a`
as since 1.2 release of 2023/03/01, etoc has been using `\expanded`,
which made it into pdflatex only in TeXLive 2019, so if Doxygen ships
the etoc 1.2a with deeplevels option, it adds this constraint.  Which perhaps
is already there from other LaTeX support files, I have not checked.

I think the way out of this is to provide some LaTeX code which will
boost older etoc to support the new `deeplevels` feature. The corresponding
hack will be of a slightly different nature than the original patch of @mtc-mlx 
which after some adaptations was merged into upstream etoc.

For TeXnical reasons etoc 1.2a implements "deeplevels" as a package
option which has to be given at package loading time.  And LaTeX
provides no (official) means to know the date string of a package prior to loading
it.  Although "deeplevels" is not implemented in older etoc, it will not
cause a pdf error but only a warning (i.e. not changing the pdflatex binary
return status) to load such old etoc with this option.

So the strategy should be to load etoc with the option, check after that
if it is supported, and if not hack the support in it a posteriori.

This hack may modify etoc in sub-optimal ways as long as it impacts
things which do not matter for Doxygen usage of etoc. 

I will do this, not today which is a bit hectic for me but probably over the week-end

jfbu:
@albert-github coming back as follow-up to your post where you mentioned LaTeX formats issues because this helped me realize that  I completely overlooked another thing which is that `etoc 1.2` executes `\NeedsTeXFormat{LaTeX2e}[2020/10/01]` (at that date the format of `.toc` files changed slightly and was unified to avoid having to test if `hyperref` is used or not).  This is more annoying than `\expanded` (which anyhow is used only in `\etocsetlevel` and in matters related the "local list of figures/tables" that Doxygen does not make use of, and the `\etocsetlevel` is the thing to redefine anyhow with old etoc, so we will have training for that).  

Thus,  with LaTeX formats earlier than `2020/10/01` we have anyhow to make do with old etoc.  I think we can make a redefinition of `\etocsetlevel` which will work with all versions of etoc, and then it looks unneeded for Doxygen to ship etoc.  Simply check after loading it if  it has implemented `deeplevels` (which will be the case with `1.2a, 2023/05/01` and then the required LaTeX format itself requires `\expanded` anyhow), and if it has not implemented `deeplevels` do some a posteriori hack of `\etocsetlevel` and/or internals and of course avoid `\expanded` usage in that step.

Although I do try my best, I have a tendancy to be very verbose in LaTeX discussions, so sorry for general noise.  I am confident making  a patch which works with all etoc versions is feasible but only needs a bit of care. Only thing I may need some reassuring is confirmation that Doxygen users do not want to fiddle with LaTeX things, so that if the hack into old etoc has some consequences to slightly bridle obscure aspects only needed by expert LaTeX users, then it is not a problem because all LaTeX stuff is from Doxygen itself not from end user right?

jfbu:
> Although I do try my best, I have a tendancy to be very verbose in LaTeX discussions, so sorry for general noise. I am confident making a patch which works with all etoc versions is feasible but only needs a bit of care.

Typical of the kind of noise I can generate.  (obviously the fever I got some days back and which lingered until today has indeed obscured parts of my brain as I have experimented in other areas over the week)  I had *completely forgotten* that `etoc 1.2` of two months ago radically change internal things and that this explains why the patch from @mtc-mlx had any chance to work at all to start with.  With earlier versions of `etoc` this would have been *much more* complex.  Sorry about that!  No, ok, there is only one way, I have to provide you with a special etoc which will clone the `1.2a` which originated in my interaction with @mtc-mlx but which will be modified to *not* make a requirement on the LaTeX format, and does *not* use `\expanded`... the latter is easy, the former a little less as I have to resurrect some old codebranch which tested how many arguments `\contentsline` was expected to have.  For easier follow-up maybe I should open a public repo, but I will push there only the `etoc.sty` not the `etoc.dtx`.  Then I can maintain a `doxy` branch specific for your needs.  Will keep you posted.

jfbu:
@albert-github last question: is Doxygen using LaTeX `hyperref` always? if yes this simplifies a bit the thing for me in tailoring an `etoc 1.2a-doxy`.

jfbu:
> > is doxygen using LaTeX `hyperref` always?
> 
> I don't get the question, what would be the alternative to `hyperref`?

oh simply not using it! Sorry for noob question here, some people do produce PDF without hyperlinks...

albert-github:
@jfbu Doxygen has the possibility to create also documents without hyperlinks (the setting `PDF_HYPERLINKS = NO`) but normally hyperlinks are used with doxygen.

jfbu:
> I think that the main idea is that the new etoc should support more than the levels (-2 ...6) it does now in a way so that it can be used with all LaTeX distributions (well the later might be a bit overdone, but maybe at least 10 years).

@albert-github I will when I will have a functional brain look at this; if it is not too unreasonable a constraint I may thus modify also the official CTAN upload so that I can revert the LaTeX format requirement.  Very old etoc used only pure Knuth TeX but at some point started using `\numexpr` or like things which I think started being a requifrement of LaTeX around 2010;  but the business with `\expanded` is not crucial, it is more the thing with the `.toc` format.

jfbu:
> .

I will check your codebase, what I need to know is whether LaTeX package `hyperref` is always loaded, this avoids `etoc` code to have to test for it (which is always annoying because one has to delay to the `\begin{document}` hooks to do such checks).

albert-github:
Yes the package hyperref is always loaded when `PDF_HYPERLINKS=YES`
```
    % Hyperlinks (required, but should be loaded last)
    \ifpdf
      \usepackage[pdftex,pagebackref=true]{hyperref}
    \else
      \ifxetex
        \usepackage[pagebackref=true]{hyperref}
      \else
        \usepackage[ps2pdf,pagebackref=true]{hyperref}
      \fi
    \fi
```
this code is inside the doxygen "preprocessing" `%%BEGIN PDF_HYPERLINKS` which brings me to the question:
- does `etoc` nowadays require that `hyperref` is loaded? or does `etoc` load it when it is not yet loaded?

We can easily put the `etoc` part inside a `%%BEGIN PDF_HYPERLINKS`` block when necessary.

albert-github:
@jfbu Forgot: the above comment code is from templates/latex/header.tex

jfbu:
> this code is inside the doxygen "preprocessing" `%%BEGIN PDF_HYPERLINKS` which brings me to the question:
> 
>     * does `etoc` nowadays require that `hyperref` is loaded? or does `etoc` load it when it is not yet loaded?

`etoc` works both with and without `hyperref`, it needs to do some internal adjustments according to the situation.  For this  it adds a code to the at begin document hook to update its `\ifEtoc@hyperref` conditional.

templates/latex/doxygen.sty
```diff
@@ -583,6 +583,25 @@
 % possibility to have sections etc. be within the margins
 % unfortunately had to copy part of book.cls and add \raggedright
 \makeatletter
+\newcounter{subsubsubsection}[subsubsection]
+\newcounter{subsubsubsubsection}[subsubsubsection]
+\newcounter{subsubsubsubsubsection}[subsubsubsubsection]
+\newcounter{subsubsubsubsubsubsection}[subsubsubsubsection]
```
jfbu, line 589: @mtc-mlx it looks as if a "sub" is missing for bracketed term

jfbu:
> I think that the main idea is that the new etoc should support more than the levels (-2 ...6) it does now in a way so that it can be used with all LaTeX distributions (well the later might be a bit overdone, but maybe at least 10 years).

@albert-github I have reverted for 1.2a the etoc requirement of LaTeX 2020-10-01 and the availability of `\expanded` so one will be able to use etoc 1.2a and its `deeplevels` option with old LaTeX.

albert-github:
@jfbu as soon if the code is available at CTAN (or in another way) I can start to test it.

jfbu:
@albert-github I plan on uploading to CTAN this evening (CEST) or tomorrow. However I am a bit cautious as I have been 30% functional since Monday (the upload incorporates a few pending items also on documentation side that were to be treated).  

But here is a gist with the a priori final etoc 1.2a so you can get it, 

[etoc 1.2a gist](https://gist.github.com/jfbu/935040579b481b2da162072f3c5251d0)

(you don't need documentation as the architecture is already all in place in this already merged PR, all you need to do is to ship etoc 1.2a with Doxygen and invoke it via `\usepackage[deeplevels]{etoc}` in your latex templates).  About this PR, I observed a [typo](https://github.com/doxygen/doxygen/pull/9936#pullrequestreview-1406444912) above.

jfbu:
sorry I had to scrap the gist as I inadvertendly lifted some privacy; will update

jfbu:
Here is [secret gist](https://gist.github.com/jfbu/40bc693be69dc22309c672df1b9119e7) well not so secret now that it is here, but it contains the `etoc.sty 1.2a` which probably is final or quasi final.

albert-github:
@jfbu Thanks I did some tests (doxygen documentation and doxygen tests) and didn't see any new problems.
To get it properly working with older versions of LaTeX the updated etoc package had to be provided as a local package and the load line had to be:
```
  \IfFormatAtLeastTF{2023/05/01}{\usepackage[deeplevels]{etoc}}{\usepackage[deeplevels]{etoc_doxygen}}
```
I created a proposed patch #10016 that I made Work In Progress so when the official `etoc.sty` is released and it contains differences to the current final / quasi final version these updates can be incorporated as well as eventual comments from your side.

albert-github:
@jfbu looks like there are some problems with older versions of LaTeX see the PR #10016, I'll have to investigate...

jfbu:
@albert-github (not yet checked the problems you mention right now) I updated the [not so secret gist](https://gist.github.com/jfbu/40bc693be69dc22309c672df1b9119e7) for these changes:
- rewrote manually the automatic commented-out header to match situation,
- used `etoc_doxygen` in the `\ProvidesPackage` line to avoid noise in LaTeX build logs,
- (as no other package does `\usepackage{etoc}` no potential issue there),
- some changes in the code I had done upstrem (mainly swapping two hunks in relation to how hyperlinks to the source code work from inside the etoc.pdf documentation)

jfbu:
@albert-github Is there any way to recover doxygen_manual.log file from (one of) the failed jobs?

albert-github:
I'm looking into that problem. Hopefully I can find something before the end of the day and I will post the results.

I've updated the etoc_doxygen.sty (thanks for the update and the fact that it now provides`etoc_doxygen`.

jfbu:
@albert-github not sure if this is relavant to oxygen_manual but perhaps remove `\batchmode`
https://github.com/doxygen/doxygen/blob/bc42dee774f000aeb3a47c4f3af0919dc4dd3ae4/templates/latex/header.tex#L3
or perhaps rather
https://github.com/doxygen/doxygen/blob/bc42dee774f000aeb3a47c4f3af0919dc4dd3ae4/doc/doxygen_manual.tex#L15
and any other relevant place from [those](https://github.com/search?q=repo%3Adoxygen%2Fdoxygen%20batchmode&type=code) so that even if we don't get log perhaps we can see console output.

I do have a TL2019 it is only other thing I have readily available but it will perhaps not match the Debian one.

albert-github:
- problem with removing the `\batchmode` is that the process will stop and wait for an answer. I'm not 100% sure whether with GitHub Actions the process will just stop or will hang.
- I expect that TL2019 version will give the same problems so it is definitely worth a try.

jfbu:
> * problem with removing the `\batchmode` is that the process will stop and wait for an answer. I'm not 100% sure whether with GitHub Actions the process will just stop or will hang.

ah yes, but there is ``--halt-on-error`` then.  I forgot to mention it.



> * I expect that TL2019 version will give the same problems so it is definitely worth a try.

I already did my local etoc tests on a (final) TL2019 and no issue emerged (on my macos).  I did not git clone doxygen and attempt to build its PDF documentation because I was afraid of collaterals about dependencies and necessity to build doxygen itself.  I have a MacPorts doxygen 1.9.3 installed but when I tried this January ``sudo port install doxygen +docs`` something went bezerk and at some point the system was asking me if I would have the kindness to allow ``Python3.7.exe``  to accept incoming network connexions, which was a bit strange as I am not on Windows.  Sure this is only issues of the MacPorts port, but this made me wary of attempting a homemade build.  You can easily spend hours once you get started on something like this only to conclude that your system is too old for some dependency at some point.

Starting around about May 15th, I will have access to a computer with all TeXLives from 2012 to 2019 so I can then more easily use etoc in old LaTeX.  But still, those will be only etoc tests and are very unlikely to fail (despite me saying the opposite in Info message to user on old LaTeX).

jfbu:
> > * problem with removing the `\batchmode` is that the process will stop and wait for an answer. I'm not 100% sure whether with GitHub Actions the process will just stop or will hang.
> 
> ah yes, but there is `--halt-on-error` then. I forgot to mention it.

hmm, unfortunately it looks as if `--halt-on-error` actually does not even print the error encountered to log.  My bad, sorry.

jfbu:
sorry about confused last comment but there is some bug in `pdftex` binary itself because in some circumstances it does not display the TeX error message causing the ``--halt-on-error`` to stop the build.  But in general it does something like this:

```
$ pdflatex --halt-on-error temp.tex
This is pdfTeX, Version 3.141592653-2.6-1.40.25 (TeX Live 2024) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
(./temp.tex
LaTeX2e <2022-11-01> patch level 1
L3 programming layer <2023-03-30>
(/usr/local/texlive/2023/texmf-dist/tex/latex/base/article.cls
Document Class: article 2022/07/02 v1.4n Standard LaTeX document class
(/usr/local/texlive/2023/texmf-dist/tex/latex/base/size10.clo)) (./etoc.sty
(/usr/local/texlive/2023/texmf-dist/tex/latex/kvoptions/kvoptions.sty
(/usr/local/texlive/2023/texmf-dist/tex/latex/graphics/keyval.sty)
(/usr/local/texlive/2023/texmf-dist/tex/generic/ltxcmds/ltxcmds.sty)
(/usr/local/texlive/2023/texmf-dist/tex/latex/kvsetkeys/kvsetkeys.sty))
(/usr/local/texlive/2023/texmf-dist/tex/latex/tools/multicol.sty))
(/usr/local/texlive/2023/texmf-dist/tex/latex/l3backend/l3backend-pdftex.def)
(./temp.aux)
! Undefined control sequence.
l.54 \error
           
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on temp.log.
```
Turns out I was initially testing with a `temp.tex` file interacting with user in some ways (via `\show`)  and there the `--halt-on-error` had unexpected behavior.

albert-github:
- Any suggestion how to run things is always welcome
- Pity that the TL2019 didn't show the results
- I just tried to `cat` the log file but this failed, have to do more tests ...

jfbu:
While I am here making your life miserable, let me add the following remark: I noted in
https://github.com/doxygen/doxygen/blob/bc42dee774f000aeb3a47c4f3af0919dc4dd3ae4/doc/Makefile.latex#L17-L23

only two passes.  I don't know to what extends local tables of contents, if any, are used there, but even if there are none and if the main table of contents is at start of document and is more than one page, then two latex passes may be not enough, one may need three:
- on first pass, latex creates external `.toc` file
- on second pass, the main TOC will be typeset; this may or may not depending on how the pages are numbered and whether the TOC is less that one page long, or less than two pages long if typeset with  a chapter heading in a document class with chapters, end up modifying page numbers.
- only on third pass will thus in certain circumstances be the page numbering for referecnes be correct.

(if the document uses `book` class, and chapters start only on odd pages, and TOC is typeset as a chapter, then one is relatively safe, because the TOC has to go beyond occupying 2 pages to shift page numbers; besides perhaps the page numbering is restarted anew after TOC so no problem then --- but if the document has many additional to the main TOC, "local" tables of contents as made possible by etoc, then it is really likely a third pass is needed).

if you add to this usage of many `\localtableofcontents` from etoc package, then it is the more likely you need three passes.  It may work with two passes but I would make three for being sure.  Sadly for reasons explained earlier I feel a bit unable to build the doxygen own doc myself locally so I can't confirm a third pass is really needed... it all depends on internals of cross-referencing in the document, whether it really uses many `\localtableofcontents` and whether those are sometimes with numerous items.

Actually I would use `latexmk -pdf` (or `latexmk -pdfxe` if using xelatex).  

Actually if I understand correctly, the production of pdf via pdflatex for user projects does use three passes generally https://github.com/doxygen/doxygen/blob/bc42dee774f000aeb3a47c4f3af0919dc4dd3ae4/src/latexgen.cpp#L401-L424
and even has some auto-detection of circumstances requiring more passes.  It would be more robust to use [latexmk](https://ctan.org/pkg/latexmk) tool here (which comes with all LaTeX distributions).

albert-github:
I succeeded in getting the information from the log file and it looks like, at least, the `\IfFormatAtLeastTF` was missing.

jfbu:
> * problem with removing the `\batchmode` is that the process will stop and wait for an answer. I'm not 100% sure whether with GitHub Actions the process will just stop or will hang.

I suggested ``--halt-on-error``  as option to pdflatex binary, but I could have proposed to replace `\batchmode` by `\nonstopmode`:

> With `\nonstopmode` TeX will scroll through errors and, in the case of the kind of error that cannot be recovered from, it will make an emergency stop, aborting the run.[...]
> The `\batchmode` differs only from `\nonstopmode` in that it gives messages only to the log file, not to the terminal.

jfbu:
> I succeeded in getting the information from the log file and it looks like, at least, the `\IfFormatAtLeastTF` was missing.
ah good news you got that info

that's logical!  but I thought you had already mentioned ``\@ifl@t@r\fmtversion{YYYY/MM/DD}{TRUE}{FALSE}``

albert-github:
Yes I did but this was in the "normal" part and it looks like we now need it in the `doxygen_manual` part as well.
I added the line but it looks like the `\frmversion` is now know:
```
! Undefined control sequence.
\IfFormatAtLeastTF ->\t@r 
                          \fmtversion 
l.134 \IfFormatAtLeastTF
                        {2023/05/01}{\usepackage[deeplevels]{etoc}}{\usepack...

```
Do you know where `\fmtversion` is defined / which package to include?

albert-github:
Regarding the comment https://github.com/doxygen/doxygen/pull/9936#issuecomment-1529029526 I think the remarks are correct, though the file `Makefile.latex` is, as far as I know, not used anymore, everything is now handled through `CMakeLists.txt` and here (quite recently, c5fedc957cd (albert-github      2023-04-25 16:20:24 +0200)) I updated the number of passes .

jfbu:
> Yes I did but this was in the "normal" part and it looks like we now need it in the `doxygen_manual` part as well. I added the line but it looks like the `\frmversion` is now know:
> 
> ```
> ! Undefined control sequence.
> \IfFormatAtLeastTF ->\t@r 
>                           \fmtversion 
> l.134 \IfFormatAtLeastTF
>                         {2023/05/01}{\usepackage[deeplevels]{etoc}}{\usepack...
> ```

Very strange.  The undefined one is not `\fmtversion` but `\t@r`, the last one indicated on the first line.
But you have `\@ifl@t@r` in your code, which is correct. How can `@ifl@` piece have disappeared and leave TeX with tokenizing `\t@r`????.  It is not a problem of catcode of `@` you have the correct `\makeatletter`.  As TeX would say, I'm stymied.  Does the doc/doxygen_manual.tex undergo some treatment by template engine?

> 
> Do you know where `\fmtversion` is defined / which package to include?

`\fmtversion` has always been defined by LaTeX as far as I know. (I will go check and will edit to complete the information).  The error message of TeX is not about`\fmtversion`.  It only shows `\fmtversion` as next in token stream.

albert-github:
I thought the `doc/doxygen_manual.tex` does not undergo any treatment but I see now it the `CMakeLists.txt`:
```
configure_file(${PROJECT_SOURCE_DIR}/doc/doxygen_manual.tex ${PROJECT_BINARY_DIR}/doc/doxygen_manual.tex)
```
it is just there to copy the file, but the `configure_file` does something more...

I've updated again.

jfbu:
> Regarding the comment [#9936 (comment)](https://github.com/doxygen/doxygen/pull/9936#issuecomment-1529029526) I think the remarks are correct, though the file `Makefile.latex` is, as far as I know, not used anymore, everything is now handled through `CMakeLists.txt` and here (quite recently, [c5fedc9](https://github.com/doxygen/doxygen/commit/c5fedc957cd7cf98434b5035d9eed13b3d2a794b) (albert-github 2023-04-25 16:20:24 +0200)) I updated the number of passes .

I have overlooked [doc/CMakeLists.txt](https://github.com/doxygen/doxygen/blob/master/doc/CMakeLists.txt) relevance.  Good thing that it does three passes always!

I see now your message.  Perhaps the configure was about strings marked for translations? (wild guess I am not familiar with your project at all!).

I will be bold and says that next failure will not be etoc related ;-)

albert-github:
Blame on me now I made a typo...

jfbu:
ah yes I see now from [official doc of CMake](https://cmake.org/cmake/help/latest/command/configure_file.html) that ``configure_file`` is quite invasive! The @VAR@ part is montruous for `.tex` and even worse `.sty` files...
```
Copies an <input> file to an <output> file and substitutes variable values
referenced as @VAR@ or ${VAR} in the input file content. Each variable
reference will be replaced with the current value of the variable, or the
empty string if the variable is not defined. Furthermore, input lines of the form
```

albert-github:
It looks like the incorrect runs went all through and that all the problems found were doxygen related.

Thank you very much for the help in debugging all the errors that were in the doxygen part.

jfbu:
great news!  The Ubuntu Debian uses the LaTeX kernel of 2020-02-02 so theoretically etoc has sent a message to the log to the effect that is is "old" LaTeX.  On the other hand, the `\expanded` was already available as it was added to PDFTeX engine at TL2019 release (somewhere in March 2019, so Debian has a more mature TL2019 from 2020). So here etoc is not tested in absence of `\expanded`, but I did it by lying to etoc about `\expanded` state and it was ok.

If you can get a hold of the pdf as artefact build it would good to check that hyperlinks from TOCs are correct.

jfbu:
This debugging episode and minutes occupied at staring at CIs logs makes me think perhaps I should let etoc issue at end of document a Warning when it detects that a new local TOC has been added that it is necessary to do one more LaTeX run... (as longtable does in some situations).

albert-github:
@jfbu I think that a warning like "Rerun to get etoc-references right" would be quite nice.

jfbu:
@albert-github the good thing about my limited knowledge of etoc which I have demonstrated these last 72hours  (I have some excuses as I had done a big etoc piece in February and was basically done for 2023 or at least until summer)  is that I can discover that I anticipated in the past your suggestion of the present: (from the comments which one finds in etoc.pdf in the source code)
```
% I hesitated with \etocrelease{1.08i} (edit: this is 2016/09/29) to write something to aux file in order to let
% \LaTeX\ prompt the user for extra pass, after insertion of some new
% \csa{localtableofcontents}, but finally I prefer to only trick \LaTeX\ into
% telling about undefined references.
% ....
% Method used here seems to work fine also with latexmk: it does not seem to induce it into
% making too many runs.
```
So there are already some situations where etoc will force a 
```
LaTeX Warning: Reference `Unknown toc ref foo.  Rerun LaTeX' on page 1 undefine
d on input line 29.
```
which will cause a
```
LaTeX Warning: There were undefined references.
```
But this is related to some advanced syntax of etoc which is not used by Doxygen.

And, in fact the key file to observe is the `.toc` file. Any change to it must trigger a new pass. When you manually work with a document gradually adding `\localtableofcontents` the next pass changes the`.toc` file and the second pass changes it only if page numbers go shifted from the `\localtableofcontents` having fetched many entries and being very long. And if this happens the `.toc` file changes.

As I systematically use `latexmk` in projects where automated builds are done, I trust it to detect such changes so I never worried too much about mentioning them to user.   But looking at the longtable warnings made me wonder if I should.

Finally, if you have a monolitichic document already completely done and start from build dir with no auxiliary files left-overs from earlier runs, 3 passes is the (minimal) magic number.  Which is however not always enough in LaTeX in general. I think that Sphinx project (they do not use  `etoc`) used a hardcoded number of 5 or 6 passes around 2016 prior to switching to [`latexmk`](https://ctan.org/pkg/latexmk).

For documents with `hyperref` usage, some extra "rerun" file checks are done by LaTeX:
```
Package rerunfilecheck Warning: File `temp.out' has changed.
(rerunfilecheck)                Rerun to get outlines right
(rerunfilecheck)                or use package `bookmark'.
```
See [`rereunfilecheck`](https://ctan.org/pkg/rerunfilecheck).  This adds some overhead but it is not noticeable.  Besides one does not have the choice as it is loaded by `hyperref`.  It probably is perhaps overkill when one is using `latexmk` which also I think does md5 checksums, not only of auxiliary files but also of *all* dependencies!  So with `latexmk` is you update your LaTeX installation and have some project where you kept the auxiliary files of LaTeX and `latexmk` from a run 6 months back, it will detect if it is needed to rerun.

jfbu:
@mtc-mlx @albert-github `etoc 1.2a` (identical to last iteration as incorporated in #10016) has been pushed to CTAN (not visible there until perhaps 24 or 48 hours).  Thanks for ping on this Doxygen context.

jfbu:
It will be announced this way (I adapted the thanks here to the @ forms)

- Drop the requirement, which was added at 1.1a, of a LaTeX kernel at
least as recent as 2020-10-01.

- Drop the requirement, which was added at 1.2, of availability of the
`\expanded` primitive.

So etoc 1.2a should work on old LaTeX installations; mind that it has
been only minimally retro-tested.

- "deeplevels" boolean option: adapted from a patch courtesy of 
@mtc-mlx  in the context of the Doxygen project (cf PR doxygen#9936).
Thanks!

With this option, available numerical levels range from -1 (usually
assigned to `part'; or -2 for memoir `book' sectioning) up to 12 rather
than up to 6.

Refer to PDF documentation for the 1.2a change log entry details.
Also, an important feature of the 1.2 **experimental** addition
`\locallistof{figures,tables}` has been documented in their devoted
PDF section.

albert-github:
- rerun OK so this should be triggered already. In "normal" doxygen runs there is already a rerun detection and there is a maximum of 8 reruns (to prevent endless reruns). In the documentation build  there are 3 runs, so we have to keep this potential problem in our minds or have a look at the possibility to use `make.bat` / `Makefile` here as well.
- I already see the new version on CTAN

albert-github:
@jfbu

The new etoc_doxygen.sty has been integrated into the master version of doxygen.
After the integration an external user runs always a spelling checker as well (as he uses the master version) and reported 2 spelling errors:
```
 etoc_doxygen.sty line 17:
   orginal -> original  (commented line (TeX))

 etoc_doxygen.sty line 1632:
   mimick -> mimic  (commented line (TeX))
```

I already created a proposed pull request #10034

jfbu:
Thanks for ping @albert-github ; duly incorporated upstream

albert-github:
@jfbu 
Around the lines 14660 of the file incorporated in doxygen we see:
```
\newcommand*\etoc@ruledtoci[2][\etocdefaultnbcol]{%
    \etocruledstyle[#1]{#2}%
^[[O    \tableofcontents*}
\newcommand*\etoc@local@ruledtoc[2][\etocdefaultnbcol]{%
    \etocruledstyle[#1]{#2}%
    \localtableofcontents}
```
I'm not sure this is also the case in the upstream version or that this has been incorporated during some edit on our side, we will correct it on our side, this notice is just for the upstream version.

jfbu:
@albert-github  Thanks for ping! It is not in original upstream file and I don't see it either in the gist from which etoc_doxygen.sty was provided in the first place (https://gist.github.com/jfbu/40bc693be69dc22309c672df1b9119e7#file-etoc_doxygen-sty-L1466).  I do not get it from a test with copy paste from this gist as visited via my browser, and neither from zip download...

albert-github:
@jfbu thank you for the lightning fast response, good to know that is not in the upstream version must have entered the doxygen code base by accident.

flagarde:
@albert-github Maybe some pre-commit have check for this stufs

PR #9937, opened 2023-03-21
Unknown generated markdown id

albert-github:
When having an example like:
```
Pipenv: Python Development Workflow for Humans
==============================================

- @ref "pipenv-python-development-workflow-for-humans" "Pipenv"
```

Based on review
- added optional text label to new `\ianchor`
- properly skipping `@ianchor `
- adjusting handling of page command in combination with `@ianchor` (in commentscan.l)

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -1338,15 +1338,16 @@ STopt  [^\n@\\]*
 
   /* --------- handle arguments of page/mainpage command ------------------- */
 
-<PageDocArg1>[^\n]*"\\ilinebr @ianchor"\{[^\]\n]*\}{B}{FILE} { // special case where the Markdown processor has rewritten
+<PageDocArg1>[^\n]*"\\ilinebr @ianchor/"\{[^\]\n]*\}{B}{FILE} { // special case where the Markdown processor has rewritten
                                                                // "@page label Title" as
                                                                // "@page md_label Title\ilinebr @ianchor{Title} label"
-                                          QCString text = yytext;
-                                          int start = text.find('{');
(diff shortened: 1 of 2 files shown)


src/commentscan.l
```diff
@@ -1338,15 +1338,16 @@ STopt  [^\n@\\]*
 
   /* --------- handle arguments of page/mainpage command ------------------- */
 
-<PageDocArg1>[^\n]*"\\ilinebr @ianchor"\{[^\]\n]*\}{B}{FILE} { // special case where the Markdown processor has rewritten
+<PageDocArg1>[^\n]*"\\ilinebr @ianchor/"\{[^\]\n]*\}{B}{FILE} { // special case where the Markdown processor has rewritten
```
doxygen, line 1341: I see a `/` appear in `@ianchor` so I do not expect this rule will ever match, probably breaking the fix for issue #9254. Did you test this?

albert-github: I tested it on the current problem.
I just now tested it on the problem from #9254 and see that there is a problem again  e.g. in pages.html

Correct after fix for 9254:

image


After my "fix":
image

So, unfortunately, I have to dive into it again ...

doxygen: @albert-github As I tried to explain the rule at line 1341 above needs to match the rewritten `@page` command in Markdown to fix issue #9254, and because if the literal `/` that is introduced in the rule it no longer matches...

You probably meant to write this instead (I moved the `/` outside the `""`):

```.l
<PageDocArg1>[^\n]*"\\ilinebr @ianchor"/\{[^\]\n]*\}{B}{FILE} {
```

PR #9938, opened 2023-03-21
bug 615500 Program Icon under Windows

albert-github:
Analogous to the doxywizard adding an icon to the the other executables. (Icon is less useful as the executables are command line executables, but in file viewers it has an advantage).

The chosen icon is the colored `d` from the doxygen color logo, which is converted to an icon.


doxygen:
@albert-github I've decided to use the official doxygen icon, which was already used for doxywizard. It is a multi resolution icon with alpha mask, so it looks better on different screens/views/backgrounds.

albert-github:
Wouldn't it be better to adjust the CMakeLists.txt of the wizard as well so we have everywhere the same code and we don't need the icons on different directories?

doxygen:
yes, I agree. I'll look into it.

PR #9948, opened 2023-03-24
Ignore c11 static assert

sebhub:


sebhub:
> Shouldn't the "_Static_assert" also be mentioned in the function `skipLanguageSpecificKeyword` in code.l (around line 3798)?

The test case did pass without adding `_Static_assert` to `skipLanguageSpecificKeyword`. I don't know the code good enough to judge if this is necessary. If I run the test case in GDB and set a break point to skipLanguageSpecificKeyword() it is not hit. What do I have to do so that this function is called?

albert-github:
In the `skipLanguageSpecificKeyword ` function we see:
```
...
    case SrcLangExt_Java:
      retval = (non_java_keywords.find(keyword) != non_java_keywords.end());
      break;
...
```

so it is hit in case we have java code.

sebhub:
Ok, but Java has no _Static_assert keyword. This keyword is specific to C11. In C23 it it was deprecated, however, Doxygen should still ignore it.

albert-github:
The variable is `non_java_keywords` so as java doesn't have this keyword I think it should be added in this list.

sebhub:
Ok, now I understand. I will try to improve the test case and update the pull request.

sebhub:
Thanks for pointing out the bug in my fix. I added a test case which shows that skipLanguageSpecificKeyword() definitely needs the suggested update.

Issue #9950, opened 2023-03-24
PlantUML is not fully supported in markdown fenced code blocks 


MrNice2001:
Markdown files can be included in Doxygen. Additionally, PlantUML can be used within a markdown file in fenced code blocks, e.g. 

```
```plantuml
@startuml
A --> B
@enduml
` ``
```
This works for _@startuml_. 

But for other PlantUML engines (e.g. _@startjson_ ) the graphs are not generated: 
```
```plantuml
@startjson
{
"Graph": "not visible"
}
@endjson
` ``
```

The configuration file contains ([Reference: Doxygen help](https://www.doxygen.nl/manual/commands.html#cmdstartuml)): 
```
ALIASES               += startjson=@startuml{json}
ALIASES               += endjson=@enduml  
```
but this works only for PlantUML which is not fenced in a code block: 
```
@startjson
{
"Graph": "not visible"
}
@endjson
```

I would expect that all [PlantUML engines](https://www.doxygen.nl/manual/commands.html#cmdstartuml) (bpm, wire, dot, ditaa, salt, math, latex, gantt, mindmap, wbs, yaml, creole, json, flow, board, git, hcl, regex and ebnf ) are supported within code blocks. 

**To Reproduce**
I have attached an example to reproduce the issue. It includes a Doxygen configuration file and a markdown file. The missing diagram can be obsereved in _html/md_main.html_. 

**Versions**
PlantUML v1.2023.4
```
java -Djava.awt.headless=true -jar /home/user/Downloads/plantuml-1.2023.4.jar -version

PlantUML version 1.2023.3 (Thu Mar 09 18:30:18 CET 2023)
(GPL source distribution)
Java Runtime: OpenJDK Runtime Environment
JVM: OpenJDK 64-Bit Server VM
Default Encoding: UTF-8
Language: en
Country: US
 
PLANTUML_LIMIT_SIZE: 4096

Dot version: dot - graphviz version 2.43.0 (0)
Installation seems OK. File generation OK
```

Doxygen v1.9.6
[markdown_plantuml_example.zip](https://github.com/doxygen/doxygen/files/11063188/markdown_plantuml_example.zip)

```
/bin/doxygen  --version

1.9.6 (4586b5cfaa3d46d51f6a51882951d15644c49edf)

```

Operation System
```
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.2 LTS
Release:        22.04
Codename:       jammy
```

**Additional context**
[...]

PR #9951, opened 2023-03-24
issue #9950 PlantUML is not fully supported in markdown fenced code blocks

albert-github:
Implemented the possibility to use the different engines in the fenced code blocks.


doxygen:
@albert-github I think the algorithm to insert the image name can be optimized, see https://github.com/doxygen/doxygen/commit/183479a59ad028c058cbb224579f2faff4f9dc09. Please check if it still does what you had in mind.

PR #9952, opened 2023-03-24
Use groups to associate documentation entries

sebhub:
This patch set enables the documentation of static functions, enums, enum fields, typedefs, defines, and variables though documentation entries in a separate file using group membership relations. This can be used to augment the Doxygen produced documentation with outputs from other tools (code metrics, test results, static analysis, design to requirements traceability). In a first Doxygen run only the tagfile is produced. Then the tagfile is used to associate tool generated results with documentation items and a Doxygen documentation file is generated. This file is added to a second Doxygen run which now contains the augmented documentation.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -7324,6 +7324,33 @@ static void addEnumDocs(const Entry *root,MemberDefMutable *md)
   }
 }
 
+//----------------------------------------------------------------------
+// Search for the name in the associated groups.  If a matching member
+// definition exists, then add the documentation to it and return TRUE,
+// otherwise FALSE.
+
+static bool tryAddEnumDocsToGroupMember(const Entry *root,const QCString &name)
+{
+  for (const auto &g : root->groups)
+  {
+    const GroupDef *gd = Doxygen::groupLinkedMap->find(g.groupname);
+    if (gd)
+    {
+      MemberList *ml = gd->getMemberList(MemberListType_decEnumMembers);
+      if (ml)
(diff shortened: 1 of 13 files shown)


doxygen:
@sebhub Can you add an example of how your changes are intended to be used? It is a bit hard to tell for me from just looking at the commit description and the code that was changed.

sebhub:
I tried to explain the use case in the pull request comment. We use Doxygen to pre-qualify the open-source real-time operating system RTEMS according to ECSS software development standards:

https://rtems-qual.io.esa.int/

These standards require a traceability from requirements to the software design and vice versa. The requirements are contained here for example:

https://git.rtems.org/rtems-central/tree/spec/rtems/task/if/construct.yml

In the first Doxygen run, we just create the tagfile. In the tagfile look up for example `rtems_task_construct` to get the reference to the HTML documentation. We generate a large Doxygen documentation file which contains things like this:

```
/**
 * @fn rtems_status_code rtems_task_construct(const rtems_task_config *config, rtems_id *id)
 *
 * @par Traceability
 * This design element is related to
 * [spec:/rtems/task/if/construct](file:///opt/rtems/rtems-6-sparc-gr740-smp-5/doc/ts/icd/html/requirements_and_design.html#spec-rtems-task-if-construct).
 */
```

You could also add additional information generated from other tools to this, for example code metrics, code coverage, static analysis results, etc. All members are in groups, so this patch uses groups to make the association. We have lots of duplicated names in test code for example.

PR #9954, opened 2023-03-26
Update translator_gr.h

nickreserved:
Update up to 1.9.7.
Keep in mind, I did not try a compilation.

### DIFF
--- a/src/translator_gr.h
+++ b/src/translator_gr.h
@@ -2342,6 +2342,301 @@ class TranslatorGreek : public TranslatorAdapter_1_9_4
       }
       return result;
     }
+//////////////////////////////////////////////////////////////////////////
+// new since 1.9.4
+//////////////////////////////////////////////////////////////////////////
+
+    virtual QCString trPackageList()
+    { return "Λίστα πακέτου"; }
+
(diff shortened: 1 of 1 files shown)


albert-github:
Nice.

I found 2 small problems:
the line
```
class TranslatorGreek : public TranslatorAdapter_1_9_4
```
should be replaced by
```
class TranslatorGreek : public Translator
```
as the Greek translation is now fully up to date

The lines 2323 - 2343 should be removed as the function `trCompoundType` would now be double  defined, the lines entered by means of the commit:
```
commit d7aa4a2a863c62a5ad1258fbc8047f5c2fc396ef
Author: albert-github <albert.tests@gmail.com>
Date:   Tue Jan 18 18:05:11 2022 +0100

    Union / Struct indicated as Class in the Xref list

    In the Xref lists (like Todo list) the Union / Struct and other compound elements are shown like:
    ```
    Class S_bb
    ```
    instead of:
    ```
    Struct S_bb
    ```

    The relevant translations have been added and an educated layman guess has been made for some translations (for the up to date translation).
```

(looks though that the layman's guess was not to bad.)

PR #9926, opened 2023-03-17
Fix enum field association for global enums

sebhub:
Make sure that the enum and contained enum fields originate from the  same file.  This fixes the following test case:

File "a.c":
```
/**
 * @file
 *
 * @brief a.c 
 */

/**
 * @brief E in a.c 
 */
enum E { 
  /** 
   * @brief A in a.c 
   */  
  A
};
```
File "b.c":
```
/**
 * @file
 *
 * @brief b.c 
 */

/**
 * @brief E in b.c 
 */
enum E { 
  /** 
   * @brief B in b.c 
   */  
  B
};
```
Without this fix, each enum E documentation contained the enum fields A and B.

Issue #3728, opened 2018-07-02
Enums with the same name in different C files being merged (Origin: bugzilla #615045)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.6.3* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[doxygen-merge-enum.zip](http://ftp.stack.nl/pub/users/dimitri/bugzilla/attachments//bug615045_158111_doxygen-merge-enum.zip)_ (ID 158111)

On 2010-04-07 11:52:46 +0000, Bernhard wrote:

> Created attachment 158111
> example files, configuration, doxygen 1.6.3 output
> 
> I have several C files, each with an enum with the same name. Doxygen merges all the enums together, giving the enum values of all enums in every enum description. It does not matter whether the enums are typedef'd or not.
> 
> Code:
> 
> test1.c
> 
> /// Example enum of test1.c.
> typedef enum TestEnum {
>     A = 0,  ///< First of test1.c.
>     B,      ///< Second of test1.c.
> } TestEnum;
> 
> 
> test2.c
> 
> /// Example enum of test2.c.
> typedef enum TestEnum {
>     C = 0,  ///< First of test2.c.
>     D,      ///< Second of test2.c.
> } TestEnum;
> 
> 
> test3.c
> 
> /// Example enum of test3.c.
> enum TestEnum {
>     E = 0,  ///< First of test3.c.
>     F,      ///< Second of test3.c.
> };
> 
> 
> I have attached a zip file containing the example files, configuration, and doxygen output.
> 
> The behaviour seems to be quite the same as described in the already fixed bug reports # 364673 and # 522193.

On 2013-01-04 11:11:22 +0000, Alfonso wrote:

> I found the same problem in Doxygen 1.8.3. I can add that this can cause that the same id is given two enum values in the xml output, which is problematic. More details below. 
> 
> Example, I define 
> 
> /**
>  * \brief One enum.
>  */
> enum my_enum {
> 	/** Enum field 0. */
> 	MY_FIELD_0 = 0,
> 	/** Enum field 1. */
> 	MY_FIELD_1 = 1,
> };
> 
> in two files (driver1.h and driver2.h) and each is added to a different group. When I generate html output, the enums are placed in the right group, but the values have been merged, i.e. in a group the enum is documented as:
> 
> enum my_enum
> Enumerator
> MY_FIELD_0 Enum field 0.
[...]

PR #9965, opened 2023-03-30
Add test case for enum typedef

sebhub:
This test case shows a bug.  The typedef in file more_099_b.c wrongly references the enum in file 099_a.c.

See also: #3728, #9926.


sebhub:
Ok, makes sense. Everyone interested in fixing the bug can pick up the test case if needed.

Issue #7087, opened 2019-06-28
Invalid (X)HTML in case of \mscfile, \dotfile, \diafile


albert-github:
In case of the usage of the commands `\dotfile`, `\mscfile`, `\diafile` the resulting html is invalid (as shown by the xmllint command together with the default doxygen dtd):

```
xmllint --path .../testing/dtd --noout  --nonet --postvalid html/*html
Element div is not declared in p list of possible children
Document html/aa_8h.html does not validate
```

The example is with `\mscfile` but the problem also occurs with `\dotfile` and `\diafile`: [msc_5.zip](https://github.com/doxygen/doxygen/files/3339225/msc_5.zip)

albert-github:
I've just pushed a proposed patch, pull request #9968

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.7.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #9968, opened 2023-04-06
issue #7087 Invalid (X)HTML in case of \mscfile, \dotfile, \diafile

albert-github:
Adding closing and opening paragraph tag before / after the  `\mscfile` etc. as a `\mscfile` can only appear inside a paragraph and would always be a "new" paragraph.

(Extended example: [example_2.tar.gz](https://github.com/doxygen/doxygen/files/11170345/example_2.tar.gz))

### DIFF
--- a/src/htmldocvisitor.cpp
+++ b/src/htmldocvisitor.cpp
@@ -1785,6 +1785,7 @@ void HtmlDocVisitor::operator()(const DocDotFile &df)
 {
   if (m_hide) return;
   if (!Config_getBool(DOT_CLEANUP)) copyFile(df.file(),Config_getString(HTML_OUTPUT)+"/"+stripPath(df.file()));
+  m_t << "</p>";
   m_t << "<div class=\"dotgraph\">\n";
   writeDotFile(df.file(),df.relPath(),df.context(),df.srcFile(),df.srcLine());
   if (df.hasCaption())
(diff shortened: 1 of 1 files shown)


src/htmldocvisitor.cpp
```diff
@@ -1785,6 +1785,7 @@ void HtmlDocVisitor::operator()(const DocDotFile &df)
 {
   if (m_hide) return;
   if (!Config_getBool(DOT_CLEANUP)) copyFile(df.file(),Config_getString(HTML_OUTPUT)+"/"+stripPath(df.file()));
+  m_t << "</p>";
```
doxygen, line 1788: This assumes the element is always already in a paragraph, which does not have to be the case. In other places, `forceEndParagraph()` and `forceStartParagraph()` are used to conditionally add a `</p>` or `<p>`.

albert-github: I have not found a case where we are not akready in a paragraph.
The `forceStartParagraph()` doesn't work everywhere, especially not when nothing is after the `\mscfile` line like:
```
...
  *
  * nothing after:
  * \mscfile with_msc.txt "With msc" width=5cm
 */
```

albert-github: I've found a case that in the new situation will fail:
```
/**
  * - \mscfile with_msc.txt "With msc" width=5cm
  * - \mscfile with_msc.txt "With msc" width=5cm
  */
void fie1();
```
but is still OK in case we use `forceStartParagraph()` and `forceEndParagraph()`.

so it looks like the current patch is indeed a bit to simple and `forceStartParagraph()` (and `forceEndParagraph()`?) should be adjusted to reflect the situation as given with the situation:
```
...
  *
  * nothing after:
  * \mscfile with_msc.txt "With msc" width=5cm
 */
```

new example (old example also lacked `MSCFILE_DIRS=.`): [example_3.tar.gz](https://github.com/doxygen/doxygen/files/11177053/example_3.tar.gz)

albert-github: It looks like it has, partly, to do with the line in htmldocvisitor.cpp:
```
return; // only whitespace at the end!
```
but there also looks like to be some interaction with the line:
```
      m_t << "<p" << htmlAttribsToString(p.attribs()) << ">";
```
in `void HtmlDocVisitor::operator()(const DocPara &p)`

Though I couldn't find a solution.

doxygen: I expect the specific DocNodes need to be added to `mustBeOutsideParagraph()` for it to work

albert-github: I did a test and this indeed seems to be the solution, I'll update the PR.

albert-github: Updated.

PR #9974, opened 2023-04-10
bug 630661 Highlighting lines in a dot graph

albert-github:
Based on the method as used by plantuml when `!pragma svginteractive true`  is set in doxygen a similar implementation is made:
- when hovering over a node all nodes and edges not directly connected to this node are blurred
- code used `svg.min.js` and loaded it from the internet, the file has been downloaded (so doxygen can still be used offline)  and adjusted (removed the reference line `//# sourceMappingURL=svg.min.js.map`)

To accomplish this we need unique edge identifications (hence a counting of the edges per graph)  and also the node identifications have to be in a way that a node id cannot be an abbreviation of another (e.g. `Node1` is an abbreviation of `Node10`).


src/dotfilepatcher.cpp
```diff
@@ -27,20 +27,146 @@
...
+"  if (mapOfLists[name].length &gt; 0) {\n"
+"    if (!mapOfLists[name].includes(item)) {\n"
+"      mapOfLists[name].push(item);\n"
+"    }\n"
+"  } else {\n"
+"    mapOfLists[name] = [item];\n"
+"  }\n"
+"}\n"
+"\n"
+"function main() {\n"
```
doxygen, line 69: @albert-github It looked like this `main()` function was not really needed, so I removed it. Please let me know if that still breaks something I didn't see.

doxygen:
@albert-github Please check https://github.com/doxygen/doxygen/commit/d43e255a58ee85e3d9e46348903cb5c5a835a1dd for the changes I made.

src/dotfilepatcher.cpp
albert-github: At first glance I don't see a problem.

PR #9975, opened 2023-04-11
Allow multiple newlines in a code span.

albert-github:
Currently a maximum of 1 newline is allowed in a code span, so something like:
```
`void fie(int t,
 int x,
 int y,
 int z)`
```
results in a warning.
With this patch multiple newlines are allowed, though a newline followed a new newline or a line with with just spaces  will terminate the code span as an incomplete code span.


albert-github:
I'll look into the problem.

albert-github:
Fixed.

Issue #7746, opened 2020-05-02
mismatch between page title in sqlite3 and other (at least XML) outputs at 1.8.16+


abathur:
**Additional context**
It may help to have context first: I've found time between projects to update/package/release a Python API I wrote to leverage (and test) the improvements I contributed to sqlite3gen back in #6482 (released in 1.8.15) and noticed a new break (in my module's tests) when using 1.8.16+. 

Its test suite runs Doxygen on the files in doxygen/examples and validates some entities in the database against the XML output for the same entity.

**Describe the bug**
I'm not sure if this is the only manifestation of the issue, but I identified it when a query for the page `mypage1` defined in `examples/group.cpp` found no result. Here's what the XML looks like in this test set:
```xml
<!-- xml/group__group5_mypage1.xml -->
<?xml version='1.0' encoding='UTF-8' standalone='no'?>
<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="1.8.17">
  <compounddef id="group__group5_mypage1" kind="page">
    <compoundname>mypage1</compoundname>
    <title>This is a section in group 5</title>
    <briefdescription>
    </briefdescription>
    <detaileddescription>
<para>Text of the first section </para>
    </detaileddescription>
  </compounddef>
</doxygen>

<!-- /xml/group__group5_mypage2.xml -->
<?xml version='1.0' encoding='UTF-8' standalone='no'?>
<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="1.8.17">
  <compounddef id="group__group5_mypage2" kind="page">
    <compoundname>mypage2</compoundname>
    <title>This is another section in group 5</title>
    <briefdescription>
    </briefdescription>
    <detaileddescription>
<para>Text of the second section </para>
    </detaileddescription>
  </compounddef>
</doxygen>
```

In the database, these pages look like:
```
sqlite> select * from def where kind='page' and name like '%mypage%';
rowid       refid                  kind        name        summary                     
----------  ---------------------  ----------  ----------  ----------------------------
228         group__group5_mypage1  page        mypage0     This is a section in group 5
229         group__group5_mypage2  page        mypage@     This is another section in g
```
The refids are correct, but the names are not.

**Expected behavior**
The names set as `mypage0` and `mypage@` should instead be `mypage1` and `mypage2`.

[...]

PR #7750, opened 2020-05-04
sqlite3gen: defend text binds against junk data

abathur:
As explored in #7746, some junk data is still finding its way into the sqlite3 tables.

There's probably a more elegant solution, but for now I've passed _static==FALSE for more/most calls to bindTextParameter to ensure that sqlite3 copies the strings on bind.

I used `time --verbose` to measure runs before and after this change, and don't see a marked impact on the maximum resident set size for processing a given config and set of files.

Below I've included output from `sqlite3 doxygen_sqlite3.db -column -header ".width 5 5 0 40" "select rowid, kind, refid, name from def"` run against a recent master:

```
rowid  kind   refid                                      name                                    
-----  -----  -----------------------------------------  ----------------------------------------
4      varia  classA_1a086d3a4efc697dba0601b9fef3d082ad                                          
6      enume  classAfterdoc__Test_1adab0cd7ad3b4875e245                                          
7      varia  classAfterdoc__Test_1a9287a08830e5cdfd9c7                                          
8      funct  classAfterdoc__Test_1a57ba94e9039ee90a1b1                                          
12     enume  classAutolink__Test_1aeb611627c332d067bde                                          
13     varia  classAutolink__Test_1a8de85603114bc9b9e53                                          
14     funct  classAutolink__Test_1a278d631f9943428c05b                                          
15     funct  classAutolink__Test_1a393ea281f235a2f603d                                          
16     funct  classAutolink__Test_1acf783a43c2b4b6cc9dd                                          
17     funct  classAutolink__Test_1a03bf46c8e2b73368003  ~Autolink_Test                          
21     varia  classB_1a26c70b64fe7cf17fcced7755ecff7537                                          
22     varia  classC_1a4ef972d28b73ff78eba3ab4f54c3b449                                          
30     varia  structCar_1ab8ff28306286da5a8b14fa9bdccaa                                          
31     funct  structObject_1a71225073d06a793b9a6ea9263e                                          
33     funct  structObject_1a924ee0cecc906d148022b3f0d6                                          
34     funct  structVehicle_1a6891d3d28853bc3fdd075596d                                          
35     funct  manual_8c_1a840291bc02cba5474a4cb46a9b956                                          
36     funct  structVehicle_1a4dcbcba43792dcd673a552b14                                          
37     varia  structObject_1a1b6037fba835e83243ababce42                                          
39     varia  structCoordStruct_1a183d7226fc5a8470ce9b9                                          
[...]

PR #9992, opened 2023-04-21
fix sqlite3gen regressions from 592aaa4f17

abathur:
This removes the `_static` argument from `bindTextParameter` and last two call sites using it.

While SQLITE_STATIC (which tells sqlite it doesn't need to copy the value) was fine for string literals, it looks like we should always pass SQLITE_TRANSIENT (which tells sqlite to copy the value at bind time) when using QCString.

(Without this, sqlite could end up inserting junk data into fields that were still using the `TRUE` value for this argument.)

I noticed these causing errors like the below in Python (here's the [CI build](https://github.com/abathur/doxy_db/actions/runs/4761192056/jobs/8462203813#step:4:1224) it came from):

```
    self._meta = sql.Statement(self).table("meta", "id").prepare()().fetchone()
E   sqlite3.OperationalError: Could not decode to UTF-8 column 'schema_version' with text '�'
```

I confirmed in CI (https://github.com/abathur/doxy_db/actions/runs/4767390612/jobs/8475561221) that applying this change on top of 1.9.6 fixes the error.

### DIFF
--- a/src/sqlite3gen.cpp
+++ b/src/sqlite3gen.cpp
@@ -859,14 +859,14 @@ class TextGeneratorSqlite3Impl : public TextGeneratorIntf
 };
 
 
-static bool bindTextParameter(SqlStmt &s,const char *name,const QCString &value, bool _static=FALSE)
+static bool bindTextParameter(SqlStmt &s,const char *name,const QCString &value)
 {
   int idx = sqlite3_bind_parameter_index(s.stmt, name);
   if (idx==0) {
     err("sqlite3_bind_parameter_index(%s)[%s] failed: %s\n", name, s.query, sqlite3_errmsg(s.db));
     return false;
   }
-  int rv = sqlite3_bind_text(s.stmt, idx, value.data(), -1, _static==TRUE?SQLITE_STATIC:SQLITE_TRANSIENT);
+  int rv = sqlite3_bind_text(s.stmt, idx, value.data(), -1, SQLITE_TRANSIENT);
   if (rv!=SQLITE_OK) {
     err("sqlite3_bind_text(%s)[%s] failed: %s\n", name, s.query, sqlite3_errmsg(s.db));
     return false;
@@ -930,7 +930,7 @@ static int insertPath(QCString name, bool local=TRUE, bool found=TRUE, int type=
 static void recordMetadata()
 {
   bindTextParameter(meta_insert,":doxygen_version",getFullVersion());
-  bindTextParameter(meta_insert,":schema_version","0.2.1",TRUE); //TODO: this should be a constant somewhere; not sure where
+  bindTextParameter(meta_insert,":schema_version","0.2.1"); //TODO: this should be a constant somewhere; not sure where
   bindTextParameter(meta_insert,":generated_at",dateToString(DateTimeType::DateTime));
   bindTextParameter(meta_insert,":generated_on",dateToString(DateTimeType::Date));
   bindTextParameter(meta_insert,":project_name",Config_getString(PROJECT_NAME));
@@ -2437,7 +2437,7 @@ static void generateSqlite3ForPage(const PageDef *pd,bool isExample)
   // + title
   bindTextParameter(compounddef_insert,":title",title);
 
-  bindTextParameter(compounddef_insert,":kind", isExample ? "example" : "page",TRUE);
+  bindTextParameter(compounddef_insert,":kind", isExample ? "example" : "page");
 
   int file_id = insertPath(pd->getDefFileName());
 


src/sqlite3gen.cpp
```diff
@@ -930,7 +930,7 @@ static int insertPath(QCString name, bool local=TRUE, bool found=TRUE, int type=
 static void recordMetadata()
 {
   bindTextParameter(meta_insert,":doxygen_version",getFullVersion());
-  bindTextParameter(meta_insert,":schema_version","0.2.1",TRUE); //TODO: this should be a constant somewhere; not sure where
+  bindTextParameter(meta_insert,":schema_version",QCString("0.2.1")); //TODO: this should be a constant somewhere; not sure where
```
albert-github, line 933: Why is `TRUE` gone here and in line 2440 as the default is `FALSE` for this parameter?

abathur: The _short_ answer is just that I still see the same symptom (schema_version is junk/empty) with the TRUE present. (Here's a fresh CI job, https://github.com/abathur/doxy_db/actions/runs/4765677175/jobs/8471719090, which uses doxygen built from https://github.com/abathur/doxygen/commit/3d4906a0fbd2ebfef08c8ce89ab57ccad3c5dc73.)

The default for `_static` causes `bindTextParameter` to pass `SQLITE_TRANSIENT` to `sqlite3_bind_text`, which in turn tells sqlite to go ahead and make a copy of the data because the pointer could change.

Passing `TRUE` tells it to pass `SQLITE_STATIC`, which tells sqlite that it doesn't need to make a copy because the value won't change.

It seems plausible to me that changes in 592aaa4f17 rendered the `_static` argument moot (i.e., `SQLITE_TRANSIENT` may now always be correct), but I don't write much C++ and don't completely understand all of the underlying motives and implications of that refactor.

abathur: If you're confident that it's correct, I'm happy to update the PR to remove the `_static` argument and update `bindTextParameter` to always pass `SQLITE_TRANSIENT` to `sqlite3_bind_text`.

abathur: Thanks for picking at this.

I'll respond in a bit more detail later for posterity, but I was just parroting some conversions from `"blah"` -> `QCString("blah")` that I saw elsewhere in 592aaa4, and it looks like all I actually need here is to fix the _STATIC/_TRANSIENT bit. I'll force push to update soon.

abathur: > It is a difficult problem and I also reer to the commits:
> 
> ```
> commit ee9a4679ae630f8f808619c5450f6b2f5a860994
> Author: Travis A. Everett <travis.a.everett@gmail.com>
> Date:   Tue May 5 09:07:15 2020 -0500
> 
>     sqlite: flip _static default in bindTextParameter
> 
> commit 2a6f9d50d606c59e86fe99e4304f056fd7f1032c
> Author: Travis A. Everett <travis.a.everett@gmail.com>
> Date:   Mon May 4 17:50:47 2020 -0500
> 
>     fix #7746 with liberal use of _static==FALSE
> 
>     Something in the refactor at 960bae7 caused junk data to start showing up in
>     some sqlite3 columns. Calling bindTextParameter with _static==FALSE causes it
>     to call sqlite3_bind_text with SQLITE_TRANSIENT instead of SQLITE_STATIC,
>     which in turn causes sqlite3 to copy the string at bind time.
> ```

For broader context, when I did the refactoring work on sqlite3gen that got merged in 2018, there were quire a few places where junk data could end up in some database fields. My confidence in C++ is minimal, so I was trying to be conservative and only tried to fix fields I had seen junk data in at some point myself. 

IIRC in 2020 I was fixing up some build failures that had crept in, and in the process I noticed more of these fields with junk data because I had been to conservative (see discussion in https://github.com/doxygen/doxygen/pull/7750). Initially that PR was just explicitly passing _static=FALSE to more of the calls and in https://github.com/doxygen/doxygen/pull/7750#issuecomment-624046822 you encouraged me to invert the default to minimize how many call sites override it.

>  * It is a bit strange to me that you need the explicit conversion to `QCString` as I thought this would be done automatically (by means of `QCString( const char *str ) : m_rep(str?str:"") {}`, taken from qcstring.h)

Good catch. I was basically just parroting some of the transformations that were already in 592aaa4f17, such as:
- https://github.com/doxygen/doxygen/commit/592aaa4f17d73ec8c475df0f44efaea8cc4d575c#diff-c9367a6d87716cac78bb0ce2a1018084931694345c1c27e61ffbc71a79531674L2864-R2867
and
- https://github.com/doxygen/doxygen/commit/592aaa4f17d73ec8c475df0f44efaea8cc4d575c#diff-c9367a6d87716cac78bb0ce2a1018084931694345c1c27e61ffbc71a79531674L3073-R3074

I was reading these notes in the commit as supporting that:
```
    - The implicit convert from 'QCString' to 'const char *' is removed
    - Strings parameters use 'const QCString &' as much as possible in favor
      over 'const char *'
```

But now that you call it out, I realize they say nothing about the inverse conversion.

> * do you see the problems with `_static` is `TRUE` and with `FALSE`?
>   This is a bit hard to get from the debug output you give as it is regarding `schema_version` and I expect that this would be one of the first things to check / retrieve.

Just with `TRUE`, I think. It's hitting on `schema_version` because the tool I'm updating ([doxy_db](https://github.com/abathur/doxy_db) asserts what schema version it's compatible with, so it's failing before it does much.

> might it be that the `SQLITE_STATIC` part has problems with reading the data?

My mental model for what it's doing is that when `SQLITE_STATIC` is used at bind time on something that could actually change before the statement is finally executed, it'll hold on to the memory location and then later insert whatever is currently memory where it expected the value to be. (These are wrong in the generated database.) Always using `SQLITE_TRANSIENT` with QCStrings at least appears to fix this (without explicitly converting to QCString) in testing on my end.

abathur:
After force-pushing, I also updated the first post with a new explanation and updated the link to a passing CI job based on this change.

Issue #10010, opened 2023-04-27
SHOW_HEADERFILE doesn't work well for headers with no file extension


jwakely:
**Describe the bug**

With this config:

```
SHOW_HEADERFILE=YES
EXTENSION_MAPPING      = no_extension=C++ \
                         .h=C++
```

 the description for every C++ class defined in a header file `foo.h` will show:

    #include <foo.h>

This happens whether the classes have `\headerfile` in their comments or not. And if you use `@headerfile foo` it will show `#include <foo>` instead. Great.

But C++ classes defined in a file with no extension, e.g. `bar`, will not show their header file by default. You have to use `\headerfile` to get them to display the `#include <bar>` info. This is inconsistent.

The point of `\headerfile` is to be able to "correct" the header shown for `SHOW_HEADERFILE`, when the file that the class is actually defined is some implementation details like `detail/foo.h` and the public API says to include `<bar>` instead. So you use `\headerfile bar` to make the class display the right header.

But when the class is already defined in the same header that users should actually include, you usually don't need to specify `\headerfile`, you just let Doxygen infer it correctly and display `#include <foo.h>`. This doesn't work if the file has no extension, so you need to add an explicit `\headerfile bar` to every class in file `bar`. Why? Is it because a file with no extension is being treated as a source file, not a header file?

**Expected behavior**

A class defined in a file with no extension should infer its `\headerfile` to be the filename, and show that in the `SHOW_HEADERFILE` info.

If the problem is that an extension-less file like `bar` is assumed to be more like `bar.cc` than `bar.h`, could we maybe get a way to say that a given file is header-like not source-like? Maybe an extension (no pun intended) to `EXTENSION_MAPPING` so that we can say `no_extension=C++hdr` instead of `no_extension=C++`?

Or maybe something on the `\file` for `bar` that could say "treat this file like a header, and so every class defined in the file has an implied `\headerfile bar`. Maybe implementing #9334 would allow that, by giving a `header-file` or `\headerfile` in the `\file` comment block.

**Screenshots**

![issue10007](https://user-images.githubusercontent.com/1254480/234913661-78f00975-75c7-4e15-88fe-c06d5a8d30e8.png)

This shows the docs for three structs side by side. `S` and `S2` are both in a file called `bar` with this content:

```c++
/// @file bar
/// This is a header file called `<bar>` (with no extension).

/// A struct in file `<bar>`
struct S { };

/// @brief Another struct in file `<bar>`
/// @headerfile bar
[...]

PR #10014, opened 2023-04-29
issue #10010 SHOW_HEADERFILE doesn't work well for headers with no file extension

albert-github:
Implemented that when a file has no extension and the mapping for files without extension is set to C++ such a file is seen as header file (thanks @jwakely )


jwakely:
Thank you!

Issue #10017, opened 2023-04-30
search_opensearch.php wrongly indicates that each suggestion has exactly 1 result


Krinkle:
**Describe the bug** (Actual behaviour):

<https://doc.wikimedia.org/mediawiki-core/master/php/search_opensearch.php?v=json>
> `["foo", ["footer", "footers", "footerhtml", …, "fooey"],["1 result", "1 result", "1 result", … "1 result"]]`

There is also a warning in the server-side PHP error.log:
```
PHP Warning:  array_key_exists(): The first argument should be either a string or an integer)
```

**Expected behavior**

> `["foo", ["footer", "footers", "footerhtml", …, "fooey"],["33 results", "4 results", "5 results", … "1 result"]]`

**Version**
1.9.6

albert-github:
Please don't add a link to a pull request that closes the issue.
When merged doxygen will close the issue at the next release after the merge.

Krinkle:
@albert-github Ack, I've updated the commit message to say "Ref" instead of "Fixes" to avoid automatic closure.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.7.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10018, opened 2023-04-30
search: Fix incorrect result count in opensearch suggestions

Krinkle:
The descriptions displayed under the OpenSearch suggestions in a web browser always said "1 result". This also emitted a warning on the server-side:

> PHP Warning:  array_key_exists(): The first argument should be either a string or an integer)

See also https://github.com/doxygen/doxygen/issues/10017.

### DIFF
--- a/templates/html/search_functions.php
+++ b/templates/html/search_functions.php
@@ -301,6 +301,16 @@ function report_results(&$docs)
   echo "</div>\n";
 }
 
+/**
+ * @param string $query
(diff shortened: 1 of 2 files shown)


hashar:
@albert-github / @doxygen may you potentially give this one a look? Krinkle has spotted that from the PHP error log of our documentation host ( https://doc.wikimedia.org/mediawiki-core/master/php/ ).

PR #10021, opened 2023-05-01
Enable possible other LaTeX error modes

albert-github:
Besides the supported `\batchmode` there are a number of other "error" modes for LaTeX available, they can be enabled as well.

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -681,13 +681,26 @@ static QCString makeIndex()
   return result;
 }
 
+static QCString latex_batchmode()
+{
+  switch Config_getEnum(LATEX_BATCHMODE)
+  {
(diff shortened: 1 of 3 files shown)


src/latexgen.cpp
```diff
@@ -681,13 +681,26 @@ static QCString makeIndex()
   return result;
 }
 
+static QCString latex_batchmode()
+{
+  switch Config_getEnum(LATEX_BATCHMODE)
```
doxygen, line 686: This switch statement looks a bit weird without parenthesis; it probably works only because Config_getEnum is macro that already comes with the parenthesis.

albert-github: Indeed looks a bit strange and the macro (config_get...) already contains the brackets.

Updated.

Issue #10015, opened 2023-04-30
Add warning when file set for USE_MDFILE_AS_MAINPAGE isn't found


dereckson:
**Describe the bug**

If the file set for USE_MDFILE_AS_MAINPAGE isn't found, for example because it's not set in INPUT, there is no warning in doxygen output.

**Expected behavior**
warning: README.md not found, but set in USE_MDFILE_AS_MAINPAGE

**To Reproduce**

```
mkdir -p test/src
cd test
touch README.md
cat > Doxyfile
INPUT = src
USE_MDFILE_AS_MAINPAGE = README.md
```

**Version**
1.9.6

albert-github:
I've just pushed a proposed patch, pull request #10022

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.7.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10022, opened 2023-05-01
issue #10015 Add warning when file set for USE_MDFILE_AS_MAINPAGE isn't found

albert-github:
Add check on existing and specified mainpage

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -11692,6 +11692,27 @@ void searchInputFiles()
 }
 
 
+static void checkMarkdownMainfile()
+{
+  if (Config_getBool(MARKDOWN_SUPPORT))
+  {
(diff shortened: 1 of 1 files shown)


divenal (commented):
After upgrading to a version with this change, we get a warning about the README.md file not being found, even though it gets processed correctly.
I think the problem might be that this doesn't use the search path to find the file, just uses the filename directly as the (possibly relative) path.

albert-github:
@divenal 

Without further information it is not possible to tell what is the problem at your side, so:

albert-github:
I tried with the current doxygen 1.13.2 version I get and here:
- I get the warning
- In the HTML output I don't see the text and the mentioned grep does not show the text either

So looks like the problem has been solved in the current doxygen 1.13.2 version

PR #10016, opened 2023-04-30
Usage of updated etoc package

albert-github:
For pull request #9936 an updated version of the LaTeX `etoc` package is required. To be able to run the code also with older versions of LaTeX (i.e. versions that not yet have these updates) the updated package is provided as a local package.

PR #10024, opened 2023-05-01
Fix latex counter reset and subsubsection

mtc-mlx:
Fix remaining issues from #9936. This does **not** replace #10016.


albert-github:
I checked it against the problems I saw in [#9936](https://github.com/doxygen/doxygen/pull/9936#issuecomment-1517506542) regarding the missing paragraph titles, these are now present.

PR #10025, opened 2023-05-02
Incorrect svg file generated

albert-github:
- a number of, simple, svg files were missing missing the closing `</svg>` tag giving an invalid svg file (see e.g. https://doxygen.github.io/doxygen/dir_534aeb5b2351b2c83787cc7391511c04.html when using Chrome)
- added `type` attribute to `script` tag (more svg standard conforming)

### DIFF
--- a/src/dotfilepatcher.cpp
+++ b/src/dotfilepatcher.cpp
@@ -111,7 +111,7 @@ static const char svgZoomFooter2[] = R"svg(
 [data-mouse-over-selected='true']  { opacity: 1.0; }
 ]]>
 </style>
-<script><![CDATA[
+<script xlink:type="simple"><![CDATA[
 document.addEventListener('DOMContentLoaded', (event) => {
   highlightEdges();
   highlightAdjacentNodes();
@@ -380,7 +380,7 @@ bool DotFilePatcher::run() const
           }
           if (useNagivation)
           {
-            t << "<script type=\"text/javascript\">\n";
+            t << "<script xlink:type=\"simple\">\n";
             t << "var viewWidth = " << width << ";\n";
             t << "var viewHeight = " << height << ";\n";
             if (graphId>=0)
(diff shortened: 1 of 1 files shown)


src/dotfilepatcher.cpp
```diff
@@ -111,7 +111,7 @@ static const char svgZoomFooter2[] = R"svg(
 [data-mouse-over-selected='true']  { opacity: 1.0; }
 ]]>
 </style>
-<script><![CDATA[
+<script xlink:type="simple"><![CDATA[
```
doxygen, line 114: where did you find that xlink:type is more standard conforming? I found [here](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:type) it is actually a deprecated tag.

albert-github: I found it in the svg definitions. It is possible that the `xlink:type` is deprecation for the svg 2.0 version but doxygen is using svg 1.1  (as is e.g. `dot`).

doxygen: I read [here](https://www.w3.org/TR/SVG11/linking.html) that: "In SVG 1.1, only simple links are available. Links are simple links by default, so the attribute xlink:type="simple" is optional and may be omitted on simple links. ". So I think it is better to leave out the xlink:type="simple" (smaller file size).

albert-github: Yes only the value "simple" is possible, though when checking against the dtd with xmllint I get a warning.
In the dtd I fined everywhere. n respect to title, %XLINK.pfx;type ( simple ) #FIXED 'simple' so because of the warning I added the attribute.

doxygen:
@albert-github I've used type="application/ecmascript" now as mentioned in the SVG 1.1 standard. Let me know if you still see problems and how I can reproduce them.

albert-github:
I don't see this problem now, there are some other problems but I will look into them later on (I have some ideas for them):
- problem wit http <-> https
- element a: validity error : Element a does not carry attribute xlink:href


Regarding the "how I can reproduce them", I downloaded the `dtd`s ([dtd_svg.tar.gz](https://github.com/doxygen/doxygen/files/11414881/dtd_svg.tar.gz) ) and used them with xmllint (analogous to using the "normal" `dtd` with the tests).
When necessary I can add these files (as a separate directory or as part of the `dtd` directory to the testing directory).

PR #10041, opened 2023-05-11
bug 784736 Words that are the same to class names are hyperlinked in references in the HTML output

albert-github:
Temporary disable the `AUTOLINK_SUPPORT` for the "citelist" page so information is consistent between LaTeX (PDF) output and other output formats (and no automatic hyperlinks appear in the bibliography).


albert-github:
Note: the solution only works as long the printing of the pages is done single threaded (as it is done now) as in a multi threaded environment the different threads might interact.

PR #10060, opened 2023-05-19
Setting correct version in LaTeX documentation

albert-github:
Due to the fact that on `tex` files it is not possible to use the normal environment replacement strategy of cmake the `@VERSION@` does not get replaced anymore. Using the possibility of retrieving environment variables (see also: https://tex.stackexchange.com/a/583976/44119)


albert-github:
Will have to check what is wrong.

albert-github:
Problem looks like Debian / version dependent (I already asked on tex.stackexchange.com)

albert-github:
Superseded y #10061

PR #10061, opened 2023-05-20
Setting correct version in LaTeX documentation

albert-github:
Due to the fact that on `tex` files it is not possible to use the normal environment replacement strategy of cmake the `@VERSION@` does not get replaced anymore. An internal LaTeX possibility (see: https://tex.stackexchange.com/a/583976/44119) looks like to be LaTeX version dependent, so a Cmake-Python solution

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -202,9 +202,9 @@ add_custom_target(run_doxygen
 add_custom_target(doxygen_pdf
         COMMENT "Generating Doxygen Manual PDF."
         COMMAND ${CMAKE_COMMAND} -E remove refman.tex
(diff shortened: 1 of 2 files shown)


albert-github:
This proposed pull request replaces #10060

doxygen:
@albert-github Isn't there a simpler cmake only solution for this? e.g. something like mentioned here: https://discourse.cmake.org/t/what-is-the-best-way-to-search-and-replace-strings-in-a-file/1879

albert-github:
I think the `file` / `string` functions only work on the initial level and cannot be used in "rule" or have dependencies.

doxygen:
> I think the `file` / `string` functions only work on the initial level and cannot be used in "rule" or have dependencies.

I think you are right. Found a cmake only solution here, but it requires generating a separate cmake file and then calling it:
https://stackoverflow.com/a/75761149/784672 so it does not improve things much.

PR #10062, opened 2023-05-21
cleanup qcstring indentation and style

zchrissirhcz:
This PR try cleanup work in qcstring.h and qcstring.cpp, by:
- indentation adjusted, use 2 space, keep the same as most of other c/cpp code
- replace `0` with `nullptr` for better readability

### DIFF
--- a/src/qcstring.cpp
+++ b/src/qcstring.cpp
@@ -433,12 +433,12 @@ void qstrfree( const char *str )
 
 char *qstrncpy( char *dst, const char *src, size_t len )
 {
-    if ( !src )
-	return 0;
-    strncpy( dst, src, len );
-    if ( len > 0 )
-	dst[len-1] = '\0';
-    return dst;
+  if ( !src )
+    return nullptr;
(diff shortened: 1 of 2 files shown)


src/qcstring.cpp
```diff
@@ -470,20 +470,27 @@ int qstricmp( const char *s1, const char *s2 )
...
     }
-    for ( ; len--; s1++, s2++ )
-    {
-        char c = toLowerChar(*s1);
-        int res = c-toLowerChar(*s2);
-	if ( res!=0 ) // strings are not equal
-	    return res;
-	if ( c==0 ) // strings are equal
-	    break;
+    if (c2!=0 ) {
```
doxygen, line 488: This should probably have been `c2==0`

zchrissirhcz: Fixed now.

zchrissirhcz:
My previous commit of modifying `qstricmp()` and `qstrnicmp()` was not correct. Droped and force pushed now.

PR #10063, opened 2023-05-21
parser update to vhdl2017

maddox11:
parser update to VHDL IEEE 1076-2017


albert-github:
Nice big piece of work.
At the bottom I see that the parts 2002 and 2008 have been removed, are all constructs from 2002 / 2008 also included now in this version or does it mean that 2002 / 2008 compliant sources cannot be run anymore?

maddox11:
parser is compliant with VHDL 1987/93/2002/2008/2017

Issue #10076, opened 2023-05-23
\include and Python


couet:
**Describe the bug**
In the ROOT reference guide (build with Doxygen) we are using `\include` to show the code of tutorials.
Some are in C++ some are in Python. The highlighting looks nice for C++ but does not work very well with 
some Python examples we have. Like the one mentioned here:
https://github.com/root-project/root/issues/12851

**Expected behavior**
The correct highlight of Python code.

**Screenshots**
See https://github.com/root-project/root/issues/12851

**Version**
1.9.5

albert-github:
I can see in the string literals some problems but which other problems do you see?
The source and output from the project mentioned is a moving target so:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

couet:
Yes, the problem is with the string literal. I will test your patch. Thanks

couet:
By the way, I have a question about `\include`: How does it decide which language is the file included / with the file name extension?

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

albert-github:
Regarding the questing about the `\include` (https://github.com/doxygen/doxygen/issues/10076#issuecomment-1560557820) this is done in the same way as "normal" files i.e. based on the extension.

couet:
Maybe it can be mentioned here: https://www.doxygen.nl/manual/commands.html#cmdinclude ?

albert-github:
Is an idea, but should in that case also be added to other commands like `\snippet`, `\example`, ...

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.8.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

couet:
It is fixed in 1.9.8. Thanks

PR #10077, opened 2023-05-23
issue #10076 \include and Python

albert-github:
In general the source browser for python code showed some strange effects in case of a number of spaces at the beginning of a line inside a triple quoted block. Instead of staying inside the triple quoted block a jump was done to the outside of the block as  the general rule `<*>[ \t]+` was executed (and a jump to condition `Body` occurred) instead  of the a handling inside the triple quoted block.

Constructed example: [example.tar.gz](https://github.com/doxygen/doxygen/files/11546627/example.tar.gz)


couet:
This PR fixes the issue:
Screenshot 2023-05-24 at 09 46 06

Issue #10104, opened 2023-05-31
Incorrect handling of unicode characters


albert-github:
When trying to verify whether the problem of #5454 was solved another problem emerged, regarding the characters used.
When having a vhdl file (**MainAddressDecoder.vhd**):
```

--! @file
--! @brief Hlavní adresový dekodér pro platformu Z102
--!
```
we get the warnings:
![image](https://github.com/doxygen/doxygen/assets/5223533/9e737e32-4d27-45b2-afd4-aab4145308c7)

sorry for the screenshot but in a file it looks like:
```
MainAddressDecoder.vhd:3: warning: Lexical error, Encountered: 'Ã' after: ''
MainAddressDecoder.vhd:3: warning: Lexical error, Encountered: '­' after: ''
MainAddressDecoder.vhd:3: warning: syntax error 'adresov'
```

and the Html output is:
![image](https://github.com/doxygen/doxygen/assets/5223533/319c236f-6ab6-4251-8754-893a7a0e2f31)

When taking an equivalent cpp file (**MainAddressDecoder.cpp**):
```

/// @file
/// @brief Hlavní adresový dekodér pro platformu Z102
///
```
We don't get a warning and the output looks correct:
![image](https://github.com/doxygen/doxygen/assets/5223533/75187124-c26a-4318-972b-23a4b7b0a90a)

When having the "block comments " in VHDL like (**MainAddressDecoder1.vhd**):
```
/*! @file
    @brief Hlavní adresový dekodér pro platformu Z102
*/
```
We get the warning:
```
MainAddressDecoder1.vhd:1: warning: syntax error '/'
```
and no output at all.
(analogous cpp is OK)

@maddox11 do you have an idea?

albert-github:
I forgot to attach a small example, although I was referring to it: [example.tar.gz](https://github.com/doxygen/doxygen/files/11642697/example.tar.gz)

albert-github:
Was just thinking, might it be a Windows specific problem and maybe even depending on the compiler version or the bitness?

All tests with 1.9.8 (b3f1ffa5d8e2794ef6f8f0c6128cbb45eddbadeb)
- My default old 32-bit old Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27034 for x86: problem
- The 64-bit old Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27034 for x64: no problem
- GitHub Actions (64-bit) release and debug: no problem
[...]

PR #10114, opened 2023-06-06
issue #10104 Incorrect handling of unicode characters

albert-github:
Correcting code in respect to differences in char type between Java (i.e. char is unsigned) and C++ (char is signed) (see also https://github.com/javacc/javacc/pull/265)


albert-github:
In https://github.com/javacc/javacc/pull/265 it was suggested to use `-f unsigned-char` (and for MSVC `/J`) though I don't think this is a good solution.

albert-github:
Based on the discussion here, in #10104 and https://github.com/javacc/javacc/pull/265 an alternative approach has been implemented.

This PR is obsolete.

PR #10116, opened 2023-06-07
Switch dark / light mode on ESC key

albert-github:
In the darkmode some users have a problem to find the possibility to switch back to the light mode. Adding a key release event on the ESC-key that toggles between the dark and light mode.

### DIFF
--- a/templates/html/darkmode_toggle.js
+++ b/templates/html/darkmode_toggle.js
@@ -100,6 +100,11 @@ class DarkModeToggle extends HTMLElement {
                 $(window).resize(function(){
                     addButton();
                 })
+                $(document).keyup(function(e) {
+                    if (e.key === "Escape") { // escape key maps to keycode `27`
+                       DarkModeToggle.userPreference = !DarkModeToggle.userPreference
+                   }
+                })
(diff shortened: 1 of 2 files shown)


templates/html/darkmode_toggle.js
```diff
@@ -100,6 +100,11 @@ class DarkModeToggle extends HTMLElement {
                 $(window).resize(function(){
                     addButton();
                 })
+                $(document).keyup(function(e) {
+                    if (e.key === "Escape") { // escape key maps to keycode `27`
```
doxygen, line 104: `search.js` also has an escape key check to abort the search, i.e.

```javascript
else if (e.keyCode==27) // Escape out of the search field
```

how do this toggle interact with that?

albert-github: It doesn't interact good as when hitting the ESC key whilst in the search box it also sees the ESC key for the toggle and handles both. So some smart detection would probably be in place i.e. the "toggle" ESC key should not kick in when ithe focus is in the search box.
Do you know of a way to do this?

albert-github: Thanks fo the pointer. The `e.stopPropagation()`  looks like the one that should be used.

Issue #10120, opened 2023-06-09
Files with same name have a path prefix


hargorin:
**Describe the bug**
I have files with the same name but in different directories. Doxygen automatically adds a prefix in front of the files with the same name. So if you go to 'Modules', only the files with the same name have the prefix. The other files in the same directory have no prefix.

e.g.
├── header
│   ├── folder_a
│   │   ├── folder_a/foo.h
│   ├── folder_b
│   │   ├── bar.h
│   │   ├── folder_b/foo.h
└── baz.h

In my original case, the prefix behaves as follows (but I cannot reproduce it in this example)
├── header
│   ├── folder_a
│   │   ├── a/foo.h
│   ├── folder_b
│   │   ├── bar.h
│   │   ├── b/foo.h
└── baz.h

**Expected behavior**
It would be good if you had the option to add the prefix for all files in these folders.
- FULL_PATH_NAMES is not an option for me as the "file list" should be a one dimensional list.

├── header
│   ├── folder_a
│   │   ├── folder_a/foo.h
│   ├── folder_b
│   │   ├── folder_b/bar.h
│   │   ├── folder_b/foo.h
└── baz.h

**To Reproduce**
See attached file

**Version**
doxygen 1.9.1

**Additional context**
[SameFileName.zip](https://github.com/doxygen/doxygen/files/11702809/SameFileName.zip)

albert-github:
This is indeed not logical as output like:
![image](https://github.com/doxygen/doxygen/assets/5223533/c078412c-ed04-4142-83cc-c461c28efd46)


gives the impression that the file `foo.h`  would reside in the subfolder `folder_b` of `folder_b` and this is not the case.

There are a few places where these constructs are used, and there the `fd->name()` should probably be replaced with `fd->docName()`

doxygen:
@hargorin Please verify that the referenced commit fixes the problem. Do not close the issue report, this will be done automatically when the next official release is out.

hargorin:
Sorry for the late reply.
I tried it yesterday and found the same behavior/problem as described in #10124.
Let me know if you want me to try it again.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.8.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10124, opened 2023-06-12
Unclear / double path for file names

albert-github:
Based on #10120  some more path problems were found
- HTML in case of `FULL_PATH_NAMES=NO` there was still a path shown
- other formats, in the file index it was unclear `FULL_PATH_NAMES=NO` which file was intended

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/11721647/example.tar.gz)


hargorin:
A question about the new fix.
If I now have two files with the same name but in different folders, the configuration FULL_PATH_NAMES = (YES/NO) has no effect.
Is this the way it is supposed to be?

PR #10125, opened 2023-06-12
Mention "Source File" in file index.

albert-github:
In the file index (e.g. LaTeX) we for the source files just the filename, so in the index it is unclear what is the case here. (In the HTML output the words "Source File" were already present).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/11723493/example.tar.gz)

### DIFF
--- a/src/filedef.cpp
+++ b/src/filedef.cpp
@@ -1064,7 +1064,12 @@ void FileDefImpl::writeSourceHeader(OutputList &ol)
       ol.endQuickIndices();
     }
     startTitle(ol,getSourceFileBase());
-    ol.parseText(name());
+    QCString nam;
+    if (Config_getBool(FULL_PATH_NAMES))
+      nam = m_docname;
+    else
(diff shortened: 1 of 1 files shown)


src/filedef.cpp
```diff
@@ -1064,7 +1064,12 @@ void FileDefImpl::writeSourceHeader(OutputList &ol)
       ol.endQuickIndices();
     }
     startTitle(ol,getSourceFileBase());
-    ol.parseText(name());
+    QCString nam;
+    if (Config_getBool(FULL_PATH_NAMES))
+      nam = m_docname;
+    else
+      nam = name();
```
albert-github, line 1071: There were some second thoughts about this construct in https://github.com/albert-github/doxygen/commit/d53f37cfd34af1d51f9715c7e6101fbbd11c3c5b#r117687209.
The solution here should be identical to the solution for #10124

PR #10132, opened 2023-06-14
Better layout of lists in output formats

albert-github:
In different output format the layout of some listing was not correct (especially in the RTF output):
- missing space after the word "directory" (whilst present after the word "file")
- missing separators in lists (the big dots), due to missing initial initial parts, i.e. strange concatenation of items (when brief was missing).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/11746377/example.tar.gz)

### DIFF
--- a/src/groupdef.cpp
+++ b/src/groupdef.cpp
@@ -914,7 +914,7 @@ void GroupDefImpl::writeNestedGroups(OutputList &ol,const QCString &title)
         ol.startMemberDeclaration();
         ol.startMemberItem(gd->getOutputFileBase(),OutputGenerator::MemberItemType::Normal);
         //ol.docify(theTranslator->trGroup(FALSE,TRUE));
-        //ol.docify(" ");
+        ol.docify(" ");
         ol.insertMemberAlign();
         ol.writeObjectLink(gd->getReference(),gd->getOutputFileBase(),QCString(),gd->groupTitle());
         ol.endMemberItem(OutputGenerator::MemberItemType::Normal);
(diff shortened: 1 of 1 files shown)


src/groupdef.cpp
```diff
@@ -946,7 +946,7 @@ void GroupDefImpl::writeDirs(OutputList &ol,const QCString &title)
       if (!dd->hasDocumentation()) continue;
       ol.startMemberDeclaration();
       ol.startMemberItem(dd->getOutputFileBase(),OutputGenerator::MemberItemType::Normal);
-      ol.parseText(theTranslator->trDir(FALSE,TRUE));
+      ol.docify(theTranslator->trDir(FALSE,TRUE)+" ");
```
doxygen, line 949: wouldn't it be better to move adding the space inside `insertMemberAlign()` for those formats that require it?

albert-github: I made it so that it is analogous to the code in `void GroupDefImpl::writeFiles(OutputList &ol,const QCString &title)` where we already have: `ol.docify(theTranslator->trFile(FALSE,TRUE)+" ");`
The change is required for LaTeX, RTF, Docbook (I didn't test it for man). For the HTML output it doesn't really hurt.
When looking where `insertMemberAlign()` is used we see that this is in:
- classdef.cpp
- conceptdef.cpp
- dirdef.cpp
- groupdef.cpp
- namespacedef.cpp
- vhdldocgen.cpp

With dirdef.cpp there is currently no space either and I don't know what the effect would be , though the fact that there is a `<span...` line a little above it I think it is here only used in HTML or tag file.
The situation in `vhdldocgen.cpp` is a bit more complicated as here we see that there is here sometimes a space before or (often) after the call to `insertMemberAlign`.

PR #10159, opened 2023-07-01
bug 591431 The encoding of the layout file is not taken into account

albert-github:
In case of the layout file also take the possible encoding in the `<?xml...` part into account.


albert-github:
An alternative code update has been made in master on GitHub (906571f37afbb3e8ebb5be3a88a35c239adefa73), so this PR is obsolete.

PR #10161, opened 2023-07-01
Fix c++20 error

dr-al:
simple-template-id not allowed for ctor in c++20

### DIFF
--- a/src/template.cpp
+++ b/src/template.cpp
@@ -1599,7 +1599,7 @@ class TemplateFilterFactory
     template<class T> class AutoRegister
     {
       public:
-        AutoRegister<T>(const QCString &key)
+        AutoRegister(const QCString &key)
         {
           TemplateFilterFactory::instance().registerFilter(key,&T::apply);
         }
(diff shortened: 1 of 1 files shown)


albert-github:
- where is this mentioned in the C++ standard?
- how does the error manifest in the build of doxygen?
  - error message
  - which settings / options used with CMake
  - which platform
  - which compiler

(I started the build to see what happens on the different GitHub Actions platforms)

dr-al:
This [section](http://eel.is/c++draft/diff.cpp17.class#2) explains the issue

I am seeing the following errors when building with gcc 13.1.0 and -std=c++20
```
/app/src/template.cpp:1602:25: error: expected unqualified-id before 'const'
 1602 |         AutoRegister<T>(const QCString &key)
      |                         ^~~~~
/app/src/template.cpp:1602:25: error: expected ')' before 'const'
 1602 |         AutoRegister<T>(const QCString &key)
      |                        ~^~~~~
      |                         )
/app/src/template.cpp:1613:80: error: too many initializers for 'TemplateFilterFactory::AutoRegister<FilterAdd>'
 1613 | static TemplateFilterFactory::AutoRegister<FilterAdd>                fAdd("add");
      |                                                                                ^
```

PR #10168, opened 2023-07-03
Show outline on focused elements with tab press

SergeyRyabinin:
## Issue
When navigating pages with `<tab>`* button, navigation elements are not focused/outlined.

## Change
Adjust CSS so that `outline: none;` from another property is not applied on focused element.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Update: *< tab > was not rendered in markdown.

### DIFF
--- a/templates/html/doxygen.css
+++ b/templates/html/doxygen.css
@@ -139,6 +139,10 @@ a.navtab {
 	font-weight: bold;
 }
 
+.tablist a:focus {
+	outline: auto;
+}
+
 div.qindex{
 	text-align: center;
 	width: 100%;


albert-github:
> When navigating pages with button, navigation elements are not focused/outlined.

Which button ?

SergeyRyabinin:
> Which button ?

Tab on the keyboard. Sorry, it was not rendered in markdown.
I'll post links this week for before/after for both PRs once we deploy the updated version.

SergeyRyabinin:
Hello,

an updated version has been deployed, could you please take a look at before/after?
[BEFORE](https://sdk.amazonaws.com/cpp/api/1.11.113/aws-cpp-sdk-s3/html/class_aws_1_1_s3_1_1_s3_client.html), [AFTER](https://sdk.amazonaws.com/cpp/api/1.11.114/aws-cpp-sdk-s3/html/class_aws_1_1_s3_1_1_s3_client.html).

The issue in before is that when navigating with a keyboard with a `TAB` key does not show outline rectangle on (all) elements currently being browsed.
We use custom CSS styling, but the issue still can be observed on a stock doxygen theming (by disabling all additional CSSs in a browser dev tools).

Best regards,
Sergey

albert-github:
Accessibility is a great thing, so I had a quick look.

In your example the "boxes" show up but in a standard doxygen run they don't show (tested on a doxygen internal documentation run, `doxygen docs_internal`). I added the extra code to the `doxygen.css`
- with the doxygen internal doxygen I don't see the tablist, looks like this is only generated when `GENERATE_TREEVIEW=NO`
- with the doxygen internal doxygen and `GENERATE_TREEVIEW=NO`I don't see the boxes, so might also be that local settings have anninfluence
- maybe there are more settings needed (or a general setting?)

PR #10169, opened 2023-07-03
Make MSearchSelect keyboard accessible.

SergeyRyabinin:
## Issue
Interactive functionality of search kind selection is not operable with the keyboard (accessibility issue)
## Change
Make MSearchSelect (aka loop button selection of search type) keyboard accessible by adjusting tabindex and handling focus in JS.

### DIFF
--- a/templates/html/htmlbase.tpl
+++ b/templates/html/htmlbase.tpl
@@ -217,7 +217,7 @@ $(document).ready(function(){initNavTree('{{ page.fileName }}{% if page_postfix 
 {% block searchInfo %}
 {% if config.SEARCHENGINE and not config.SERVER_BASED_SEARCH %}
 <!-- window showing the filter options -->
-<div id="MSearchSelectWindow"
+<div id="MSearchSelectWindow tabindex="0""
      onmouseover="return searchBox.OnSearchSelectShow()"
      onmouseout="return searchBox.OnSearchSelectHide()"
      onkeydown="return searchBox.OnSearchSelectKey(event)">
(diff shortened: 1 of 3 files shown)


SergeyRyabinin:
Hi,

Could you please take a look at before/after?
[BEFORE](https://sdk.amazonaws.com/cpp/api/1.11.113/aws-cpp-sdk-s3/html/class_aws_1_1_s3_1_1_s3_client.html), [AFTER](https://sdk.amazonaws.com/cpp/api/1.11.114/aws-cpp-sdk-s3/html/class_aws_1_1_s3_1_1_s3_client.html).

The issue reported falls under "accessibility" category, the "loop" button next to search field is accessible only with a mouse.
With my change I'm trying to make it keyboard-accessible (with a sequence of: tabbing to loop button, enter, select search kind, enter/escape to close search kind selector window) . Could you please take a look at the after state and clarify if it something that could be accepted?
> Note that the ESC-key (key code 27) is also used for toggeling between light / dark mode (HTML_COLORSTYLE=TOGGLE)

I've also checked that colorstyle switch works well with my change.

if it is something that you would accept - I will try to make it HTML4-compatible and fix CI build issues reported.

Best regards,
Sergey

templates/html/htmlbase.tpl
```diff
@@ -217,7 +217,7 @@ $(document).ready(function(){initNavTree('{{ page.fileName }}{% if page_postfix
 {% block searchInfo %}
 {% if config.SEARCHENGINE and not config.SERVER_BASED_SEARCH %}
 <!-- window showing the filter options -->
-<div id="MSearchSelectWindow"
+<div id="MSearchSelectWindow tabindex="0""
```
albert-github, line 220: Looks like an erroneous double quote (`"`) at the end

albert-github:
Regarding the " I will try to make it HTML4-compatible and fix CI build issues reported" this should be done straight away (so the effects can bee better seen), it should be corrected into `testing/dtd/xhtml1-transitional.dtd`.
It can be tested by means of: `doxygen tests TEST_FLAGS="--xhtml"`
for other options of the target tests seed `doxygen tests TEST_FLAGS="--help"`

PR #10150, opened 2023-06-26
Fix #10147 `Generating pdf with LuaLateX`

flagarde:

PR #10171, opened 2023-07-05
Sync etoc_doxygen with upstream 1.2b release of 2023/07/01

jfbu:
The diff includes the suppression of the funny (here in escape notation) `^[[0` reported at https://github.com/doxygen/doxygen/pull/9936#issuecomment-1607633248, which was not in upstream file at line 1466.

@albert-github Should I rather make a patch which does not worry about the stray extra bytes at line 1466 of (current) file, which would be removed in a separate unrelated commit?  or do I keep it this way so that merging this also fixes that problem? 

edit: I had not seen #10150 which is a PR also removing the stray bytes in passing, linking to it from here now.

Is master the correct target branch?

The upstream file of which etoc_doxygen.sty is almost identical copy apart from changing the LaTeX name from etoc to etoc_doxygen was uploaded to CTAN only yesterday and may not have propagated to CTAN mirrors yet.

The  changes from 1.2a to 1.2b are very few,  some trimming of left-overs comments, the main change is at line 1839 (in new file) to fix compatibility with another LaTeX package (`tocloft`), as a regression was added upstream at 1.2  which removed the needed `\AtBeginDocument` now re-put in place there.

edit: I see that in file templatex/latex/header.tex doxygen there is `\usepackage[titles]{tocloft}`, so the issue with `tocloft` had remained hidden here because it arose only if `tocloft` was loaded *without* the `titles` option...(and before `etoc`, as here).

### DIFF
--- a/templates/latex/etoc_doxygen.sty
+++ b/templates/latex/etoc_doxygen.sty
@@ -7,7 +7,7 @@
 %%
 %%     etoc.sty
 %%
-%% at version 1.2a of 2023/05/01.
+%% at version 1.2b of 2023/07/01.
 %%
 %% This file has been provided to Doxygen team courtesy of the
 %% author for benefit of users having a LaTeX installation not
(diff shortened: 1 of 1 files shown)


albert-github:
Thanks for the patch
- good to have the doxygen version that is compatible with the version on CTAN 
- the stray bytes, they have to be removed and it is irrelevant where this happens (or whether this happens double, as long as it happens).
- the master branch is the correct branch (this is the only branch we maintain)

PR #10172, opened 2023-07-06
don't fill plain shaped root node

makelinux:
because filling is intended for record or box shaped nodes. Plain shaped root node is highlighted with bold style in function DotNode::writeLabel.

Before:
image

After:
image


albert-github:
The gray fill color is to indicate  a special feature (see the explanation with the legend graph).
> - A filled gray box represents the struct or class for which the graph is generated.
 
So when using this it would also be necessary to adjust the first line of text.

- I assume you explicitly set the `shape` to `plain`, what is the reason?
- In the graphviz documentation we see (https://graphviz.org/doc/info/shapes.html):

  > As the figures suggest, the shapes rect and rectangle are synonyms for box, and none is a synonym for plaintext. The shape plain is similar to these two, except that it also enforces width=0 height=0 margin=0, which guarantees that the actual size of the node is entirely determined by the label. This is useful, for example, when using [HTML-like labels](https://graphviz.org/doc/info/shapes.html#html). Also, unlike the rest, we have shown these three, as well as underline, without style=filled to indicate the normal use. If fill were turned on, the label text would appear in a filled rectangle.

  I don't see from this that it is not allowed to use the fill (it might not be intended but that doesn't mean it is not allowed).
- why is the "undocumented" box in your example still gray? Due to the "plain" change shouldn't this loose the gray filling as well? (due to the cut-off I cannot see what happens here with the "Truncated" filled box)

makelinux:
I've updated the patch and  the description.

Fill is allowed for shapes in Graphviz. Boxes can be removed for aesthetic purposes accordingly subjective taste with my accepted PR: #9463

makelinux:
@albert-github , would you like please to review?

src/translator_en.h
```diff
@@ -1039,7 +1039,8 @@ class TranslatorEnglish : public Translator
         "This will result in the following graph:"
         "<p><center><img alt=\"\" src=\"graph_legend."+getDotImageExtension()+"\"></center></p>\n"
         "<p>\n"
-        "The boxes in the above graph have the following meaning:\n"
+        "For the default configuration, "
+        "the boxes in the above graph have the following meaning:\n"
```
albert-github, line 1043: Does this mean that the "legend" is in the "old" stile even when using "shape=plain" and thus having an inconsistent "legend" with the images? 
The "legend" should follow the current status (there might also be some places already where the color description doesn't fit).

albert-github (changes requested):

albert-github:
I already expressed my  opinion before and now placed a, small, review as well.

src/translator_en.h
makelinux: Do you mean legend text? For default configuration the legend image is consistent with legend text. For another style configuration legend images are inconsistent with the hard-coded legend text.
What do you suggest?

albert-github: I think that the legend (and its texts) should express the the way the graphs are generated (i.e. colors, shapes and texts).

makelinux: We have contradiction between flexible graph generation rigid text description on many languages.

Hypothetically it is possible to translate arbitrary graph style to description to all languages with help of AI.

PR #10179, opened 2023-07-12
Adding a separator to ALIASES

albert-github:
In some cases it is very inconvenient to have a comma as argument separator in an alias (e.g. when using function definitions or templates) as all internal commas have to be escaped. By introducing a separator (like e.g. `;`) this can be made more convenient, easier to maintain and more understandable.

Also placed software for alias handling in a separate file.


doxygen:
I noticed there is some ambiguity, where the chosen alias definition depends on the character code of the separator (I guess because all separators for a given alias are placed in a set).

Consider these aliases:

```
ALIASES                = \
                         "Hello{3}=comma3: 1='\1' 2='\2' 3='\3'" \
                         "Hello{2,.}=dot2: 1='\1' 2='\2'" \
                         "World{3,&}=amp3: 1='\1' 2='\2' 3='\3'" \
                         "World{2,$}=dollar2: 1='\1' 2='\2'" \
```

and this code:
```c++
/** Test
 *
 * - \Hello{1,2,3} uses ,(char44) as separator, ok
 * - \Hello{1,2.3} uses .(char46) as separator, ok
 * - \Hello{1.2,3,4} uses ,(char44) as separator, but could have been &(char46) as well
 *
 * - \World{1&2&3} uses &(char38) as separator, ok
 * - \World{1&2$3} uses $(char36) as separator, ok
 * - \World{1$2&3&4} uses $(char36) as separator, but could have been &(char38) as well
 */
class Test {};
```

gives this output:

- comma3: 1='1' 2='2' 3='3' uses ,(char44) as separator, ok
- dot2: 1='1,2' 2='3' uses .(char46) as separator, ok
- comma3: 1='1.2' 2='3' 3='4' uses ,(char44) as separator, but could have been &(char46) as well
- amp3: 1='1' 2='2' 3='3' uses &(char38) as separator, ok
- dollar2: 1='1&2' 2='3' uses as separator, ok
- dollar2: 1='1' 2='2&3&4' uses as separator, but could have been &(char38) as well

A different choice is made for the 3th and 6th case, even though the definitions are identical, only difference is the chosen separator.

Note that for some reason it is not allowed to write `"Hello{3,,}=comma3: 1='\1' 2='\2' 3='\3'" \`

albert-github:
Thanks for the review.
- problem choosing of `ALIASES` in case of ambiguity. This is indeed the problem as indicated due to the usage of `set`, at the moment the first match is chosen. Actually there should be a warning message (also on other places) but I refrained from adding these at the moment as not to disrupt current working situations.
- the problem regarding the use of the `,` as separator, I thought this was handled by means of the function `void readAliases()` around line 120 in `[src/aliases.cpp](https://github.com/doxygen/doxygen/pull/10179/files#diff-25dc0815658fbf139db91a01aa86821f19a44d170b717b8b3cc50781008a0909) but it looks indeed that, probably, in the function `expandAliasRec` (around line 318) this is not handled properly, we probably need a way to find whether the `,` was explicitly given or not.

albert-github:
Added code for explicitly added `,`.

doc/custcmd.doc
```diff
@@ -107,6 +107,15 @@ i.e.
 \endverbatim
 given the alias definition of `\l` in the example above.
 
+By default the separator for arguments in an alias ist the comma though when having e.g templates of function
```
doxygen, line 110: typo: 'ist' -> 'is'

src/aliases.cpp
```diff
@@ -329,6 +329,19 @@ static QCString expandAliasRec(StringUnorderedSet &aliasesProcessed,const QCStri
             numArgs = numArgs1;
             cmd += arg;
           }
+          if ((sep1 == ',') && it==g_aliasMap.end() && it1==g_aliasMap.end())
+          {
+            sprintf(tmp,"{%d,%c}",numArgs1,sep1);
```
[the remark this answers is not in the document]
albert-github: probably better to use the `sprintf` from `QCString`, I'll look into it (tomorrow)

src/aliases.h
```diff
@@ -0,0 +1,25 @@
...
+ *
+ */
+
+#ifndef ALIASES_H
+#define ALIASES_H
+
+#include "qcstring.h"
+
+QCString resolveAliasCmd(const QCString &aliasCmd);
+std::string expandAlias(const std::string &aliasName,const std::string &aliasValue);
```
doxygen, line 22: Looks like `expandAlias` is only needed inside `aliases.cpp` so it can be removed here and made static in the implementation.

albert-github: Thanks, All fixed.

doxygen:
@albert-github I made a number of changes, see https://github.com/doxygen/doxygen/commit/6b20feb17b8b542149d1f524982745d3e90153ba. Please check if I didn't break anything in the process (I did test as good as I could).

albert-github:
It did break on CGAL quite a bit.

In CGAL we have the construct like:
```
                         "cgalRefines=Refines" \
                         "cgalRefines{1}=<dl><dt>@cgalRefines</dt><dd>@c \1</dd></dl>" \
                         "cgalRefines{2}=<dl><dt>@cgalRefines</dt><dd>@c \1</dd><dt></dt><dd>@c \2</dd></dl>" \
                         "cgalRefines{3}=<dl><dt>@cgalRefines</dt><dd>@c \1</dd><dt></dt><dd>@c \2</dd><dt></dt><dd>@c \3</dd></dl>" \
                         "cgalRefines{4}=<dl><dt>@cgalRefines</dt><dd>@c \1</dd><dt></dt><dd>@c \2</dd><dt></dt><dd>@c \3</dd><dt></dt><dd>@c \4</dd></dl>" \
                         "cgalRefines{5}=<dl><dt>@cgalRefines</dt><dd>@c \1</dd><dt></dt><dd>@c \2</dd><dt></dt><dd>@c \3</dd><dt></dt><dd>@c \4</dd><dt></dt><dd>@c \5</dd></dl>" \
                         "cgalRefinesBare{1}=<dl><dt>@cgalRefines</dt><dd>\1</dd></dl>" \
                         "cgalRefinesBare{2}=<dl><dt>@cgalRefines</dt><dd>@c \1</dd><dt></dt><dd>\2</dd></dl>" \
```
and this, now, causes problems as it looks like it is not allowed to have an `ALIASES` without arguments and one or more `ALIASES` with arguments. 
This is in principle easy to fix (by choosing a different name, but I think it should work).

There are also a number of warnings like:
```
alpha_wrap_3.h:70: warning: File ended in the middle of a comment block! Perhaps a missing \endcode?
```

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/12159305/example.tar.gz)

It looks like that in the `commentcnv` phase the `commentcnv` process is abruptly ended.
Looks like the
```
"cgalParamDefault{1}=<li><b>%Default: </b>\1</li>" \
```
when called as
```
\cgalParamDefault{a \cgal Kernel deduced from the point type, using `CGAL::Kernel_traits`}
```
so with a `,` gives problems and doesn't revert to the default of just 1 argument (as it did
This is in principle easy to fix (by choosing `;;` as separator, note the double `;` as there are places where there is a `;` in the alias, but I think it should work).

I didn't yet check further for other type of errors (looks like CGAL has no other at the moment).

albert-github:
I see that in the set of possible separators there are a.o. `"` and `` ` `` but no `'` any special reason?
In the set of possible separators there is `^` but can't this conflict with `^^` which is a replacement for the "inline newline"?

doxygen:
@albert-github I made a fix which should hopefully fix the CGAL issue you found, and I also reduced the set of separators and added single quote as allowed separator (see https://github.com/doxygen/doxygen/commit/08e889e30cd8594ad23391de36c684e514381315)

albert-github:
- I saw the reduction of the allowed separators, still a bit strange that e.g. `_` is missing and that now `"` is removed (although as this is used with e.g. `doxygen -x` and other string operations it is quite logical). I can live quiet well with the mentioned set (especially as multicharacter-separators and combinations of separators are possible)
- problem of the `cgalParamDefault{1}` looks like to be solved
- problem with the `cgalRefines=Refines` is still present and although for CGAL I have a workaround where I just have to change the `Doxyfile` I don't think this is really possible for the problem as indicated by @eseiler in #10193 as here we have the `myalias` in the source code as well (it might even be that this is a separate problem).

albert-github:
Regarding the coverity scan [problem, the assert is correct but apparently doesn't withhold coverity from emitting its warning in aliases.cpp

albert-github:
Reran the CGAL tests with the `1.9.8 (a1f16f361285205ec15b3f90f87dfa1f4f5092da)` version and this looks OK.
In the GitHub actions we see that the generation of the documentation crashes, it looks like that the:
```
ALIASES           = LaTeX="\f({\LaTeX}\f)"
ALIASES          += TeX="\f({\TeX}\f)"
```
cause the problems.
Only way I got it up and running, at the moment, was removing the `\` in `\LaTeX` and `\TeX` but this of course ruins the formulas.
To reproduce the problem it is sufficient to use the following example: [example.tar.gz](https://github.com/doxygen/doxygen/files/12192264/example.tar.gz)

albert-github:
Found the base cause of the problem, the `ALIASES` are (self) recursive, the alias name is `LaTeX` but inside there is also the command `\LaTeX`, so it will call itself again in `expandAliasRec`. 

Looks like the following patch functions:
```
diff --git a/src/aliases.cpp b/src/aliases.cpp
index 9945d92fd8..1059613cc9 100644
--- a/src/aliases.cpp
+++ b/src/aliases.cpp
@@ -451,13 +451,13 @@ static QCString expandAliasRec(StringUnorderedSet &aliasesProcessed,const QCStri
       //printf("Alias %s not found\n",qPrint(cmd));
     }
     //printf("Found command s='%s' cmd='%s' numArgs=%d args='%s'\n", qPrint(s),qPrint(cmd),selectedNumArgs,qPrint(args));
-    if ((allowRecursion || aliasesProcessed.find(cmd.str())==aliasesProcessed.end()) &&
+    std::string qualifiedName = cmd.str()+":"+std::to_string(selectedNumArgs);
+    if ((allowRecursion || aliasesProcessed.find(qualifiedName)==aliasesProcessed.end()) &&
         it!=g_aliasInfoMap.end() && selectedNumArgs!=-1 &&
         it->second.find(selectedNumArgs)!=it->second.end()) // expand the alias
     {
       const auto &aliasInfo = it->second.find(selectedNumArgs)->second;
       //printf("is an alias with separator='%s' hasArgs=%d!\n",qPrint(aliasInfo.separator),hasArgs);
-      std::string qualifiedName = cmd.str()+":"+std::to_string(selectedNumArgs);
       if (!allowRecursion) aliasesProcessed.insert(qualifiedName);
       std::string val = aliasInfo.value;
       if (hasArgs)
```

where the most important part is the test:
```
aliasesProcessed.find(qualifiedName)==aliasesProcessed.end()
```
instead of
```
aliasesProcessed.find(qualifiedName)==aliasesProcessed.end()
```

(also tested against CGAL).

albert-github:
Looks like there is also a problem when we have multiple `ALIASES` with the same number of arguments but a different separator like:
```
ALIASES += "A{1+}=\1 the plus replacement \2"
ALIASES += "A{1&}=\1 the ampersand replacement \2"
```
when having the example:
```
TEST
====

- `ALIASES += A{1+}="\1 the plus replacement \2"`
- `ALIASES += A{1&}="\1 the ampersand replacement \2"`


values:
- `\A{A,B}` \A{A,B}
- `\A{A+B}` \A{A+B}
- `\A{A&B}` \A{A&B}
- `\A{A!B}` \A{A!B}
```

we get:
image

- everything uses the ampersand replacement (explainable as we only have 1 argument so the last one is taken)
- we have a double text (but in fact `\2` does not exist, so what is used)

Note this is independent from the proposed change in https://github.com/doxygen/doxygen/pull/10179#issuecomment-1655369145

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/12206679/example.tar.gz)

albert-github:
Another small problem:
```
ALIASES += "A{2+}=\1 the plus replacement \2"
ALIASES += "A{2&}=\1 the ampersand replacement \2"
```
and
```
TEST
====

- `ALIASES += A{2+}="\1 the plus replacement \2"`
- `ALIASES += A{2&}="\1 the ampersand replacement \2"`


values:
- `\A{A,B}` \A{A,B}
- `\A{A+B}` \A{A+B}
- `\A{A&B}` \A{A&B}
- `\A{A!B}` \A{A!B}
```
gives some warnings:
```
bb.md:9: warning: Found unexpanded alias '\A'. Check if number of arguments passed is correct.
bb.md:10: warning: Found unexpanded alias '\A'. Check if number of arguments passed is correct.
bb.md:12: warning: Found unexpanded alias '\A'. Check if number of arguments passed is correct.
```

and output:
image



Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/12206682/example.tar.gz)

doxygen:
@albert The problem with detected recursive aliases is solved in the way you suggested.

The other issues are as expected. When defining an alias with the same number of parameters as an earlier alias definition, the alias is replaced (independent of the separators used). So only the last definition is used. This was also the original behavior before separators were introduced, and is also needed to avoid ambiguity.

albert-github:
I can imagine that it is consistent with the old situation, though I think that the example as given in https://github.com/doxygen/doxygen/pull/10179#issuecomment-1656631045 is not ambiguous and probably could be solved by means of adding the separator to the `qualifiedName` before inserting it in the `aliasesProcessed`.

I constructed the mentioned example and have not yet seen a real live example where it could be used but there might be examples when there are 2 teams working on different libraries that use a different separator and when merging the libraries these definitions would now conflict. Probably there will be other issues at that moment as well and this one would be easily solved (but maybe not detected immediately when there are a lot of ALIASES).

PR #10185, opened 2023-07-18
Fix latex quotes in <pre>

mtc-mlx:
Commit 2073b3b (#9978) removed `\char` but left its backslash. `\char` should not be necessary since fontenc is used. If fontenc is not used the curly braces prevent interpretation as an umlaut (although the quote will be curled the wrong way).

## Current output:
image

## With this patch:
image

## Notes

If I understand correctly: before 2073b3b, the output was ``\char`\"``, which worked because `\"` is treated as an argument to `` ` `` instead of a command. Since the ``\char` `` part is now removed, `\"` is treated as the babel command for an umlaut.

I'm still a little confused, since babel is not used when building the Doxygen manual as far as I can tell.

On our internal CI that uses Doxygen + LuaLaTeX + Polyglossia, `\"` generates tofu, but on my local setup it generates an umlaut just like in the doxygen manual.

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -2364,7 +2364,7 @@ void filterLatexString(TextStream &t,const QCString &str,
                    break;
         case '\\': t << "\\textbackslash{}";
                    break;
-        case '"':  t << "\\\"{}";
+        case '"':  t << "\"{}";
                    break;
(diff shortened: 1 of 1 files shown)


albert-github:
Well spotted.

PR #10208, opened 2023-08-03
Possibility to exclude tests by number

albert-github:
When selecting all or a range of tests it sometimes is useful to disable one or more tests , for this purpose the option `--exclude_id` is introduced.


testing/runtests.py
```diff
@@ -697,6 +698,18 @@ def main():
         tests = sorted(glob.glob('[0-9][0-9][0-9]_*'))
     else:
         tests = list(itertools.chain.from_iterable(tests))
+
+    if args.exclude_ids:  # test ids are given by user
+        for id in list(itertools.chain.from_iterable(args.exclude_ids)):
+            x=glob.glob('%s_*'%id)
```
doxygen, line 704: I think it is better to do `x=glob.glob('{:03d}_*'.format(id))` here instead of 3 different tests, or doesn't this work?

albert-github: I wanted it to be consistent with the other code that was already present. the line like:
```
print(glob.glob('%03d_*'%int(id))`
```
works in Python 2 and 3 so would indeed make the code a bit smaller.
I'll update the PR.

albert-github: Updated.

Issue #10209, opened 2023-08-03
UML diagrams do not honor HIDE_UNDOC_MEMBERS


sjoubert:
**Describe the bug**
Having an undocumented method in a class and HIDE_UNDOC_MEMBERS does not prevent the method to show up in the class/collaboration diagram of the class (while the method is correctly absent from the rest of the documentation).

**Expected behavior**
UML diagrams should honor HIDE_UNDOC_MEMBERS

**Screenshots**
![image](https://github.com/doxygen/doxygen/assets/200310/204b9042-ca78-4e75-b0d7-2a5590b2f01b)

**To Reproduce**
```
EXTRACT_ALL = NO
HIDE_UNDOC_MEMBERS = YES
COLLABORATION_GRAPH = YES
UML_LOOK = YES
HAVE_DOT = YES
GENERATE_HTML = YES
SOURCES = Cat.hpp
```
```
/// Soft kitty, warm kitty
struct Cat
{
  /// Purr, purr, purr
  void purr();
  // Bad kitty!
  void scratch_everywhere();
};
```

**Version**
Tested with the latest Linux binary (1.9.7) from Github as well as version 1.9.4 from Debian testing.
```sh
$ doxygen --version
1.9.7 (a2f34e1ed90f83713dc347b5134920f32c455d46)
```

albert-github:
I've just pushed a proposed patch, pull request #10210

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.8.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10210, opened 2023-08-03
issue #10209 UML diagrams do not honor HIDE_UNDOC_MEMBERS

albert-github:
In case `HIDE_UNDOC_MEMBERS` is set hide undocumented members also in UML diagrams.


sjoubert:
Just tested the build artifacts, it works well, thank you.

Issue #4111, opened 2018-07-02
full indexes are missing (Origin: bugzilla #640619)


doxygen:
status NEW severity *enhancement* in component *general* for *---*
Reported in version *unspecified* on platform *Other*
Assigned to: Dimitri van Heesch

On 2011-01-26 08:04:06 +0000, bugzilla.gnome.org@poelzi.org wrote:

    Currently doxygen generates index pages with prefix paginators only and also
    scatters them on many subpages. This makes it extremely hard to find
    signatures of functions partially known or to search for a keyword in the
    name.
    You need to know a lib quite good to find documentation and that is not very
    helpful.
    
    I personally find doxygen docs more or less useless as they don't generate a
    full index like the glib doc generator.
    
    Example of a good full index:
    http://library.gnome.org/devel/glib/2.26/api-index-full.html

apolukhin:
Looks like a redy-to-use solution is now available https://lunrjs.com/

The API is quite close to the current search engine

apolukhin:
There's a ready to use solution that does not require third party languages and tools https://github.com/divideconcept/DoxygenDeepSearch

albert-github:
Code of #10218 has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.9.8.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10218, opened 2023-08-07
Implement search by each significant word from page caption

apolukhin:
This patch does the following:
* Makes the `g_searchIndexInfo` value store `lowercase(searchName) -> definitions relations` rather than `lowercase(searchName[0]) -> definitions relations`
* When the JL index is being formed, the `g_searchIndexInfo` is mapped to `searchName[0]` to repeat the old logic
* `splitSearchTokens` splits the titles into multiple words and adds each word into `g_searchIndexInfo`
* `writeJavasScriptSearchDataPage` now iterates over `search id`, that simplifies the function logic

Addresses https://github.com/doxygen/doxygen/issues/4111


apolukhin:
> looks like a lowercase conversion is gone?

It is now in the `SearchIndexInfo::add` function. Investigating the crash

apolukhin:
Found and fixed some issue with multithreaded runs. Still testing the fix. Meanwhile, I'm open to any requests/sugestions/comments/...

apolukhin:
`make docs_internal` finished well with the fix:

image

Our docs were also built well:

image_2023-08-07_20-39-01

albert-github:
Did a few other tests on the doxygen internal documentation/ When searching for `intern` we get:
image

note the `%` sign (which is present to prevent linking to the `Doxygen` class).
Searching for `doxygen` / `Doxygen` didn't give a result (regarding this header).
Searching for `%` gave
image
but this was also the case in the doxygen versions before the changes.

@doxygen
- shouldn't actually the results of the % search be omitted?
- shouldn't the results in the search panel be "filtered"?

I created a small project where I also experimented a bit with the html entities (`&...;`) and in the search panel also the html entities are shown (search for `load`): [example.tar.gz](https://github.com/doxygen/doxygen/files/12289328/example.tar.gz)

albert-github:
Looks like I found another problem.

When searching for `%doxygen in` before the changes I got:
image

and after the changes I get:
image

Here it is not so spectacular, but when searching for e.g. `overview of` it is even clearer.



I see also a positive change (at least the one that I found):
Old:
image

New:
image

where the `number` are merged

apolukhin:
Fixed search by multiple words in 45d2a9c08e64942419a9834e769042e3bcd47e60. Here's a playground if someone needs one https://userver.tech/

The `%` searches I'd rather fix in a separate PR later. IMO this one is quite big already

albert-github:
You wrote:
> IMO this one is quite big already

it is not that big, e.g. https://github.com/doxygen/doxygen/pull/9825 is bigger.


I'm not totally getting the:
> The % searches I'd rather fix in a separate PR later.

I see already a number of fixes regarding the `%` or did you mean the `%` in the search panel (with this I would agree as it is a more separate issue, and would probably also solve issues with the html entities `&...;`)

apolukhin:
> I see already a number of fixes regarding the `% or` did you mean the % in the search panel

Yes, I've experimented with `%` but removed all the `%` related code for now.

I could restore the `%` related changes if the following behavior is fine:
* the `%doxygen` search would stop working, only 'doxygen'* wold do the job. The search result would still contain the `%Doxygen internals`
* for the title `some1 %some2`
  * the `some1 some2` search would work
  * the `some1 %some2` search *not* work

albert-github:
I think that goal should be that entering in the search box the text as seen on the pages it will show up. That, temporarily, the e.g. `%` is shown in the search results is not nice, but this happens now as well and can be handled later.

All in all it remains a difficult topic.

apolukhin:
Seems that I've fixed the `%` issue:

image

`%` in the title is kept if it does not followed by word. For `\page pg_cmds_int Overview % of introduction of internal special commands` doxygen still generates:

image

apolukhin:
The `macOS Latest Release` CI run failed because of some network issue

apolukhin:
CI passed well.

Is there anything I can do to speedup the review process and to get this PR merged for the next release of doxygen?

apolukhin:
Something wrong with certificates in MacOS CI. The other tests pass well

doxygen:
@apolukhin I decided to add section titles as well and did some performance improvements/restructuring (i.e. directly store the results in the right format so `getSearchIndexMapByLetter` is no longer needed and optimize the `splitSearchTokens` function. See https://github.com/doxygen/doxygen/commit/42391fce4676eeffc64f5836c66db359355fe61b and check if I didn't accidentally break something that was working in the changes you made.

apolukhin:
Looks great!

Noted a small typo the prevents searching for a word after some special chars. See https://github.com/doxygen/doxygen/pull/10243

PR #10224, opened 2023-08-10
Return correct updated message for translators

albert-github:
Return correct updated message for translators: Czech, Spanish, Greek and Swedish


albert-github:
Obsolete, fix is include in 6d80fc7e5d03c259b1a7280972e0b28884217655  (Add support for C++20 modules)

PR #10225, opened 2023-08-10
Show not up to date message  for translator

albert-github:
In:
```
Commit: 4c47b238edf4f94894ff70f6b32904e142f159a7 [4c47b23]
Date: Saturday, February 26, 2022 10:44:11 PM
Downgrade stale translator warning to message
```
The text was downgraded to a message and in this way dependent on the setting of `QUIET` whilst other configuration settings are given. Also the text was written to `stdout` instead of `stderr`.

### DIFF
--- a/src/language.cpp
+++ b/src/language.cpp
@@ -69,6 +69,8 @@
 
 Translator *theTranslator=0;
 
+void config_warn(const char *fmt, ...);
+
 void setTranslator(OUTPUT_LANGUAGE_t langName)
 {
   switch (langName)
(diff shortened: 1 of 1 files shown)


src/language.cpp
```diff
@@ -69,6 +69,8 @@
 
 Translator *theTranslator=0;
 
+void config_warn(const char *fmt, ...);
```
doxygen, line 72: a bit tricky to depend on internal functions of `config_impl.l`. Maybe better to add a `err_non_fatal`? (like `err_full` is now, but maybe that should have had a `handle_warn_as_error()` as well?)

albert-github: It is indeed a bit tricky.
I thought about something like `err_non_fatal` as well but I decided to do it this way as I wanted the message to have the config look and feel as well as the place (so not into the warnings log file). Maybe a solution would be to have a separate include file or something like it.

doxygen:
@albert-github I now added the warning functions to the ConfigImpl class as static members.

albert-github:
Looks OK, thanks.

PR #10226, opened 2023-08-12
More flexible include graphs regarding show / hide

albert-github:
For the `INCLUDE_GRAPH` and the `INCLUDED_BY_GRAPH` it is possible to get a warning like:
```
warning: Included by graph for 'includes.h' not generated, too many nodes (268), threshold is 250. Consider increasing DOT_GRAPH_MAX_NODES.
```

This warning is quite disruptive especially in case `WARN_AS_ERROR` is set. The only way to prevent the warning and its consequences is to increase the `DOT_GRAPH_MAX_NODES` meaning there there might be some very large graphs (also at other places) or the set `INCLUDED_BY_GRAPH=NO` meaning, potential, loosing quite a few useful graphs. For call / caller graphs it is possible to steer the graph creation process on a 1 to 1 base (commands like `\callgraph` and `\hidecallgraph`).

Introducing for the include graphs:
- `\includegraph` and `\hideincludegraph`
- `\includedbygraph` and `\hideincludedbygraph`

Note: we needed  to set in the `layout_default.xml`:
```
    <includegraph visible="yes"/>
    <includedbygraph visible="yes"/>
```
as otherwise `\includegraph`and `\includedbygraph` won't work.

Basic example: [example.tar.gz](https://github.com/doxygen/doxygen/files/12327074/example.tar.gz)


albert-github:
https://github.com/doxygen/doxygen/pull/10226#issuecomment-1683782710 That would mean introducing a new category of warnings and I think the `WARN_AS_ERROR` is just for this case to signal that something in a build is not 100% clean and that the user might want to take some action. I think that by means of the extra commands (although again extra commands ....) this problem can elegantly be solved by the user.

PR #10264, opened 2023-08-27
add a use_sys_sqlite3 config option

### LINKED ISSUE #10263: please add a use_sys_sqlite3 config option
While working at packaging 1.9.8 for Debian, I noticed that you have vendored (bundled) spdlog and sqlite3 in `deps/` dir.

There dependencies are available on all platforms supported by Debian, so this is not required for us (but I understand your point of course).

Debian is against vendoring so we'll repackage doxygen stripping these source files (that we already have in the corresponding packages), so for us the source tree has the dirs `deps/spdlog` and `deps/sqlite3` missing.

With `use_sys_spdlog` config option I can tell CMake to skip the former, but it errors out because it still lacks the latter.

So for us it would be really helpful if you could add a `use_sys_sqlite3` option and patch `deps/CMakeLists.txt` so:
```diff
 if (NOT use_sys_spdlog)
 	add_subdirectory(spdlog)
 endif()
+if (NOT use_sys_sqlite3)
+	add_subdirectory(sqlite3)
+endif()
-add_subdirectory(sqlite3)
```

simevo:
fixes #10263


albert-github:
See also comments with the issue #10263

doxygen:
see https://github.com/doxygen/doxygen/commit/28609fecd41d885e54fa170e499a0e5b55def2c2 for a more complete fix

albert-github:
Pull request is obsolete by commit [28609fe](https://github.com/doxygen/doxygen/commit/28609fecd41d885e54fa170e499a0e5b55def2c2) and the proposed Pull Requests #10266.

PR #10285, opened 2023-09-05
bug_692887 groups display empty Detail Description

albert-github:
As indicated in the issue the `documentation()` is not empty (it contains just white space), stripping this out fixes the problem. (done also for the other parts, to be on the safe side)


albert-github:
Example based on #10335  [example.tar.gz](https://github.com/doxygen/doxygen/files/12777232/example.tar.gz)

PR #10286, opened 2023-09-05
add bibtex to testing readme as dependency

pyrophorol:
building doxygen before graphical environment, so no tex at this point
with 1.9.6 and 1.9.7 test 012 fails, as far as I can tell it is because I don't have bibtex

### DIFF
--- a/testing/README.txt
+++ b/testing/README.txt
@@ -52,6 +52,8 @@ The runtests.py has the following dependencies on 3rd party tools:
 - python  to run the script
 - xmllint to normalize the XML output
 - diff    to show the differences in case a test fails
+- LaTeX   for test 12 and when testing with --pdf flag (with adequate packages)
+- perl    for test 12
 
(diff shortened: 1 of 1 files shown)


albert-github:
Indeed `bibtex` is needed but this is normally part of the LaTeX distribution, and I don't think just adding `bibtex` would be sufficient. Furthermore it should be mentioned that a LaTeX distribution with the adequate packages (shouldn't be listed) is needed when using the `--pdf` possibility of the test procedure.
For the non `--pdf` tests` it is also necessary to have `perl` installed as test 12 runs a perl script.

As a side note I see I forgot to add the description of the new option, see proposed pull request #10288, to the readme file.

pyrophorol:
thanks for the addition of --exclude_id, that is exactly one thing I came to the readme looking to find; I'll use that once I update to 1.9.8

anyway I modified what I wrote; but feel free to rewrite it if you don't like how I wrote it; whatever is easiest for you

PR #10287, opened 2023-09-05
legend: fix description of documented items

makelinux:
for configurations:
- DOT_NODE_ATTR = shape=plaintext
- DOT_NODE_ATTR = shape=plain

Preview of  DOT_NODE_ATTR = shape=plaintext:
image


src/translator_en.h
```diff
@@ -1039,12 +1039,14 @@ class TranslatorEnglish : public Translator
...
-        "The boxes in the above graph have the following meaning:\n"
+        "The nodes in the above graph have the following meaning:\n"
         "</p>\n"
         "<ul>\n"
         "<li>%A filled gray box represents the struct or class for which the "
         "graph is generated.</li>\n"
-        "<li>%A box with a black border denotes a documented struct or class.</li>\n"
+        + (Config_getString(DOT_NODE_ATTR).contains("shape=plain")?
+          "<li>Name in regular font ":"<li>%A box with a black border ") +
+        "denotes a documented struct or class.</li>\n"
```
albert-github, line 1049: Line 1042 you changed "boxes" to "nodes" but in line 1048 you used "box" again, also in the original text there are more occurrences of "box".
I also think that "node" / "nodes" is a bit an abstract term for the users that read the the diagram whilst wit the word "box" / "boxes" they straight away have a visual feedback.
(It is always difficult to get the right terms).

makelinux: A text "A box with a black border" is the old text from previous line 1047. It is printed for default configuration "shape=box".
The patch follows rule: "Preserve the default behaviour".
The patch writes "Name in regular font" only when node configuration is "shape=plain" or "shape=plaintext".
The rest of nodes are represented with filled boxes anytime. For "shape=plain" without border.

PR #10290, opened 2023-09-06
Warning about internal doxygen inconsistency for python packages

albert-github:
In the Ansible package (through Fossies) we get warnings like:
```
.../lib/ansible/module_utils/facts/virtual/base.py:27: warning: Internal inconsistency: scope for class ansible::module_utils::facts::virtual ::base::Virtual not found!
```
Due to the naming of python the classes in doxygen it is possible that those names will contain reserved names as the path is used in the identification of the class (here `virtual`). When a `:` is in front of the word it is a class name and no space should be appended.


albert-github:
Seen the problems as shown in #10304 
I've just pushed a proposed patch, pull request #10308 which reverts #10290

albert-github:
This PR has been reverted by means of #10308

Issue #10292, opened 2023-09-08
Not generate png for rtf automatically


zeerd:
**Describe the bug**

RTF(By Microsoft Word 2016) could not use a SVG picture.
I am drawing some pictures with PlantUML and DOT and want SVG under html document.
So, I set DOT_IMAGE_FORMAT to svg.

When I want both html and rtf, the pictures of PlantUML will generate twice. SVG for html and PNG for rtf.
But, pictures of DOT generate SVG only once for both html and rtf which rtf could not use.

**Expected behavior**

Could we generate PNG for rtf when DOT_IMAGE_FORMAT is svg?

**Screenshots**

```
Running plantuml with JAVA...
Generating PlantUML png Files in rtf
Generating PlantUML svg Files in html
Running dot...
Patching output file 1/1
type lookup cache used 71/65536 hits=104 misses=71
symbol lookup cache used 80/65536 hits=122 misses=80
finished...
```

**To Reproduce**

```
/**
@startuml
a -> b : c
@enduml
@dot
digraph A { a }
@enddot
*/
```

**Version**

```
1.10.0 (a3cba2660e0464c4a584377ccd80f5830add7277)
```

```
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.6 LTS
Release:	20.04
Codename:	focal
```

**Stack trace**

none

**Additional context**

none

albert-github:
@zeerd 

As you closed the proposed PR #10293 
- is the issue mentioned here still an issue with current Microsoft Word versions? (I don't have a recent version of Microsoft Word so I cannot test it).

zeerd:
I don't use `doxygen` and `Word` for a long time. Just cleaning old repos. If there is no others report it, might be not a problem any more.

albert-github:
@zeerd thanks for the answer

Well without testing it is of course hard to tell.

PR #10293, opened 2023-09-08
force to use png for rtf when DOT_IMAGE_FORMAT is svg

zeerd:
For #10292

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -237,15 +237,21 @@ void writeDotGraphFromFile(const QCString &inFile,const QCString &outDir,
     term("Output dir %s does not exist!\n",qPrint(outDir));
   }
 
+  QCString jobFormat = Config_getEnumAsString(DOT_IMAGE_FORMAT);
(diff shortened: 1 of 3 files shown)


albert-github:
The problem does not only occur with plantuml images but also with other images that are generated in svg format (e.g. call / caller graphs) and as far as I can tell they are not included here.
I think it would be better to see whether there is a way to include svg images natively into rtf documents.
Note that when looking with OpenOffice I do see the svg image (I just tested it with a call graph)

zeerd:
I am using the Microsoft Word(2016) which could not include svg.
But, OpenOffice is a good idea that I missed.

I have no OpenOffice and tried libreOffice(6.4.7.2) which could not include svg, too.

And I find that Microsoft Word would zoom large png pictures out to suit the page, but libreOffice will not.

Seems there is not an easy solution for anything, sigh .

zeerd:
I am not sure how to make `call/caller graphs`, but `inherit_graph` and `coll_graph` for c++ could swap between png and svg by the new tag now.

PR #10297, opened 2023-09-11
Incorrect showing of hidden call graph

albert-github:
When having:
** aa.cpp **
```
/// \file

void fie() {}

void fie_bare() {fie();}

void fie_hide() {fie();}

void fie_show() {fie();}
```

and ** aa.h **
```
/// the called fie
void fie();

/// the fie
void fie_bare();

/// the hide fie
/// \hidecallgraph
void fie_hide();

/// the show fie
/// \callgraph
void fie_show();
```
with
```
HAVE_DOT = YES
CALL_GRAPH = YES
```
This will lead to the fact that `fie_hide` still shows a call graph although it is explicitly disabled. (analogous for caller graph, referenced by relation and references relation).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/12575464/example.tar.gz)

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -293,9 +293,13 @@ class MemberDefImpl : public DefinitionMixin<MemberDefMutable>
     virtual void setFromAnonymousScope(bool b) override;
     virtual void setFromAnonymousMember(MemberDef *m) override;
     virtual void enableCallGraph(bool e) override;
+    virtual void enableCallGraph(bool e1, bool e2) override;
     virtual void enableCallerGraph(bool e) override;
+    virtual void enableCallerGraph(bool e1, bool e2) override;
     virtual void enableReferencedByRelation(bool e) override;
+    virtual void enableReferencedByRelation(bool e1, bool e2) override;
     virtual void enableReferencesRelation(bool e) override;
+    virtual void enableReferencesRelation(bool e1, bool e2) override;
     virtual void setTemplateMaster(MemberDef *mt) override;
     virtual void addListReference(Definition *d) override;
     virtual void setDocsForDefinition(bool b) override;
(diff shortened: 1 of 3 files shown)


src/memberdef.cpp
```diff
@@ -4810,24 +4814,52 @@ void MemberDefImpl::enableCallGraph(bool e)
   if (e) Doxygen::parseSourcesNeeded = TRUE;
 }
 
+void MemberDefImpl::enableCallGraph(bool e1, bool e2)
+{
+  if (e1 == e2) m_hasCallGraph=e1;
+  else m_hasCallGraph=!Config_getBool(CALL_GRAPH);
```
doxygen, line 4820: is the `!` here correct? Feels counterintuitive to assign the negation of `CALL_GRAPH` to `m_hasCallGraph`

albert-github: Yes it is correct, at startup the value of `m_hasCallGraph` (actually through `callGraph` in the `Entry` class) is `Config_getBool(CALL_GRAPH)` as here `e1` and `e2` differ it means that a setting contrary to `Config_getBool(CALL_GRAPH)` is used and thus should here `!Config_getBool(CALL_GRAPH)` be used.

doxygen: Ok, I see the logic now, I think it may be clearer to replace the `enableCallGraph(bool,bool)` with a `mergeEnableCallGraph(bool)`, like so:

```c++
    void mergeEnableCallGraph(bool other)
    {
      if (Config_getBool(CALL_GRAPH))
      {
        enableCallGraph(m_hasCallGraph && other); // enabled if neither deviate from config value
      }
      else
      {
        enableCallGraph(m_hasCallGraph || other); // enabled if either deviate from config value
      }
    }
```

albert-github: Makes sense. When starting the change for the the `||`, the easiest was to change from 1 to 2 arguments and place the logic in the function. When looking at the code it shows that always am other value is combined with the current value and so a "merge" is more appropriate.
Regarding the the logic in the merge function multiple formulations are possible and all of them have their typical drawback in interpreting them.

I've updated the proposed PR

PR #10303, opened 2023-09-12
More flexible inline sources regarding show / hide

albert-github:
For different graphs graphs it is possible to steer the creation process on a 1 to 1 base (commands like \callgraph and \hidecallgraph).

Introducing for the inline sources:
```
    \showinlinesource and \hideinlinesource
```

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/12587279/example.tar.gz)


albert-github:
Based on review for #10297 adjusted code here as well

PR #10317, opened 2023-09-19
Add missing C++20 module file extension

sjdalessandro:
Add missing file extension for Clang importable module units: `.ccm`. See [file name requirement](https://clang.llvm.org/docs/StandardCPlusPlusModules.html#file-name-requirement) in Clang documentation.


albert-github:
Also at least the documentation file doc.starting.doc should be adjusted, especially the table in the the paragraph "Step 1: Creating a configuration file", it looks like that when implementing the C++modules with the extension `.cppm`, `.cxxm` and `c++m` the table hasn't been updated either, should probably also still be done (@doxygen?).

PR #10321, opened 2023-09-21
Casting of enum in types.h

albert-github:
Replace the `static_cast` in the return of `ETYPE` define
- setting type for `enum` to `int`
- remove `static _cast`s
- setting mask `CategoryMask`to `0x0FFF0000` as only 5 bits are currently used and otherwise one can get: ``` warning C4369: 'EntryType::CategoryMask':  enumerator value '4294901760' cannot be represented as 'int', value is '-65536' ```

### DIFF
--- a/src/types.h
+++ b/src/types.h
@@ -520,8 +520,8 @@ enum class VhdlSpecifier
 class EntryType
 {
   public:
-#define ETYPE(x,bits)                                                                            \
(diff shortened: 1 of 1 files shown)


albert-github:
Looks like this does not result in a better solution (warnings with the CLang compilere are back).

PR #10323, opened 2023-09-23
Prevent reading tag file multiple times

albert-github:
Prevent the reading of a tag file multiple times (as already done for e.g INPUT files) as this might lead to warnings like (example from CGAL):
```
.../doc_tags/BGL.tag:2569: warning: Duplicate anchor fig__regularization_fig found
```

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -9797,7 +9797,7 @@ static void readTagFile(const std::shared_ptr<Entry> &root,const QCString &tagLi
     return;
   }
 
-  if (std::find(Doxygen::tagFileSet.begin(), Doxygen::tagFileSet.end(), fi.absFilePath().c_str()) != Doxygen::tagFileSet.end()) return;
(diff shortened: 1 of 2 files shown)


src/doxygen.cpp
```diff
@@ -145,6 +145,7 @@ FileNameLinkedMap    *Doxygen::mscFileNameLinkedMap = 0;     // msc files
 FileNameLinkedMap    *Doxygen::diaFileNameLinkedMap = 0;     // dia files
 StringUnorderedMap    Doxygen::namespaceAliasMap;            // all namespace aliases
 StringMap             Doxygen::tagDestinationMap;            // all tag locations
+std::vector<QCString> Doxygen::tagFileVector;                // all tag file names
```
doxygen, line 148: a `std::set` is more efficient than a `std::vector` for checking if something is already processed (O(log(N)) vs O(N)).

albert-github: I don't think this will be noticeable as typical the number of tag files will be small (with CGAL I saw one subproject with 113 tag files and all other had less than 20). Also the set of tag files is just read once.

Adjusted the code from `std::vector` to `std::set`

PR #10328, opened 2023-09-26
Publish doxmlparser on PyPI

edersondisouza:
In order to make `doxmlparser` readily available for users, it's important to publish it - being a Python module - to PyPI. This way, it can be included on `requirements.txt` files for Python dependencies.

This PR adds a new github job that, when there's a new relase of doxygen, publishes `doxmlparser` to PyPI. As I'm not a doxygen maintainer, this PR is incomplete, so I'm keeping it as draft now.

What is missing is the `PyPI API token`, that, once created, needs to be added to this repository as a secret.

It would be awesome if a doxygen maintainer could create this token. It should be simple, just a matter of creating an account on [PyPI](https://pypi.org/account/register/), create the API token and adding it as secret named `PYPI_API_TOKEN` to this repository.

Indeed, to test this package, I've created a package for TestPyPI from my fork - one can see it [here](https://test.pypi.org/project/doxmlparser/). I'll remove it once, hopefully, this PR is merged.

I'm ready to help if there's any issues, or look for alternatives to get the PyPI token.

### DIFF
--- a/.github/workflows/doxmlparser-publish.yml
+++ b/.github/workflows/doxmlparser-publish.yml
@@ -1,0 +1,34 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.x'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+
+    - name: Build package
+      run: python -m build -w addon/doxmlparser/
+
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        packages-dir: addon/doxmlparser/dist/
+        user: __token__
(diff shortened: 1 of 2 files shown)


albert-github:
As the doxmlparser is dependent on the doxygen version for which it is generated
- I don't think the results of each update of the doxygen master version should be pushed to PyPI, but only the results of a release
- will there be a history of, future, versions in PyPI so that the user can choose the right version (often the newest version will do but sometimes one needs an older version, due to restructuring, new mandatory items, removed items ... ).

Please remove the automatic  closing link, which is placed there automatically by GitHub because `` has been written in the PR text (doxygen will take care on closing issues when the PR is integrated at the moment of an official release):
image

doxygen:
I've created the PyPi account and set the token. Anything else needed before this can be merged?

edersondisouza:
> I've created the PyPi account and set the token. Anything else needed before this can be merged?
> 
> It would be good to publish the `doxmlparser` package under the same version as doxygen, so it is clear which version it is compatible with (the version is stored in a `VERSION` file in the root of the repo). I guess that still needs to be tweaked into the `setup.py` script?

That makes sense, will update this PR.

edersondisouza:
v2:
  - Use doxygen VERSION as doxmlparser version.

Issue #10306, opened 2023-09-13
Some intra-page links include the HTML file name


pawelszramowski:
**Describe the bug**
I observed that intra-page links (links to elements on the same page) generated by [`\ref`](https://www.doxygen.nl/manual/commands.html#cmdref) include the HTML file name.

For example, in [page1.html](https://www.doxygen.nl/manual/examples/page/html/page1.html) generated from [this example](https://www.doxygen.nl/manual/commands.html#cmdpage), intra-page links created by [`\tableofcontents`](https://www.doxygen.nl/manual/commands.html#cmdtableofcontents) are correct:

```html
<li class="level2"><a href="#subsection1">The first subsection</a></li><li class="level2"><a href="#subsection2">The second subsection</a></li>
```

whereas the links generated by `\ref` unnecessarily start with `page1.html`:

```html
This page contains the subsections <a class="el" href="page1.html#subsection1">The first subsection</a> and <a class="el" href="page1.html#subsection2">The second subsection</a>.
```

Intra-page links should not include the file name because:
1. It is not necessary for the links to work (links are relative to the current page).
2. It creates a dependency between the file name and content (if the file is renamed then the file content is no longer correct; not that it makes much sense to rename a single documentation page).
3. When the HTML page is printed to PDF in some web browsers, the links in the PDF file lead to elements within the original HTML file rather than within the PDF file. Mozilla Firefox is affected while Chromium-based Microsoft Edge works correctly.

**Expected behavior**
The HTML file name is not included in intra-page links, e.g. the aforementioned example is changed to:

```html
This page contains the subsections <a class="el" href="#subsection1">The first subsection</a> and <a class="el" href="#subsection2">The second subsection</a>.
```

**Screenshots**
None

**To Reproduce**
See bug description

**Version**
All?

**Stack trace**
N/A

**Additional context**
None

albert-github:
For the test I used the current (1.9.8) doxygen version, an empty Doxyfile and the example as indicated: [example.tar.gz](https://github.com/doxygen/doxygen/files/12596608/example.tar.gz)

1. it is indeed not necessary in this case but due to the general setup (of e.g. the `\ref` command) it is, probably, hard to prevent / quite a bit of work. So this would be an enhancement
2. It indeed is a problem when renaming a file, but this does not make sense and when referring from other files these links would be broken anyway. 
[...]

PR #10338, opened 2023-10-02
issue #10306 Some intra-page links include the HTML file name

albert-github:
In case the current filename to write to and the filename in the link (and the link has an anchor) don't write out the filename in the link. One has to be careful with external references to tag files.

(Checked the results against the doxygen internal documentation and against CGAL)

### DIFF
--- a/src/htmldocvisitor.cpp
+++ b/src/htmldocvisitor.cpp
@@ -280,8 +280,8 @@ static QCString htmlAttribsToString(const HtmlAttribList &attribs, QCString *pAl
 //-------------------------------------------------------------------------
 
 HtmlDocVisitor::HtmlDocVisitor(TextStream &t,OutputCodeList &ci,
-                               const Definition *ctx)
-  : m_t(t), m_ci(ci), m_ctx(ctx)
+                               const Definition *ctx,const QCString &fn)
+  : m_t(t), m_ci(ci), m_ctx(ctx), m_fileName(fn)
 {
   if (ctx) m_langExt=ctx->getDefFileExtension();
 }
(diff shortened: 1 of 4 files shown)


doxygen:
@albert-github I've factored out the link construction code into a `createHtmlUrl()` function to avoid code duplication. See https://github.com/doxygen/doxygen/commit/66e3e94112ed9814d575f16cf4e41aa4d06cb7df and let me know if I overlooked anything.

albert-github:
@doxygen,

I think that in htmlgen.cpp the lines around 1709:
```
  m_t << createHtmlUrl(m_relPath,ref,true,
                       Config_getString(HTML_OUTPUT)+"/"+fn,f,anchor);
```
should read
```
  m_t << createHtmlUrl(m_relPath,ref,true,
                       Config_getString(HTML_OUTPUT)+"/"+fn,fileName(),anchor);
```
around line 1721:
```
  m_t << createHtmlUrl("","",true,Config_getString(HTML_OUTPUT)+"/"+fn,f,anchor);
```
should read
```
  m_t << createHtmlUrl("","",true,Config_getString(HTML_OUTPUT)+"/"+fn,fileName(),anchor);
```

and in htmldocvisitor.cpp around line 2149:
```
  m_t << createHtmlUrl(relPath,ref,true,
                       Config_getString(HTML_OUTPUT)+"/"+fn,
                       file,
                       anchor);
```
should read
```
  m_t << createHtmlUrl(relPath,ref,true,
                       Config_getString(HTML_OUTPUT)+"/"+fn,
                       m_fileName,
                       anchor);

```

albert-github:
Even my previous comment is not 100% correct as now it would mean that `fileName()` / `m_fileName` (i.e. the "target") would be added and this contains a path like also the "src" contains a path, we just need the filename that is in `fn`

doxygen:
I see what you mean. New fix https://github.com/doxygen/doxygen/commit/67591cb7310cfec07131b0d6f7a179d7c3051c02.

albert-github:
Looks OK now, did a check against CGAL and didn't see anything problematic in this respect.

doxygen:
@albert-github Looks like there is some regression. The "Go to source link" is broken here: https://doxygen.github.io/doxygen-docs/d4/d3a/filedef_8h.html (was still ok with 1.9.8). Haven't confirmed if this is the cause, but it is likely. I'll investigate this.

Issue #10356, opened 2023-10-14
CopyDoc target is parsed incorrectly


MikeWalrus:
**Describe the bug**
#10223 causes some of the CopyDoc targets to be parsed incorrectly.

**To Reproduce**
mainpage.dox:
```
/**
\mainpage Home
\copydoc page_home
\page page_home Overview

Detail
*/
```

Doxyfile:
```
INPUT = mainpage.dox
```

doxygen warning:
```
mainpage.dox:3: warning: @copybrief or @copydoc target 'page_home\copydetails' not found
```

Enabling more printing reveals that `extractCopyDocId` is given the input `'page_home\copydetails page_home'`.

**Version**
1.9.8

albert-github:
I've just pushed a proposed patch, pull request #10357

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

MikeWalrus:
Thanks for the quick fix! @albert-github
My only question is why do `\copybrief page_home` and `\copydetails` got concatenated like that?
I've checked `commentscan.l` and it seems that it puts `\copybrief` into the brief context and `\copydetails` into the details context.

albert-github:
Due to the fact that that the brief and detailed section were concatenated and before there was as removal of begining "white space"

MikeWalrus:
That makes senses. Thanks!

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.10.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10357, opened 2023-10-14
issue #10356 CopyDoc target is parsed incorrectly

albert-github:
An id should not only end at a space like character but also at the start of a command.


Paukageo10 (approved):
https://bossglobalelectronics.github.io/doxysphinx/docs/linking_Onestore to_doxygen.html

PR #10367, opened 2023-10-18
Fix indexing words with non-ASCII letters

bolshakov-a:
`Xapian::QueryParser` converts words to lowercase using `Unicode::tolower` (see https://github.com/xapian/xapian/blob/v1.4.23/xapian-core/queryparser/queryparser.lemony#L1244), so doxyindexer should do the same so that a term originally containing uppercase letters could be found.

Besides that, a stemming algorithm should probably be selected using correct language, but this change allows finding words at least in the form which they are written in.

### DIFF
--- a/addon/doxysearch/doxyindexer.cpp
+++ b/addon/doxysearch/doxyindexer.cpp
@@ -18,7 +18,6 @@
 #include <cstdlib>
 #include <iostream>
 #include <string>
-#include <algorithm>
 #include <sstream>
 #include <fstream>
 #include <iterator>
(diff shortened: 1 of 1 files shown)


addon/doxysearch/doxyindexer.cpp
```diff
@@ -1,6 +1,7 @@
 /******************************************************************************
  *
  * Copyright (C) 1997-2015 by Dimitri van Heesch.
+ * Copyright (C) 2023      by A. Bolshakov, Renga Software LLC.
```
doxygen, line 4: Is there a need to add this copyright line? 
I prefer not to keep track of all contributions in this way.
See also [this article](https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects)

bolshakov-a: I think, not at all. It just seemed to me that it is one of GPL requirements.
GPL v. 2, p. 2a:
>  You must cause the modified files to carry prominent notices stating that you changed the files and the date of any change.

GPL v. 3, p. 5a:
> The work must carry prominent notices stating that you modified it, and giving a relevant date.

Issue #10377, opened 2023-10-23
Wrong syntax for output of non-C++ deprecation lists for enum members


StefanBattmer:
**Describe the bug**
When e.g. building a C# or Python documentation the namespace syntax for enum members marked as being deprecated will show up using C++ style namespace syntax in the deprecation list

**Expected behavior**
Instead of Bla::blub::myEnum I would like to see Bla.blub.myEnum

**Screenshots**
![wrongSyntax](https://github.com/doxygen/doxygen/assets/33477553/fe063dc6-90a7-4a62-9110-a61f4ff565ef)

**To Reproduce**
See attached zip archive
[DoxygenNamespaceIssue.zip](https://github.com/doxygen/doxygen/files/13069727/DoxygenNamespaceIssue.zip)

**Version**
While we encountered this using Doxygen 1.9.6 this has also been checked using the latest version 1.9.8 on both Windows and Linux.

albert-github:
I've just pushed a proposed patch, pull request #10378

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.10.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10378, opened 2023-10-23
issue #10377 Wrong syntax for output of non-C++ deprecation lists for enum members

albert-github:
The member names should be converted to the syntax in respect to the used language


StefanBattmer:
Looks good to me! Thanks for the (as always) fast response!

Issue #10351, opened 2023-10-10
Search subpart research after a search


morgan-mathiaut:
**Describe the bug**
In HTML output
When you select a subpart for a search ("Classes" for example), this subpart is reset after a search.

**Expected behavior**
Search subpart is keep after a search

**To Reproduce**
Go to http://eigen.tuxfamily.org/dox/
Select "Classes" on search subpart
Enter "Array" in search
Go to the "Array" doc page
Selected subpart "Classes" is reset to "all"

**Version**
1.9.8

albert-github:
Doxygen has besides also other search possibilities, see https://www.doxygen.nl/manual/searching.html probably you can accomplish with e.g. with an external / server side search the things you would like to see.

morgan-mathiaut:
Thanks to your impressive quick reply :)
Here a small video :
https://github.com/doxygen/doxygen/assets/140616075/074956ac-d3b0-4ea5-9ec4-885b13dba20f
The espected behavior is : keep subpart "Classes" after a search and don't reset it (to "all")
I hope that it's clearer.

Thanks a lot

doxygen:
__[implementation note]__: this means this value should be stored in localstorage (for Chrome) or in a cookie (other browsers), like we do for the darkmode setting (`darkmode_toggle.js`).

albert-github:
@doxygen or should the structure be used that is used in`resize.js`?

doxygen:
@albert-github Yes, it is basically the same idea indeed (i.e. localStorage for Chrome, and cookies otherwise). This dual approach is needed to get it to work with `file://` style URLs.

albert-github:
@doxygen that it should be stored in localStorage / cookie is clear. I'm not completely getting the difference, should the create / read / erase version of `darkmode_toggle.js` be used or the read/write version of `resize.js` ?

Are you looking into it or should I have a look in the coming week?

doxygen:
@albert-github there is no real difference. We probably need a third instance of such functions for this new property. You can have a go at it if you like.

albert-github:
I've just pushed a proposed patch, pull request #10381

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.10.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
[...]

PR #10381, opened 2023-10-23
issue #10351 Search subpart research after a search

albert-github:
Implemented cookie that stores the selected category

### DIFF
--- a/templates/html/search.js
+++ b/templates/html/search.js
@@ -22,6 +22,9 @@
 
  @licend  The above is the entire license notice for the JavaScript code in this file
  */
+var cookie_namespace = 'doxygen';
+var cookie_name = '_search_grp';
+
 function convertToId(search)
 {
   var result = '';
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I added [this commit](https://github.com/doxygen/doxygen/commit/390499d40f8cb9ea1e3440a30f33c77ead85e7db) so that the name is stored in the cookie and not the id. Indices may change (for instance if project didn't have a namespace and then one is added, all indices after the namespace in the list shift by 1).

albert-github:
Makes sense, though in case the name doesn't exist anymore and the value 0 (i.e. All) is used wouldn't it make sense to write this value back into the cookie?

doxygen:
I think this also happens. When the name stored in the cookie isn't found, 0 is returned, and selected. This should correspond to the 'All' item, which is written to the cookie unconditionally as part of the selection.

albert-github:
Indeed I was thinking about the setting at the end of the search.js but here, through small layer of functions, the writeSettings is called.

PR #10383, opened 2023-10-25
Handling of relative include path in preprocessor

albert-github:
In the 7z package we get a warning like:
```
error: NamespaceDefImpl::insertMembers(): member 'CCoderReleaser' with unexpected type id 8 and class scope '' inserted in namespace scope 'NCompress::NArj::NDecoder'!
```
this because of the fact that the include files like:
```
#include "../../../C/CpuArch.h"
```
are "not found" in the preprocessor and thus a number of macros are not evaluated (7z heavily relies on macros), so we check now the absolute path and the path as done before so the `INCLUDE_PATH` still can be applied.

(Found by Fossies for the 7z package)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -2851,7 +2851,7 @@ static bool expandExpression(yyscan_t yyscanner,QCString &expr,QCString *rest,in
 
         if (replaced) // expand the macro and rescan the expression
         {
-          //printf(" replacing '%s'->'%s'\n",expr.mid(p,qPrint(len)),qPrint(expMacro));
+          //printf(" replacing '%s'->'%s'\n",qPrint(expr.mid(p,len)),qPrint(expMacro));
           QCString resultExpr=expMacro;
           QCString restExpr=expr.right(expr.length()-len-p);
           processConcatOperators(resultExpr);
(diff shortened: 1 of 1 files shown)


src/pre.l
```diff
@@ -3439,7 +3439,9 @@ static void readIncludeFile(yyscan_t yyscanner,const QCString &inc)
     std::unique_ptr<FileState> fs;
     bool alreadyProcessed = FALSE;
     //printf("calling findFile(%s)\n",qPrint(incFileName));
-    if ((fs=findFile(yyscanner,incFileName,localInclude,alreadyProcessed))) // see if the include file can be found
+    fs=findFile(yyscanner,absIncFileName,localInclude,alreadyProcessed); // see if the absolute include file can be found
+    if (!fs && !alreadyProcessed) fs=findFile(yyscanner,incFileName,localInclude,alreadyProcessed); // see if the include file can be found
```
doxygen, line 3443: Why was this not resolved by `determineAbsoluteIncludeName()`? Don't think it is correct to look relative from where doxygen is run. If this is wanted also `INCLUDE_PATH=.` would also do the trick, I'd expect.

albert-github: The Adding of `INCLUDE_PATH=.` does not solve the problem.
The `determineAbsoluteIncludeName()` crates an absolute path but the result (`absIncFileName`) was not given to the `findFile` routine, here the `incFileName` was given.
Probably just changing `incFileName` into `absIncFileName` would have solved the issue as well (so in the patch `if (!fs && !alreadyProcessed) fs=findFile(yyscanner,incFileName,localInclude,alreadyProcessed);` could be removed), but I wanted to play save.

doxygen: I don't like the extra `findFile()` call. After some time nobody understands anymore why it is there. So I've taken the risk of removing it.

albert-github: I think that this should be OK, tested 7z and looks OK.

PR #10403, opened 2023-11-10
Miscounting in case of markdown links

albert-github:
In PR #8111 a number of miscountings were corrected but in case of
```
# Release process

Example PR: [GH-3139]

And now @error_5

[gh-3139]: https://github.com/doxygen/doxygen/pull/3139
```
the reported line was 6 instead of 5
This has been corrected.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13320659/example.tar.gz)

(Found by Fossies for the black package)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -1145,7 +1145,6 @@ int Markdown::processLink(const char *data,int offset,int size)
   {
     whiteSpace = true;
     i++;
-    nl++;
     // skip more whitespace
     while (i<size && data[i]==' ') i++;
   }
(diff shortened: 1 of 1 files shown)


src/markdown.cpp
```diff
@@ -1145,7 +1145,6 @@ int Markdown::processLink(const char *data,int offset,int size)
   {
     whiteSpace = true;
     i++;
-    nl++;
```
doxygen, line 1148: I guess this is removed because the whitespace here is not a part of the output (i.e. it is a separator between the `[..]` and `(..)` parts). But then lines 1152 and 1153 also do not make much sense, as `nl` is initialized to 0 at line 1133 and then not modified anymore until it is set to 0 again on line 1153.

albert-github: Indeed they don't make much sense anymore now the `nl++` is gone, adjusted.

PR #10409, opened 2023-11-13
Potential buffer overflow (strncat)

albert-github:
From the draft 2023 C standard:
> char *strncat(char * restrict s1, const char * restrict s2, size_t n);
>
> The strncat function appends not more than n characters (a null character and characters that
> follow it are not appended) from the array pointed to by s2 to the end of the string pointed to by
> s1. The initial character of s2 overwrites the null character at the end of s1. A terminating null
> character is always appended to the result. If copying takes place between objects that overlap,
> the behavior is undefined.

and

> char *strncpy(char * restrict s1, const char * restrict s2, size_t n);

> The strncpy function copies not more than n characters (characters that follow a null character are
> not copied) from the array pointed to by s2 to the array pointed to by s1.377) If copying takes place
> between objects that overlap, the behavior is undefined

### DIFF
--- a/libversion/fullversion.cpp
+++ b/libversion/fullversion.cpp
@@ -9,11 +9,12 @@ const char *getFullVersion(void)
   if (!init)
   {
     strncpy(fullVersionString,getDoxygenVersion(),BUF_SIZE-1);
+    fullVersionString[BUF_SIZE-1]='\0';
     if (strlen(getGitVersion())>0)
     {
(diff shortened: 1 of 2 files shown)


albert-github:
As far as I remember the fact that it looks more like C code is that it would have been easier to use in C like projects or projects that call these functions. Most important is that the potential buffer flows are removed. I think it is good to modernize code.

Issue #10418, opened 2023-11-17
Missing Resource on How to use Doxygen for "Normal Documentation"


cuppajoeman:
**Describe the bug**
On the main page of the site, I was informed that doxygen can help me in 3 ways. The third way is this:

>  You can also use doxygen for creating normal documentation (as I did for the doxygen user manual and web-site). 

I headed over to the documentation to read about this, and to my suprise I wasn't able to find any information on how I can make normal documentation. 

**Expected behavior**
I expected to easily see how to make normal documentation

**To Reproduce**
Go to the main site and try to figure out how to create normal documentation

**Notes**
Doing some searching online I found this SO question: https://stackoverflow.com/questions/9502426/how-to-make-an-introduction-page-with-doxygen which helps a bit.

I looked in the doxyfile for this repository and and looks like [this line](https://github.com/doxygen/doxygen/blob/e29c4a5e86941ed6b6c2437fcdf82263a2679268/doc/Doxyfile#L38) seems to provide many of the pages of the site

But the most useful thing as of yet has been looking in https://github.com/doxygen/doxygen/blob/master/doc/index.doc.

I found the last word of the [main page](https://www.doxygen.nl/index.html) (available) and then found the corresponding place in the index.doc file [here](https://github.com/doxygen/doxygen/blob/e29c4a5e86941ed6b6c2437fcdf82263a2679268/doc/index.doc#L58) , after that line we can see that there's a newline and then the source for the [documentation](https://www.doxygen.nl/manual/index.html) tab starts.

If there is really no resource available on the main site that tells us how to create normal documentation, I was hoping the creator or someone could help give some tips on how we can do this and then eventually add it to the main site.

cuppajoeman:
After using doxygen for a bit and figuring out how things work, your response makes more sense, but I still have a few points.

One example of creating "normal documentation" might be a usage/tutorial/examples section which might explain how to use the documented functions together to create sample programs, this may also incorporate adding custom elements to the navbar to grab the user's attention.

Even though it might be obvious now how normal documentation works, I think my younger self would have appreciated a section dedicated on how to actually do the relevant tasks that come with creating normal documentation.

Also taking a look online for questions/resources relating to this topic, there are some:

* https://stackoverflow.com/questions/3052036/how-to-include-custom-files-in-doxygen
* http://entrenchant.blogspot.com/2009/09/doxygen-gotchas.html
[...]

Issue #8184, opened 2020-11-19
Bad parsing of CMakeLists.txt


mcandre:
**Describe the bug**

When I try to doxygen my C project, then doxygen goes down a rabbit hole trying to parse CMakeLists.txt and fails early.

**Expected behavior**

doxygen should completely ignore the syntax in CMakeLists.txt and focus on actual source code files that contribute to a public API.

**To Reproduce**

CMakeLists.txt snippet:

```cmake
execute_process(COMMAND cat /etc/*release* OUTPUT_VARIABLE ETC_RELEASE)
```

Trace:

```console
$ doxygen Doxyfile
Searching for include files...
Searching for example files...
Searching for images...
Searching for dot files...
Searching for msc files...
Searching for dia files...
Searching for files to exclude
Searching INPUT for files to process...
Searching for files in directory /home/andrew/go/src/github.com/mcandre/toys/c/args
Reading and parsing tag files
Parsing files
Preprocessing /home/andrew/go/src/github.com/mcandre/toys/c/args/CMakeCache.txt...
Parsing file /home/andrew/go/src/github.com/mcandre/toys/c/args/CMakeCache.txt...
Preprocessing /home/andrew/go/src/github.com/mcandre/toys/c/args/CMakeLists.txt...
/home/andrew/go/src/github.com/mcandre/toys/c/args/CMakeLists.txt:117: error: Reached end of file while still inside a (nested) comment. Nesting level 1 (probable line reference: 22) (warning treated as error, aborting now)
Exiting...
```

Hint hint, `/*release*` is NOT part of a multiline comment. CMakeLists.txt uses a [completely different](https://cmake.org/cmake/help/v3.1/manual/cmake-language.7.html#comments) multiline comment syntax. In fact, doxygen shouldn't even be *opening* CMakeLists.txt, let alone trying to parse it with a C/C++ AST.

**Version**

1.8.17

As a workaround, I am skipping populating my debugging variable `ETC_RELEASE`.

albert-github:
Doxygen does not know about the meaning of the names of files. The `*.txt` files are enabled through the `INPUT` and `FILE_PATTERNS` settings. Also of relevance is the way `.txt` files are processed i.e. as C files (this is due to historical reasons), and can be changed by means of `EXTENSION_MAPPING`.

Furthermore in doxygen files can be excluded by means of the `EXCLUDE` and `EXCLUDE_PATTERNS`. 

I think the problem is usage.

[...]

Issue #7218, opened 2019-08-22
Doxygen parsing word documents


mirkovich0:
Altough it clearly states on the doxyfile:
```
# If left blank the following patterns are tested:*.c, *.cc, *.cxx, *.cpp,
# *.c++, *.java, *.ii, *.ixx, *.ipp, *.i++, *.inl, *.idl, *.ddl, *.odl, *.h,
# *.hh, *.hxx, *.hpp, *.h++, *.cs, *.d, *.php, *.php4, *.php5, *.phtml, *.inc,
# *.m, *.markdown, *.md, *.mm, *.dox, *.py, *.pyw, *.f90, *.f, *.for, *.tcl,
# *.vhd, *.vhdl, *.ucf, *.qsf, *.as and *.js.
FILE_PATTERNS          =
```

It tries to parse a MS-Word .doc that I have on my docs folder inside the code.

It's also a problem the fact that Doxygen **parses it** and creates a lot of noise:

[Screen capture](https://www.dropbox.com/s/fmk0csr9hma9xfo/Screenshot%20from%202019-08-22%2010-54-36.png?dl=0)

albert-github:
- Please don't use external images, but embed them in the question.
- Which version of doxygen are you referring to (is, probably, not of influence but is good to record)?

When you don't want to process the just specify the extensions you want to process or if the `*.doc` files are on a special directory exclude that directory.

Looks like that e.g. `.doc` and `.txt` are not included in the list, I think the list should be revised.

mirkovich0:
> Please don't use external images
Sorry. I didn't found how to embed an image.
> Which version of doxygen are you referring to
Forgot to put that info. It's the 1.8.11. Not the last one, but I think the last one that ships with Ubuntu.

Yes, I solved my problem excluding the directory, but I think the list may need to be revisited like you said.

albert-github:
1.8.11 is already a bit older (December 2015), I would anyway try to update (will not solve the current issue).

At the bottom of my input window it says "Attach files by dragging & dropping, selecting or pasting them.", I have to note that this is on windows though.

mirkovich0:
>1.8.11 is already a bit older (December 2015), I would anyway try to update (will not solve the current issue).

Running `sudo apt install doxygen` i get:

>doxygen is already the newest version (1.8.11-1ubuntu0.1).

I'll update later building it myself.

[...]

PR #10420, opened 2023-11-18
add .doc file to default FILE_PATTERNS

cuppajoeman:
It's mentioned [here](https://stackoverflow.com/a/6336368/6660685) that the .doc format is supported by doxygen, along with md, dox and txt.

When you add a directory containing doc files, they won't be found because .doc is not included in the FILE_PATTERNS by default, but md, dox would be as they are in the FILE_PATTERNS, so you have to manually add them to the list to make things work.

Not sure if this is done on purpose or just forgotten about, either way I think it'd be useful, so I'm making a PR. 

PS: .txt files are also not included here, so let me know if you think they should be


albert-github:
The mentioned stack overflow answer is quite old. In the mean time the `.doc` entered by means of #7218 / #7220 (Release 1.8.17) and was removed again by means of #8184 / #8200 (Release 1.9.0).

It is not sufficient to just update the `config.xml` also the general documentation should be updated (i.e. `starting.doc`)

I think seen the discussion with #8184 (and possible confusion with the Microsoft Word doc extension) it should not be in the default settings.


Note: When `*.doc` is added to the file patterns it is by default still parsed a `.c` type file.

cuppajoeman:
The conflict with microsoft word doc files makes sense.

You'll see that my recent omments are relating to how we can improve new user's experiences with doxygen. I myself am a relatively new user to doxygen, and I'm trying to improve this aspect of doxygen through my own lens (which I hope will also apply to others).

The reason why I created this PR at all comes down to how a new user might use the doxygen website to learn about how they themselves can create a website like the [doxygen website](https://www.doxygen.nl/index.html).

When a new user visits the documention for doxygen and and follow the getting started guide, they will most likely generate the default `Doxyfile`.

When a new user realizes they can add "normal" documentation, they may create a `.doc` file, the reason why they may create a `.doc` file is because if you look in the source code for this project itself, they are used to create the "normal" documentation.

When they try and build their documentation, they will not see any output because `.doc` files will are not accounted for in the FILE_PATTERNS, which may be confusing.

Due to this I think I'll make a PR including the info mentioned in https://github.com/doxygen/doxygen/issues/10418 but also write about the fact that `.doc` files are not included by default because of the word conflict.

On the other hand, there might be some value for "normal" documentation files be visible to doxygen right out of the box, but for now I think a note will suffice. 

Please let me know your thoughts

albert-github:
> You'll see that my recent comments are relating to how we can improve new user's experiences with doxygen. I myself am a relatively new user to doxygen, and I'm trying to improve this aspect of doxygen through my own lens (which I hope will also apply to others).

This makes sense and is appreciated


> When a new user realizes they can add "normal" documentation, they may create a `.doc` file, the reason why they may create a `.doc` file is because if you look in the source code for this project itself, they are used to create the "normal" documentation.

Maybe we should change here the `.doc` extension to `.dox`, though not just changing `.doc` to `.dox` is sufficient but also some processing software has to be changed, but still this change would make sense. I'll have a go at it.

> Due to this I think I'll make a PR including the info mentioned in https://github.com/doxygen/doxygen/issues/10418 but also write about the fact that .doc files are not included by default because of the word conflict.

see the remark above regarding `.doc`.

doxygen:
I agree we should move away from using the .doc extension for doxygen documentation. Using .dox is probably the best option.

albert-github:
Regarding the `.doc` to `.dox` I've just pushed a proposed patch, pull request #10425

PR #10421, opened 2023-11-18
Give functions override attribute

albert-github:
Give virtual functions override attribute to prevent sonarcloud warnings.


doxygen:
@albert-github I think it is better to squash the commits in this case. I can do it while merging.

albert-github:
@doxygen indeed in this case it probably is better to squash the commits, I have no experience with squashing commits so please do so while merging.

PR #10441, opened 2023-11-23
Incorrect warning and output with `\image` followed by HTML command

albert-github:
When having:
```
@image html PCore-Block-Diagram.png width=100px <b>bold</b>
```
we get a warning like:
```
dox_main.md:16: warning: found </b> tag without matching <b>
```

and in the output we see as running text: `bbold`

When having instead of a HTML command a doxygen command or a symbol (`&...;`) we don't get a warning. The HTML command should be seen as end of the underlying command and not be handled at this moment (not the code is also used on other places hence the conditional).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13450905/example.tar.gz)

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -963,6 +963,7 @@ void DocParser::defaultHandleTitleAndSize(const int cmd, DocNodeVariant *parent,
 
   // parse title
   tokenizer.setStateTitle();
+  tokenizer.setStateTitleSize(true);
   int tok;
   while ((tok=tokenizer.lex()))
   {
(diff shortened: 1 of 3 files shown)


albert-github:
Was wrong Albert, but I see it anyway when setting it to fixed.

I think it is OK and more consistent as in the old situation when there was normal text there was a warning and with a doxygen command or a `&...;` html entity there was no waning, these should also be considered as just text. The html commands `<...>` are also just text so it is good that all is handled in the same way.

PR #10446, opened 2023-11-26
Fix non-reproducible search index

### LINKED ISSUE #10445: Search engine index generation is non-deterministic
**Describe the bug**

`html/search/*` files are non-deterministic when key collision occurs (`name` and `title` values are equal).

**Expected behavior**
Search engine index doesn't change between builds.

**To Reproduce**

Install Nix tooling, run `nix build nixpkgs#pipewire.doc --rebuild` on two different machines.
This will result in some files inside `html/search` directory being different.

**Version**
doxygen 1.9.7, NixOS 23.11

**Additional context**

I was able to fix this behavior by replacing `std::sort` with `std::stable_sort` here:
https://github.com/doxygen/doxygen/blob/c87ad3cbfbfe67a14e5078250178478719815dbb/src/searchindex_js.cpp#L489-L496

However, I'm not sure if that's a correct solution.

ivan770:
Closes #10445

### DIFF
--- a/src/searchindex_js.cpp
+++ b/src/searchindex_js.cpp
@@ -487,7 +487,11 @@ void createJavaScriptSearchIndex()
     for (auto &[name,symList] : sii.symbolMap) // for each symbol in the index
     {
       // sort the symbols (first on search term, and then on full name)
-      std::sort(symList.begin(),
+      //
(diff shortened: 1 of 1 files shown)


ivan770:
CI failure seems unrelated

Issue #7688, opened 2020-04-07
Include mardown file in mainpage.dox


fjp:
Hi, I think this issue is related to #6478. I am trying to include a README.md markdown file into my doxygen mainpage.dox with the following result:

![image](https://user-images.githubusercontent.com/1286618/78648592-76c49180-78bc-11ea-9f25-91f1afb50999.png)

I started using `\include{doc}` but this doesn't include the text as shown in the image above. 
Using the obsolete `\includedoc` the content is included but the markdown heading is not formatted correctly.

Using markdown inside mainpage.dox is working great though. 

Is it possible to include such a formatted markdown or would this require an enhancement?

Here is the mainpage.dox:

```
/**
\mainpage
\htmlinclude manifest.html

\includedoc README.md

\include{doc} README.md

\include labelid

## Test heading

test text 
``` 

The content of the README.md is:

```
## Hardware Interfaces               {#labelid}

Test text inside README.
```

I am using doxygen verison 1.8.13

albert-github:
With the above snippets it looks to me that they are not 100% complete:
- missing `*/` in `mainpage.dox`
- settings in your doxygen configuration file
- missing file `manifest.html`
- missing file `labelid`

It looks like I can reproduce the problems you see and the number of warnings look like to be a bit less and more consistent with version 1.8.17 (and with the current master), but to get a consistent test possibility:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem?  Please don't add external links as they might not be persistent.

albert-github:
Note also that the syntax `\include{doc}` was not yet supported with 1.8.13 (introduced with 1.8.15)

fjp:
Hi @albert-github thanks for the infos and sorry for the incomplete example I just copied part of the mainpage.dox file. 
`manifest.html` is part of [ROS](http://wiki.ros.org/rosdoc_lite)  which has a python command line tool `rosdoc_lite` that is a wrapper for doxygen. They use a `Doxyfile` template. 

[...]

PR #10451, opened 2023-11-27
issue #7688 Include markdown file in mainpage.dox

albert-github:
In the comment https://github.com/doxygen/doxygen/issues/7688#issuecomment-1822840875 a problem was identified regarding HTML type of comments (`<!--... -->`), they are excluded from processing now similar to the code in commentscan.l

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -231,6 +231,9 @@ SLASHopt [/]*
                                       if (yyextra->lang!=SrcLangExt_Cpp) REJECT;
                                       copyToOutput(yyscanner,yytext,(int)yyleng);
                                     }
+<*>"\"<!--\""                       {
+                                      copyToOutput(yyscanner,yytext,(int)yyleng);
+                                    }
 <*>"<!--"                           {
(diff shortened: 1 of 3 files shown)


src/commentcnv.l
```diff
@@ -229,6 +231,32 @@ SLASHopt [/]*
                                       if (yyextra->lang!=SrcLangExt_Cpp) REJECT;
                                       copyToOutput(yyscanner,yytext,(int)yyleng);
                                     }
+<*>"<!--"                           {
```
doxygen, line 237: using `<*>` is risky, as it would also apply outside of comment blocks, i.e. when in the Scan state, which is undesired as someone could write

```c++
int k = "hey <!-- look at this";
```

albert-github: The `<*>` is indeed a bit dangerous.
I tested the given example together git this patch and it looks like it is working correctly.

I think it would probably be better not to use `<*>` but use a method analogous the one used with e.g. `\htmlonly` (i.e. consider it as a verbatim block).

I'll have a look at it....

albert-github:
This proposed pull request is superseded by  #10457

PR #10478, opened 2023-12-11
Incorrect warning due to import statement

albert-github:
With the Objective C example:
**aa.m**
```
#import "bb.h"
```

**bb.h**
```
#ifndef INCLUDED_OOFUNCTIONATTRIBUTES_h
#endif
```
we get the warning:
```
bb.h:2: Error: More #endif's than #if's found.
```
as the first line of the imported file is read as "comment", by delaying the read this is prevented.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13634350/example.tar.gz)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -1210,17 +1210,18 @@ WSopt [ \t\r]*
                                         }
 <Include>[^\">\n]+[\">]                 {
                                           yyextra->incName+=yytext;
-                                          readIncludeFile(yyscanner,yyextra->incName);
                                           if (yyextra->isImported)
                                           {
                                             BEGIN(EndImport);
                                           }
                                           else
                                           {
+                                            readIncludeFile(yyscanner,yyextra->incName);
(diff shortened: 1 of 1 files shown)


albert-github:
Note: I chose the given solution as not to change what, I think, would be the meaning, maybe it is also possible to change:
```
<Include>[^\">\n]+[\">]                 {
                                          yyextra->incName+=yytext;
                                          readIncludeFile(yyscanner,yyextra->incName);
                                          if (yyextra->isImported)
                                          {
                                            BEGIN(EndImport);
                                          }
                                          else
                                          {
                                            BEGIN(Start);
                                          }
                                        }

```
to
```
<Include>[^\">\n]+[\">]                 {
                                          yyextra->incName+=yytext;
                                          readIncludeFile(yyscanner,yyextra->incName);
                                          BEGIN(Start);

                                        }
```
and also remove all references to the state `EndImport`

Issue #10485, opened 2023-12-15
Parsing error with C++20 requires clause if at the end of the function declaration


lieser:
**Describe the bug**
If a C++20 requires clause is at the end of the function declaration it seems to mess up the parsing and leads to strange follow errors.

**Expected behavior**
Doxygen can correctly parse the requires clause.

**To Reproduce**

[doxygen-requires.zip](https://github.com/doxygen/doxygen/files/13684030/doxygen-requires.zip)
(minimal example from https://github.com/randombit/botan/blob/3dcf8af5a77e9e7643cc58cd5498d3950c1470ef/src/lib/rng/rng.h)

```c++
#include <concepts>
#include <span>
#include <type_traits>

/**
* An interface to a cryptographic random number generator
*/
class RandomNumberGenerator {
   public:
      /**
      * Incorporate some additional data into the RNG state.
      */
      template <typename T>
      void add_entropy_T(const T& t)
         requires std::is_standard_layout<T>::value && std::is_trivial<T>::value
      {
         return;
      }

   protected:
      /**
      * Generic interface to provide entropy to a concrete implementation and to
      * fill a given buffer with random output. Both @p output and @p input may
      * be empty and should be ignored in that case. If both buffers are
      * non-empty implementations should typically first apply the @p input data
      * and then generate random data into @p output.
      *
      * This method must be implemented by all RandomNumberGenerator sub-classes.
      *
      * @param output  Byte buffer to write random bytes into. Implementations
      *                should not read from this buffer.
      * @param input   Byte buffer that may contain bytes to be incorporated in
      *                the RNG's internal state. Implementations may choose to
      *                ignore the bytes in this buffer.
      */
      virtual void fill_bytes_with_input(std::span<uint8_t> output, std::span<const uint8_t> input) = 0;
};
```

Results in the error:
```
/mnt/c/work/test/doxygen-requires/test.h:36: warning: Found ';' while parsing initializer list! (doxygen could be confused by a macro call without semicolon)
```

**Version**

Version: 1.9.8 (c2fe5c3e4986974eb2a97608b24086683502f07f)
Platform: Ubuntu 22.04.3 LTS
Binary from https://www.doxygen.nl/download.html

**Additional context**

A workaround is to put the requires clause after the template parameter list, which at least newer versions of doxygen can parse (see also https://github.com/doxygen/doxygen/issues/8980)

albert-github:
[...]

PR #10486, opened 2023-12-15
issue #10485 Parsing error with C++20 requires clause if at the end of the function declaration

albert-github:
Handle also things like `::value`


lieser:
Thanks for the quick fix, can confirm that with the `1.10.0 (a98c2db1fa7ed1737985a3b977aece9766c1c41e*)` build from the CI the problems I reported in #10485 are gone.

PR #10487, opened 2023-12-16
Do not lowercase search titles for code and capitalize on page terms

apolukhin:
This PR addresses the lower-casing of functions in search box so that:

image

Becomes the same as in source code:

image

It also capitalizes the words from captions so that:

image

becomes:

image

### DIFF
--- a/src/searchindex_js.cpp
+++ b/src/searchindex_js.cpp
@@ -53,6 +53,12 @@ void SearchTerm::makeTitle()
   else if (std::holds_alternative<const SectionInfo *>(info))
   {
     title = std::get<const SectionInfo *>(info)->title();
+
+    // Capitalizing the word as this is not a code entity
+    std::string letter = getUTF8CharAt(word.str(),0);
+    // Uppercase letter could have different size
+    word.remove(0, letter.size());
+    word.prepend(convertUTF8ToUpper(letter));
   }
   else
   {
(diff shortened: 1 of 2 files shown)


albert-github:
For e.g. C++ this might be true but for case insensitive languages (e.g. Fortran) this gives problems.
It is not always clear to the user of the documentation which words a capitalized / where words are capitalized and thus it is more user friendly when it finds words in both cases.

- does the search engine have problems when typing things in uppercase?

apolukhin:
Search is still case insensitive

image

image

image

Only the representation in search box is changed. All the other logic left intact

albert-github:
That is a bit reassuring, though the e.g. Fortran problem remains (it is bad practice when using different cases but still it happens).

A question for your implementation, when having the function `toString` and `ToString` (quite bad coding, but possible) we get:
image

Is this still the case?

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13692414/example.tar.gz)

apolukhin:
Still fine

image

Capitalization affects only representation of pages in search box

apolukhin:
Quick search gave no matching results on "fortran forward declare function". So probably it is fine for Fortran to show the case as in source file, as the function could be declared only once. The same function should not appear in docs multiple times in different cases

albert-github:
In Fortran there is indeed nothing like forward declaration but I'm not sure how it will all go with (not yet implemented) submodules (and currently I'm not using Fortran anymore, so getting a bit rusty).

One nice example is with `interfaces`, see attached example ([example_ftn.tar.gz](https://github.com/doxygen/doxygen/files/13692636/example_ftn.tar.gz)) and here we have a split interface named once `alc` and once `ALC`, currently in the Fortran lexer everything is converted to lowercase and in the search bar we only see `alc` how is this after the change (seen the early stage of the conversion in the lexer probably also just lowercase).

apolukhin:
After the change we still see only `alc`:
image

As you mentioned, looks like it is lower-cased in lexer

apolukhin:
Do I need to change something or the PR looks fine?

PR #10490, opened 2023-12-18
Incorrect line count with `\include{doc}`

albert-github:
In case we have:
**foo.h**
```
/// \file
///
/// The level 1
///
/// provoking an error \error_5_1_h
///
/// \include{doc} kaboom.md

/// \brief the fie 1
/// \details the fie 1
/// \error_11_h
void foo_1() {}
```
**kaboom.md**
```
# Uh-oh 1

Seems Doxygen isn't playing nice with slightly longer paths... \error_3
```
we get the warnings:
```
.../foo.h:5: warning: Found unknown command '\error_5_1_h'
.../kaboom.md:3: warning: Found unknown command '\error_3'
.../foo.h:14: warning: Found unknown command '\error_11_h'
```
though the last one should read
```
.../foo.h:11: warning: Found unknown command '\error_11_h'
```

Problem is due to the fact that `\include{doc}` adds lines but the correction of the number of lines is done based for the current comment block so subsequent comment blocks don't know about it, this has to be handled in the scanner that extracts the comment blocks. (code analogous to the doctokenizer)

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13702684/example.tar.gz)

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -281,6 +281,13 @@ LOGICOP   "=="|"!="|">"|"<"|">="|"<="|"&&"|"||"|"!"|"<=>"
 BITOP     "&"|"|"|"^"|"<<"|">>"|"~"
 OPERATOR  "operator"{B}*({ARITHOP}|{ASSIGNOP}|{LOGICOP}|{BITOP})
 MODULE_ID ({ID}".")*{ID}
+LINENR    {B}*[1-9][0-9]*
+FILEICHAR [a-z_A-Z0-9\\:\\\/\-\+=&#@]
+FILEECHAR [a-z_A-Z0-9\-\+=&#@]
+FILECHARS {FILEICHAR}*{FILEECHAR}+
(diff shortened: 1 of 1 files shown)


src/scanner.l
```diff
@@ -6901,6 +6908,31 @@ NONLopt [^\n]*
                                           yyextra->nestedComment=0;
                                           BEGIN(DocCopyBlock);
                                         }
+<DocBlock>{CMD}"ifile"{B}+"\""[^\n\"]+"\"" {
```
doxygen, line 6911: Do we need to take care of escaped `\ifile` and `\iline` commands as well? I don't expect all cases can be handled by the rule on line 6889 as the match here is potentially longer.

albert-github: No I don't think so as there is an explicit rule for `{CMD}{CMD}` (around 22 lines upwards):
```
<DocBlock>({CMD}{CMD}){ID}/[^a-z_A-Z0-9] { // escaped command
                                          yyextra->docBlock << yytext;
                                        }

```

Issue #10484, opened 2023-12-15
Extract comment blocks and signature


tdegeus:
## Question

I would like to interact with doxygen on a more basic level than e.g. the `xml` output:

**I would like to just have the 'docstrings' (as they are written, i.e. without highlighting) and the function signature as used by doxygen.**

## Motivation
- Debug e.g. broken links. 
- Set the C++ doc-blocks as docstrings for `pybind11` extensions. 

## Example

Consider:
```cpp
namespace name {
/**
 * @brief My first function.
 * @param a This is a parameter.
 */
int foo(int a);

/**
 * This is a class
 */
class Bar
{
public:
    /**
     * @brief Some counter
     * @return integer
     */
    int count();
};
} // namespace name
```

I would like to be able to extract the signature as used by doxygen. For example, I would like to get the following YAML file (though any other format is fine):
```yaml
name::foo(int): |
    @brief My first function.
    @param a This is a parameter.

name::Bar: |
    This is a class

name::Bar::count(): |
    @brief Some counter
    @return integer
```

albert-github:
I don't get what you want

- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem? Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).
- Please also specify the **full** doxygen version used (`doxygen -v`).

albert-github:
It would be very difficult to show this information as doxygen does no store the information in this way.

I don't get why an e.g. yaml file with the original information would solve the issues / make it easier to find them?
In the xml file all the relevant information is present and otherwise a warning should have been thrown (or are there cases where a link is expected but no  warning is thrown?)

tdegeus:
Indeed, I have had cases where a link was not created where I did expect it (on implicit creation by putting e.g. `name::foo()` in the above example). It has sometimes been hard to figure out what I had to do to correct. 

My strong motivation is the second listed. My C++ codes have Python bindings written using pybind11. It is very nice to have good docstrings as they are very useful as many editors show them while coding. However, I do not want to write them twice to avoid out-dating them, and I want to keep them in the C++ code not to break my doxygen documentation of it. So what I would like to do is to automatically parse them when I compile the Python API.

Indeed the XML has the info. However, the information that I want is at that point somewhat convoluted. In particular, I am keen to have the doc-block pre formatting.

doxygen:
[...]

PR #10496, opened 2023-12-19
addon: list symbols as JSON

tdegeus:
Fixes https://github.com/doxygen/doxygen/issues/10484

@doxygen @albert-github Thanks for your pointers. With it, I propose a small add-on that lists basic (or a bit less basic) information on which symbols are found where. I would appreciate it if it could be added here! A motivation that I have crossed many discussions and partial attempts to do exactly this, and that I belief that such information is hard to get from the common output mechanisms of doxygen. 

Please let me know what you think.

P.S. In the case that this 'example' is considered, it would be good to consider merging with `doxyapp`. However, I think the purpose of the current 'example' does not fit too well with the 'app' character of `doxyapp`.


doxygen:
@tdegeus Since we already have doxyapp and doxyparse to maintain now, I did not want to maintain a third, very similar example. So it is ok to replace the doxyapp functionality with this 'list' functionality if you believe this is more useful for more people. I don't like to introduce another third party dependency, so ideally the json output can be written without this dependency.

tdegeus:
@doxygen It is indeed fine to extend the `doxyapp` functionality, and I completely understand that you'd like to keep examples to a minimum. Would you be supportive of a command-line option to bypass the interactive loop:
```
while (1)
  {
    printf("> Type a symbol name or\n> .list for a list of symbols or\n> .quit to exit\n> ");
    (void)fgets(cmd,256,stdin);
    QCString s(cmd);
    if (s.at(s.length()-1)=='\n') s=s.left(s.length()-1); // strip trailing \n
    if (s==".list")
      listSymbols();
    else if (s==".quit")
      exit(0);
    else
      lookupSymbols(s);
  }
```
?

Another question is, suppose that one wants to use the example to write one's own application, how can one interact with doxygen? I.e. is there a CMake target and shared-library installation support?

tdegeus:
@doxygen I added the discussed functionality to `doxyapp` making is a bit more general without being backward incompatible. Also, as requested, I got rid of the 3rd party dependency

PR #10497, opened 2023-12-19
Problem with `\fC` on man pages

albert-github:
In https://github.com/phmarek/fsvs/issues/19 "uses non-portable 'C' font, causing warnings from groff" this problem was menioned and it can be reproduced by means of:
```
/// \file
/// the file description

/// The function description
///
/// some `description`
void fie();
```

and when giving: `man man/ma3/aa.h.3 > /dev/null` we get:
```
troff:<standard input>:24: warning: cannot select font 'C'
```

The chosen way is the least intrusive, an alternative would have been to change all occurences of `\fC` to `\fR` (i.e. to the regular font).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13718341/example.tar.gz)

### DIFF
--- a/src/mangen.cpp
+++ b/src/mangen.cpp
@@ -318,6 +318,7 @@ static QCString buildFileName(const QCString &name)
 void ManGenerator::startFile(const QCString &,const QCString &manName,const QCString &,int,int)
 {
   startPlainFile( buildFileName( manName ) );
+  m_t << ".do ftr C R\n";
   m_firstCol=TRUE;
 }
 


doxygen:
@albert-github With this change I get on my Mac the warning: "This manpage is not compatible with mandoc(1) and might display incorrectly."

albert-github:
Probably an old version om mandoc?
What happens when you use:
```
.if \n(.g .if n \{\
. do ftr C R
.\}
```
instead of 
```
. do ftr C R
```

doxygen:
@albert-github That doesn't make a difference for me. Note that MacOSX is BSD based, not GNU/Linux based, so mandoc is probably using `roff` and not the GNU version `groff` for formatting.

For me `man mandoc` says:

HISTORY
     The mandoc utility first appeared in OpenBSD 4.8.  The option -I appeared in OpenBSD 5.2, and -aCcfhKklMSsw in OpenBSD 5.7.

albert-github:
- Were you able to reproduce the original problem when there were no changes?
- On Cygwin I don't have `mandoc` so I cannot test it.
- As written in the PR text:
  > The chosen way is the least intrusive, an alternative would have been to change all occurences of \fC to \fR (i.e. to the regular font).

  So probably best to revert to that solution, I'll create a PR.

albert-github:
I've just pushed a proposed patch, pull request #10501 that reverts this update and implements the alternative.

Issue #10498, opened 2023-12-19
#cmakedefine01 macros are not documented


vardarirrgan1:
https://github.com/doxygen/doxygen/issues/3519 addressed macros generated by CMake using `#cmakedefine`. CMake also has a related feature called `#cmakedefine01`. Ref https://cmake.org/cmake/help/latest/command/configure_file.html

Could we get `#cmakedefine01` to be treated the same as `#cmakedefine`?

**Version**
1.9.8

albert-github:
For testing purposes:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem (with the `#cmakdefifine01` and prefereably also with `@cmakedefine`)? Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

albert-github:
I tried to create a small example and came up with:
**CMakeLists.txt**
```
project(cmakedefine)
cmake_minimum_required(VERSION 3.22)

set(HAVE_GETOPT "1")
set(ZERO_HAVE_GETOPT "0")

configure_file(${CMAKE_SOURCE_DIR}/aa.h.in ${CMAKE_SOURCE_DIR}/aa.h)
```
**aa.h.in**
```
/// \file
/// some tests for `#cmakedefine`

#cmakedefine01 HAVE_GETOPT
#cmakedefine01 ZERO_HAVE_GETOPT
#cmakedefine01 NOT_HAVE_GETOPT
```
running cmake wil give:
**aa.h**
```
/// \file
/// some tests for `#cmakedefine`

#define HAVE_GETOPT 1
#define ZERO_HAVE_GETOPT 0
#define NOT_HAVE_GETOPT 0
```

so the best doxygen can do would be:
```
/// \file
/// some tests for `#cmakedefine`

#define HAVE_GETOPT 0
#define ZERO_HAVE_GETOPT 0
#define NOT_HAVE_GETOPT 0
```
(note the difference for `HAVE_GETOPT`)

vardarirrgan1:
I should have provided a more complete description. As in issue #3519, the use case is to document a file before substitution instead of after. When documenting preprocesser definitions, Doxygen currently treats `#cmakedefine` as if it were `#define`. My request is to do the same with `#cmakedefine01`

### Example

aa.in.h:
```
/// \file
/// some tests for `#cmakedefine`

/// A CMake configured macro
#cmakedefine HAVE_GETOPT

[...]

PR #10502, opened 2023-12-20
issue #10498 `cmakedefine01 macros are not documented

albert-github:
Handle `#makedefine01` even when found in a file (normally it should have been handled by CMake so we can only assume it to be defined with a value 0).

My example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13730233/example.tar.gz)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -416,6 +416,7 @@ WSopt [ \t\r]*
 %x      DefName
 %x      DefineArg
 %x      DefineText
+%x      CmakeDefName01
 %x      SkipCPPBlock
 %x      SkipCComment
 %x      ArgCopyCComment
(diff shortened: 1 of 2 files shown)


src/pre.l
```diff
@@ -957,6 +958,12 @@ WSopt [ \t\r]*
                                           yyextra->yyColNr+=(int)yyleng;
                                           BEGIN(DefName);
                                         }
+<Command>"cmakedefine01"{B}+            {
+                                          yyextra->potentialDefine += substitute(yytext,"cmakedefine01","     define");
```
vardarirrgan1, line 962: Is this substitution supposed to preserve the number of characters, as above? If so, it looks like it needs two more spaces.

albert-github: At first I thought it as well but I think that it is about where the `define` starts.

albert-github: I rethought about it and it is indeed better to replace `cmakedefine01` with `     define  ` although I didn't see any problems (tested `GENERATE_XML=YES`).

Issue #10504, opened 2023-12-22
A list inside an alias on a markdown page repeats the alias indefinitely


emmenlau:
**Describe the bug**
I have the following alias in my `doxygen.cfg`:
```
ALIASES += highlight_green{1}="<div style='border: 2px solid green; background-color: #FFFFEE; margin: 10px 0px 20px 0px; padding: 10px;'>\1</div>"
```

Generally it works nicely on markdown pages, so I can i.e. write:
```
@highlight_green{Here is a relevant aspect!}
```

However I'd like to also put lists inside the alias, like this:
```
@highlight_green{
 - Consider this aspect: A
 - Also consider the other aspect: B
}
```

When I do that, the resulting page shows all text following the highlight also with the same highlight!

**Expected behavior**
Only the list inside the alias should be rendered with the alias formatting.

**Screenshots**
Here is the screenshot of the problem:
![image](https://github.com/doxygen/doxygen/assets/1098087/53ceb907-a298-47dd-81fa-e0227d66e89b)

**To Reproduce**
Here is the full example: [doxygen-issue-10504.zip](https://github.com/doxygen/doxygen/files/13753539/doxygen-issue-10504.zip)

**Version**
1.9.8 (https://github.com/doxygen/doxygen/commit/c2fe5c3e4986974eb2a97608b24086683502f07f) installed from the official website, running on Ubuntu 22.04 x86_64.

albert-github:
> Here is the full example:

Where or am I too fast?

emmenlau:
> > Here is the full example:
> 
> Where or am I too fast?

Hahaha, you are **really** fast! Thanks! I added the issue number to the zip file, therefore the upload was delayed :-)

emmenlau:
Also, I can work around the problem by using an HTML unordered list, like so:
```
@highlight_green{
<ul>
<li>List entry A.</li>
<li>List entry B.</li>
</ul>
}

This line is not highlighted, which is great!
```

albert-github:
Indeed, the mentioned problem is also noted in the warnings:
```
.../input/frontpage.md:13: warning: found </div> at different nesting level (5) than expected (2)
.../input/frontpage.md:16: warning: end of comment block while expecting command </div>
```
this is something that should be investigated (in the doxygen code).

Note: you don't generated LaTeX but the alias could be a bit more sophisticated / working for other formats as well by changing:
```
[...]

PR #10519, opened 2023-12-27

albert-github:
When the HTML tag as an end tag and matches the tag just before the list it is handled as an end list as well.

Extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13780491/example.tar.gz)


albert-github:
Something weird happened in my local repository, discard this PR

PR #10520, opened 2023-12-27

albert-github:
When the HTML tag is an end tag and matches the tag just before the list it is handled as an end list as well.

Extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13780563/example.tar.gz)

### DIFF
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -5380,6 +5380,47 @@ reparsetoken:
           }
           else // found an end tag
           {
+            const DocNodeVariant *n=parent();
+            int indent = 0;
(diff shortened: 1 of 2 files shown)


Repairman-Running:
> Repairman_RunningHTML Repairman_RunningRepairman_Running
> 
> Repairman_Running: [example.tar.gz](https://github.com/doxygen/doxygen/files/13780563/Repairman_Running/example.tar.gz)

[Phone Manager User Agreement.txt](https://github.com/doxygen/doxygen/files/13796624/Phone.Manager.User.Agreement.txt)

Issue #10516, opened 2023-12-27
@copybrief command does not take effect with 1.10.0


st-masa:
**Describe the bug**

From 1.10.0, @copybrief command does not quote @brief command text of target @page. 

**Expected behavior**
Correctly quotes target @brief text.

**Screenshots**
![image](https://github.com/doxygen/doxygen/assets/79039413/7a7a1468-ded9-405d-a39b-a03576992e08)

**To Reproduce**
Extract attached [copybriefSample.zip](https://github.com/doxygen/doxygen/files/13778767/copybriefSample.zip) and execute doxygen 1.10.0.

```doxygen ./Doxyfile_1_10_0```

**Version**
* 1.10.0 (Windows)

**Comment by albert-github** [comment]
I've just pushed a proposed patch, pull request #10521

**Comment by st-masa** [comment]
@albert-github , Thank you for the fix. I have confirmed the \copybrief command works properly.
One more thing, when target \brief text is specified as follows, copied brief seems to include line break as well.
While line break is stripped with previous version.

```
/**
 @page  ExamplePage
 @brief   
     This is example page.

 */
```

**Comment by albert-github** [comment]
@st-masa 
Thanks for the quick testing.

You wrote:
> While line break seems to be stripped with previous version.

I tried this second example and after a long search I found a version where the line break was not present, it was version 1.8.15, in the 1.8.16 version the line break was present.

I'm not sure whether or not this should be fixed / is intended, though it is now inconsistent with e.g. a brief description with a function.

The line break issue is an issue not related to the original issue here though.

@doxygen 
- what to do?

**Comment by st-masa** [comment]
Sorry, "previous version" that I mentioned was a bit ambiguous.
Anyway my last comment is off-topic and can workaround by modifying \brief command text.

**Comment by albert-github** [comment]
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

**Comment by st-masa** [comment]
Thanks for the fix. I have checked the change 44ad700.

With the following page comment, \copybrief does not work.

Blank line after \page command:
```
/**
 @page  ExamplePage

 @brief  This is example page.
 */
```

Other command after \page command:
```
/**
 @page  ExamplePage
 @tableofcontents
 @brief  This is example page.
 */
```

While using the following comment that \brief command follows just after the next line, \copybrief works correctly.
```
/**
 @page  ExamplePage
 @brief  This is example page.
 */
```

albert-github:
[...]

PR #10521, opened 2023-12-28
issue #10516 @copybrief command does not take effect with 1.10.0

albert-github:
For better error / warning handling a line with file / line information was added to the brief description. When a new brief line arrives it was checked whether or not it could be the brief description by checking against the empty string, the later was not the case anymore and hence such a line should not be taken into consideration for determining the empty-ness of the original brief description.

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -3841,26 +3841,23 @@ static inline void setOutput(yyscan_t yyscanner,OutputContext ctx)
         }
         else
         {
-          static const reg::Ex nonBrief(R"( *[\\@]ifile *\"[^\"]*\" *[\\@]ilinebr *[\\@]iline *(\d+) *[\\@]ilinebr *\n)");
-          static const reg::Ex nonBrief1(R"( *[\\@]ifile *\"[^\"]*\" *[\\@]ilinebr *[\\@]iline *\d+ *[\\@]ilinebr *)");
+          static const reg::Ex nonBrief(R"( *[\\@]ifile *\"[^\"]*\" *[\\@]ilinebr *[\\@]iline *(\d+) *[\\@]ilinebr *([\n]?))");
           std::string str = yyextra->current->brief.str();
           reg::Match match;
-          reg::Match match1;
           if (reg::match(str,match,nonBrief)) // match found
           {
-            yyextra->current->brief = yyextra->current->brief.left(yyextra->current->brief.length()-1);
-            // set warning line correct
-            yyextra->current->brief += "\\iline " + QCString().setNum(1 + static_cast<int>(std::stoul(match[1].str()))) + " \\ilinebr ";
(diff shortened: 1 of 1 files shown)


src/commentscan.l
```diff
@@ -3821,29 +3822,55 @@ static inline void setOutput(yyscan_t yyscanner,OutputContext ctx)
...
+        if (yyextra->current->brief.stripWhiteSpace().isEmpty()) // we only want one brief
+          // description even if multiple
+          // are given...
+        {
+          foundMatch = true;
+        }
+        else
+        {
+          static const reg::Ex nonBrief(R"( *[\\@]ifile *\"[^\"]*\" *[\\@]ilinebr *[\\@]iline *(\d+) *[\\@]ilinebr *\n)");
+          static const reg::Ex nonBrief1(R"( *[\\@]ifile *\"[^\"]*\" *[\\@]ilinebr *[\\@]iline *\d+ *[\\@]ilinebr *)");
```
doxygen, line 3845: Looks like this is a longer match, but it is checked below later in the `else` case, so I wonder if there is ever a possibility `nonBrief` is set to `false` while `nonBrief1` is `true`? if not then this can be optimized away.

albert-github: I added the `nonBrief1` as safety measure. The `nonBrief` (i.e. the one with the `\n` ending`) normally is used.

doxygen: I think the 'safety' case can never be reached, i.e. it is dead code.

doxygen: Looks like I missed the extra `\n` for `nonBrief`. Still, I'm not very happy with the additional regex.

albert-github: I agree that the extra regex is not nice.
I didn't find, quickly, the place where the `\n` is added, maybe the end of the regex should read `(\n)?` and later on we test for the `\n`

But maybe something like:
```
/**
 @page  ExamplePage
 This is the example page. The second line.

 */
```
which gives in 1.9.8:
image

and without the fix:
image

and with the fix this doesn't change, looks like we don't even enter the tests with the regex.

The later case I have to check.
- Should this check also be done in the light of this issue?
- should we change the regex to a version with `(\n)?` and rets for the `\n` already?

qusaibb:
> للحصول على خطأ/تحذير أفضل، تمت إضافة سطر يحتوي على معلومات الملف/الخط إلى الوصف المختصر. عندما يصل سطر مختصر جديد، يتم التحقق مما إذا كان يمكن أن يكون وصفًا مختصرًا أم لا عن طريق التحقق من السلسلة الفارغة، ولم يعد الأمر كذلك بعد الآن، وبالتالي لا ينبغي أخذ هذا السطر في الاعتبار لتحديد خلوص السطر الوصف المختصر الأصلي.

PR #10522, opened 2023-12-29

albert-github:
When the HTML tag is an end tag and matches the tag just before the list it is handled as an end list as well.


albert-github:
Maybe this has to be discarded but it is a test to get the repo again correct

PR #10535, opened 2024-01-03
Abort doxygen if latex cannot generate images from formulas

### LINKED ISSUE #10534: Doxygen does not abort if the subcommand for generating images from formulas detects errors
**Describe the error**
If inserted formulas contain errors, only a message is currently displayed in the log, but generation continues. Since a large number of doxygen messages are output in larger projects, the message in question can easily be overlooked, so that the actual problem is not immediately recognizable.

**Expected behavior**
If there are errors in the contained formulas that lead to a termination of the `latex` application, doxygen should abort the generation after outputting the error message.

**Screenshots**
The problem is visible in the fact that the generated HTML output contains the original latex source text and not the desired image.

**Reproducible**
The problem can be reproduced with the following example
[testcase-formular-exit.zip](https://github.com/doxygen/doxygen/files/13818040/testcase-formular-exit.zip)

1. Unpack the zip file
2. Run `/usr/bin/doxygen Doxyfile ; echo $?`, which will return 
```
error: Problems running latex. Check your installation or look for typos in _formulas.tex and check _formulas.log!
0
```

where doxygen should return a value != 0 in this case.

**Version**
Version 1.9.2
Platform: Linux
Distribution openSUSE Leap 15.5
Source: Package manager

**Additional context**
With the current method, detection of the problems listed is only possible through special post-processing of the generated logs, which is not very robust. When integrating doxygen into a build system such as cmake or a CI system, errors are usually only recognizable through the program exit code, so that a corresponding adjustment makes the corresponding problems visible.

rhabacker:
Fixes #10534


albert-github:
See my remarks with the corresponding issue #10534

PR #10537, opened 2024-01-05
Make the menu bar stick to the top of the screen when scrolling

albert-github:
In the past an attempt was made (see fcb8cfbcd67d95680ccde950c03fd7e67acdb5b3) to get the top part of the screen fixed in case of `GENERATE_TREEVIEW=NO`. This method was not working and after a small time has been reverted. In the current implementation:
- adds a tag to the HTML file in case of `GENERATE_TREEVIEW=NO` (`div id="doc-content">`) that was also used in case of `GENERATE_TREEVIEW=YES` so `resize.js` can handle the scrollbar
- add end tag to the `footer.html` in case of `GENERATE_TREEVIEW=NO`
- always use `resize.js` and `navtree.css`
- adjust `resize.js` so it can be used for the `GENERATE_TREEVIEW=NO` case (also some javascript arrow functions, `=>` changes had to be reverted for CHM generation).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/13840216/example.tar.gz)

### DIFF
--- a/src/ftvhelp.cpp
+++ b/src/ftvhelp.cpp
@@ -871,28 +871,6 @@ void FTVHelp::generateTreeViewScripts()
 
   // generate navtree.js & navtreeindex.js
   generateJSNavTree(p->indentNodes[0]);
-
-  // copy resize.js & navtree.css
-  auto &mgr = ResourceMgr::instance();
-  {
-    std::ofstream f = Portable::openOutputStream(htmlOutput+"/resize.js");
-    if (f.is_open())
-    {
-      TextStream t(&f);
-      t << substitute(
-             substitute(mgr.getAsString("resize.js"),
-                "$TREEVIEW_WIDTH", QCString().setNum(Config_getInt(TREEVIEW_WIDTH))),
-                "$PROJECTID",      getProjectId());
-    }
-  }
-  {
-    std::ofstream f = Portable::openOutputStream(htmlOutput+"/navtree.css");
-    if (f.is_open())
-    {
-      TextStream t(&f);
-      t << HtmlGenerator::getNavTreeCss();
-    }
-  }
 }
 
 // write tree inside page
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -406,16 +406,13 @@ static QCString substituteHtmlKeywords(const QCString &str,
       generatedBy = theTranslator->trGeneratedBy();
       break;
   }
+  treeViewCssJs = "<link href=\"$relpath^navtree.css\" rel=\"stylesheet\" type=\"text/css\"/>\n";
   if (treeView)
   {
-    treeViewCssJs = "<link href=\"$relpath^navtree.css\" rel=\"stylesheet\" type=\"text/css\"/>\n"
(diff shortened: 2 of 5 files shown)


doxygen:
@albert-github Works very nicely, with one difference. In my original 'sticky' change, the title area slid out of view once one would scroll down, and slide back into view if you scroll up to the top of the page. The idea behind this is to not waste any vertical space. This is what the 'sticky' menu normally does and what makes it different from a 'fixed' menu. 

Luckily, something similar can be easily added via the following patch:
```.diff
diff --git a/templates/html/resize.js b/templates/html/resize.js
index 99099a7b88..da6df631e9 100644
--- a/templates/html/resize.js
+++ b/templates/html/resize.js
@@ -91,8 +91,17 @@ function initResizable(treeview) {
   content = $("#doc-content");
   footer  = $("#nav-path");
   sidenav = $("#side-nav");
-  if (treeview)
-  {
+  if (!treeview) {
+    title   = $("#titlearea");
+    titleH  = $(title).height();
+    content.on("scroll", function() {
+      if (content.scrollTop()>titleH) {
+        title.slideUp(200);
+      } else {
+        title.slideDown(200);
+      }
+    });
+  } else {
     navtree = $("#nav-tree");
     $(".side-nav-resizable").resizable({resize: function(e, ui) { resizeWidth(); } });
     $(sidenav).resizable({ minWidth: 0 });

```
What do you think? Any concerns?

albert-github:
The space of a page is of some concern, especially when the breadcrum exists of multiple lines.

I do have a problem.
When looking at e.g.html16/aa_8h.xhtml and scrolling down (don't resize the window, when necessary do it before the original opening) we see a blank space, that I made red for demonstration purposes, at the bottom:

image

doxygen:
@albert-github I see what you mean. I've corrected it like this:
```.diff
diff --git a/templates/html/resize.js b/templates/html/resize.js
index 99099a7b88..cf4a580486 100644
--- a/templates/html/resize.js
+++ b/templates/html/resize.js
@@ -91,8 +91,18 @@ function initResizable(treeview) {
   content = $("#doc-content");
   footer  = $("#nav-path");
   sidenav = $("#side-nav");
-  if (treeview)
-  {
+  if (!treeview) {
+    title   = $("#titlearea");
+    titleH  = $(title).height();
+    content.on("scroll", function() {
+      slideOpts = { 'duration': 200, 'step' : function() { resizeHeight(treeview); }};
+      if (content.scrollTop()>titleH && title.css('display')!='none') {
+        title.slideUp(slideOpts);
+      } else if (content.scrollTop()<=titleH && title.css('display')=='none') {
+        title.slideDown(slideOpts);
+      }
+    });
+  } else {
     navtree = $("#nav-tree");
     $(".side-nav-resizable").resizable({resize: function(e, ui) { resizeWidth(); } });
     $(sidenav).resizable({ minWidth: 0 });

```

albert-github:
Looks OK to me now. 

My personal preference is to still have the information, but as written before "The space of a page is of some concern, especially when the breadcrum exists of multiple lines.".

PR #10572, opened 2024-01-19
doxycfg: fix linking to spdlog

eli-schwartz:
e.g. trace.cpp (part of the doxycfg static library) includes the spdlog headers and thus needs its compile definitions as well.


src/CMakeLists.txt
```diff
@@ -211,6 +211,9 @@ add_library(doxycfg STATIC
     debug.cpp
     trace.cpp
 )
+target_link_libraries(doxycfg PRIVATE
+    spdlog::spdlog
+)
```
albert-github, line 216: Did you see anywhere some problems when compiling / linking?
I see the line `spdlog::spdlog` also with the `target_link_libraries(doxygen`, wouldn't be better to
- remove it here

or 
- have it there just so all are listed in one place?


from a consistency / dependency point of view it is probably best to have it with `doxycfg` though.

eli-schwartz: Again, it's missing the compile args. This is most easily noticed on my system because, while spdlog does happen to be in the system include directory -- and thus works without any include directories specified -- it is missing the define which makes spdlog use the system shared library. I noticed as a result that it recompiled the parts of the spdlog ABI which it used, and ended up not linking to spdlog at runtime.

It is probably best to list it in both places because doxymain will include trace.h, and also needs the compile args. It may be able to be removed from the apps after removing PRIVATE.

eli-schwartz: Testing this will be a huge pain of course since every time you run `ninja` the doxygen build files regenerate a ton of files with code generators, then recompile them. There's no such thing as a clean build, every `ninja` recompiles the same 100 files again and again.

eli-schwartz: I took some time to investigate the dependency propagation model and usage.

It is *incorrect* to use spdlog anywhere other than doxycfg and doxymain, but it is needed in both those places.

Patch has been updated.

albert-github: I'm a bit confused as I see here:
```
target_link_libraries(doxycfg PRIVATE
```
but a bit further down (around line 327) I see the same lines included, shouldn't the later one be:
```
target_link_libraries(doxymain PRIVATE
```

eli-schwartz: Yes, sorry, copy/paste typo.

PR #10578, opened 2024-01-20
Fix some typos in options description

DelazJ:


src/config.xml
```diff
@@ -2217,7 +2217,7 @@ hr.footer {
       <value name="DARK" desc="always generate dark mode output"/>
       <value name="AUTO_LIGHT" desc="automatically set the mode according to the user preference, use light mode if no preference is set (the default)"/>
       <value name="AUTO_DARK" desc="automatically set the mode according to the user preference, use dark mode if no preference is set"/>
-      <value name="TOGGLE" desc="allow to user to switch between light and dark mode via a button"/>
+      <value name="TOGGLE" desc="allow user to switch between light and dark mode via a button"/>
```
albert-github, line 2220: I think better formulation would be "allows user to ...", so an extra `s` as well

DelazJ: "Allows" was also my first option, from reading the file. If OK, I can walk through the file and align the <value> description with the third person, when possible.

albert-github: As this was changed I think it should be corrected. I don't know if there are more places where there is a, non essential problem like this.

PR #10581, opened 2024-01-21
Incorrect line number and filename after `\include{doc}` and `\snippet{doc}`

albert-github:
After the commands `\include{doc}` and `\snippet{doc}` the information in warnings from the comment scanner points to the includer file and the line numbers keep increasing even when the warning comes from code in the included file. This is due to the fact that the commands are already handled in the comment converter and the, added, `\ifile` and `\iline` commands are not taken into account in the comment scanner. Also the line number from a `\snippet{doc}` command is not correct, it doesn't refer  to the line of the snippet but to the beginning of the file (due to the hard coded `\iline 1`).

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14000863/example.tar.gz)

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -1140,7 +1140,7 @@ SLASHopt [/]*
                                         yyextra->inBufPos   = fs->oldFileBufPos;
                                         yyextra->includeCtx = fs->oldIncludeCtx;
                                         yyextra->incPrefix  = fs->oldIncPrefix;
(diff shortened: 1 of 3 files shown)


doxygen:
@albert-github I've tried to merge this commit with master, but I'm not sure if it is correct. Can you take a look?

albert-github:
@doxygen, you added the `SectionHandling::Escape` for the `ifile` and `iline` commands as a first step this is OK though I have to check how this works out when the `\include{doc}` is used in a `@section` part. I think it might be that it (most likely) should be `SectionHandling::Replace`, with some extra code, or (less likely) `SectionHandling::Invisible`.
I;ll dive into it ...

albert-github:
I adjusted the code.

Example:

PR #10583, opened 2024-01-21
Improve warnings

albert-github:
Improve a number of warnings in respect to the use starting char and placing single quotes around the text part.

### DIFF
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -629,7 +629,7 @@ void DocSecRefList::parse()
         case CMD_ENDSECREFLIST:
           return;
         default:
-          warn_doc_error(parser()->context.fileName,parser()->tokenizer.getLineNr(),"Illegal command '%s%s' as part of a \\secreflist",
+          warn_doc_error(parser()->context.fileName,parser()->tokenizer.getLineNr(),"Illegal command '%c%s' as part of a \\secreflist",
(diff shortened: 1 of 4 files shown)


src/docnode.cpp
```diff
@@ -3214,58 +3210,59 @@ void DocPara::handleCite()
   parser()->tokenizer.setStatePara();
 }
 
-void DocPara::handleEmoji()
+void DocPara::handleEmoji(const QCString &cmdName, const int tokIn)
```
doxygen, line 3213: The `const` is kind of redundant for value types, and I would prefer to resolve the `TK_COMMAND_CHAR(tokIn)` at the caller side and pass in a char, to prevent accidental mixing of `tokIn` and other `tok`'s.

albert-github: Indeed const is redundant, would you like to have here a `const char *` or a `char`?
Currently `TK_COMMAND_CHAR(tokIn)` returns a string.

(I already used `tokIn` to prevent this potential mixing.

doxygen: Maybe best to change `TK_COMMAND_CHAR` to return a char and then pass the char and use `%c` in the warnings.

albert-github: Rereading the comment I see a `char` is  preferred.

PR #10585, opened 2024-01-22
Show emoji in HTML treeview

albert-github:
Show emoji in the HTML treeview as emoji and not as  text

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14013838/example.tar.gz)

### DIFF
--- a/src/textdocvisitor.cpp
+++ b/src/textdocvisitor.cpp
@@ -39,7 +39,7 @@ void TextDocVisitor::operator()(const DocSymbol &s)
 
 void TextDocVisitor::operator()(const DocEmoji &s)
 {
-  // the TextDocVisitor is only invokated for the JS part of the HTML output
+  // the TextDocVisitor is only invoked for the JS part of the HTML output
   const char *res = EmojiEntityMapper::instance().unicode(s.index());
   if (res)
   {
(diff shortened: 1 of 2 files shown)


src/textdocvisitor.cpp
```diff
@@ -39,10 +39,29 @@ void TextDocVisitor::operator()(const DocSymbol &s)
 
 void TextDocVisitor::operator()(const DocEmoji &s)
 {
-  const char *res = EmojiEntityMapper::instance().name(s.index());
+  // the TextDocVisitor is only invokated for the JS part of the HTML output
```
doxygen, line 42: typo: invokated->invoked

src/util.cpp
```diff
@@ -5635,7 +5637,8 @@ QCString parseCommentAsText(const Definition *scope,const MemberDef *md,
       if (result.at(i)==',' ||
           result.at(i)=='.' ||
           result.at(i)=='!' ||
-          result.at(i)=='?')
+          result.at(i)=='?' ||
+          result.at(i)=='}')    // good for UTF-16 characters and } otherwise also hood point to stop string
```
doxygen, line 5641: typo: hood -> good

doxygen:
@albert-github I think there is an issue with emoticons in section titles when using latex output as shown by your example (with and without this change), I get:

```
! Argument of \@sect has an extra }.
<inserted text> 
                \par 
l.3 ...tion \+:+1\+: \doxygenemoji{\+:+1\+:}{+1}}}
                                                  \label{pg_all_autotoc_md1}

```

albert-github:
@doxygen 
Regarding the `\emoji` command and also the `\ref`, `\subpage` and `\link` commands in LaTeX there are some problems. 
I found a solution already and working on a patch.

PR #10607, opened 2024-01-29
Better warning when in Fortran a program unit or routine is not closed

albert-github:
Analogues to #10587 we also need warnings for program units and as it can be nested we need a stack to record what is still open.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14087996/example.tar.gz)

(Found by Fossies for the rlab package)


src/fortranscanner.l
```diff
@@ -85,6 +86,13 @@ typedef yyguts_t *yyscan_t;
 enum ScanVar { V_IGNORE, V_VARIABLE, V_PARAMETER, V_RESULT};
 enum InterfaceType { IF_NONE, IF_SPECIFIC, IF_GENERIC, IF_ABSTRACT };
 
+struct blockState
+{
+  blockState(QCString str, int lineNr) : blockString(str), blockLineNr(lineNr) {}
+  QCString blockString;
+  int      blockLineNr = -1;
+};
+std::stack<blockState> g_blockStack;
```
doxygen, line 95: Global variables will create data races and crashes when using multiple threads (e.g. `NUM_PROC_THREADS!=1`). 
Assuming this state is per file, you need to put this variable in `fortranscannerYY_state` and access it via `yyextra`.

albert-github: I think the commit e87e708 from February 17, 2024 fixes this problem.

doxygen:
@albert-github made [some changes](https://github.com/doxygen/doxygen/commit/4d83858e593edc36dddd0a768546399187112f7b). Please check if I didn't break anything.

Issue #10631, opened 2024-02-06
Spurious .PP on Value headings in Macro Definition Documentation in man page 


duncan-roe:
If you download List.3.txt and run `man ./List.3.txt`, you will see the errant **Value:** lines straight away.
Download the other 2 and remove *.txt* suffices. Running `doxygen` will create *man/man3/List.3* (and link pages).
[doxyfile.txt](https://github.com/doxygen/doxygen/files/14173020/doxyfile.txt)
[linux_list.h.txt](https://github.com/doxygen/doxygen/files/14173028/linux_list.h.txt)
[List.3.txt](https://github.com/doxygen/doxygen/files/14173034/List.3.txt)

duncan-roe:
Aaargh! I've just realised things are much worse than just the *.PP*. The preceding **#define** line runs the first argument into the macro name. E.g. **#define INIT_LIST_HEADptr** should be **#define INIT_LIST_HEAD ptr**.
Would you like a separate issue raised for this? It is more serious for me as it messes up my post-processing script that gathers all items into a comma-separated list on the **NAME** line.

albert-github:
The problem:
```
#define INIT_LIST_HEADptr
```
is even more serious, it should actually read:
```
#define INIT_LIST_HEAD(ptr)
```
note the missing round brackets, as shown with the HTML output:
![image](https://github.com/doxygen/doxygen/assets/5223533/8b8316b0-1f66-4dc2-bc4f-59490639f402)

- The round bracket / space problem looks like to be introduced in the current master (looks like it is in the latest doxygen release, 1.10.0 still OK)
- The `.PP` problem looks like to be introduced in the 1.9.6 version

I think the current example will do.

albert-github:
The problem of the missing round brackets  is caused by:
```
Commit: 9a9fff28c9ecf451f8538eb81594d3dabd1d9ee0 [9a9fff2]
Date: Thursday, February 1, 2024 7:47:11 PM

Parameter names were shown at wrong place for function pointer types (fix)
```
as some methods for rtf and man are empty.


I've just pushed a proposed patch, pull request #10635

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.11.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
[...]

PR #10635, opened 2024-02-07
issue #10631 Spurious .PP on Value headings in Macro Definition Documentation in man page

albert-github:
Original .PP problem
- have the `.PP` start at a new line
- for RTF the word `Values` was included in the code block due to a missing `\par`

Regarding the problem with the missing round brackets:
- implement some of the original empty functions for RTF and man output

Extended examples: [example.tar.gz](https://github.com/doxygen/doxygen/files/14197269/example.tar.gz)

### DIFF
--- a/src/mangen.cpp
+++ b/src/mangen.cpp
@@ -114,6 +114,7 @@ ManCodeGenerator::ManCodeGenerator(TextStream *t) : m_t(t)
 
 void ManCodeGenerator::startCodeFragment(const QCString &)
 {
+  *m_t << "\n";
   *m_t << ".PP\n";
   *m_t << ".nf\n";
 }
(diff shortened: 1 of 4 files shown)


duncan-roe:
Hi Albert,
I applied your patch and the result looks to be as you intended. 
One nit: Why have .PP after the heading? It looks much better (to me anyway) to have it after the macro definition.
I made this change manually and present old and new versions for your inspection.
Macros with 1-line definition are fine as_is (e.g. *list_entry*).
Cheers ... Duncan.
[INIT_LIST_HEAD.3.orig.txt](https://github.com/doxygen/doxygen/files/14205251/INIT_LIST_HEAD.3.orig.txt)
[INIT_LIST_HEAD.3.new.txt](https://github.com/doxygen/doxygen/files/14205268/INIT_LIST_HEAD.3.new.txt)

albert-github:
Thanks for the quick test and the suggestion with the current and a proposed solution.

**new doxygen** version

image

**proposed** version

image

The layout is always a matter of discussion / personal preferences, but here I agree as in the proposed version the there is a better separation after the value of the e.g. define.

I've adjusted the PR.

PR #10637, opened 2024-02-08
Possibility to, locally, generate man output files from tests

albert-github:
Creating an easy possibility to, locally, also generate man output.


albert-github:
Thanks for merging, though looks like a similar / better version was already incorporated (#11389) on February 5, 2025. So best is to revert this (#10637) PR.

Issue #10636, opened 2024-02-08
weird behaviour with \hideinheritancegraph


TommyPec:
**Describe the bug**
hideinheritancegraph is sometimes skipped.

Ok, this needs some more explanation - as I fear we're using an undocumented (and brittle) feature.
The class `DerivedClassNamespace` have some documentation in the regular place (before the class definition), and some in an external file (doc/introspected-doxygen.h`).
The reason is because in ns-3 we generate part of the class documentation after the build with an introspection.

Now, I found out that:
1. If the classes are in a namespace the class documentation is merged, BUT (and here is the problem) if the class does have a \hideinheritancegraph, it does not have any effect **UNLESS** I put it in the introspected part.
2. If the class are not in a namespace (not shown in the attached tarball), the class documentation sometimes is merged, sometimes not. I've not been able to reproduce it consistently.
3. hideinheritancegraph in the templated class `TemplatedClass` works in the example, but it does not work when building the equivalent class in ns-3 (is named `Ptr`, which makes me mad because its inheritance graph contains more than 400 nodes, and I'd love to skip it. Otherwise it generates a warning from dot not being able to process it - which is not a big issue, but it might be the sign of an undefined behaviour.

**Expected behavior**
To not have the inheritance graph.

**Screenshots**

<img width="673" alt="Screenshot 2024-02-07 at 22 57 25" src="https://github.com/doxygen/doxygen/assets/169026/7e564795-aab6-4fe4-9979-45a21a5551b1">

<img width="551" alt="Screenshot 2024-02-07 at 22 59 35" src="https://github.com/doxygen/doxygen/assets/169026/34946d4b-be5c-4088-997d-9a4334003fcb">

**To Reproduce**

Open the attached archive. Use `doxygen doc/doxygen.conf`.
Build with `g++ --std=c++17 src/*.cc`, generate docs with `doxygen dox/doxygen.conf`

[DoxyTestGraphs.tgz](https://github.com/doxygen/doxygen/files/14204410/DoxyTestGraphs.tgz)

**Version**
1.11.0 (77cff916c1edfc5de8667e4de683c5dca6764ae2), macOS ARM

**Stack trace**
n/a

**Additional context**
Sorry for the weird issues... and please, don't say that we can't have the class documentation in two places, it's very useful.
However, we'd be totally ok to have to add a keyword to indicate that a piece of documentation is the continuation of another one.

albert-github:
[...]

PR #10638, opened 2024-02-08
issue #10636 weird behavior with `\hideinheritancegraph`

albert-github:
Implemented an explicit / implicit switch so the last explicit setting of `\inheritancegraph / `\hideinheritancegraph` will be used.

NOTE: this is just an implementation for the inheritancegraph and when acceptable it should be extended to other "show" / "hide" constructs.


src/classdef.cpp
```diff
@@ -5045,9 +5045,9 @@ QCString ClassDefImpl::inheritanceGraphFileName() const
   return m_impl->inheritFileName;
 }
 
-void ClassDefImpl::setTypeInheritanceGraph(CLASS_GRAPH_t e)
+void ClassDefImpl::setTypeInheritanceGraph(CLASS_GRAPH_t e, bool explicitSet)
```
doxygen, line 5048: Since this function only does something when `explicitSet` is `true` why call `setTypeInheritanceGraph` at all unless `explicitSet` is true (and then omit the parameter)?

src/entry.h
```diff
@@ -199,6 +199,7 @@ class Entry
     bool directoryGraph;      //!< do we need to draw the directory graph?
     bool collaborationGraph;  //!< do we need to draw the collaboration graph?
     CLASS_GRAPH_t inheritanceGraph; //!< type of inheritance graph?
+    bool inheritanceGraphExplicit;  //!< type of inheritance graph set implicit or explicit?
```
doxygen, line 202: I'd prefer the slightly longer`inheritanceGraphExplicitlySet` as a more explicit name.

albert-github:
I adjusted the code based on the 2 remarks
Next step will be to implement it for the other "show" / "hide" constructs.

doxygen:
@albert-github Looks like there is an issue with this change. Before the change the default setting for e.g. `callGraph` was stored in `Entry`, possibly changed but then always passed to `MemberDef`. Now, this passing only happens if either `\callgraph` or `\hidecallgraph` are present. The default stored in `MemberDef` is `false`, so if no `\callgraph` or `\hidecallgraph` are set, no callgraphs are generated even though the `CALL_GRAPH` option is enabled. So I think the solution is to store the `CALL_GRAPH` setting in `MemberDef` and change `enableCallGraph` into `overrideCallGraph` that can then change the default if a `\callgraph` or `\hidecallgraph` is explicit set. Same problem for other settings. 

I want to do a bit of restructuring, so I will take this along.

albert-github:
> The default stored in `MemberDef` is `false`

This doesn't look good, proposed solution does look like a valid idea.

PR #10647, opened 2024-02-11
cmake: always install man pages, only install for the binaries installed

eli-schwartz:
There are currently two problems with the manpage installation:

- Installing man pages is conditional on "build_doc", which otherwise controls whether the PDF and HTML is generated. However, the man pages are small and statically checked into git -- their purpose is not to fully document doxygen, but provide versions of the `--help` text which don't require running the program, and can be searched and indexed. Distributors *always* expect software to install the man pages, even when documentation is disabled.

- All the manpages are installed together, for 4 different programs. But depending on build configuration, 3 of those programs might not be built. It makes no sense to document the `--help` text for programs which are not installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,9 +223,10 @@ if (build_doc_chm)
     endif ()
 endif ()
 
+# always parse doc directory to at least install man pages
+add_subdirectory(doc)
 if (build_doc)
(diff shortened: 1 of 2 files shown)


albert-github:
I was thinking: 
- wouldn't it make more sense to place the man files on a separate directory?

PR #10652, opened 2024-02-13
Prevent `size_t` constructs to wrap around

albert-github:
When variables of `size_t` are subtracted  it is possible that the (integer) result gets negative but when `size_t` is unsigned this might wrap around with the effect that doxygen will terminate in an unpredictable way. Reformulation with additions.


albert-github:
Problem surfaced whit CGAL when a construct was offered after doxygen comment conversion that was not correct (this should be handled too) but doxygen should not unpredictable terminate (at different places or even sometimes run correctly).

PR #10661, opened 2024-02-17
Use icons instead of words for search result types

zyphlar:
The default external search behavior outputs words like

1. object <a href="#">MyInterface(int foo)</a>

for each search result, however for any sizable project this gets visually very cluttered while adding relatively little understanding. Iconography is probably a better way of helping people skim for what they want.

This PR is a work in progress and feedback/changes/rewrites welcome, but I think it would be valuable for everyone. The use of Google Fonts is optional, any similar symbol would be just as useful, I just wanted to minimize the number of asset files to worry about while maximizing the number of available free icons at my disposal.

Here's what the PR looks like:

## Stock, with some css tweaks:

```css
  .material-symbols-outlined {
    font-size: 1em;
  }
  .searchresult .material-symbols-outlined {
    margin-left: 0.5em;
    vertical-align: middle;
  }
```

Screenshot 2024-02-17 002551

## Or with doxygen-awesome-css in dark mode:

```css
  .material-symbols-outlined {
    font-size: 1em;
  }
  .searchresult .material-symbols-outlined {
    margin-left: 0.5em;
    vertical-align: bottom;
  }
```

Screenshot 2024-02-17 002631


albert-github:
I'm not sure whether a reference to the google fonts is a good choice (but you already wrote:  The use of Google Fonts is optional, any similar symbol would be just as useful, I just wanted to minimize the number of asset files to worry about while maximizing the number of available free icons at my disposal.):
- potential licensing / copyright issues
- need for an internet connection, one could have a local server running to do tests etc. or be inside a closed network that has no connection to the outside world.

I think it would be better to use a local set of icons, which would also give the possibility (through redefinition /  settings) to use ones own preferred icons.

Furthermore it might be better to have the icon before the text as now it is a bit "ragged".

Note: as you mention it is still work in progress maybe an idea to make this more clear by setting the PR into WIP mode (right hand side of the PR page).: 
image

templates/html/extsearch.js
```diff
@@ -78,16 +78,60 @@ function searchFor(query,page,count) {
       $.each(data.items, function(i,item){
         let prefix = tagMap[item.tag];
         if (prefix) prefix+='/'; else prefix='';
+
```
albert-github, line 81: Please check whether the list is complete, (see also the doc/extsearch.doc and src/searchindex.cpp), current list is (updated February 16, 2024):
```
          source, function, slot,
          signal, variable, typedef, enum, enumvalue, property, event, related,
          friend, define, file, namespace, concept, group, package, page, dir, module,
          constants, library, type, union, interface, protocol category, exception, class,
          struct, service, singleton
```

src/htmlgen.cpp
```diff
@@ -420,6 +420,8 @@ static QCString substituteHtmlKeywords(const QCString &str,
     if (!serverBasedSearch)
     {
       searchCssJs += "<script type=\"text/javascript\" src=\"$relpath^search/searchdata.js\"></script>\n";
+    } else {
+      searchCssJs += "<link rel=\"stylesheet\" href=\"https://fonts.googleapis.com/css2?family=Material+Symbols+Outlined:opsz,wght,FILL,GRAD@24,400,0,0\" />";
```
doxygen, line 424: I don't really like the external dependency. May break in the future and prevents off-line browsing. Better to bundle the set of icons with doxygen.

templates/html/extsearch.js
```diff
@@ -78,16 +78,60 @@ function searchFor(query,page,count) {
...
+        icon = "asterisk";
+
+        if (item.type == "dir")
```
doxygen, line 84: Suggest to use a `map` instead of `if..else if..else if`, i.e.

```js
const myMap = {
    "key 1" : "value 1", 
    "key 2" : "value 2", 
    ...
    "key n" : "value n"
};
```

zyphlar:
Thanks for the feedback! All good points. It may be possible to bundle a WebFont with appropriate licensing to reduce the internet dependency.

As for putting the icons on the left vs the right, I started out on the left but it visually clashed with the numbering, as if there were two bullets. To me the right hand side sort of has a "external link information, here's what you'll get if you click" feel, kinda like Google's "cached" or other link metadata. I don't have a strong opinion though, anyone with a good sense for design is welcome to improve. I wonder if the numbered bullets are even that important? Or maybe the icons float to the left of the number like a pre-bullet?

albert-github:
> As for putting the icons on the left vs the right, I started out on the left but it visually clashed with the numbering, as if there were two bullets. ... 

I'm not getting this, how does this clash? I was thinking something like: `<number> <icon> <text>`
Maybe you could shown an image of the "left side problem"

doxygen:
> > As for putting the icons on the left vs the right, I started out on the left but it visually clashed with the numbering, as if there were two bullets. ...
> 
> I'm not getting this, how does this clash? I was thinking something like: `<number> <icon> <text>` Maybe you could shown an image of the "left side problem"

The number also doesn't add that much, so the icon could also serve as the "bullet" instead of a number.

zyphlar:
@albert-github @doxygen I finally got around to working on this again, here's what it looks like with the "left hand side number clash problem" and with no numbers, just icons. I think no numbers just icons looks best, personally, though I wonder if people may prefer to have the option. To get rid of the external dependency without a whole bunch of SVG wrangling and file copying and duplication between light and dark and relative path nonsense, I checked out the Google Font license and copied the font file and its license into the repo. If we want to try and find two dozen other icons and not use Google at all that's OK but webfont technology is pretty slick, `<span class="material-symbols-outlined">icon_name</span>` generates a day or night colored icon with zero javascript, and it's Apache licensed.

Left hand side icon with numbers:

Screenshot from 2024-05-17 20-31-03

No numbers just icons:

Screenshot from 2024-05-17 20-32-14

Stock (no awesome-css):

Screenshot from 2024-05-17 22-21-40

I also print the result type in the `title` attribute so that arbitrary icons may be less confusing:

Screenshot from 2024-05-17 20-39-00

Please see the updated PR, depending on your decisions I think after cursory review this is pretty good to go!

zyphlar:
Also fun fact, this woff file contains a few instances of the bytes `0x0d0a` which means that git may interpret it as text and try to replace it with `0x0a` ... in other words, turn `\r\n` into `\n` in a binary file.

Anyway I think I fixed that issue lol :)

zyphlar:
@albert-github this should be ready if it looks good to you

src/htmlgen.cpp
Croydon: Also privacy concerns and legal headaches that come with it, depending on the country of residence. Please don't load external resources.

zyphlar:
@albert-github @doxygen hi! I think this is an unambiguously positive improvement, I've been running it on my site this whole time and it's pretty good/informative/clean. With each doxygen upgrade I recompile this branch to re-add the improvement since it looks old and frustrating without it :grin:

doxygen:
@zyphlar I've merged the changes and reduced the size of the font by a factor of 1000 by only including the icons needed and only by codepoint not name, using the following script [minify_material_symbol_font.sh](https://github.com/user-attachments/files/30970264/minify_material_symbol_font.sh). Please check if the result works for you.

Issue #9021, opened 2022-01-10
Support markdown links to sections/labels in a markdown file


lahwaacz:
Doxygen 1.9.3 supports [linking to other markdown pages](https://www.doxygen.nl/manual/markdown.html#md_page_header) with the syntax

    See [the other page](other.md) for more info.

It would be cool to also support

    See [a section in the other page](other.md#section) for more info.

albert-github:
A section names is an unique name so it does not require a file name, one can directly link to it like `#section` or `\ref section` instead of the `other.md#section`.

So I think this is already possible or did I miss something?

lahwaacz:
> A section names is an unique name so it does not require a file name, one can directly link to it like `#section` or `\ref section` instead of the `other.md#section`.

That is not obvious, especially when your output format is HTML where named anchors are unique per-document rather than per-project.

Markdown documents are also stand-alone, which has support even in Doxygen (since `[the other page](other.md)` works). Expecting uniqueness of named anchors across all Markdown documents in a project is rather restrictive, and useless.

It also does not work for named anchors written in HTML as `<a name="section"></a>`.

albert-github:
In your original post you were talking about `section`, due to the lack of an example I had to assume that you meant the doxygen `\section` command (or `\page` or `\*section`), this doesn't seem to be the case.

We need an example which makes it clear where you want to link to, so:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

lahwaacz:
Sure: [doxygen_test.zip](https://github.com/doxygen/doxygen/files/7841597/doxygen_test.zip)

albert-github:
When I run your example I get:
```
page2.md:3: warning: multiple use of section label 'section_1' while adding section, (first occurrence: .../page1.md, line 3)
```
so still the problem with the section labels.

That the links like `md_page2.html#section_1` work with the doxygen HTML output is a bit of luck and only works for the HTML type of output but not for LaTeX / PDF, Docbook, RTF, ....

lahwaacz:
> so still the problem with the section labels.

As I said, Markdown documents are a priori standalone, so the problem is with Doxygen requiring uniqueness where it is not needed.

> That the links like md_page2.html#section_1 work with the doxygen HTML output is a bit of luck and only works for the HTML type of output but not for LaTeX / PDF, Docbook, RTF, ....

[...]

PR #10663, opened 2024-02-17
Fix invalid generated link when using Markdown GitHub style link to a section in another file

jnkvca:
Fix for a bug in HTML documentation generated from Markdown files using GitHub style. When I have two Markdown pages and I want to link from the first one to the second one's subsection directly, I would write something like this:

```md
This is the [link](second-page.md#subsection).
```

This works on GitHub and other Markdown previewers such as the one in VS Code, but in Doxygen generated HTML this does not generate a correct link. The link is treated as a regular URL so basically it tries to open `.../out/html/second-page.md#subsection` link on my generated documentation, which does not exist.

I know I can rewrite my Markdown to look like this:

```md
This is the [link](#subsection).
```

And this would work correctly with Doxygen - it opens `.../out/html/md_second-page.html#subsection` which indeed points to the subsection of my second page that I wanted it to point. However this does not work on GitHub nor does it work on other Markdown previewers. For consistency I think that the first way of linking to sections that are in different files should also be supported because I and most likely other people who use the GitHub Markdown style option in Doxygen want the Markdown files to work not just with Doxygen generated documentation but in other Markdown previewers and editors as well.

My change makes the parser recognize `second-page.md#subsection` links as `#subsection` links, so there is no additional logic to verify that the `#subsection` exists on that specific page so technically any page (or text) can be specified in place of `second-page.md` part and the link would still work the same. This is not necessarily that problematic as Doxygen makes the section IDs unique. So for example if we had two sections named "Subsection" and even if they're in their own separate Markdown files, one will get the ID "subsection" and the other will get the ID "subsection-1". So to link to the second one you must use the "#subsection-1" regardless from which page you're trying to link to it, else you're going to make a link to the first "Subsection" section. I didn't want to touch any logic with link resolution myself, but I think that maybe when using GitHub Markdown style the parser should first try to resolve the `#subsection` links to sections named "Subsection" in the current file and only later look elsewhere. Similarly if the link is specified like `second-page.md#subsection` it should attempt to link to that section in that specific page first. This would make the logic more consistent with the GitHub Markdown logic.

In any way, here is the minimal example showcasing the bug which gets fixed by my PR: 
~[markdown_link_bug_minimal_example.zip](https://github.com/doxygen/doxygen/files/14319788/markdown_link_bug_minimal_example.zip)~ (EDIT: wrong file provided by accident, scroll down for correct one)

Also, this PR possibly also fixes the problems in #9021.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -1529,7 +1529,7 @@ int Markdown::Private::processLink(const std::string_view data,size_t offset)
     }
     else if ((lp=link.find('#'))!=-1 || link.find('/')!=-1 || link.find('.')!=-1)
     { // file/url link
-      if (lp!=-1 && !isURL(link) && Config_getEnum(MARKDOWN_ID_STYLE)==MARKDOWN_ID_STYLE_t::GITHUB)
+      if (lp==0 || (lp>0 && !isURL(link) && Config_getEnum(MARKDOWN_ID_STYLE)==MARKDOWN_ID_STYLE_t::GITHUB))
       {
         out+="@ref \"";
         out+=AnchorGenerator::addPrefixIfNeeded(link.mid(lp+1).str());
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -1527,12 +1527,12 @@ int Markdown::Private::processLink(const std::string_view data,size_t offset)
       }
       out+="\"";
     }
-    else if (link.find('/')!=-1 || link.find('.')!=-1 || link.find('#')!=-1)
+    else if ((lp=link.find('#'))!=-1 || link.find('/')!=-1 || link.find('.')!=-1)
     { // file/url link
-      if (link.at(0) == '#')
+      if (lp!=-1 && !isURL(link))
       {
         out+="@ref \"";
-        out+=AnchorGenerator::addPrefixIfNeeded(link.mid(1).str());
+        out+=AnchorGenerator::addPrefixIfNeeded(link.mid(lp+1).str());
         out+="\" \"";
         out+=substitute(content.simplifyWhiteSpace(),"\"","&quot;");
         out+="\"";
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -1529,7 +1529,7 @@ int Markdown::Private::processLink(const std::string_view data,size_t offset)
     }
     else if ((lp=link.find('#'))!=-1 || link.find('/')!=-1 || link.find('.')!=-1)
     { // file/url link
-      if (lp!=-1 && !isURL(link))
+      if (lp!=-1 && !isURL(link) && Config_getEnum(MARKDOWN_ID_STYLE)==MARKDOWN_ID_STYLE_t::GITHUB)
       {
         out+="@ref \"";
         out+=AnchorGenerator::addPrefixIfNeeded(link.mid(lp+1).str());


albert-github:
As far as I can tell this does not work properly. 
With the patch:
- the first link points to the "#subsection" on the first page and not tho the "#subsection" of the second page.
- the second link in the old situation points to "index.html#subsection", this could be improved so it would point to "#subsection" (the later has already been done for a number of other links, see #10306 / #10338).

jnkvca:
Ah I see, I have accidentally posted the wrong files. Sorry about that, here are the correct ones:
[markdown_link_bug_minimal_example_correct.zip](https://github.com/doxygen/doxygen/files/14322428/markdown_link_bug_minimal_example_correct.zip)

The example wasn't supposed to have duplicate section names, that was me just testing how it would work in that case. But having duplicate section names breaks links altogether because Doxygen will assign unique IDs to duplicate sections, for example having multiple sections named "Subsection" will produce IDs "subsection", "subsection_1", "subsection_2" etc (most likely in the order that the input files were specified). So using "#subsection" or "anything.md#subsection" (with this change) from any file will always link to the first one, if you wanted to link to any of the other ones you have to use "#subsection_1" or "anything.md#subsection_1". I don't like this behavior but as long as the section names are unique the behavior will be identical to GitHub (if there are duplicate section names, the behavior is broken regardless of if you use this patch).

Ideally, this is how it should work (when GitHub Markdown style is used):
- if you specify "#subsection" it should look for a section "Subsection" in the current file, regardless of whether its ID is "subsection" or "subsection_1" etc.
- if you specify "anything.md#subsection" it should look for a section "Subsection" in anything.md file, regardless of other files containing Subsections with ID "subsection"

Tl;dr: this patch adds the support for using link formats such as "anything.md#subsection", but does not change the old behavior of Doxygen when duplicate section names are present. I could probably even look into fixing the linking behavior in GitHub mode to match the behavior I described above when I have some extra time (if you agree that is how it should work).

albert-github:
Indeed GitHub uses more local anchors whilst doxygen uses the explicit defining and global approach (which is in my opinion better and also more flexible).

As the given approach is only valid for the GitHub approach this might also be necessary to incorporate this in the test in the code (i.e. probably the `if (lp!=-1 && !isURL(link))` part).

jnkvca:
Added the config check.

src/markdown.cpp
```diff
@@ -1529,7 +1529,7 @@ int Markdown::Private::processLink(const std::string_view data,size_t offset)
     }
     else if ((lp=link.find('#'))!=-1 || link.find('/')!=-1 || link.find('.')!=-1)
     { // file/url link
-      if (lp!=-1 && !isURL(link))
+      if (lp!=-1 && !isURL(link) && Config_getEnum(MARKDOWN_ID_STYLE)==MARKDOWN_ID_STYLE_t::GITHUB)
```
doxygen, line 1532: Looks like this still changes the behavior in case `MARKDOWN_ID_STYLE` is not `GITHUB`. Probably the expression should be:
```c++
  if (lp!=-1 && (Config_getEnum(MARKDOWN_ID_STYLE)==MARKDOWN_ID_STYLE_t::DOXYGEN) || !isURL(link)) { ... }
```

jnkvca: The intended logic is if `#` character is found and either `MARKDOWN_ID_STYLE` is `DOXYGEN` or link is not a valid URL, correct? I'm asking because of missing parenthesis.

If so, then valid URLs that contain a `#` in them (for example `https://example.com/index.html#some-heading-id` will not be recognized as URLs in Doxygen style mode and instead it will attempt to link them to a section named `some-heading-id`. You can confirm this with the example files I posted.

jnkvca: Maybe this is what you want:
```cpp
if (lp==0 || (lp>0 && !isURL(link) && Config_getEnum(MARKDOWN_ID_STYLE)==MARKDOWN_ID_STYLE_t::GITHUB)) { ... }
```

The left part of the `||` is the old logic and the right part is the new added logic.

jnkvca: @doxygen @albert-github Pushed the change, you guys can check if it seems valid to you. Link parsing in Doxygen mode should remain unchanged with this commit, only GitHub parsing is affected now.

PR #10667, opened 2024-02-18
Fix table overlaping footer

### LINKED ISSUE #6532: pdflatex: parameter table reaches into footer
Using doxygen 1.8.9.1 all parameter tables contain page-breaks such that they fit into the bodies of the pages.

In higher versions of doxygen (here: used the latest version 1.8.14) the page-break in parameter tables is sometimes missing, such that the table reaches into the footer of the page.

**Table reaches into the footer:**
doxygen_pdflatex_issue_-_table_reaches_into_footer

**Table contains a page break:**
doxygen_pdflatex_issue_-_pagebreak_in_table

The zip archive contains the doxygen output generated for a header which demonstrates the issue:

[doxygen_pdflatex_issue.zip](https://github.com/doxygen/doxygen/files/2432117/doxygen_pdflatex_issue.zip)

jreybert:
This bug lies in Latex longtable since ages. doxygen, and  pandoc, suffer this bug since. See #6532.

The issue is that a table may overlap the footer. There are even cases where the bottom of a table is cropped at the end of the page.

The culprit is the longtable latex package. This bug is known for a long time, and a patch has been proposed since 2003. This package is not maintained anymore.

- https://github.com/jgm/pandoc/issues/2889#issuecomment-245879494
- https://www.latex-project.org/cgi-bin/ltxbugs2html?pr=tools/3512

The proposed patch has been applied without any conflict.

The result is that there is not more table overlap. See example in pull request #10667.

Fixes #6532

### DIFF
--- a/templates/latex/longtable_doxygen.sty
+++ b/templates/latex/longtable_doxygen.sty
@@ -164,7 +164,7 @@
   \ifdim \dimen@>\z@\vfil\break\fi
       \global\@colroom\@colht
   \ifvoid\LT@foot\else
-    \advance\vsize-\ht\LT@foot
+    \global\advance\vsize-\ht\LT@foot
     \global\advance\@colroom-\ht\LT@foot
     \dimen@\pagegoal\advance\dimen@-\ht\LT@foot\pagegoal\dimen@
     \maxdepth\z@
@@ -193,9 +193,13 @@
     \LT@final@warn
   \fi
   \endgraf\penalty -\LT@end@pen
+  \ifvoid\LT@foot\else
+    \global\advance\vsize\ht\LT@foot
+    \global\advance\@colroom\ht\LT@foot
+    \dimen@\pagegoal\advance\dimen@\ht\LT@foot\pagegoal\dimen@
+  \fi
   \endgroup
   \global\@mparbottom\z@
-  \pagegoal\vsize
   \endgraf\penalty\z@\addvspace\LTpost
   \ifvoid\footins\else\insert\footins{}\fi}
 \def\LT@nofcols#1&{%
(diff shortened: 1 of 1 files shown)


jreybert:
This example shows tables overlaping footer.

[mainpage.md](https://github.com/doxygen/doxygen/files/14324708/mainpage.md)

There is no more overlap after the patch.

albert-github:
- Did you check the doxygen_manual.pdf after this fix?
- Did you check the individual doxygen tests when generating the  the pdf output (and keeping the results)?

jreybert:
> * Did you check the doxygen_manual.pdf after this fix?
> * Did you check the individual doxygen tests when generating the  the pdf output (and keeping the results)?

Nope. Is there any documentation which summarize how to contribute? Which tests should be run and how?

albert-github:
Unfortunately there is no instruction for it.

>  forget my dumb question, I am right now running make doc and make tests...

it is not a dumb question (especially in respect to the options for the `make tests ...`.


Most tests are automatically run but to check the LaTeX output properly one has to do it manually and visually.
- doxygen manual `make docs` which results (a.o) in a file `latex/doxygen_manual.pdf`, which should also be present as an artifact at https://github.com/doxygen/doxygen/actions/runs/7952202926?pr=10667
- the tests, these have to be run locally to keep the output and be amble to inspect it
  - `make tests TEST_FLAGS="--pdf --keep` (use e.g also `--pool=4` to make the generation a bit faster, it will be in parallel)
  - check testing/test_output_*/latex/refman.pdf

jreybert:
Note for myself. I'll try to improve some README to build and test the doxygen project:

```console
$ cmake . -B build -Dbuild_doc=ON
```

I struggle with installing some dependency, e.g `newunicodechar.sty` on Rocky8. I'll give it a try on a Debian system later.

doxygen:
I checked the manual and tests. Didn't spot any issues.

PR #10679, opened 2024-02-22
Formula doesn't render due to empty lines.

albert-github:
When having a formula like:
```
  @f[

      A = \left(\begin{array}{cc}
                                  Ixx & Ixy \\
                                  Ixy & Iyy \\
           \end{array}\right)

  @f]
```
this doesn't render in LaTeX but renders under MathJax. The problem is the starting and ending empty lines inside the formula.

Stripping away the starting and ending empty lines.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14375378/example.tar.gz)

(originally found by Fossies for the rawtherapeepackage)

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -462,6 +462,8 @@ struct commentscanYY_state
   OutputContext    inContext;              // are we inside the brief, details or xref part
   bool             briefEndsAtDot = FALSE; // does the brief description stop at a dot?
   QCString         formulaText;            // Running text of a formula
+  QCString         formulaPreText;         // Start command of a formula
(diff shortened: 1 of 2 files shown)


mosra:
I'm hitting the same issue after upgrading to 1.12 on my end. I can work around it by removing the extra empty lines in my code, but it'd be nice if this fix could get merged.

Thank you! :)

doxygen:
@albert-github I restructured the code a bit (see [this commit](https://github.com/doxygen/doxygen/commit/589554349c563b34812b52cb1ecb165df7a352d9)) such that existing function `stripEmptyLines()` could be reused. Let me know if I overlooked something.

PR #10687, opened 2024-02-23
Spelling correction Dutch translation

albert-github:
Spelling correction Dutch translation


albert-github:
I also see a number of times the translation:
```
is gegenereerd op grond van
```
wouldn't it be better to use:
```
is gegenereerd op basis van
```

PR #10715, opened 2024-03-07
latexgen: fix doxy*section references

jreybert:
This commit fixes an advanced usage of doxygen, more especially latex doxylink usage.

Standard doxylink is a simple hyperlink:
```latex
\newcommand{\doxylink}[2]{
  \mbox{\hyperlink{#1}{#2}}
}
```

This simple reference is OK for digital pdf. But, in case of printed document, it lacks location information.

We extended doxylink like this, to add section number and title:
```latex
\renewcommand{\doxylink}[2]{
  \hyperlink{#1}{#2}, in \ref{#1} \nameref{#1}
}
```

The latex compiles well. But the generated latex is wrong. In the pdf, we observe that the hyperlink is good, but the `ref` and `nameref` point to the previous section.

Digging into generated .tex files, we found the following construct:
```latex
\Hypertarget{header_8h_ad8a2dfa8cbec508ec4cf97b2c5452797}\label{header_8h_ad8a2dfa8cbec508ec4cf97b2c5452797}
\index{header.h@{header.h}!bar@{bar}}
\index{bar@{bar}!header.h@{header.h}}
\doxysubsubsection{\texorpdfstring{bar()}{bar()}}
```

We can see that the `\label` is placed before `\doxysubsubsection`. doxylink points to `label`. And in this case, `\label` is part of the previous `doxy*section`.

This commit fixes this behavior, by setting the `label` after the `doxy*section`.

The resulted generated .tex is the following:
```latex
\index{header.h@{header.h}!bar@{bar}}
\index{bar@{bar}!header.h@{header.h}}
\doxysubsubsection{\texorpdfstring{bar()}{bar()}}
{\footnotesize\ttfamily
\Hypertarget{header_8h_ad8a2dfa8cbec508ec4cf97b2c5452797}\label{header_8h_ad8a2dfa8cbec508ec4cf97b2c5452797}
```

This behavior can be tested with a simple header file like this:
```c
/** @file header.h */

/** foo function */
void
foo();
/** bar function, see @ref foo */
void
bar();
```

Nota bene:
* We can see that `\Hypertarget` and `\label` are now embedded in `{\footnotesize\ttfamily` environment. That seems OK, and make the diff very simple.
* this commit fixes our specific usecase. It does not seem to break anything in pdf generation, but I may miss something.
* futhermore, this has not been tested with other generations: html, manpage, docbook...

### DIFF
--- a/src/docbookgen.cpp
+++ b/src/docbookgen.cpp
@@ -936,6 +936,10 @@ void DocbookGenerator::endDoxyAnchor(const QCString &,const QCString &)
 {
 DB_GEN_C
 }
+void DocbookGenerator::addLabel(const QCString &,const QCString &)
+{
+DB_GEN_C
+}
 void DocbookGenerator::startMemberDocName(bool)
 {
 DB_GEN_C
--- a/src/docbookgen.h
+++ b/src/docbookgen.h
@@ -220,6 +220,7 @@ class DocbookGenerator : public OutputGenerator
                          const QCString &anchor,const QCString &name,
                          const QCString &args);
     void endDoxyAnchor(const QCString &fileName,const QCString &anchor);
+    void addLabel(const QCString &,const QCString &);
     void writeLatexSpacing(){DB_GEN_EMPTY}
     void writeStartAnnoItem(const QCString &,const QCString &,
                             const QCString &,const QCString &){DB_GEN_NEW};
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -1675,6 +1675,10 @@ void HtmlGenerator::endDoxyAnchor(const QCString &,const QCString &)
 {
 }
 
+void HtmlGenerator::addLabel(const QCString &,const QCString &)
+{
+}
+
 void HtmlGenerator::startParagraph(const QCString &classDef)
 {
   if (!classDef.isEmpty())
--- a/src/htmlgen.h
+++ b/src/htmlgen.h
@@ -203,6 +203,7 @@ class HtmlGenerator : public OutputGenerator
                          const QCString &anchor,const QCString &name,
                          const QCString &args);
     void endDoxyAnchor(const QCString &fName,const QCString &anchor);
+    void addLabel(const QCString &,const QCString &);
     void writeLatexSpacing() {}
     void writeStartAnnoItem(const QCString &type,const QCString &file,
                             const QCString &path,const QCString &name);
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -1565,15 +1565,19 @@ void LatexGenerator::startDoxyAnchor(const QCString &fName,const QCString &,
     if (!anchor.isEmpty()) m_t << "_" << anchor;
     m_t << "}";
   }
+}
+
+void LatexGenerator::endDoxyAnchor(const QCString &/* fName */,const QCString &/* anchor */)
+{
+}
+
+void LatexGenerator::addLabel(const QCString &fName, const QCString &anchor)
+{
   m_t << "\\label{";
   if (!fName.isEmpty()) m_t << stripPath(fName);
   if (!anchor.isEmpty()) m_t << "_" << anchor;
   if (m_insideTableEnv) m_t << "}";
   m_t << "} \n";
-}
-
-void LatexGenerator::endDoxyAnchor(const QCString &/* fName */,const QCString &/* anchor */)
-{
 }
 
 void LatexGenerator::writeAnchor(const QCString &fName,const QCString &name)
--- a/src/latexgen.h
+++ b/src/latexgen.h
@@ -191,6 +191,7 @@ class LatexGenerator : public OutputGenerator
     void endMemberDoc(bool);
     void startDoxyAnchor(const QCString &,const QCString &,const QCString &,const QCString &,const QCString &);
     void endDoxyAnchor(const QCString &,const QCString &);
+    void addLabel(const QCString &,const QCString &);
     void writeChar(char c);
     void writeLatexSpacing() { m_t << "\\hspace{0.3cm}"; }
     void writeStartAnnoItem(const QCString &type,const QCString &file,
--- a/src/mangen.cpp
+++ b/src/mangen.cpp
@@ -542,6 +542,10 @@ void ManGenerator::startDoxyAnchor(const QCString &,const QCString &manName,
     }
 }
 
+void ManGenerator::addLabel(const QCString &,const QCString &)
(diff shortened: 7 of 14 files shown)


albert-github:
I think it looks like a small improvement, but still not 100% optimal.
In the situation before the fix we get when clicking on the `@ref foo`:

image


In the new situation we get:

image


Before we still were more or less in the old function, but in the new situation we are inside the new paragraph, it would be nicer when the **2.1.1.2 foo()** would be on top like (just got it by scrolling the page):

image

though I don't see a quick solution for this.

I extended the example slightly (Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14523440/example.tar.gz))

And in the old situation I got:
image

and in the new situation:
image


So missing the top of the "foo".

In my opinion the current solution is less good than the old one.
(the solution should also look at the consequences for the other output formats like RTF, Docbook)

jreybert:
Thanks @albert-github for the review. Indeed, as I susptected, there are some side effects.

For the pdf/latex one: IMO this is more important to have correct section name/numbering (when extending `\doxylink`) than landing click position.

For the html though, this side effect is killer. I don't know doxygen source code enough, so there may be more elegant solution, but doing something like this may work:
```c
        // considering something like doctype exists in this context
        if (docytype == latex) {
          ol.startMemberDoc(ciname,name(),memAnchor,name(),memCount,memTotal,showInline);
          ol.startDoxyAnchor(cfname, cname, memAnchor, doxyName, doxyArgs);
        }
        else {
          ol.startDoxyAnchor(cfname, cname, memAnchor, doxyName, doxyArgs);
          ol.startMemberDoc(ciname,name(),memAnchor,name(),memCount,memTotal,showInline);
        }
```

albert-github:
The original code is 
```
          ol.startDoxyAnchor(cfname, cname, memAnchor, doxyName, doxyArgs);
          ol.startMemberDoc(ciname,name(),memAnchor,name(),memCount,memTotal,showInline);
```
and doxygen has some functions for, in a clever way, disabling / enabling output types, the code would be :
```
          ol.disable(OutputType::Latex);
          ol.startDoxyAnchor(cfname, cname, memAnchor, doxyName, doxyArgs);
          ol.enable(OutputType::Latex);
          ol.startMemberDoc(ciname,name(),memAnchor,name(),memCount,memTotal,showInline);
          ol.disableAllBut(OutputType::Latex);
          ol.startDoxyAnchor(cfname, cname, memAnchor, doxyName, doxyArgs);
          ol.enableAll();
```

albert-github:
> For the pdf/latex one: IMO this is more important to have correct section name/numbering (when extending \doxylink) than landing click position.

I don't understand this as I didn't see a problem with the "section name/numbering".

jreybert:
> I don't understand this as I didn't see a problem with the "section name/numbering".

This is the root cause of this merge request, as explained in the description. We try to extend `doxylink` command, to add some context for pdf link (for example, when the documentation is printed).

I extended the example, with the file `redef_doxylink.tex`:
```latex
\renewcommand{\doxylink}[2]{%
      \hyperlink{#1}{#2}, in \ref{#1} \nameref{#1}%
}
```

[doxyexample.tar.gz](https://github.com/doxygen/doxygen/files/14525790/doxyexample.tar.gz)

With original doxygen:
image
[refman.pdf](https://github.com/doxygen/doxygen/files/14525735/refman.pdf)

As you can see, `bar function, see foo, in 2.1.1.3 bar2()`, reference to `foo` is wrong, it points to `2.1.1.3 bar2()`. It should point to `2.1.1.4 foo()`.

With merge request doxygen:
image
[refman.pdf](https://github.com/doxygen/doxygen/files/14525720/refman.pdf)

The reference to foo is correct.

albert-github:
Thanks for the explanation, the original behavior is indeed not what one wants (doesn't show up with the default doxygen version, but when using some advanced redefinitions this can cause problems, maybe also with other output formats when they use some redefinitions).

jreybert:
> So missing the top of the "foo".

OK, I tested the html output. I understand now that when you say "missing", it means that it is hidden when arriving on the link target. But the top actually exists. So this is the same result than for the pdf: the anchor has moved, as it is within the section, pdf/html jump below the section title.

The merge request is made like this: the anchor MUST be within the section, so that `\ref` and `\refname` are correct. 

Maybe the correct solution would be to separate `Hypertarget` and `label`, like this:
```latex
\Hypertarget{header_8h_ad8a2dfa8cbec508ec4cf97b2c5452797}
\index{header.h@{header.h}!bar@{bar}}
\index{bar@{bar}!header.h@{header.h}}
\doxysubsubsection{\texorpdfstring{bar()}{bar()}}
\label{header_8h_ad8a2dfa8cbec508ec4cf97b2c5452797}
```

albert-github:
Sorry about the confusion about "missing" indeed I meant "scrolled off the page" (like it is also with the LaTeX version).
By head I don't think `Hypertarget` and `label` have to be at the same place.
Some experiments should be mad to see whether the output is correct (with and without the redefinition of `doxylink`)

jreybert:
Before I go deeper, are you OK with this merge request ?

albert-github:
With the explanation given about the problem with LaTeX I think the PR will make sense when the change indicated in https://github.com/doxygen/doxygen/pull/10715#issuecomment-1983622220 is used (though I don't decide).

jreybert:
The commit has been updated, and the commit message too. The actual behavior seems correct:
* the hypertarget is placed before the section, than when we click in pdf or html, the landing is BEFORE the section.
* the label is placed AFTER the section, then `ref` and `refname` are correct.

With this new commit, latex is the only output updated. All the other generator outputs are unchanged.

This commit fixes this behavior, by splitting the function `startDoxyAnchor` into
`startDoxyAnchor` and `addLabel`. Then, the `Hypertarget` is placed before the `doxy*section`,
and the `label` is placed after.

We want the `Hypertarget` to be placed before the `doxy*section`. It allows, when the user click,
to land just before the section title, and then see the section title.

The resulted generated .tex is the following:
```latex
\Hypertarget{header_8h_a49a4b11e50430aa0a78de989ea99e082}\index{header.h@{header.h}!bar@{bar}}
\index{bar@{bar}!header.h@{header.h}}
\doxysubsubsection{\texorpdfstring{bar()}{bar()}}
{\footnotesize\ttfamily \label{header_8h_a49a4b11e50430aa0a78de989ea99e082}
```

albert-github:
I only read the adjusted code and it looks OK to me, only LaTeX is affected, all other output formats implement `addLabel` as an empty function.

PR #10724, opened 2024-03-10
Trying to fix #10327

### LINKED ISSUE #10327: undocumented method redirection
**Describe the bug**

// file doxygen.h
#pragma once

// Magique doc
struct Magique
{
    /*!
     * @brief doc for Magique::estDo()
     * @return void
     */
    void estDo();
};

//-------------------
// file C.h
struct c
{

    void Magique();
};

C::Magique is undocumented but in class C html description ( source code _c_8h_source.html), **C::Magique** redirect to class **Magique**
**Expected behavior**
C::Magique must not be clickable

**Screenshots**
class_C_html
redirection

**To Reproduce**
[testDoxy.zip](https://github.com/doxygen/doxygen/files/12725899/testDoxy.zip)

**Version**
1.9.8 (c2fe5c3e4986974eb2a97608b24086683502f07f)
windows binary

whatacold:
Hi,

This would fix the problem, but I'm not 100% sure if it will cause other problems. Just try it.

### DIFF
--- a/src/code.l
+++ b/src/code.l
@@ -69,7 +69,7 @@ typedef yyguts_t *yyscan_t;
 #include "moduledef.h"
 
 // Toggle for some debugging info
-//#define DBG_CTX(x) fprintf x
+//#define DBG_CTX(x) do {fprintf(stderr, "line %d %s() :", __LINE__, __func__); fprintf x;} while(0)
 #define DBG_CTX(x) do { } while(0)
 
 #define YY_NO_UNISTD_H 1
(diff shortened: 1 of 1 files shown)


albert-github:
Thanks for the attempt, but I think that it might fix the appearance of the very special case presented but (also seen the complexity of `generateClassOrGlobalLink`) there are other cases that now will have problems as there are no links generated anymore for this part.

whatacold:
OK, thanks for the extra info.

> there are other cases that now will have problems as there are no links generated anymore for this part.

This is what I'm uncertain with, I will try to figure it out.

albert-github:
Note: that the `generateClassOrGlobalLink` also exists in other `*code.l` in a bit different form, maybe some changes will be necessary there later on as well.

PR #10737, opened 2024-03-14
Incorrect handling command character in comment conversion

albert-github:
In case we have a construct like `///@` that should be handled by commentcnv this was not done correctly as the command character was taken a bit early.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14603474/example.tar.gz)


albert-github:
Example for second commit: [example2.tar.gz](https://github.com/doxygen/doxygen/files/14603670/example2.tar.gz)

PR #10746, opened 2024-03-17
Unique labels for `\include{doc}`

albert-github:
With the `raise` option with the `\internal{doc}` command it was possible to raise the level of the used sections to the sub-sections specified. Problem would still be that the same `label_id` would have been used which would give warnings like:
```
 warning: multiple use of section label 'cmd_1' while adding section
```
by creating also the `label` option this can be prevented, the specified value will be, recursively, appended to the given `label_id`. This would not only happen for section ids but also for anchor ids.

- option handling in commentcnv
  - more flexibility in the order in which the options can be used (only a fixed order was possible so far).
  - add handling of the new option `label`
- comentscan handling of the `label` with the `\iraise` command
- updating documentation
  - for the new option
  - making `\includedoc` also have the possibility to use the `raise` and `label` options.
- updating the internal documentation of `\iraise` to accommodate the `label_id`.

(everything also applies the `\snippet` type of commands)

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14628213/example.tar.gz)


doxygen:
@albert-github I've changed the `label` option to `prefix` which then also adds a prefix instead of a suffix. This makes the behavior more like a specifying a namespace. I've also added an `\iprefix` command that is inserted to also change `\ref` and `\link` targets when looking for sections in a page that was included when the `prefix` option is used. Otherwise, local references in an included page would break because the section label was changed, but a reference to it in the same page was not.

Let me know if you see any issues.

Issue #10733, opened 2024-03-12
`std::source_location` not found using clang-assisted parsing


tobiasleibner:
**Describe the bug**
When using clang-assisted parsing with a compilation database, C++20's `std::source_location` from the `source_location` header cannot be found, although it is available during normal compilation without doxygen. It also works if doxygen's default parser is used.

**To Reproduce**
Execute:
```shell
git clone https://github.com/tobiasleibner/DoxygenSourceLocationReproducer.git
cd DoxygenSourceLocationReproducer
mkdir build && cd build
cmake -DCMAKE_CXX_COMPILER="/usr/bin/clang++" ..
make reproducer_lib
make doxygen_docs
```
Output:
```
Generating file sources...
error: DoxygenSourceLocationReproducer/src/example.h:15:13: error: use of undeclared identifier 'std' [clang]
error: DoxygenSourceLocationReproducer/src/example.h:18:5: error: use of undeclared identifier 'std' [clang]
error: DoxygenSourceLocationReproducer/src/example.h:15:45: error: use of undeclared identifier 'std' [clang]
error: DoxygenSourceLocationReproducer/src/example.cpp:7:18: error: use of undeclared identifier 'std' [clang]
error: DoxygenSourceLocationReproducer/src/example.h:15:13: error: use of undeclared identifier 'std' [clang]
error: DoxygenSourceLocationReproducer/src/example.h:18:5: error: use of undeclared identifier 'std' [clang]
error: DoxygenSourceLocationReproducer/src/example.h:15:45: error: use of undeclared identifier 'std' [clang]
Generating code for file DoxygenSourceLocationReproducer/src/example.h...
```

**Expected behavior**
No error during doc generation.

**Version**
```
$ doxygen --version
1.10.0 (ebc57c6dd303a980bd19dd74b8b61c8f3f5180ca)
```
Tested on Ubuntu 22.04
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.4 LTS
Release:        22.04
Codename:       jammy
```
Doxygen was installed by
```
wget https://www.doxygen.nl/files/doxygen-1.10.0.linux.bin.tar.gz
tar -xf doxygen-1.10.0.linux.bin.tar.gz
cd doxygen-1.10.0
sudo make install
```

**Additional context**
[...]

PR #10757, opened 2024-03-19
Fix clang-assisted parsing with clang >= 14

tobiasleibner:
See #10733. To fix clang-assisted parsing with recent clang versions, we can check for `--` as the second to last argument and then pass the filename separately.

### DIFF
--- a/src/clangparser.cpp
+++ b/src/clangparser.cpp
@@ -235,8 +235,23 @@ void ClangTUParser::parse()
 
   // let libclang do the actual parsing
   //for (i=0;i<argv.size();i++) printf("Argument %d: %s\n",i,argv[i]);
-  p->tu = clang_parseTranslationUnit(p->index, nullptr,
-                                     argv.data(), static_cast<int>(argv.size()), p->ufs.data(), numUnsavedFiles,
+  const char * filename_arg = nullptr;
+  auto num_args = argv.size();
+  // Usually, the last compile command (last entry of argv) is the filename of the source
+  // file to parse. It does not matter to clang_parseTranslationUnit if we pass the file name
+  // separately in its second argument or if we just pass it a nullptr as the second
+  // argument and pass the file name with the other compile commands.
+  // However, in some cases (e.g., starting from Clang 14, if we are parsing a header file, see
+  // https://github.com/doxygen/doxygen/issues/10733), argv includes a "--" as second to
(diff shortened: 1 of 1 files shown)


tobiasleibner:
The current version does not work since we cannot assume that `--` is the second to last argument as we are adding user-defined options and `-w`. I will see if I can fix this.

tobiasleibner:
The current version works in my local testing, but we probably should include some error handling (e.g. if the last entry of `argv` is not the file name for some reason). How do you do this in your code base?

src/clangparser.cpp
```diff
@@ -140,6 +140,26 @@ void ClangTUParser::parse()
...
+    // argument and pass the file name with the other compile commands.
+    // However, in some cases (e.g., starting from Clang 14, if we are parsing a header file, see
+    // https://github.com/doxygen/doxygen/issues/10733), the compile commands returned by
+    // getCompileCommands include a "--" as second to last argument (which is supposed to make it
+    // easier to parse the argument list). If we pass this "--" to clang_parseTranslationUnit below,
+    // it returns an error. To avoid this, we remove the file name argument (and the "--" if present)
+    // from argv and pass the file name separately.
+    if (std::string(argv[argv.size() - 1]) != fileName.str()) {
+      std::cout << fileName.str() << " vs. " << argv[argv.size() - 1] << std::endl;
+    }
```
albert-github, line 155: Looks like debug statements, should not be present anymore / commented..

src/clangparser.cpp
```diff
@@ -234,8 +250,8 @@ void ClangTUParser::parse()
   }
 
   // let libclang do the actual parsing
-  //for (i=0;i<argv.size();i++) printf("Argument %d: %s\n",i,argv[i]);
-  p->tu = clang_parseTranslationUnit(p->index, nullptr,
+  for (i=0;i<argv.size();i++) printf("Argument %d: %s\n",i,argv[i]);
```
albert-github, line 253: Looks like a debug statement, should not be present anymore / commented..

src/clangparser.cpp
tobiasleibner: Removed.

src/clangparser.cpp
tobiasleibner: Commented again.

Issue #10762, opened 2024-03-21
Macro define a class name to a different one causes extra whitespace to appear


scott-zhong:
**Describe the bug**
Macro define a class name to a different one causes extra whitespace to appear when `MACRO_EXPANSION = YES` is used. The problem started with 1.8.12 and continues with Doxygen trunk (https://github.com/doxygen/doxygen/commit/cbb48edb43715e3420a69d69c21e6e3c4f998894).

```
#define A B

/**
 * A docs
 */
class A
{
public:
    /**
     * Type of food.
     */
    enum Food {
        /**
         * apple
         */
        apple  
    };
};

/**
 * Test docs
 */
template <class T>
struct Test
{
};

/**
 * Test specialized docs
 */
template<>
struct Test<A>
{
    /**
     * Test specialized ctor
     */
    Test(A::Food food)
        : food_(food) {}

    A::Food food_;
};
```

There is an extra whitespace going from Doxygen 1.8.11 to 1.8.12 and later (`B::Food` -> `B ::Food`).

**Screenshots**
Doxygen 1.8.11:
![image](https://github.com/doxygen/doxygen/assets/32584126/fef26cce-0e2e-4fc8-8ec0-32cd095c252e)

Doxygen 1.8.12 - Doxygen trunk (https://github.com/doxygen/doxygen/commit/cbb48edb43715e3420a69d69c21e6e3c4f998894):
![image](https://github.com/doxygen/doxygen/assets/32584126/f0b0a598-8a65-46bd-a7f9-10ffde544130)

**To Reproduce**
[testcase.zip](https://github.com/doxygen/doxygen/files/14676036/testcase.zip)

**Expected behavior**
No extra whitespace in the generated documentation.

scott-zhong:
The whitespace also appears with templates as well.

```
/**
 * Base docs
 */
template <class Type>
class Base
{
public:
    /**
     * A typedef
     */
    typedef typename Type::element_type element_type;
};

/**
 * MyCharType docs
 */
class MyCharType
{
public:
    /**
     * A typedef
     */
    typedef char element_type;
};

/**
 * Derived docs
 */
class Derived : public Base<MyCharType>
{
};
```

Doxygen 1.8.11:
![image](https://github.com/doxygen/doxygen/assets/32584126/128e59bf-d628-436c-95f4-cff9f9d07498)

Doxygen 1.8.12:
[...]

PR #10771, opened 2024-03-29
issue #10762 Macro define a class name to a different one causes extra whitespace to appear

albert-github:
Problem looks like a side effect of
```
00ee930a1d73e11885197102c54fd4c8141127da is the first bad commit
commit 00ee930a1d73e11885197102c54fd4c8141127da
Date:   Sun Mar 13 10:12:46 2016 +0100

    reimplemented removeRedundantWhiteSpace() to improve performance

 src/util.cpp | 385 +++++++++++++++++++++++++++++++++++------------------------
 1 file changed, 231 insertions(+), 154 deletions(-)
```

the preprocessor should not add extra spaces to the result.

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -2983,6 +2983,7 @@ static bool expandExpression(yyscan_t yyscanner,QCString &expr,QCString *rest,in
       i=p+l;
     }
   }
+  expr = expr.stripWhiteSpace();
   //printf("<expandExpression(expr='%s',rest='%s',pos=%d,level=%d)\n",qPrint(expr),rest ? qPrint(*rest) : "", pos,level);
   return TRUE;
 }


albert-github:
I don't think the extra fix of 8dd81733da56bc4258261573fe00d03113979c41  is correct.
When having a nested expansion this might lead to problems as not the individual parts are evaluated properly, this can be shown with:
```
#define A B
#define C D
#define E C::A
```
and 
```
    Test(E::Food food)
```
This leads with 8dd81733da56bc4258261573fe00d03113979c41  to:
image

Note the extra space between `D` and ` ::B`.

and with the original fix d4535fa3841d8bc1d3c38457b6c424f330632753  to:
image

(I'm not sure whether this is all 100% valid C++ code though it is about the idea).


Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/14820904/example.tar.gz)

doxygen:
@albert-github I wanted to minimize the impact, because spaces are sometimes needed (that is why `yyextra->nospaces` is there). Here is an example that makes the preprocessor add a space
```c
#define A() int
#define P A()X
void f(P) {}
```
Without extra space this produces `void f(intX) {}` where the preprocessor would normally produce `void f(int X) {}` 

Note that doxygen processes this wrongly regardless of the changes, so more work is needed.
See also [this topic](https://stackoverflow.com/questions/37796947/spaces-inserted-by-the-c-preprocessor) for more discussion.

albert-github:
Interesting indeed looks like there is more to it. At first glance it looks like that when using round brackets there is a space added.
(Looks like the problem is present already for a long time, when not from the beginning).

PR #10782, opened 2024-04-05
PERLMOD_PYTHON to use python in perlmod (it generates simple document for class / member functions as example)

cheoljoo:
PERLMOD_PYTHON is added in config.xml because of supporting python from perl output.
- perlmod makes DoxyDocs.pm.  PERLMOD_PYTHON will make DoxyDocs.py. 
- it will import DoxyDocs to use parsing data.
- doxygen_perlmod_python examples
 	- https://github.com/cheoljoo/doxygen_perlmod_python/blob/main/README.md
 	- i made separate repository to add example in related to PELMOD_PYTHON (python code).
 		- make  : make markdown 
 		- make testserver : make html
 		- make testmako : make html with different type using mako python module
 	- all tests have error because plantuml size is zero.  if you want to remove the error , use plantuml.jar from https://plantuml.com/ko/download

# modified files
- config.xml : add PERLMOD_PYTHON
- perlmodgen.cpp : add more generated file when PERLMOD_PYTHON is YES.   
	- Makefile(changed)  put2cpp.py(added)  README.md(added)  sample.py(added) in perlmod directory

# PERLMOD_PYTHON
- developers want to use parsing data from doxygen.
- they want to do something with python. 
- set **GENERATE_PERLMOD = YES**   and **PERLMOD_PYTHON = YES**

## make python_doc  in output/perlmod
- doxy2py.pl : DoxyDocs.pm converts to DoxyDocs.py.
  - DoxyDocs.py is loaded in python program.
- sample.py makes markdown (output.md) file as final output.
  - output.md includes table for classes and member functions.
  - if doxygen comments include plantuml , output.md includes plantuml   if you set PLANTUML_JAR_PATH = ./plantuml.jar and this file exists.
- output 
	- output.md : show markdown including plantuml with vscode (useful extension : markdown preview enhanced)

## make code2cpp in output/perlmod
- i made to put basic code files in perlmodgen.cpp conveniently. 
- all codes with related to PERLMOD_PYTHON are created by perlmodgen.cpp.
- put2cpp.py insert converted c++ code into perlmodgen.cpp from files in input_perlmodgen_files directory.
- `--perlmodgen` option : location of perlmodgen.cpp. python codes in inputdir will be updated.
- generated code will be inserted between // perlmodPython PerlModGenerator [key] <start> and // perlmodPython PerlModGenerator [key] <end>
- output    ( `--perlmodgen=../../../../src/perlmodgen.cpp` )
	- ../../../../src/perlmodgen.cpp.mod : you can replace perlmodgen.cpp with it.


doxygen:
@cheoljoo Thanks for your interest in Doxygen and for creating the pull request. 

I see this basically adds a couple of static files to the output, and these then use the output of Doxygen as input. I think this can better be hosted as a separate project on GitHub. By including it in Doxygen, the maintenance burden moves to us, and we are already stretched supporting the current output formats.

Did you know, Doxygen has a doxmlparser addon that provides a Python interface to the XML output produced by doxygen? I would expect/hope it offers a better way to interface with the doxygen output from Python than using the Perl module.

cheoljoo:
@doxygen ,  I will check doxmlparser.
- I made separate github to support PYTHON with PERLMOD. 
  - https://github.com/cheoljoo/doxygen_perlmod_python
- DoxyDocs.py (python) supports simpler parsing database like DoxyDocs.pm (perl).
- it supports plantuml server instead of png.  (it will have less running time.)
thanks for your advice.

Issue #10766, opened 2024-03-23
C# cref or \ref doesn't work without full namespace, even for classes in same namespace


runevision:
**Describe the bug**
Referencing another class (in this case called TopLayerDependency) in the same namespace with `<see cref="TopLayerDependency"/>` does not create a link when the class name is not qualified with the (common) namespace. In the output I see
`unable to resolve reference to 'TopLayerDependency' for \ref command`.

**Screenshots**

![MissingLink](https://github.com/doxygen/doxygen/assets/6685642/30e12cea-9f31-4602-99e6-5a88a7bae11c)

**To Reproduce**
Default Doxyfile without modifications used to reproduce the issue.

Place these two files in the folder Doxygen is run from:

File GenerationSource.cs
```
namespace Runevision.LayerGen {
    /// <summary>
    /// Unity component that creates a <see cref="TopLayerDependency"/>. <br/>
    /// Unity component that creates a \ref TopLayerDependency. <br/>
    /// Unity component that creates a <see cref="Runevision.LayerGen.TopLayerDependency"/>. <br/>
    /// Unity component that creates a \ref Runevision.LayerGen.TopLayerDependency.
    /// </summary>
    public class GenerationSource : MonoBehaviour { }
}
```

File TopLayerDependency.cs
```
namespace Runevision.LayerGen {
    /// <summary>
    /// All generation starts with one or more top level dependencies at the root.
    /// </summary>
    public class TopLayerDependency { }
}
```

**Expected behavior**
I would expect class names in the same namespace to be turned into links without having to specify the namespace.

**Version**
1.10.0 (GIT-NOTFOUND)
Doxygen installed via brew
macOS 12.5.1

**Additional context**
None of my other cross references get turned into links either (for example references from one method to another method in the same class). It seems like something is very broken about Doxygen reference resolution for xmldoc and/or C#, but I think this report here is the simplest demonstration, so let's start with that. With any luck, the fix for this will fix the other ones too.

albert-github:
I've just pushed a proposed patch, pull request #10795

zhangzq:
The same problems exist in Python. I need to write whole path, i.e. package.sub_package.file_name.class_name.function_name to resolve the function.

albert-github:
@zhangzq The python problem should also be fixed with the patch #10795, did you try it?
Otherwise:
[...]

PR #10795, opened 2024-04-13
issue #10766 C# cref or \ref doesn't work without full namespace, even for classes in same namespace

albert-github:
Also check against the case that we have a `.` as separator (i.e. C# or Java) in the context name.


doxygen:
@albert-github I reworked the code so that the context is corrected at the start and the language to use is also stored. In this way additional (expense) calls to `resolveRef()` and `resolveLink()` are avoided. See [this commit](https://github.com/doxygen/doxygen/commit/15e5493cc4e79ec44d55d1ef1f4d5d2bc08290f4).

albert-github:
Indeed calls like `resolveRef()` and `resolveLink()` are expensive. 
Looks all OK.

Issue #10240, opened 2023-08-18
Multi-line doc not recognized in Python class field


lelegard:
In Python, multi-line doc is recognized on class, class method, but not class fields.

Sample code:
~~~
##
# Class multi-line doc
#
class C:
    ##
    # Constructor multi-line doc
    #
    def __init__(self):
        ## field one-line doc
        self.a = None
        ##
        # field multi-line doc
        #
        self.b = None
        ## field one-line doc
        self.c = None
~~~

For fields, only one-line doc are recognized. The doc for field `b` is not recognized:
~~~
$ ../doxygen-master/build/bin/doxygen 
test.py:11: warning: Member b (variable) of class test.C is not documented.
$
$ ../doxygen-master/build/bin/doxygen --version
1.9.8 (27ba0983fdcc4b7b836e06fef90e05400d8351c8)
~~~

Looking at the generated HTML, `b` has no documentation.

Is this expected?

This has been reproduced with version 1.9.4, 1.9.5, 1.9.6, 1.9.7, and current master of the repo.

Reproducer: [doxybug4.zip](https://github.com/doxygen/doxygen/files/12379399/doxybug4.zip)

albert-github:
I've just pushed a proposed patch, pull request #10798

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.11.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10798, opened 2024-04-14
issue #10240 Multi-line doc not recognized in Python class field

albert-github:
Improve handling of start of a special comment block. (Looks like it should not be corrected at other places)


doxygen:
@albert-github You mention "Looks like it should not be corrected at other places", but I wonder why this is the case. It doesn't look any different from the other 2 cases.

doxygen:
@albert-github I think I already understand it. You basically disabled processing special comments while parsing the function body (they will be picked up while processing individual members). I've now made that more explicit by removing the lexer rule completely.

PR #10805, opened 2024-04-16
Updated Polish translation to 1.11.0

marecl:
Added translation for `trImportant()`.  

Leaving as draft in case new strings appear before release 1.11.0.


albert-github:
I think is is better to remove the draft part, as the chance of new strings before this issue is integrated is next to zero.

marecl:
> I think is is better to remove the draft part, as the chance of new strings before this issue is integrated is next to zero.

If that's the case then it's ready to merge

PR #10809, opened 2024-04-19
Specify documentation except for listed languages.

albert-github:
With the commands `\~` and `\~LanguageId` it is possible to specify comments for one or all languages. When one wants to write the documentation for the languages except for some specified this needs a complete set of commands, with this enhancement this can be done with one command.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/15040892/example.tar.gz)

### DIFF
--- a/doc/commands.dox
+++ b/doc/commands.dox
@@ -4099,18 +4099,20 @@ class Receiver
   because doxygen uses it to detect Javadoc commands.
 
 <hr>
-\section cmdtilde \\~[LanguageId]
+\section cmdtilde \\~[LanguageId] or \\~'{'LanguageId [,LanguageId]*'}'
   \addindex \\~
   This command enables/disables a language specific filter. This can be
   used to put documentation for different language into one comment block
   and use the \ref cfg_output_language "OUTPUT_LANGUAGE" tag to filter out only a specific language.
-  Use \c \\~language_id to enable output for a specific language only and
+  Use \c \\~language_id to enable output for a specific language only,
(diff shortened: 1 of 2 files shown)


doc/commands.dox
```diff
@@ -4099,18 +4099,20 @@ class Receiver
   because doxygen uses it to detect Javadoc commands.
 
 <hr>
-\section cmdtilde \\~[LanguageId]
+\section cmdtilde \\~[LanguageId] or \\~'{'LanguageId [,LanguageId]*'}'
```
doxygen, line 4102: I have some concerns.

1) Is there a use-case for this? I can only think about using it as fallback telling some translation is missing.
2) I think the `\~lang` and `\~{lang}` look a bit too similar, and it is not intuitively clear one is the negation of the other, maybe `\~!{lang}` would be more clear.
3) From a usability point of view, I would prefer something like `\~dutch Nederlands \~german Deutsch \~else Fallback to e.g. english text \~`.

albert-github: 1. No there is not a direct use case, but it is a bit strange that one cannot easily signal the fallback telling some translation is missing.
2. They are indeed quite similar, `\~!{lang}` would indeed be more clear.
3. I thought about this as well but decided against it as
   - one has to keep track of which language have already been selected and reset the list when `\~` is given / end of a comment block
   - the current solution gives also a possibility to have `\~dutch Nederlands \~german Deutsch \~{dutch} some "Fallback" even for a previous handled language (probably quite unusual / uncommon but gives more flexibility) \~`.

I think best is to implement the solution mentioned under 2.

Issue #10827, opened 2024-04-25
member variable with type annotation and not used anywhere of python class  is missed


zhangzq:
**Describe the bug** 

The member variable of python class is missed, if:
1. with type annotation
2. not appeared elsewhere.

Below is a short example. The `a` will show in public member of docs, but `b` is missed.

```python
class Test:
    def __init__(self):
        ## first var
        self.a:int = 1

        ## second var
        self.b:int = 2

    def test(self):
        self.a = 3
```
 

**To Reproduce** 

```python
class Test:
    def __init__(self):
        ## first var
        self.a:int = 1

        ## second var
        self.b:int = 2

    def test(self):
        self.a = 3
```

with default doxyfile.

**Expected behavior** 

The class has two public members, `a` and `b`, now `b` is missed.

**Version** 
1.11.0 (74f8b41aead51679bf364056b4fd091cf828a7cf)

albert-github:
Not only `b` is missing but also the documentation from `a` as we get the warning:
```
test.py:13: warning: Member a (variable) of class test.Test is not documented.
```

Problem here is the "typing" which originally was missing in python and now is subsequently added.

albert-github:
Code has been integrated and corrected in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.11.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #10834, opened 2024-04-29
issue #10827 member variable with type annotation and not used anywhere of python class is missed

albert-github:
Handle type annotation for class variables

Extended example: [example.tar.gz](https://github.com/doxygen/doxygen/files/15148570/example.tar.gz)

### DIFF
--- a/src/pyscanner.l
+++ b/src/pyscanner.l
@@ -126,7 +126,7 @@ static void initParser(yyscan_t yyscanner);
 static void initEntry(yyscan_t yyscanner);
 static void newEntry(yyscan_t yyscanner);
 static void addEntry(yyscan_t yyscanner);
-static void docVariable(yyscan_t yyscanner,const char *name);
+static void docVariable(yyscan_t yyscanner,const QCString &name);
 static void newVariable(yyscan_t yyscanner);
 static void addVariable(yyscan_t yyscanner);
 static void newFunction(yyscan_t yyscanner);
@@ -544,6 +544,14 @@ ID        [a-z_A-Z%]+{IDSYM}*
                         docVariable(yyscanner,&yytext[5]);
                         newEntry(yyscanner);
                       }
+    "self."{IDENTIFIER}{B}:{B}{IDENTIFIER}/{B}"=" { // with typing
+                        QCString inp(yytext);
+                        int i=inp.find(':');
+                        docVariable(yyscanner,inp.mid(5,i-5).stripWhiteSpace());
+                        yyextra->current->type = inp.mid(i+1).stripWhiteSpace();
+                        yyextra->startInit = false;
+                        newEntry(yyscanner);
+                      }
     "cls."{IDENTIFIER}/{B}[,)] {
                         DBG_CTX((stderr,"Found class method variable %s in %s at %d\n",&yytext[4],qPrint(yyextra->current_root->name),yyextra->yyLineNr));
                         docVariable(yyscanner,&yytext[4]);
@@ -552,6 +560,14 @@ ID        [a-z_A-Z%]+{IDSYM}*
     "cls."{IDENTIFIER}/{B}"=" {
                         DBG_CTX((stderr,"Found class method variable %s in %s at %d\n",&yytext[4],qPrint(yyextra->current_root->name),yyextra->yyLineNr));
                         docVariable(yyscanner,&yytext[4]);
+                        newEntry(yyscanner);
+                      }
+    "cls."{IDENTIFIER}{B}:{B}{IDENTIFIER}/{B}"=" { // with typing
+                        QCString inp(yytext);
+                        int i=inp.find(':');
+                        docVariable(yyscanner,inp.mid(4,i-4).stripWhiteSpace());
+                        yyextra->current->type = inp.mid(i+1).stripWhiteSpace();
+                        yyextra->startInit = false;
                         newEntry(yyscanner);
                       }
     {TRIDOUBLEQUOTE}  { // start of a comment block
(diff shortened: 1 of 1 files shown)


albert-github:
I saw that a reworked version was present, so any type hint expression is allowed, but it doesn't look like this is working correctly it looks like it is not addressing the problem of the issue #10827 at all.
I think the code:
```
  for (auto &v : yyextra->current_root->children())
  {
    if (v->name == name)
    {
      v->section=EntryType::makeVariable();
      yyextra->current = v;
      return;
    }
  }
```
in in the `docVariable` routine wass accidently removed.
(a quick test showed that this would work on the provided extended example with this PR and on the original example).

doxygen:
@albert-github I didn't fully understand what this was for, and the problem I see with it is that `v` is a child of `current_root` and by assigning it to `yyextra->current` it will later be added again to `current_root` when `newEntry()` is called, causing potential memory issues (probably not a crash because `std::shared_ptr` was used to hold the children). Probably to get it correct, `v` should be removed from `current_root`.

doxygen:
@albert-github Please check if [this commit](https://github.com/doxygen/doxygen/commit/fb19cf7b928ff6269b5409c49b3fb3d906d778e0) correctly addresses the problem.

albert-github:
@doxygen no unfortunately it doesn't work
- documentation fails
- extended example gives quite a few warnings.

all these are gone when adding the code as indicated to `docVariable`.

I tried to remove the `v` but didn't succeed as I also had similar doubt about memory leaks, but the documentation states (https://en.cppreference.com/w/cpp/memory/shared_ptr):
> std::shared_ptr is a smart pointer that retains shared ownership of an object through a pointer. Several shared_ptr objects may own the same object. The object is destroyed and its memory deallocated when either of the following happens:
>
>    the last remaining shared_ptr owning the object is destroyed;
>    the last remaining shared_ptr owning the object is assigned another pointer via [operator=](https://en.cppreference.com/w/cpp/memory/shared_ptr/operator%3D) or [reset()](https://en.cppreference.com/w/cpp/memory/shared_ptr/reset).

so it looks to me that deleting should not be necessary.

albert-github:
I did a small revisit and added the line:
```
 v->section=yyextra->current->section;
```
to `newEntry` just before `found=true;` (just adding to the both "doc" setting blocks doesn't work.

This gives a better result less warnings:
```
p1.py:7: warning: Member a (variable) of class p1.p1_P1 is not documented.
p1.py:8: warning: Member b (variable) of class p1.p1_P1 is not documented.
p1.py:14: warning: Member a (variable) of class p1.p1_P2 is not documented.
p1.py:15: warning: Member b (variable) of class p1.p1_P2 is not documented.
p1.py:16: warning: Member c (variable) of class p1.p1_P2 is not documented.
p1.py:17: warning: Member d (variable) of class p1.p1_P2 is not documented.
p1.py:28: warning: Member a (variable) of class p1.p1_P3 is not documented.
p1.py:29: warning: Member b (variable) of class p1.p1_P3 is not documented.
p1.py:30: warning: Member c (variable) of class p1.p1_P3 is not documented.
p1.py:31: warning: Member d (variable) of class p1.p1_P3 is not documented.
p7.py:7: warning: Member a (variable) of class p7.p7_P1 is not documented.
p7.py:8: warning: Member b (variable) of class p7.p7_P1 is not documented.

p7.py:14: warning: Member a (variable) of class p7.p7_P2 is not documented.
p7.py:15: warning: Member b (variable) of class p7.p7_P2 is not documented.
p7.py:16: warning: Member c (variable) of class p7.p7_P2 is not documented.
p7.py:36: warning: Member a (variable) of class p7.p7_P3 is not documented.
p7.py:37: warning: Member b (variable) of class p7.p7_P3 is not documented.
p7.py:38: warning: Member c (variable) of class p7.p7_P3 is not documented.
```
The first set of warnings is correct but the last 6 are incorrect.

Furthermore when looping at the `p8_P3 ` class we see:

image

so some double entries although they should be merged.

doxygen:
@albert-github You didn't attach the p1..p7 examples you mentioned, but please check if [this commit](https://github.com/doxygen/doxygen/commit/88d7a74c1533561111ce9d64ed45786f675eab7e) helps.

Regarding the `std::shared_ptr`. Yes, it works to have an Entry `v` added to a list twice (both will then point to the same object and cleanup will only delete it once), but it also means that the object `v` will later be processed twice, which may still lead to problems. So I think it is better to guarantee that all objects are only added once. I'll see if I can replace `std::shared_ptr` by `std::unique_ptr` at some point, which would enforce this.

albert-github:
@doxygen 
The examples are all included in the attachment with the original PR, see https://github.com/doxygen/doxygen/pull/10834#issue-2268798664

image

The warnings now look all OK (are the same as with my original patch).
When running the current master (1.11.0 (5c38906c81f5707db134501f25df9f2e17c18255)) I see on the resulting page html/classp8_1_1p8___p3.html

image

whilst with the original patch:
image

in other words the variable `d` has been, in my opinion, correctly been  marked `int` in the original patch and isn't in the master version.

The current master version has also some better results e.g. on page classp3_1_1p3___p1.html in the master version I see:

image

whilst with the original patch the references to source code are missing:
image

doxygen:
@albert-github See https://github.com/doxygen/doxygen/commit/f45a560a7cace2c3dcb4c0ab0c3e426133d3f558 for a fix for the `p3.py` issue with variable `d`'s type hint.

doxygen:
@albert-github Your last example showed an issue that was not limited to Python. When using

```c++
int i;

/** @var i
 * Docs
 */
```

The `int` was missing from the definition, because it was overwritten by the `@var` which didn't mention it. [This commit](https://github.com/doxygen/doxygen/commit/1d674af1933d88038d469271df60bf6dee5a947d) should address this.

albert-github:
I didn't see any other problems.

Nice side effect with the `@var` for the other languages.

Issue #10851, opened 2024-05-04
Unable to build documentation from "master" branch


SchweinDeBurg:
I've just got the following error message:
```
[274/275] Generating Latex and HTML documentation.
FAILED: doc/CMakeFiles/run_doxygen /home/elijah/.local/rpmbuild/BUILD/doxygen/x86_64-redhat-linux-gnu/doc/CMakeFiles/run_doxygen 
cd /home/elijah/.local/rpmbuild/BUILD/doxygen/x86_64-redhat-linux-gnu/doc && /usr/bin/cmake -E env VERSION=1.11.0 INDEX_DOC=index_html.dox GENERATE_HTML=YES GENERATE_LATEX=NO /home/elijah/.local/rpmbuild/BUILD/doxygen/x86_64-redhat-linux-gnu/bin/doxygen && /usr/bin/cmake -E env VERSION=1.11.0 INDEX_DOC=index_latex.dox GENERATE_HTML=NO GENERATE_LATEX=YES /home/elijah/.local/rpmbuild/BUILD/doxygen/x86_64-redhat-linux-gnu/bin/doxygen
error: Problems running latex. Check your installation or look for typos in _formulas.tex and check _formulas.log!
error: Problems running latex. Check your installation or look for typos in _formulas_dark.tex and check _formulas_dark.log!
ninja: build stopped: subcommand failed.
```
[_formulas.log](https://github.com/doxygen/doxygen/files/15206590/_formulas.log)
[_formulas_dark.log](https://github.com/doxygen/doxygen/files/15206591/_formulas_dark.log)

albert-github:
When specifying "master" version please also specify the full version i.e. `doxygen -v` as the master version is a moving target.

When looking in the log file I see as first line:
```
This is pdfTeX, Version 3.14159265-2.6-1.40.17 (TeX Live 2016) (preloaded format=latex 2023.7.12)
```

It looks like that the used LaTeX version is too old (Tex Live 2016).

What happens when you change in the original `doc/doxygen_manual.tex` (around line 52):
```
\usepackage{ifpdf}
\ifpdf
```
to
```
\usepackage{iftex}
\ifpdftex
``

(In this case: Between 2 runs don't forget to remove the html and latex directory in, probably, /home/elijah/.local/rpmbuild/BUILD/doxygen/x86_64-redhat-linux-gnu)

SchweinDeBurg:
> When specifying "master" version please also specify the full version

[...]

PR #10852, opened 2024-05-05
issue #10851 Unable to build documentation from "master" branch

albert-github:
The commands like `\ifpdftex` were not supported in the older versions of the LaTeX distributions, commands like `\ifPDFTeX` are, renaming.


albert-github:
I've just pushed a proposed patch, pull request #10852

PR #10857, opened 2024-05-09
QT help problems with tag files

albert-github:
The QT assistant has problems with files that are not directly included into its input file (i.e. the qch file).
- This happens when we use tag files, hence `TAGFILES` will be disabled when `GENERATE_QHP` is set
- When there are in the tree items that do not refer to a file (so `ref=""`) this also gives problems. hence replacing these with a placeholder `doxygen_blank` that is just an empty file and will be added when necessary.

Example: [example.tar.gz](https://github.com/doxygen/doxygen/files/15263524/example.tar.gz)

### DIFF
--- a/src/configimpl.l
+++ b/src/configimpl.l
@@ -2143,6 +2143,13 @@ void Config::checkAndCorrect(bool quiet, const bool check)
       err("GENERATE_QHP=YES requires QHP_VIRTUAL_FOLDER to be set. Using 'doc' as default!\n");
       Config_updateString(QHP_VIRTUAL_FOLDER,"doc");
     }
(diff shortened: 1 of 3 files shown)


doxygen:
@albert-github Evaluating a static boolean is not thread-safe, so I replaced it by `std::call_once()`

albert-github:
Indeed I forgot about the multi threading (the adding of the HTML code was necessary for xmllint).
I also see that the 'blankFileName` variable was removed and replaced by a fixed string, I used the construct with the variable:
- Be consistent with the other outputs like crawl. sitemap etc.
- Less error prone when use on multiple places (in this case this is not the case and I don't yet see a case where this should be necessary).

PR #10874, opened 2024-05-18
Update BUILD.txt with dependencies/docker

zyphlar:
I don't see anywhere that describes all the dependencies required to build, so figured it'd be good to mention them here.

### DIFF
--- a/BUILD.txt
+++ b/BUILD.txt
@@ -1,6 +1,7 @@
 Doxygen uses cmake (http://www.cmake.org/) to build executables for various platforms.
-At least cmake version 3.14 is required.
-G++, Python, Flex, and Bison are also required.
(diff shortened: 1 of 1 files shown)


BUILD.txt
```diff
@@ -1,5 +1,7 @@
 Doxygen uses cmake (http://www.cmake.org/) to build executables for various platforms.
 At least cmake version 3.14 is required.
+G++, Python, Flex, and Bison are also required.
```
albert-github, line 3: I think it would be better to refer to the manual (chapter Installation) otherwise the same information is scattered on multiple places.
The reference to docker would be OK.

zyphlar: OK, updated

PR #10888, opened 2024-05-22
Use CMake's FindIconv (see #10887)

### LINKED ISSUE #10887: Use upstream FindIconv
Similar to https://github.com/doxygen/doxygen/commit/e5c75032987e4b45b04ac978df1e28751f1d08c6 I think it is a good idea to switch to upstream FindIconv

CMake's FindIconv is available since CMake 3.11, so no CMake version bump required https://cmake.org/cmake/help/latest/module/FindIconv.html

Croydon:
By the way, I have found that the Windows CI is always ending up using the lib binary files in git, even though that the CI is installing libiconv via another way. Example CI log: https://github.com/doxygen/doxygen/actions/runs/9178874312/job/25239722835#step:25:110. I am not going to change anything there as I consider it off-topic. Generally, I don't think that those built libraries should be in git and used. There should be enough options nowadays to get libiconv on Windows installed.

I renamed winbuild -> deps/iconv_winbuild to make it clearer what it is about. I had to give it sub-directories to let standard CMake's FindIconv find the right library.

See also #10887


Croydon:
> doxygen leaves the issues open and will automatically close the issue with a new release , please remove this automatic closing "merge" link.

Done. Thanks for pointing it out, have forgotten about it in the meantime.

I will comment on / look into the rest later.

Croydon:
Just for protocol, I have a WIP branch to improve CI overall, but I unearthed several bugs in packages providing dependencies in the process. I will probably return to work on that eventually, but it might be months.

For now, everything merged so far should be fine for Doxygen, so it isn't really time-sensitive.

PR #10899, opened 2024-05-26
issue #10989 bison: error: '%name-prefix' and '%define api.prefix' cannot be used together

albert-github:
- remove the deprecated `-p` option of bison
- updating the minimum version of flex
- removing old test for flex against version >= 2.5.33


dl8dtl:
Thanks, looks good to me.

Issue #10907, opened 2024-05-29
Cite command creates case sensitivity on name


BrianSipos:
**Describe the bug**
I am using the [`\cite`](https://www.doxygen.nl/manual/commands.html#cmdcite) command from a few places in source markup and in markdown files to reference documents from a BibTeX file.
The BibTeX file uses lower cased keys for its entries.
When the first `\cite` command encountered uses a specific case and matches one of the entries in a case-insensitive way that specific case then becomes the only way that later `\cite` commands will match.

**To Reproduce**
I have a BibTeX file with an entry copied from [rfc9172](https://datatracker.ietf.org/doc/rfc9172/bibtex/) and several uses of `\cite RFC9172` and `\cite rfc9172`.

This causes the following warnings from doxygen and broken links from the produced HTML docs.
The reference cited with upper case key _does_ have a proper link and the entry _is_ present in the Bibliography page.
```
warning: \cite command to 'rfc9172' does not have an associated number
```

**Expected behavior**
If the BibTeX key matching is case-insensitive it should apply to all uses of `\cite` not just the first one.
Otherwise, the matching should be documented as case-sensitive and properly apply to all uses.

**Version**
Doxygen 1.9.1 from package on Ubuntu 22.04.

albert-github:
- Where is it defined that the bibtext references are case independent?
- The doxygen version 1.9.1 is a bit older (January 8, 2021), the current doxygen version is 1.11.0 (though I don't think this will make a difference.

Furthermore to be able to reproduce the problem:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem? Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

albert-github:
I've just pushed a proposed patch, pull request #10908

BrianSipos:
I have been scouring documentation at https://www.bibtex.org/ and https://www.ctan.org/pkg/bibtex and https://www.latex-project.org/help/documentation/ and cannot actually find any definition of how the citation key is supposed to be matched. Though all examples I have seen use identical case.

The actual confusing part to me is not that there is case sensitivity (that I could guess) but that the first match from doxygen source is _not_ case sensitive, while later matches are sensitive based on the result of the first match. So for the example below I would expect that only the `xyz` citation would not result in a warning/failure.

[...]

PR #10909, opened 2024-05-29
issue #10907 Cite command creates case sensitivity on name

albert-github:
Relaxing the restriction regarding the case sensitivity of the citation label. The name `CITEREF` is a doxygen name so it can be changed as well (tests).

### DIFF
--- a/src/cite.cpp
+++ b/src/cite.cpp
@@ -79,14 +79,14 @@ void CitationManager::insert(const QCString &label)
 {
   p->entries.insert(
       std::make_pair(
-        label.str(),
-        std::make_unique<CiteInfoImpl>(label)
+        label.lower().str(),
+        std::make_unique<CiteInfoImpl>(label.lower())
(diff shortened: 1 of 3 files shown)


src/cite.cpp
```diff
@@ -79,14 +79,14 @@ void CitationManager::insert(const QCString &label)
...
-        std::make_unique<CiteInfoImpl>(label)
+        label.lower().str(),
+        std::make_unique<CiteInfoImpl>(label.lower())
       ));
 }
 
 const CiteInfo *CitationManager::find(const QCString &label) const
 {
-  auto it = p->entries.find(label.str());
+  auto it = p->entries.find(label.lower().str());
```
doxygen, line 89: I think the `.lower()` calls below can be avoided by using `find()` and `insert()` everywhere instead of `p->entries.find()` and `p->entries.insert()`.

Issue #8900, opened 2021-11-18
Doxygen does not support emphasis of file extensions


RochaStratovan:
I am trying to add emphasis to a file extension. It renders fine in a markdown viewer. Doxygen doesn't render it correctly.

Example markdown
```
2. If a script is applicable to ***both*** Windows and Linux it will not have a
   trailing **.bat** or **.sh** extention.
```

<br>

Markdown render example 1

![image](https://user-images.githubusercontent.com/40575252/142476226-b3d44edc-31d5-475f-acf0-601ee1b7b8cd.png)

<br>
Markdown render example 2

![image](https://user-images.githubusercontent.com/40575252/142476339-adc326fb-81ca-4fa3-823d-58132d255ed7.png)

<br>
Doxygen render

![image](https://user-images.githubusercontent.com/40575252/142476432-709601fb-1543-4484-a797-58cd65e864c8.png)

albert-github:
Most likely a bug in the doxygen markdown interpreter, though a `.` always has a special handling, but seems to be a problem when it is at the first, non-blank, position after the start of a style element like bold, emphasis, struikethrough.


Note: always mention the doxygen version used (`doxygen -v`), the differences in the doxygen settings compared to the default settings (`doxygen -x Doxyfile`), so in general (for the future):
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.
- Please also specify the **full** doxygen version used (`doxygen -v`).

albert-github:
From the documentation:

> 5.3.4 Emphasis and strikethrough limits
> Unlike standard Markdown and Github Flavored Markdown doxygen will not touch internal underscores or stars or
tildes, so the following will appear as-is:
>    `a_nice_identifier`
> Furthermore, a * or _ only starts an emphasis and a ~ only starts a strikethrough if
> • it is followed by an alphanumerical character, and
> • it is preceded by a space, newline, or one the following characters `<{([,:;`
>
> An emphasis or a strikethrough ends if
> • it is not followed by an alphanumerical character, and
> • it is not preceded by a space, newline, or one the following characters `({[<=+-\@`


The span of the emphasis or strikethrough is limited to a single paragraph.

So the problem looks like it is well documented and thus usage.

@doxygen should we allow the punctuation characters as suggested by the Merriam-Webster?

From https://en.wikipedia.org/wiki/Alphanumeric:
[...]

Issue #10696, opened 2024-02-28
Italic doesn't italicize parentheses


Hi-Angel:
**Describe the bug**

While writing non-code markdown documentation I noticed that Doxygen has failed to render many `*`-symbols all over the text. Upon reducing this to a minimal testcase it turns out Doxygen fails with it meets this syntax: `some text *(some remark about the text)*`. As I always italize remarks, these asterixes are all over the place.

**Expected behavior**

In the opened browser window the `(italic)` part of text is italic.

**To Reproduce *(in terms of terminal commands)***

```haskell
λ cat mainpage.md
# Test

This is *(italic)*.
λ doxygen -g Doxyconfig


Configuration file 'Doxyconfig' created.

Now edit the configuration file and enter

  doxygen Doxyconfig

to generate the documentation for your project
λ sed -i 's/^USE_MDFILE_AS_MAINPAGE\s*=/USE_MDFILE_AS_MAINPAGE = mainpage.md/g' Doxyconfig
λ doxygen Doxyconfig
[…]
λ xdg-open html/index.html
```

**Version**
* doxygen version `1.10.0 (GIT-NOTFOUND)`, from the Archlnux repo
* `lsb_release -a`
    ```
    LSB Version:    n/a
    Distributor ID: Arch
    Description:    Arch Linux
    Release:        rolling
    Codename:       n/a
    ```

albert-github:
The problem is originating from `markdown.cpp`:
```
int Markdown::Private::processEmphasis(std::string_view data,size_t offset)
{
  AUTO_TRACE("data='{}' offset={}",Trace::trunc(data),offset);
  const size_t size = data.size();

  if ((offset>0 && !isOpenEmphChar(data.data()[-1])) || // invalid char before * or _
      (size>1 && data[0]!=data[1] && !(isIdChar(data[1]) || extraChar(data[1]) || data[1]=='[')) || // invalid char after * or _
      (size>2 && data[0]==data[1] && !(isIdChar(data[2]) || extraChar(data[2]) || data[2]=='[')))   // invalid char after ** or __
  {
    return 0;
  }
```
and 
```
#define extraChar(c) \
  (c=='-' || c=='+' || c=='!' || \
   c=='?' || c=='$' || c=='@' || \
   c=='&' || c=='*' || c=='%')
```

@doxygen Is there a reason why `(` as not in this list of allowed extraChar characters?
Especially as `[` is explicitly added  (could be in the #define!)

[...]

Issue #8902, opened 2021-11-18
Doxygen markdown doesn't work for quoted values


RochaStratovan:
I have markdown that has a quoted value that we want emphasized. It works fine in standard markdown renders, but does not work in Doxygen renders.

I tried emphasizing just the contents, and then the contents and the quotes. Neither works.

Example markdown
```
"Internal Emphasis"

"External Emphasis"
```

Markdown Render Example:

![image](https://user-images.githubusercontent.com/40575252/142497452-4cabcec5-552f-4660-ab04-1e4b5442d8cc.png)

Doxygen Render Example:

![image](https://user-images.githubusercontent.com/40575252/142497567-99890360-23bd-4cb3-ae1a-edb35c622099.png)

albert-github:
In your example I don't see any settings for `bold` so looks like this has been omitted.
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.
- Please also specify the **full** doxygen version used (`doxygen -v`).

albert-github:
I think we have here, nearly, the same situation as with #8900 though we now have `"` as non alphanumeric character, that as such can again need some extra processing.

@doxygen what to do with this?

Hi-Angel:
Should be fixed by https://github.com/doxygen/doxygen/pull/10910

albert-github:
Looks indeed to be fixed by #10910, so solved in the mean time.

PR #10910, opened 2024-05-29
markdown: fix emphasis with extra characters

Hi-Angel:
This fixes a number of problems with emphasis in markdown, specifically #10696 #8900 #8902. 

With these changes applied out of [this @RochaStratovan's comment](https://github.com/doxygen/doxygen/issues/8900#issuecomment-974549812) the situation with problematic rendering looks currently like this:

Screenshot_20240530_221854

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -78,11 +78,12 @@ enum class ExplicitPageResult
    (c>='0' && c<='9') || \
    (static_cast<unsigned char>(c)>=0x80)) // unicode characters
 
+// is character allowed right at the beginning of an emphasis section
 #define extraChar(c) \
   (c=='-' || c=='+' || c=='!' || \
    c=='?' || c=='$' || c=='@' || \
    c=='&' || c=='*' || c=='%' || \
-   c=='[')
+   c=='[' || c=='(')
 
 // is character at position i in data allowed before an emphasis section
 #define isOpenEmphChar(c) \
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -81,7 +81,8 @@ enum class ExplicitPageResult
 #define extraChar(c) \
   (c=='-' || c=='+' || c=='!' || \
    c=='?' || c=='$' || c=='@' || \
-   c=='&' || c=='*' || c=='%')
+   c=='&' || c=='*' || c=='%' || \
+   c=='[')
 
 // is character at position i in data allowed before an emphasis section
 #define isOpenEmphChar(c) \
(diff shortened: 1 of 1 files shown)


RochaStratovan:
This is a great start!

However it doesn't quite get all the issues reported with #8900 

The issues that I see missing are highlighted in the following image.

image

Thank you for the improvements so far!

Hi-Angel:
@RochaStratovan right, I decided to fix the low-hanging fruit for now. The ones you highlighted in yellow require more digging *(because they're broken elsewhere)*.

The PR should fix #8900 however, because the report is about a word italic + bold combinations not working for a word starting with dot. The ones you highlighted are indeed mentioned in comments, but I think they deserve a separate report, especially seeing how those bugs reside elsewhere in the code.

Hi-Angel:
…not saying I wouldn't want to fix everything, because honestly all of these are likely pretty simple bugs. But I kind of swamped with other personal stuff right now and only have that much time.

RochaStratovan:
Good points, @Hi-Angel , thank you for the help/progress so far. I appreciate it.

albert-github:
@Hi-Angel 

- In general I think as well that indeed a set of excluded characters would be better.
- I'm wondering actually why `{` is not in the list.
- I'm wondering a bit why closing brackets aren't in the list (might have to do with constructs like: `fie(int **);`).
- I looked in the ASCII table and saw some more characters that are not in the list, might also have a special reason.

doxygen:
If I remember correctly, one of the reasons I was careful with the allowed characters is to prevent breaking auto-links, e.g.

```c++
/** @file
 *  See foo(int *x,int *y) for more info.
 */

/** A function */
void foo(int *x ,int *y) {}
```

Probably the current implementation is too conservative, although at the same time, writing `foo(int *,int* y)` will break things (with or without this pull request) so it is hard to combine doxygen's markup features that are predating Markdown, with the Markdown syntax 100% of the time.

Hi-Angel:
FTR, given that the changes discussed reside inside function called `processEmphasis`, I think it's fair to assume that if the function got called then auto-linking *(if there was any)* already broke, otherwise that text would've been skipped from processing by `processEmphasis`.

doxygen:
@Hi-Angel I've fix another case to better support backticks. The markup for text within double quotes remains untouched however. Changing this will have a bigger impact.

As for the way doxygen parses the comments; Markdown processing is used as a preprocessing step and will internally translate the markup into doxygen's natively supported markup commands. Auto-links are resolved in that later step. So ideally the Markdown processor should detect auto-links and leave them untouched.

Hi-Angel:
Hello, sorry, I just noticed the issues enlisted weren't closed when `fixed but not released` label was removed. I closed one of them as I was the author, but I'm afraid the other two has to be closed by someone with maintainer permissions 😊

Issue #10917, opened 2024-06-02
Doxywizard in proposed patches doesn't run


thom-sch:
**Describe the bug**
Calling doxywizard.exe delivered with proposed patches results in an error message at least under windows (other OSes not tested) - see attached screenshot.

**Screenshots**
![grafik](https://github.com/doxygen/doxygen/assets/44623251/69c44980-170d-4181-b861-6dda9ec714f1)

**To Reproduce**
Start doxywizard.exe from any proposed patch

**Expected behavior**
...just a flaw in my eyes

**Version**
- Windows 10 (probably other versions, too)
- Any proposed patch (official releases work fine!)

Regards from Meerbusch
Thomas

albert-github:
The official release of doxygen starts without problems and gives as version information:
![image](https://github.com/doxygen/doxygen/assets/5223533/ce728624-48fb-4d25-8625-e8d21921fae8)

The version from the artifacts (https://github.com/doxygen/doxygen/actions/runs/9336693228) doesn't start and gives:
![image](https://github.com/doxygen/doxygen/assets/5223533/fe3dd787-ea1f-4216-9b77-dfc1874b174f)

@doxygen Is there a difference in the build between the official doxygen release and the GitHub Artifacts (dynamic versus static linking? Maybe the `-Dwin_static=YES` in the official doxygen release build)

As a side note:
My locally build 32 bit version does start (just a little bit older Qt than with the official release i.e. 5.14.0)
My locally build 64 bit version does start (just a little bit newer Qt than with the official release i.e. 6.1.2)

thom-sch:
an additional information:
Replacing the proposed patch version with an "official" doxygen.exe runs without any problem.

Regards from Meerbusch
Thomas

albert-github:
Did you mean doxygen.exe or doxywizard.exe?

I see a big difference in the size of the GitHub Actions doxywizard (673,280 bytes) and the one from the official 11.0.0 release (15,468,544 bytes), definitely indicating a difference between a dynamic and a static build.

albert-github:
I've just pushed a proposed patch, pull request #10918 

When it succeeds this will probably result in a bit a larger artifact so we have to see whether we want this or not (though my experience is that after some time the artifact are removed automatically so should not be a real big deal).

thom-sch:
> Did you mean doxygen.exe or doxywizard.exe?

oh, sorry - of course doxywizard.exe... :)

albert-github:
[...]

PR #10918, opened 2024-06-02
issue #10917 Doxywizard in proposed patches doesn't run

albert-github:
Adding static linking option so the doxywizard might start.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -70,12 +70,14 @@ jobs:
             os: windows-latest,
             build_type: "Debug", cc: "cl", cxx: "cl",
             build_gen: "NMake Makefiles"
+            cmake_extra_opts: "-Dwin_static=YES"
           }
         - {
             name: "Windows Latest MSVC Release",
             os: windows-latest,
             build_type: "Release", cc: "cl", cxx: "cl",
             build_gen: "NMake Makefiles"
+            cmake_extra_opts: "-Dwin_static=YES"
           }
     steps:
     - name: Checkout doxygen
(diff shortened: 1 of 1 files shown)


albert-github:
Does not yet work ...

doxygen:
@albert-github The problem is that the [qt github action](https://github.com/jurplel/install-qt-action) used to install qt uses the [normal installer](https://ddalcino.github.io/aqt-list-server/) which installs a dynamically linked executables.

Probably it will work if you (as a user) first do these steps that are done by the github action as well:
```
curl -O https://download.qt.io/official_releases/online_installers/qt-unified-windows-x64-online.exe
./qt-unified-windows-x64-online.exe \
  --accept-licenses \
  --default-answer \
  --confirm-command install \
  qt.qt5.5152.win64_msvc2019_64
```

albert-github:
I don't get the line:
> Probably it will work if you (as a user) first do these steps that are done by the github action as well

Do you mean:
- that before I can locally use executable with GitHub Actions I should run the menioned command?

or

It there a typo and on GitHub Actions this command should be added and where (before or after the qt action install)?

doxygen:
> I don't get the line:
> 
> > Probably it will work if you (as a user) first do these steps that are done by the github action as well
> 
> Do you mean:
> 
>   * that before I can locally use executable with GitHub Actions I should run the menioned command?
> 

This is what I meant.

albert-github:
Pity that there is no direct way to solve the issue on GitHub Actions as now individual users have to install Qt (which might interfere with their other installations of Qt).

doxygen:
@albert-github Pity indeed. That's why I use a statically built version of Qt for the official release. One workaround may be to bundle the Qt DDLs together with the doxywizard binary. The dynamic loader will look for DLLs in the same directory as the executable first, I believe. The [dependencies](https://github.com/lucasg/Dependencies) tool can be used to test which libraries an executable uses.

albert-github:
> I'm going to give the unified executable a try

Looks like I already had a Qt 5.15.2  installation lying around (from 2021?) and when adding the bin directory of this to the PATH everything was working.

Issue #10902, opened 2024-05-28
Improve Markdown Support - blockquotes with special markers parse wrongly


igormironchik:
**Describe the bug**

Blockquote with special mark for highlighting (like `[!NOTE]`) parses wrongly.

> [!WARNING]
>
> Have a look at emphasis in second paragraph too. You have there issue too.

**Screenshots**

![quote](https://github.com/doxygen/doxygen/assets/11455533/1f89e97c-b69e-4a81-b7bf-179ba0c4fee5)

**To Reproduce**

[Doxygen.zip](https://github.com/doxygen/doxygen/files/15469937/Doxygen.zip)

**Expected behavior**

I'd like to see all other stuff from highlighted blockquote parsed and rendered correctly in it.

**Version**

```
1.12.0 (3d494140a674a8344c165a559d841166bbf3b80f*)
```

```
No LSB modules are available.
Distributor ID: Neon
Description:    KDE neon 6.0
Release:        22.04
Codename:       jammy
```

I've built `Doxygen` from sources.

albert-github:
So you would like to have something like:

![image](https://github.com/doxygen/doxygen/assets/5223533/1db4f5b4-6eaa-4ba1-b049-d3476d3442e2)

(the fact that ``*`markdown-it (Rust)` measurement done with `markdown_it::plugins::extra`*`` is not as emphasis text is already mentioned in another issue).

The problem is that the construction is replaced by a `\note` command and this ends with an empty line ( in this case this is the line with just `>` in the example).

I see 2 workarounds:
```
> [!NOTE]
> *This measurement done with test file in [markdown-it (Rust)](https://github.com/markdown-it-rust/markdown-it)*

> [!NOTE]
> *`markdown-it (Rust)` measurement done with `markdown_it::plugins::extra`*

> [!NOTE]
> **Keep in mind that `markdown-it (Rust)` doesn't support all GFM extensions, so it's not honest
> to compare. `md4qt` supports GitHub's autolinks extension, that `markdown-it (Rust)` doesn't, and
> this requires ~500 microseconds, for example...**
```
note: the extra real empty line and the replacement of the `>` line with a new `> [!NOTE]`

and 
```
> [!NOTE]
> *This measurement done with test file in [markdown-it (Rust)](https://github.com/markdown-it-rust/markdown-it)*<br>
> *`markdown-it (Rust)` measurement done with `markdown_it::plugins::extra`*<br>
[...]

PR #10920, opened 2024-06-03
issue #10902 Improve Markdown Support - blockquotes with special markers parse wrongly (2)

albert-github:
Based on the discussion in the issue it is better that directly after the "GitHub Markdown style alerts" there is only 1 `<br>` in case of an "empty line"

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -2712,6 +2712,7 @@ size_t Markdown::Private::writeBlockQuote(std::string_view data)
   const size_t size = data.size();
   QCString startCmd;
   int isGitHubAlert = false;
+  int isGitHubFirst = false;
   while (i<size)
   {
     // find end of this line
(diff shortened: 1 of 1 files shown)


src/markdown.cpp
```diff
@@ -2790,8 +2796,13 @@ size_t Markdown::Private::writeBlockQuote(std::string_view data)
     if (curLevel != 0 || !isGitHubAlert)
     {
       QCString txt = data.substr(indent,end-indent);
-      if (txt.stripWhiteSpace().isEmpty() && !startCmd.isEmpty()) out += "<br><br>\n";
+      if (txt.stripWhiteSpace().isEmpty() && !startCmd.isEmpty())
+      {
+        if (!isGitHubFirst) out += "<br>";
```
doxygen, line 2801: This starts to look more and more fiddly, can we not solve this in CSS by using some margins and maybe a different class for the GitHub alerts (if really needed)?

albert-github: It indeed looks like a bit fiddly but there are some considerations to do it this way:
- the problem does not only show in HTML but also in e.g. LaTeX so just a css approach is not sufficient
- the "empty line" is not always present and thus the first `<br>` can be at a different place and would be harder to detect with css (when it is anyway possible at that moment).

I didn't see a better solution than adding this `isGitHubFirst` variable.

PR #10927, opened 2024-06-05
Document all possible values of the `dir` attribute within the `param` command

EduPonz:
This PR documents some missing possible values for the direction attribute of the `param` command that were introduced in:
* #7881

### DIFF
--- a/doc/commands.dox
+++ b/doc/commands.dox
@@ -2066,8 +2066,11 @@ ALIASES  = "english=\if english" \
   present in the function declaration or definition.
 
   The \c \\param command has an optional attribute, \<dir\>, specifying the direction
-  of the parameter. Possible values are "[in]", "[out]", "[inout]", "[outin]",
-  "[in,out]", "[out,in]", "[in out]", and "[out in]"; note the [square] brackets in this description.
+  of the parameter. Possible values are "[in]", "[out]", and "[in,out]"; note the [square] brackets
+  in this description.
+  For the bidirecional values, directions "in" and "out" can be specified in any order, and they can
+  either be written altogether, or separated with a comma (`,`) or a space.
+  That means that for example values "[outin]" or "[in out]" are also valid.
   Note that it is also possible to put whitespace between the command and the \<dir\>.
   When a parameter is both input and output, [in,out] is used as attribute.
   Here is an example for the function \c memcpy:
--- a/doc/commands.dox
+++ b/doc/commands.dox
@@ -2066,9 +2066,9 @@ ALIASES  = "english=\if english" \
   present in the function declaration or definition.
 
   The \c \\param command has an optional attribute, \<dir\>, specifying the direction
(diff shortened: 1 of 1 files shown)


Mario-DL (approved):

doc/commands.dox
```diff
@@ -2066,9 +2066,9 @@ ALIASES  = "english=\if english" \
   present in the function declaration or definition.
 
   The \c \\param command has an optional attribute, \<dir\>, specifying the direction
-  of the parameter. Possible values are "[in]", "[in,out]", and "[out]",
-  note the [square] brackets in this description. Note that it is also possible to put
-  whitespace between the command and the \<dir\>.
+  of the parameter. Possible values are "[in]", "[out]", "[inout]", "[outin]",
+  "[in,out]", and "[in out]"; note the [square] brackets in this description.
+  Note that it is also possible to put whitespace between the command and the \<dir\>.
```
albert-github, line 2071: You are a bit faster than I was (I was completing another issue) but also prepared a patch with the text:
```
  The \c \\param command has an optional attribute, \<dir\>, specifying the direction
  of the parameter. Possible values are "[in]", "[in,out]", "[out,in]", and "[out]",
  note the [square] brackets in this description. Also supported are "[inout]", "[outin]"
  as well as the situation where the comma (`,`) is replaced by blanks.
  Note that it is also possible to put whitespace between the command and the \<dir\>.
```
your patch is missing the, uncommon, `"[out in]"` possibility

EduPonz: > You are a bit faster than I was (I was completing another issue) but also prepared a patch with the text:

I though I might as well :).

> your patch is missing the, uncommon, "[out in]" possibility

I've applied the missing ones. If you prefer we can go in your direction instead of specifying the all explicitly; I'll do whatever you prefer!

albert-github: Best is to distill the best of both.

doc/commands.dox
```diff
@@ -2066,8 +2066,11 @@ ALIASES  = "english=\if english" \
   present in the function declaration or definition.
 
   The \c \\param command has an optional attribute, \<dir\>, specifying the direction
-  of the parameter. Possible values are "[in]", "[out]", "[inout]", "[outin]",
-  "[in,out]", "[out,in]", "[in out]", and "[out in]"; note the [square] brackets in this description.
+  of the parameter. Possible values are "[in]", "[out]", and "[in,out]"; note the [square] brackets
+  in this description.
+  For the bidirecional values, directions "in" and "out" can be specified in any order, and they can
+  either be written altogether, or separated with a comma (`,`) or a space.
+  That means that for example values "[outin]" or "[in out]" are also valid.
```
albert-github, line 2073: Nice

EduPonz:
@albert-github I think this is ready, what do you think?

albert-github:
I think so too, it is up to @doxygen to integrate (or not).

Issue #10928, opened 2024-06-06
Can't be built with LLVM 19 due to removal of generic char_traits implementation


d125q:
Doxygen can't be built with LLVM 19 because there's no `char_traits<unsigned char>`.  (See [D157058](https://reviews.llvm.org/D157058).)  Using C++20 with `char8_t` [here](https://github.com/doxygen/doxygen/blob/3ce1c919a8cf86fb2a57571c382346e4523ae665/CMakeLists.txt#L168) and [here](https://github.com/doxygen/doxygen/blob/3ce1c919a8cf86fb2a57571c382346e4523ae665/src/qcstring.h#L95) makes it build, but there are some warnings from the logger about using `char8_t *` with an `%s` format specifier.

albert-github:
- Which version of doxygen did you try to build?
- probably relevant reference: https://en.cppreference.com/w/cpp/string/char_traits
- "logger" 
  - with the "logger" I assume you mean the `deps/spdlog`
  - what are the warnings, are they about `Char` (i.e. with capital `C`)?
    if so I think you should file a report at https://github.com/gabime/spdlog/issues

d125q:
Sorry for not providing a whole lot of information in the first place.

I am building from HEAD; commit 3ce1c91 at the time of writing.  The error is:
```
$LLVM_PREFIX/include/c++/v1/string:780:43: error: implicit instantiation of undefined template 'std::char_traits<unsigned char>'
```

Applying the following patch
```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f1103d6..af394ec 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,8 +81,8 @@ if (build_wizard)
     endif()
 endif()
 
-# use C++17 standard for compiling
-set(CMAKE_CXX_STANDARD 17)
+# use C++20 standard for compiling
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS ON)
 
@@ -165,7 +165,7 @@ if (CMAKE_SYSTEM_NAME MATCHES "Windows")
 endif()
 
 # needed for JavaCC
-set(JAVA_CC_EXTRA_FLAGS "-DJAVACC_CHAR_TYPE=\"unsigned char\"")
[...]

PR #10929, opened 2024-06-07
issue #10928 Setting JAVACC_CHAR_TYPE

albert-github:
`JAVACC_CHAR_TYPE` was already set in the main `CMakeLists.txt`  but its value was not used here.

### DIFF
--- a/src/qcstring.h
+++ b/src/qcstring.h
@@ -92,7 +92,7 @@ inline int qstricmp_sort( const char *str1, const char *str2 )
 
 int qstrnicmp( const char *str1, const char *str2, size_t len );
 
-using JavaCCString = std::basic_string<unsigned char>;
+using JavaCCString = std::basic_string<JAVACC_CHAR_TYPE>;
 
 /** This is an alternative implementation of QCString. It provides basically
  *  the same functions but uses std::string as the underlying string type


albert-github:
Strange that the release version under windows bails out, I also tried it locally with the GitHub Actions generated extecutable and the result is reproducible with directory results are:
```
 Directory of ...\testing\test_output_001

07-Jun-24  01:32 PM    <DIR>          .
07-Jun-24  01:32 PM    <DIR>          ..
07-Jun-24  01:32 PM    <DIR>          docbook
07-Jun-24  01:32 PM    <DIR>          html
07-Jun-24  01:32 PM    <DIR>          latex
07-Jun-24  01:32 PM    <DIR>          out
07-Jun-24  01:32 PM    <DIR>          rtf
07-Jun-24  01:32 PM                 0 warnings.log
               1 File(s)              0 bytes

 Directory of ...\testing\test_output_001\docbook

07-Jun-24  01:32 PM    <DIR>          .
07-Jun-24  01:32 PM    <DIR>          ..
               0 File(s)              0 bytes

 Directory of ...\testing\test_output_001\html

07-Jun-24  01:32 PM    <DIR>          .
07-Jun-24  01:32 PM    <DIR>          ..
07-Jun-24  01:32 PM                 0 doxygen_crawl.xhtml
07-Jun-24  01:32 PM                 0 index.qhp
               2 File(s)              0 bytes

 Directory of ...\testing\test_output_001\latex

07-Jun-24  01:32 PM    <DIR>          .
07-Jun-24  01:32 PM    <DIR>          ..
               0 File(s)              0 bytes

 Directory of ...\testing\test_output_001\out

07-Jun-24  01:32 PM    <DIR>          .
07-Jun-24  01:32 PM    <DIR>          ..
               0 File(s)              0 bytes

 Directory of ...\testing\test_output_001\rtf

07-Jun-24  01:32 PM    <DIR>          .
07-Jun-24  01:32 PM    <DIR>          ..
               0 File(s)              0 bytes
 
```
The executable bails out without any warning.
I now retried and got a message:
image


I also build a local release executable (with a bit a different compiler though: Microsoft (R) C/C++ Optimizing Compiler Version 19.28.29914 for x64) but this does not give any problems.

@doxygen do you have an idea?

doxygen:
@albert-github Is this crash only after the change in the pull request? Maybe the `JAVACC_CHAR_TYPE` define is not applied for all sources? I can image that would trigger undefined behavior. Anyway, a stack trace would be helpful (maybe `-DCMAKE_BUILD_TYPE=RelWithDebInfo` also shows the problem?).

albert-github:
- The only change is here that now the `JAVACC_CHAR_TYPE` is used in this one place and this is still defined as `unsigned char` so it should not make any change
- `-DCMAKE_BUILD_TYPE=RelWithDebInfo` I doubt I will get something as I don't have the used compiler and when generating the executable on GitHub Actions I still won't be able to get into the debugger, when downloading the executable I still cannot start the debugger as their is some weird restriction / block regarding my MS account.

albert-github:
Looks like a morge general problem an not just my change as I now see that also the latest change https://github.com/doxygen/doxygen/commit/6c356988444d77aed2a045dd13e84a195d7a6433 ([Refactoring] Make MemberTypeList a class instead of enum with bit mask ) bails out.

I'm going to build some versions locally...

albert-github:
I looked at the GitHub Actions and see:

Last good one https://github.com/doxygen/doxygen/actions/runs/9382662182/job/25834846349:
```
-- The C compiler identification is MSVC 19.39.33523.0
-- The CXX compiler identification is MSVC 19.39.33523.0
```

the bad ones:
```
-- The C compiler identification is MSVC 19.40.33811.0
-- The CXX compiler identification is MSVC 19.40.33811.0
```

So my conclusion is that it might be a compiler / optimizer problem.
Quick search (MSVC 19.40.33811.0 release crashes at startup) gives a nice hit: https://developercommunity.visualstudio.com/t/Invalid-code-generation-in-release-1940/10678572?sort=newest

and maybe: https://developercommunity.visualstudio.com/t/ICE-on-invalid-requires-clause/10674362?space=62&entry=problem&sort=newest

doxygen:
So it looks like the suggested workaround is to add `/D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` to the windows build flags. Let see if https://github.com/doxygen/doxygen/commit/76a25972d7dcf4d3c6d49f3f5b839dea3b408eaf fixes it.

albert-github:
Indeed it looks like https://github.com/doxygen/doxygen/commit/76a25972d7dcf4d3c6d49f3f5b839dea3b408eaf works around the problem regarding the running though we have to keep an eye out as I think that as soon as a fix with GitHub Actions has been made we should remove the workaround again.

doxygen:
@albert-github I agree

> Indeed it looks like [76a2597](https://github.com/doxygen/doxygen/commit/76a25972d7dcf4d3c6d49f3f5b839dea3b408eaf) works around the problem regarding the running though we have to keep an eye out as I think that as soon as a fix with GitHub Actions has been made we should remove the workaround again.

I agree

PR #10934, opened 2024-06-11
Spelling correction

albert-github:
Some small spelling and consistency corrections.

### DIFF
--- a/doc/faq.dox
+++ b/doc/faq.dox
@@ -187,7 +187,7 @@ in the HEAD section of the HTML page.
 In the past (prior to version 1.9.2) doxygen used a part of Qt 2.x for various
 utility classes. These have been replaced by STL container classes in the meantime.
 
-The GUI front-end called Doxywizard is based on a modern version of Qt.
+The GUI front-end called doxywizard is based on a modern version of Qt.
 Doxygen itself can also be used without the GUI.
 
 \section faq_excl_dir How can I exclude all test directories from my directory tree?
--- a/doc/extsearch.dox
+++ b/doc/extsearch.dox
@@ -259,7 +259,7 @@ have to be decoded before they can be used.
 When invoking the search engine as shown in the previous subsection, it should reply with
 the results. The format of the reply is
 <a href="https://en.wikipedia.org/wiki/JSONP">JSON with padding</a>, which is basically
-a javascript struct wrapped in a function call. The name of function should be the name of
+a JavaScript struct wrapped in a function call. The name of function should be the name of
 the callback (as passed with the *cb* field in the query).
 
 With the example query as shown the previous subsection the main structure of the reply should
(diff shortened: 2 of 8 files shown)


doc/faq.dox
```diff
@@ -187,7 +187,7 @@ in the HEAD section of the HTML page.
 In the past (prior to version 1.9.2) doxygen used a part of Qt 2.x for various
 utility classes. These have been replaced by STL container classes in the meantime.
 
-The GUI front-end called Doxywizard is based on a modern version of Qt.
+The GUI front-end called doxywizard is based on a modern version of Qt.
```
doxygen, line 190: I find it a bit inconsistent to change Doxywizard to doxywizard and the same time javascript and plantuml to JavaScript and PlantUML. 

I found [this](https://www.sussex.ac.uk/informatics/punctuation/capsandabbr/caps) source about the topic. In particular:

> m) The brand names of manufacturers and their products are capitalized:
> 
>     Maxine has bought a second-hand Ford Escort. 
>     Almost everybody owns a Sony Walkman. 
> 
> Note: There is a problem with brand names which have become so successful that they are used in ordinary speech as generic labels for classes of products. The manufacturers of Kleenex and Sellotape are exasperated to find people using kleenex and sellotape as ordinary words for facial tissues or sticky tape of any kind, and some such manufacturers may actually take legal action against this practice. If you are writing for publication, you need to be careful about this, and it is best to capitalize such words if you use them. However, when brand names are converted into verbs, no capital letter is used: we write She was hoovering the carpet and I need to xerox this report, even though the manufacturers of Hoover vacuum cleaners and Xerox photocopiers don 't much like this practice, either.

albert-github: The main purpose of the PR is to fix some issues ("followied", subpaging reference)  and to make some other words used consistently (this was, in my opinion, not the case).

I think brand names should be written as the brand owner defines and should not be a linguistic topic.

Looking at PlantUML I looked at the writing on their official web site (https://plantuml.com/) and its usage there.
For JavaScript I looked also at different sites and the usage there.
Regarding `doxywizard` this is consistent with the usage of `doxygen`.

doxygen: My point was that we should write Doxygen and Doxywizard everywhere if we want to write it correctly.

albert-github: I tried with my patch to make al things consistent and as far as I remember a couple of years ago it was decided that the lowercase version was to be used, unfortunately I cannot find the discussion about this (only reference that I found so far was https://github.com/doxygen/doxygen/commit/4f80d144f98fc998de5118855eec73797a65bf2e).

Like I wrote before:
> I think brand names should be written as the brand owner defines and should not be a linguistic topic.

and it should also be used consistently.

Looks currently that:
- doxygen / doxywizard should be used for the executable
- Doxygen / Doxywizard should be used for the product

Besides changes made I think also:
- thoughts have to be made about the logo, though this is an "artist" impression and can be in a different case from the running text
- what to do with  `doxysearch` / `doxyindexer` / `doxyapp`
- the web site as here are also lowercase and upper case versions of the word doxygen
- in the GUI also the lowercase version of the word doxygen is used maybe here also some changes are necessary with implications for the images in the documentation
- check also the "src" code (looks like it not be consistent)
-  ...

Issue #10953, opened 2024-06-21
Java Package Lists does not include nested packages


tkaken:
### The Problem
Starting with Doxygen 1.9.4, when running Doxygen optimized for Java, the Package List does not include nested packages. The problem still is happening in 1.11.0.

### Source code used

Here is the code with the nested packages.

```java
// File: MyBase.java
package tryInheritance.nested;

public interface MyBase {
	void doThis();
}
```
```java
//File: MyImpl.java
package tryInheritance.nested;

public class MyImpl implements MyBase {

	@Override
	public void doThis() {

	}
}
```
Here is the code for a class in single level package:

```java
// File: ClassInNonNestedPackage.java
package somePackage;

public class ClassInNonNestedPackage 
{
	public ClassInNonNestedPackage() {}
	public void doThis() {}

}
```
### Screenshots of results

Here are the screen shots of HTML output made using Doxygen 1.9.3 (this works correctly):
![DoxyClassList_1_9_3](https://github.com/doxygen/doxygen/assets/5786782/2d0f37e4-c9f8-471d-91e4-6222cf3e2bc3)
![DoxyPackageList_1_9_3](https://github.com/doxygen/doxygen/assets/5786782/51d92ae4-cc17-4ede-98af-548d5f93f871)

Here are the screenshots made using Doxygen 1.9.4. Note that I got the same results using 1.11.0. (this shows the nested package not in the Package List):
![DoxyClassList_1_9_4](https://github.com/doxygen/doxygen/assets/5786782/8356e4db-2e3f-42c9-8d6d-deef8f1c19a6)
![DoxyPackageList_1_9_4](https://github.com/doxygen/doxygen/assets/5786782/d1ed63f8-40bb-4b8a-8873-6d86d51e62dc)

I  see between the two versions (1.9.3 and 1.9.4), the page title changed from Packages to Package List.

### Extra information
I have attached the zip file [DoxyConfigsAndJavaNestPackageIssueSource.zip](https://github.com/user-attachments/files/15931006/DoxyConfigsAndJavaNestPackageIssueSource.zip) which contains the source code and Doxygen config files I used to generate the output.

albert-github:
Issue was introduced by:
```
Commit: 444dcca03797987cbcd5754d4a6544a0a0436824 [444dcca]
Date: Wednesday, May 4, 2022 7:55:45 PM
Merge pull request #9245 from albert-github/feature/issue_9241

[...]

PR #10968, opened 2024-06-27
issue #10953 Java Package Lists does not include nested packages

albert-github:
In the nested package `n1.n2` the package `n1` was seen as artificial package but this is not really the case.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -725,7 +725,8 @@ static Definition *buildScopeFromQualifiedName(const QCString &name_,SrcLangExt 
       if (newNd)
       {
         newNd->setLanguage(lang);
-        newNd->setArtificial(TRUE);
+        // looks like an atrificial namespace, but is in fact part of ns1.ns2, issue 10953
+        newNd->setArtificial(false);
         // add namespace to the list
         innerScope = newNd;
       }


src/doxygen.cpp
```diff
@@ -725,7 +725,8 @@ static Definition *buildScopeFromQualifiedName(const QCString &name_,SrcLangExt
       if (newNd)
       {
         newNd->setLanguage(lang);
-        newNd->setArtificial(TRUE);
+        // looks like an atrificial namespace, but is in fact part of ns1.ns2, issue 10953
+        newNd->setArtificial(false);
```
doxygen, line 729: @albert-github I fear this change will have negative side effects, such as also creating empty pages. In general, the class is 'artificial' in the sense that is it introduced by Doxygen, without finding the definition in the code. Maybe for Java packages we need to treat things differently...

albert-github: I could not think of any side effects but is well possible.
I remember that there have been some issues with CSharp as well regarding nesting. might be (at first glance in namespacedef.cpp or (more likely) in scanner.l. In the later I see with the rule:
```
<Using>({ID}{BN}*("::"|"."){BN}*)*({ID}|{OPERATOR})
```
the comment:
```
                                          if (yyextra->insideCS) /* Hack: in C# a using declaration and
                                                                    directive have the same syntax, so we
                                                                    also add it as a using directive here
                                                                  */
```

it should be checked if this is also possible for Java packages, probably, with the rule (limited to the case that we are in Java):
```
<PackageName>{ID}(("."|"\\"){ID})*
```

doxygen: I have made a different fix, see https://github.com/doxygen/doxygen/commit/269d60d6057f67b587e491ef5b1336fd071b96a8. 
See problem also happened for C++. e.g.
```c++
namespace X
{
  /** namespace Y */
  namespace Y
  {
    /** YY */
    class YY
    {
    };
  }
}
```

albert-github: Looks OK to me.

albert-github:
Closing issue as obsolete due to fix https://github.com/doxygen/doxygen/commit/269d60d6057f67b587e491ef5b1336fd071b96a8

PR #10986, opened 2024-07-05
made top and bottom row of asterisks same length to align them

ChrisIdema:
made top and bottom row of asterisks same length to align them because symmetry is good


albert-github:
My personal opinion here is just it is a matter of taste.

Issue #10995, opened 2024-07-11
Build issue: Build picks up wrong git info


ferdnyc:
This is similar to #10643, but coming from a different angle.

**Describe the bug**
I'm building in a Fedora tree where:

1. Our packaging configs are a Git repo, say it's `/src/doxygen/`
2. The package uses `doxygen-1.11.0.src.tar.gz` (the source tarball) as its source
3. Local builds are done by extracting the tarball to `/src/doxygen/doxygen-1.11.0`
4. `git` commands inside the `doxygen-1.11.0` source directory will therefore _work_, and return data from `/src/doxygen/.git`.

**To Reproduce**
```bash
mkdir /tmp/doxygen-test
cd /tmp/doxygen-test
git init
touch dummy-file.txt
git commit -m "Initial commit" dummy-file.txt

# The output here will be something like 'heads/main-0-<shorthash>'
git describe --all --long

wget 'https://www.doxygen.nl/files/doxygen-1.11.0.src.tar.gz'
tar xf doxygen-1.11.0.src.tar.gz
cd doxygen-1.11.0
cmake -B _build -S .
cmake --build _build
```

**Expected behavior**
Doxygen builds with no Git version

**Actual behavior**
`doxygen --version` shows `1.11.0 (<the_hash_from_git_describe_above>*)`

**Version**
Building 1.11.0 from source

**Additional editorializing**
IMHO it would be better to just _look for_ a `.git` directory at the root of the Doxygen source, and disable all Git processing unless one's found. Blindly running `git` commands and making decisions based on whether or not they succeed is far too error-prone.

ferdnyc:
> IMHO it would be better to just _look for_ a `.git` directory at the root of the Doxygen source

Actually, correction — make that: "look for `.git` at the root of the Doxygen source", since in a Git [worktree](https://git-scm.com/docs/git-worktree) there will be a `.git` **file** that points to the parent source tree. Even if `.git` is a file, not a directory, `git` commands from within that tree can be considered valid.

albert-github:
It is a bit a twisted example but it, unfortunately, is possible.

I think in the first block there are a few commands missing (at least as far as I can tell:
- git add dummy-file.txt
- git config --global user.email "you@example.com"
- git config --global user.name "Your Name"

[...]

PR #10996, opened 2024-07-11
issue #10995 Build issue: Build picks up wrong git info

albert-github:
Create variable to check whether a doxygen `.git` directory exists, i.e. we have a git version of doxygen or just a distribution with the source files.


ferdnyc (approved):
Looks good to me! I tested building both:

1. in a Git worktree with the PR branch checked out. The SHA was still set correctly to the HEAD of the branch.
2. as a Fedora package, with the PR commit extracted using `git format-patch` and added to the package `.spec` file. The resulting package shows only `1.11.0` for `doxygen --version`.

PR #11008, opened 2024-07-18
quest #11007 Is it possible to add usage relations to the collaboration diagrams for properties?

albert-github:
Adding to `DOT_UML_DETAILS the possibilities:
- `NONE_FLD` same as `NONE`
- `NONE_ATTR`, Doxygen will not generate attributes on the edges between the classes in the UML graphs.
- `NONE_ATTR_FLD`, Doxygen will not generate fields with class member information nor attributes on the edges between the classes in the UML graphs.

Example: [example.tar.gz](https://github.com/user-attachments/files/16281112/example.tar.gz)

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -3888,8 +3888,12 @@ If the \c DOT_UML_DETAILS tag is set to \c NO, Doxygen will
 show attributes and methods without types and arguments in the UML graphs.
 If the \c DOT_UML_DETAILS tag is set to \c YES, Doxygen will
 add type and arguments for attributes and methods in the UML graphs.
-If the \c DOT_UML_DETAILS tag is set to \c NONE, Doxygen will not generate
+If the \c DOT_UML_DETAILS tag is set to \c NONE or \c NONE_FLD, Doxygen will not generate
 fields with class member information in the UML graphs.
+If the \c DOT_UML_DETAILS tag is set to \c NONE_ATTR, Doxygen will not generate
+attributes on the edges between the classes in the UML graphs.
+If the \c DOT_UML_DETAILS tag is set to \c NONE_ATTR_FLD, Doxygen will not generate
+fields with class member information nor attributes on the edges between the classes in the UML graphs.
 The class diagrams will look similar to the default class diagrams but using
 UML notation for the relationships.
 ]]>
@@ -3897,6 +3901,9 @@ UML notation for the relationships.
       <value name="NO" />
       <value name="YES" />
       <value name="NONE" />
+      <value name="NONE_FLD" desc='same as NONE'/>
+      <value name="NONE_ATTR" />
+      <value name="NONE_ATTR_FLD" />
     </option>
     <option type='int' id='DOT_WRAP_THRESHOLD' defval='17' minval='0' maxval='1000' depends='HAVE_DOT'>
       <docs>
--- a/src/dotnode.cpp
+++ b/src/dotnode.cpp
@@ -424,7 +424,9 @@ void DotNode::writeLabel(TextStream &t, GraphType gt) const
     t << "<<TABLE CELLBORDER=\"0\" BORDER=\"1\">";
     t << hr_start << convertLabel(m_label,true) << hr_end;
     auto dotUmlDetails = Config_getEnum(DOT_UML_DETAILS);
-    if (dotUmlDetails!=DOT_UML_DETAILS_t::NONE)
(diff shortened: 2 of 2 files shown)


src/config.xml
```diff
@@ -3888,15 +3888,22 @@ If the \c DOT_UML_DETAILS tag is set to \c NO, Doxygen will
 show attributes and methods without types and arguments in the UML graphs.
 If the \c DOT_UML_DETAILS tag is set to \c YES, Doxygen will
 add type and arguments for attributes and methods in the UML graphs.
-If the \c DOT_UML_DETAILS tag is set to \c NONE, Doxygen will not generate
+If the \c DOT_UML_DETAILS tag is set to \c NONE or \c NONE_FLD, Doxygen will not generate
 fields with class member information in the UML graphs.
+If the \c DOT_UML_DETAILS tag is set to \c NONE_ATTR, Doxygen will not generate
+attributes on the edges between the classes in the UML graphs.
+If the \c DOT_UML_DETAILS tag is set to \c NONE_ATTR_FLD, Doxygen will not generate
+fields with class member information nor attributes on the edges between the classes in the UML graphs.
```
doxygen, line 3896: I find it hard to understand these options. Especially since the documentation now says what doxygen doesn't show, not what it does show. I think the original `NO` and `YES` was meant to control if detailed information about type and arguments was shown or not. Then `NONE` was added to show no field data at all. With these new options, I'm a bit lost as to what will be shown.

albert-github: Some thoughts 

When we look at the image:
image

We see here a number of items:
- What I call fields (FLD):
  - in the class `other`
    - publicMethod();
    - privateMethod();
    - mPrivateMember;
  - in the class `main`
    - other()
    - privateMethod();
- What I call attributes (ATTR)
  - between the classes `other` and `main` (along the edge):
    - mOther;


Currently (doxygen 1.11.0) the option `NONE` will remove the fields but not the attributes.
With this proposed PR (derived from #11007) a more granular approach is possible:
- `NONE` unchanged for consistencty with the older versions
- `NONE_FLD` is identical to `NON`, added for simularity / consistency with the other added options
- `NONE_ATTR`  the attributes are removed
- `NONE_ATTR_FLD` both the attributes and fields are removed (so as actually requested in #11007)

When the `NONE` should also have removed the attributes in the past (and this is now considered as a bug) we can update the PR to reflect this (and not add the new options).

I don't get the:
> Especially since the documentation now says what doxygen doesn't show, not what it does show.

It is not that unclear to me.
Thinking  about it we could change the formulation a bit 
- `NONE` show only the attributes on the edges between the classes in the UML graphs
- `NONE_ATTR` show just the fields of the classes in the in the UML graphs.

Maybe a possibility to change the `NONE_ATTR` in `FLD`, `NONE_FLD` in `ATTR` but home to call the current `NONME` and the proposed `NONE_ATTR_FLD`?

doxygen: What you call fields, in UML are called attributes, and what you call attributes, in UML are called dependencies 😉. See [here](https://www.visual-paradigm.com/guide/uml-unified-modeling-language/uml-class-diagram-tutorial/) for instance.

So I think we have the following (existing) levels of details:
- ALL (YES): show attributes with types and parameters
- COMPACT (NO): show attributes without types and parameters
- NONE: show no attributes

For each of the detail levels, the relation labels could be shown or hidden. Instead of adding 3 new values to the existing option, it is then probably better to add a new option `DOT_UML_RELATIONS`. But I'm also not so sure that this actually solves the original problem. Didn't the user want to show the relations even when `EXTRACT_PRIVATE` is disabled?

albert-github: Thanks for correcting the terminology, the terms I used were in respect to UML a bit confusing (when there is no explanation given. Maybe even `operations` should be introduced into the possibilities. Nice referenced page).

the user stated:
> I would like to be able to generate a collaboration diagram that includes usage relations but does not show the private members and methods. 

But at the moment that `EXTRACT_PRIVATE=NO` there is "no" relation between the classes `main` and `Other` anymore based on `dependencies` but there are still relations between `main` and `Other` namely based on the return types of some `operations` (her probably `other() would use the private member) and on the `Q_PROPERTY` (and this is not the case in the example but it might also be that the arguments of the `operations` contain the other class name).

- The idea about `DOT_UML_RELATIONS` is a nice idea but indeed doesn't solve the original request this setting though would give more flexibility (and probably less confusion compared to my implementation)
- what to do with the `operations`?
- what to do with the return types and argument types?

PR #11018, opened 2024-07-22
issue #11015 In Doxygen 1.11.0, Include Graphs are not made for Java

albert-github:
Adding code so that also include graphs are shown in Java.


albert-github:
@doxygen this is a proposed patch, as the modules are geared towards C++ I'm not sure whether this has side effects or not.

Issue #11016, opened 2024-07-21
Faulty xml generated when markdown header is too deep.


ilitzroth:
**Describe the bug**
Faulty xml generated when you add markdown header that is too deep:

**test.h**
```
/*!
 * \brief BRIEF
 *
 * ### SECT3
 */
class test{
};
```

Generates this classtest.xml

```
<?xml version='1.0' encoding='UTF-8' standalone='no'?>
<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="1.11.0" xml:lang="en-US">
  <compounddef id="classtest" kind="class" language="C++" prot="public">
    <compoundname>test</compoundname>
    <includes refid="bug1_8h" local="no">bug1.h</includes>
    <briefdescription>
      <para>Encapsulates. </para>
    </briefdescription>
    <detaileddescription>
      <para>blah </para>
      <sect3 id="classtest_1autotoc_md0">
        <title>BLAH</title><para>more blah </para>
      </sect3>
    </detaileddescription>
    <location file="bug1.h" line="9" column="1" bodyfile="bug1.h" bodystart="9" bodyend="10"/>
    <listofallmembers>
    </listofallmembers>
  </compounddef>
</doxygen>
```
According to the compound.xsd detaileddescription cannot contain a sect3. It is a descriptionType

```
<xsd:complexType name="descriptionType" mixed="true">
    <xsd:sequence>
      <xsd:element name="title" type="xsd:string" minOccurs="0"/>	    
      <xsd:element name="para" type="docParaType" minOccurs="0" maxOccurs="unbounded" />
      <xsd:element name="internal" type="docInternalType" minOccurs="0" maxOccurs="unbounded"/>
      <xsd:element name="sect1" type="docSect1Type" minOccurs="0" maxOccurs="unbounded" />
    </xsd:sequence>
  </xsd:complexType>
```

**To Reproduce**
Run the provided header file through doxygen with 
```
MARKDOWN_SUPPORT       = YES
GENERATE_XML = YES
```
**Expected behavior**
A valid xml with respect to compound.xml is produced. 

**Version**
1.11.0

**Additional context**
trying to generate texinfo from the xml output of doxygen with xsd: https://www.codesynthesis.com/projects/xsd/
This comes from trying to run thirdparty headers through doxygen.

ilitzroth:
I can also add that setting 
[...]

PR #11029, opened 2024-07-25
issue #11016 Faulty xml generated when markdown header is too deep

albert-github:
- Improving HTML table of contents in respect to skipped section levels.
- adding css entries for level5 and level6


doxygen:
@albert-github I think the same effect can be achieved with less CSS code (see https://github.com/doxygen/doxygen/commit/24d1cff91b4c2e14ae549219792d3f9bf9ca5c8a)

Issue #11033, opened 2024-07-27
Illegal command '\ifile' found as part of a title section 


aotto1968:
**Describe the bug**
get an *internal* error related to `@par` and `@copybrief` 
~~~
\par EXAMPLE: \copybrief MkStringR
\snippet test.h MkStringR
~~~
- the goal was to create extra docu which refers to the `MkStringR` code
- the `@snippet` is **OK**
- the problem is the `@par` heading which is **not** what I expect
- I get also an error message on stderr
~~~
.../doxygen_bug__par_with_copybrief/test.doc:6: warning: Illegal command '\ifile' found as part of a title section
.../doxygen_bug__par_with_copybrief/test.doc:6: warning: Illegal command '\iline' found as part of a title section
~~~
- another problem: I get *TWO* error message for *ONE* bug

**Screenshots**
![image](https://github.com/user-attachments/assets/768082c2-f4da-4469-b075-a71b9ad6928f)

**To Reproduce**
use the attachment: [doxygen_bug__par_with_copybrief.tar.gz](https://github.com/user-attachments/files/16401242/doxygen_bug__par_with_copybrief.tar.gz)

**Expected behavior**
1. better error message … OR …
2. allow the `@copybrief` data be an `@par` argument

**Version**
> doxygen -V
1.12.0 (14fc65f92184f8d788d5fb2723fb28e773dc4a5a)
    with sqlite3 3.42.0.

albert-github:
I've just pushed a proposed patch, pull request #11033

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.12.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #11035, opened 2024-07-28
issue #11033 Illegal command '\ifile' found as part of a title section

albert-github:
Handle the `\ifile` and `\iline` command for the `\par` command. We also needed to properly reset the state properly as otherwise
```
\par EXAMPLE: \copybrief MkStringR

\par EXAMPLE: \copydetails MkStringR
```
would give a warning regarding the second `\par` command.

### DIFF
--- a/src/docparser.cpp
+++ b/src/docparser.cpp
@@ -1398,6 +1398,22 @@ reparsetoken:
         case CMD_IMAGE:
           handleImage(parent,children);
           break;
+        case CMD_ILINE:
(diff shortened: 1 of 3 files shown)


doxygen:
@albert-github Added `yyextra->stateStack` and `pushState()` and `popState()` to avoid the need to pass the internal state back the client (`DocParser`).

albert-github:
Is a good idea, looks OK

Issue #11006, opened 2024-07-17
Unable to get property 'style' of undefined or null reference


Dowan35:
Hello, when running doxygen.exe 1.11 with a doxyfile.dox 1.11 too i have this error message spawning randomly and not consistently in the generated .chm when oppening sections.

![Screenshot 2024-07-17 094254](https://github.com/user-attachments/assets/85628fcb-07d6-40c4-a9fa-7370003bd5fb)

The doxyfile use hhc.exe 1.31 to create the chm and I have no problems when using doxygen.exe 1.9.3 is not causing this issue but this version does not support tag 'INPUT_FILE_ENCODING' that fix the issues of the special characters 'é' 'è' replaced with '?' elements in the Namespaces and Classes auto generated doc from a .cs project. (tried with UTf8, first part is ok, second part  (namespaces and classes auto generated) does not accept special characters, this is the opposite when choosing ISO-8859-1)

So that is why I need a working 1.11 version, tried to replace the resize.js as in #11000 but I cant compile the doxygen project on my machine (no admin rights).

The issue is still here when changing DISABLE_INDEX and GENERATE_TREEVIEW to yes or no, I see that in v1.11 everytime I open a book/page, the right scroll bar blink and the displayed page goes back to the top, something that does not append in v 1.9.3, might be a problem when displaying the content, the issue does not comes from my personal content.

Thanks for your help.

UPDATE; "fixed" by executing with doxygen v 1.10 instead of 1.11

Dowan35:
I dont have anymore the .js error when compiling with doxygen 1.10. Thanks for your answer. Yes chm is an ooold thing but there are some people here that like using old things.

albert-github:
So the problem is stell present with the 1.11.0 version.

@doxygen
We currently have a jquery.js that is the minimized version of the jquery official together with some tools. 
I did a small experiment with this doxygen version where I added a number of newlines so the the line does not overflow anymore.
There also exists on the jquery website a "normal" / readable version of jquery.js and this version is a little bit bigger and also contains some comments. This version does work as well. 
Probably this "normal" version is loading a bit slower so not so good for non chm use.

Would it be possible to supply in the deps directory also a "doxygen" jquery version based on the "normal" jquery.js so that this file can be used and loaded when `GENERATE_HTMLHELP = YES`.
(just providing the file would be sufficient, I can have a look at the the changes necessary for `GENERATE_HTMLHELP = YES`).

albert-github:
[...]

Issue #10839, opened 2024-04-30
Update German translator


paule32:
Hello,
I have found a translation error in the latest version of doxygen.
On left side of the screen.png - in attached zip archive, you can see German text.
But the "Topics" fold should be "Themen".
On right side of the project screen shot, you can see the same - mixed English with German.
I would like to have **ALL** German, not mixed.
When doxygen starts, I get the message that translation for german does not effect on some reason - either ?
How can be fixed this ?
[packed.zip](https://github.com/doxygen/doxygen/files/15159146/packed.zip)

albert-github:
When you start doxygen with `OUTPUT_LANGUAGE=German` you will get the warning:
```
warning: The selected output language "german" has not been updated
since release 1.8.15.  As a result some sentences may appear in English.
```
which indicates that the newest strings are not translated and thus the English fallback is used. One of these is "Topics" a term that is introduced in the version 1.9.8

On the page https://doxygen.github.io/doxygen-docs/d2/d78/pg_trans.html you can see that there are 56 methods that have not been implemented.

This can be fixed when these translations are provided meaning:
- provide a pull request where the file src/translator_de.h has been updated

paule32:
can I change the translations or parts of it without compiling the sources ?

albert-github:
Not really, only when it is a text in the doxygen layout file.

albert-github:
I don't think this issue is completed
- there is a proposed pull request #10845 (thanks) but this one hasn't been integrated in the master version yet
- in the proposed pull request #10845 only part of the outstanding German translation issues are handled

albert-github:
Code of #10845 has been integrated in master on GitHub.

Missing translations are (From: https://doxygen.github.io/doxygen-docs/d2/d78/pg_trans.html):
```
TranslatorGerman   (TranslatorAdapter_1_8_15)  29 methods to implement (9 %)
----------------

  Implements 280 of the required methods (90 %).

  Missing methods (should be implemented):

    virtual QCString trCompoundReferenceSlice(const QCString & clName, ClassDef::CompoundType compType, bool isLocal)
    virtual QCString trConstantDocumentation()
    virtual QCString trConstants()
    virtual QCString trDataMemberDocumentation()
    virtual QCString trDataMembers()
    virtual QCString trDictionaries()
    virtual QCString trDictionaryDocumentation()
    virtual QCString trExceptionDocumentation()
    virtual QCString trExceptionHierarchy()
[...]

PR #11046, opened 2024-08-02
changed jquery.js because Script Error in CHM Project's

paule32:
Line: 33
Column: 395

commenting: /* throw e*/


albert-github:
- The file jquery.js should not be changed at templates/html as it is generated from the files on deps/jquery
  did you try:
  ```
  commit 4ff955064a0133f58f1055d52caed8c2c6eecfb8
  Author: Dimitri van Heesch <doxygen@gmail.com>
  Date:   Fri Jul 19 11:12:20 2024 +0200

    issue #11006 Unable to get property 'style' of undefined or null reference
   ```
- what is the idea behind the file `cmake-multi-platform.yml`

paule32:
I dont know where the .yml File comes from - I dont remember me, but it could be a anoying Test with the Actions.
If it is not relevant, and it is not Part of Doxygen, then you can delete it.
I had patched the jquery-3.6.0.min.js File in the deps source tree.
All check have passed, so I think, this Ticket here can be close, because it is not a Doxygen Problem, directly.

src/main.cpp
```diff
@@ -30,6 +30,10 @@
  */
 int main(int argc,char **argv)
 {
+  // do a check, if CHM HTML WebView is open ...
+  #if defined(_WIN32) || defined(_WIN64)
+  
+  #endif
```
albert-github, line 36: Something missing here ?

albert-github:
Regarding the error shown in: https://github.com/doxygen/doxygen/pull/11046#issue-2444716675
this typically shows to me that in one of the js scripts used (not the jquery.js script) there is a construct used that is not understood by the CHM viewer (this is also the one of the reasons why e.g. the "HTML_COPY_CLIPBOARD" had to be disabled).
In my opinion not the jquery.js should be adjusted but the offending script or this script / feature should be disabled.

src/main.cpp
```diff
@@ -30,9 +75,29 @@
...
+    */
   #if defined(_WIN32) || defined(_WIN64)
-  
+  const TCHAR *processName = _T("msedgewebview2.exe");
+  DWORD processId = 0;
+  if (IsProcessRunning(processName, &processId)))) {
+      int response = MessageBox(0,
+      _T("The HTML Viewer is already open.\nWould you exit, or close the viewer ?"),
+      _T("Warning"),
+      MB_ICONQUESTION | MB_OKCANCEL);
```
albert-github, line 89: My 2 cents
- Why should doxygen test whether or not a viewer is running and even without looking if a chm file is generated?
- popping up a message box in the console application doxygen is completely not done.

src/commentscan.l
```diff
@@ -3262,7 +3262,9 @@ static bool handleSection(yyscan_t yyscanner,const QCString &s, const StringVect
   else if (s=="subparagraph")    yyextra->sectionLevel=SectionType::Subparagraph;
   else if (s=="subsubparagraph") yyextra->sectionLevel=SectionType::Subsubparagraph;
   // raise it if requested
-  yyextra->sectionLevel = std::min(yyextra->sectionLevel + yyextra->raiseLevel,SectionType::MaxLevel);
+// paule32 - msvc problem with std::min
+// yyextra->sectionLevel = std::min(yyextra->sectionLevel + yyextra->raiseLevel,SectionType::MaxLevel);
+  yyextra->sectionLevel = min(yyextra->sectionLevel + yyextra->raiseLevel,SectionType::MaxLevel);
   // rewrite the update section level to the output
```
albert-github, line 3268: - What kind of problems with the compiler
- Which version of the compiler

src/main.cpp
paule32: Because, when you would like to generate a CHM HTML Project for using the Microsoft HTML Workshop Compiler, to produce WebView1/2 Windows compiled html files - for e.g. product documentation, Doxygen takes no care, if the WebViewer is already open, yet.
When it is open, you can not generate a CHM File with hhc.exe - or it is empty.

The popout of the MessageBox can be re-design, so you get a message in the message.log file ?
It was a on-the-fly idea from me, that could be improof, indeed.

src/commentscan.l
paule32: the Linux Compilers for Ubuntu Action's takes care, and they work - I thinking so, because I have 6 Workflows running well. The other Workflows for Windows crashed when you would use **std::min**.
I think there is a missing Header File for it.
I don't work with MSVC, so I can not say what the problems is, directly.
Maybe a missing header - or **using namespace std;**
I don't compile it on my Computers, it is the Workflow from Doxygen GitHub Account, that I had forked.
I run Action's directly under GitHub - it takes some minutes, but it is possible.
Because, I download a archive of Doxygen last days, and tried to Compile under Mingw64 Bit (MSYS2) under Windows, and getting a Bison 3 problem, so the build CMakefile was created, but the make Makfile produce error, so I decide to compile under GitHub - where I can test for Windows, Linux, and MacOS in (nearly) one steps.

**BTW**: You noticed me about the yml File... I delete it on my repro - I hope it is all well.

albert-github:
- tests should be run locally first
- please cancel the workflow runs of the old tests you made, they are using unnecessary GitHub Action resources and are blocking other runs
- I have no idea what you intent with the myriad of changes but they don't look good at all, what is your intention here.

doxygen:
@paule32 I have to agree with Albert here. Using `using namespace std;` is a big NOT OK anyway. This will not be accepted in master. Having Windows or GNU specific checks outside of `portable*` is another NOT OK. The `IsProcessRunning()` code is not something that belongs in doxygen (it should not be aware of processes it didn't start itself). Also squash your commits, so they do not pollute the change history.

paule32:
Hello Doxygen Maintainer,
I apologies me for trouble.
I had thinking, that the Doxygen Project, that I forked, is unter my Control.
So, I had think, that the Runners, and Actions take effected on my GitHub Account, only.
Thanks for clearify this.
I have stop, and delete the Runner Workflows, so you should have more Power for your Compile Session.
I am new to github.com (naive spoken).
I hope, the source Tree on Doxygen is not changed by me on your Side.
If so, please revert the changes - you are familar enough/much more than me.
I would follow your Advice, to compile locally, first.

Thanks for reading
paule32

doxygen:
@paule32 The fork is indeed under your control. But as soon as just make a pull request towards the official doxygen repo, it will start running the CI pipeline. I understood that you are now using the CI pipeline as a kind of remote compiler. This is not the intention. You need to compile and test locally first before you make a pull request. If possible, use at least two platforms (out of the set Linux, Windows, macOS).

doc/maintainers.txt
```diff
@@ -72,6 +72,7 @@ Benoît BROSSE: Benoit dot BROSSE at ingenico dot com
 TranslatorGerman
 Peter Grotrian: Peter dot Grotrian at pdv-FS dot de
 Jens Seidel: jensseidel at users dot sf dot net
+Jens Kallup: kallup at web dot de
```
doxygen, line 75: This is meant for language translators. You added your name, but I see no changes to `translator_de.h`. Ideally make a separate pull request where you add the missing translation to `translator_de.h` and then you can also add your name here if you plan to help out with future translations as well.

src/commentscan.l
```diff
@@ -3991,7 +4006,14 @@ static bool handleToc(yyscan_t yyscanner,const QCString &, const StringVector &o
         }
         else
         {
-          level = level<=0 ? SectionType::MaxLevel : std::min(level,SectionType::MaxLevel);
+          // add: paule32 (testing: MSVC and GNU G++)
+#ifndef __GNUC__
+          if (level <= 0) level = SectionType::MaxLevel;
+          else            level = min(level, SectionType::MaxLevel);
```
doxygen, line 4012: I don't understand which issue you have with `std::min`, but this is definitely not the way to solve it.

doc/maintainers.txt
albert-github: @doxygen related #10839

PR #11073, opened 2024-08-14
Incorrect handling of spaces before `@page` command

albert-github:
When having some initial space before the `@page` command this is not considered properly, so when having:
```
      @page test_page A Test Page123456
```
this results in something like:
```
A Test Page123456

23456
```
Note the line with `23456`

Example: [example.tar.gz](https://github.com/user-attachments/files/16611687/example.tar.gz)


doxygen:
@albert-github I modified the solution a bit, see https://github.com/doxygen/doxygen/commit/d09922abb551d089dfb9659cfb67bc3d77af5676. Result should be the same.

PR #11076, opened 2024-08-16
Update translator_gr.h

nickreserved:
update to 1.11.0


albert-github:
Is it correct that you added #11077 to replace this one as you made a small mistake in the order of statements?
(you could have corrected it in this PR).

Issue #10649, opened 2024-02-12
Building failure: ISO C++17 does not allow 'register' storage class specifier [-Wregister]


wilstoff:
**Describe the bug**
When running `cmake/make` with
```
cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -Wno-dev
```
it seems to generate a bunch of code with the register keyword that seems to be deprecated.
Which errors out when building:
```
/build/bstoffel/rpmbuild/BUILD/doxygen-1.10.0/generated_src/mscgen_lexer.cpp:899:2: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register yy_state_type yy_current_state;
        ^~~~~~~~~
/build/bstoffel/rpmbuild/BUILD/doxygen-1.10.0/generated_src/mscgen_lexer.cpp:900:2: error: ISO C++17 does not allow 'register' storage class specifier [-Wregister]
        register char *yy_cp, *yy_bp;
...
```

**Expected behavior**
I expect the code to compile appropriately

**Version**
Building version 1.10.0

**Additional context**
Running on centos 7 machine with clang 10.0.0 installed.
I have tried adjusting the C++ version to 14/11 and still has the same errors.
I'm sure there is an option I'm missing or can alter some environmental variable in order for it to generate valid C++17 code.

albert-github:
When I build doxygen I see in the resulting `mscgen_lexer.cpp` file:
```
/** The main scanner function which does all the work.
 */
YY_DECL
{
        yy_state_type yy_current_state;
        char *yy_cp, *yy_bp;
        int yy_act;
```

I'm using flex version 2.6.4 locally.
On GitHub actions we also use the CLang compiler, version 10.0.0 with flex 2.6.4 (see e.g. https://github.com/doxygen/doxygen/actions/runs/7877930596/job/21495098205) and here I don't see any problems there.


- which version of flex are you using? or are you using `lex`?
- can you post the resulting `mscgen_lexer.cpp`  file in a, compressed, tar or zip file attached to the issue?

Looking in the file `CMakeLists.txt` we see for the MacOs Darwin version:
```
    set(CMAKE_CXX_FLAGS "-Wno-deprecated-register -mmacosx-version-min=${MACOS_VERSION_MIN} ${CMAKE_CXX_FLAGS}")
[...]

PR #11113, opened 2024-09-02
issue #10649 Building failure: ISO C++17 does not allow 'register' storage class specifier [-Wregister]

albert-github:
Fixing the problems when using a  `< 2.6.0` version of flex that still issues the word `register`


CMakeLists.txt
```diff
@@ -107,6 +107,21 @@ endif()
 # produce compile_commands.json
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+if (FLEX_VERSION VERSION_LESS 2.6.0)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Dregister=")
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Dregister=")
```
doxygen, line 112: @albert-github Isn't it enough to set `CMAKE_CXX_FLAGS` and not all the build type specific settings as well? This [post](https://stackoverflow.com/a/60622717/784672) seems to suggest so.

albert-github: Indeed looks like not to be necessary. `CMAKE_CXX_FLAGS_...` are supplementary to `CMAK_CXX_FLAGS`.

PR #11122, opened 2024-09-07
bug 729344 Ignoring an items when / or \ is placed after \addindex

albert-github:
The handling of the `addindex` command  is here just a copying action and therefore no extra characters should be added so instead of the `cmdName` the original text (`yytext`) should be outputted.

Example: [example.tar.gz](https://github.com/user-attachments/files/16918986/example.tar.gz)


albert-github:
It might be that there are other commands as well for which this reasoning is valid, this should be investigated after that this proposed patch is accepted and integrated.

PR #11137, opened 2024-09-12
Fix linking to strong enum values from tagfiles if they're inside a class or namespace

mosra:
The refactor in 35711f2d14acc81cff83f81f04c566ad8c8c6fe8 (part of 1.9.5) lost this particular bit, causing the code to be only able to link to enum values from tagfiles where the enum was in global scope, not in any subnamespace.

Expanded the strong enums test with various possible combinations to ensure this doesn't regress again. In particular, without the change in `doxygen.cpp`, the following warnings are produced:

```
doxygen/testing/058_strong_enum.cpp:40: warning: unable to resolve reference to 'external::ns::Enum1::Flag' for \ref command
doxygen/testing/058_strong_enum.cpp:44: warning: unable to resolve reference to 'external::ns::Class::Enum2::Flag' for \ref command
doxygen/testing/058_strong_enum.cpp:48: warning: unable to resolve reference to 'ExternalClass::Enum1::Flag' for \ref command
```

In particular, linking to the enum did work, but to its values not, only if the enum itself was in global scope. The tagfile is hand-crafted to make visual verification of the XML output easier, hopefully there aren't any serious errors.

**Note:** I did not handle the case of linking to weak enum values by prefixing them with the enum, i.e. `@ref WeakEnum::WeakEnumFlag`, `@ref ExternalWeakEnum::ExternalWeakEnumFlag` etc. For locally defined enums this works, but for the tagfile it doesn't, as the tagfile doesn't preserve any relation between the enum and the value. (And to my knowledge it never did.) Which means it isn't really a regression that needs a fix, and I'm fine with this part not working.


doxygen:
@mosra Not clear why the CI pipeline fails to find the tag file. Needs further debugging. Maybe you can remove the part of the example that depends on it for now, if you want the pull request to be merged already.

mosra:
I'll extract the tests to a dedicated file to make them simpler (in retrospect that's what I should have done initially, the original test verifies something else entirely), and will attempt some more debugging on my end.

But I feel like there should be at least one such test for the tagfiles, because from my past experience this is an area that's prone to easy breakage. So I'll try to do my best to make that work, and only if I fail I reduce it to just that two-line change :)

Issue #11010, opened 2024-07-18
Automatic link to typedef no longer generated if typedef declared twice


abbeybreeding:
**Describe the bug**
We had a typedef declared twice, one with documentation and the other without. If the one without is first in the path, the link to the documentation for the typedef will not be generated.

**Screenshots**
1.8.9.1
![image](https://github.com/user-attachments/assets/05a7f387-03cb-4c1c-a15c-0bafe0df8d15)
1.11.0
![image](https://github.com/user-attachments/assets/586197d9-f53d-4e80-bec6-c4fe00474b34)

**To Reproduce**
[testcase.zip](https://github.com/user-attachments/files/16289294/testcase.zip)

**Expected behavior**
While our code is incorrect, and we have addressed that, it would still seem that the link to the typedef documentation would still be generated.

**Version**
This seemed to "work" as we would expect in 1.8.9.1 (see screenshot). It no longer "works" in 1.11.0 or the trunk.

**Additional context**
Various things, such as changing the path, adding a `\typedef` tag, sometimes adding a `\ref` tag with work around the issue.

albert-github:
Did a quick "bisect" and it looks like that the commit (b290399fab92df45376103d7c094f053f24495fa ):
```
Commit: b290399fab92df45376103d7c094f053f24495fa [b290399]
Date: Tuesday, May 24, 2022 8:12:07 AM
Refactoring: reimplement getDefs using symbol resolver
```

Causes the problem.

@abbeybreeding when specifying trunk / master always specify the full version as this is a moving target, so `doxygen -v`.

I ran with the current master (1.12.0 (0b557b30eee328189629b2d3cc89f4774f3ad5ed)) and the problem is still present.

Might be related to:
- #10180 
- #9379

the fix for 
- #10994 

didn't solve this issue.

abbeybreeding:
@albert-github  I'm so sorry about not adding the `doxygen -v` to the trunk version, I will make sure to do so in the future!  :-)

doxygen:
@abbeybreeding Please verify if the referenced commit fixes the problem for you. Do not close the issue, this will be done automatically when the next official release becomes available.

abbeybreeding:
@doxygen Hi Dmitri.  This does fix the example in the test case for us. Thanks to you and the team for fixing this.

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.12.0.
[...]

PR #11141, opened 2024-09-13
Properly consider namespace / class scopes when merging typedefs together

mosra:
With Doxygen 1.12, the typedefs in the following snippet get combined together even though they are in different namespaces. Which, depending on the context, results in various issues, like being unable to link to either of them, or they not even appearing in generated documentation, etc.

```cpp
namespace Root {

/** @brief A typedef */
typedef int Int;

/** @brief A using declaration */
using Float = float;

namespace Sub {

/** @brief A struct */
struct Struct {};

/** @relatedalso Struct
 * @brief A typedef in Sub that doesn't get merged with Root::Int
 */
typedef int Int;

/** @relatedalso Struct
 * @brief A using declaration in Sub doesn't get merged with Root::Float
 */
using Float = float;

}}
```

And, as a side effect, because of the `@relatedalso`, the XML then contains them marked as *functions*, causing interesting breakages in downstream tools.

```html
<?xml version='1.0' encoding='UTF-8' standalone='no'?>
<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="1.12.0" xml:lang="en-US">
  <compounddef id="classRoot_1_1Sub_1_1Class" kind="class" language="C++" prot="public">
    ...
    <sectiondef kind="related">
      ...
      <memberdef kind="function" id="classRoot_1_1Sub_1_1Class_1a7cc214a236ad3bb6ad435bdcf5262a3f" prot="public" static="no" const="no" explicit="no" inline="no" virt="non-virtual">
      <!--             ^^^^^^^^---- note this here -->
        <type>typedef int</type>
        <definition>typedef int Int</definition>
        <argsstring></argsstring>
        <name>Int</name>
        <qualifiedname>Root::Sub::Class::Int</qualifiedname>
        <briefdescription>
<para>A typedef in Sub namespace. </para>
```

I tracked this down to a7179f016c974c38b3d1725dcfcb77924da8b182 (which was fixing #11010), which got subsequently patched in 70aff92f9ecdfced4ea745c94eb98d22a90543bd. The former commit had an issue where it didn't consider actual scopes at all, the latter then partially patched this to restore the original functionality for typedefs defined in classes, but typedefs defined outside of classes got still merged even if they had a different namespace.

In this PR I partially reverted the second commit, and replaced it with a check that both candidates appear in the same scope (the same class, the same namespace, or neither, i.e. a global symbol). That made the above snippet work again. (I hope I did that correctly, let me know if not.)

That wasn't enough however, as the snippet below still produces a typedef marked as a function. The `@relatedalso` is again important here.

```cpp
using Float = float;

/** @brief A struct */
struct Struct {};

/** @relatedalso Struct
 * @brief A using declaration
 */
using Float = float;
```

The reason, in my understanding, is that if the duplicated typedefs get merged, the original second duplicate is left in the entry tree, and due to `@relatedalso` ends up in `filterMemberDocumentation()` later, which implicitly treats it as a function. To fix that, I marked it as processed, which is what `addVariable()` itself did before the two above-referenced commits, restoring the original behavior. In other words, I didn't see a reason to pass it further to another filter, given that the merge happened once already.

I added an extra test to verify both the deduplication behavior and that deduplication takes the scope correctly into account. I recommend viewing the diff with whitespace ignored, or going comit-by-commit. Let me know if you have any questions -- thanks!

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -3209,71 +3209,57 @@ static void buildTypedefList(const Entry *root)
       scope=root->parent()->name; //stripAnonymousNamespaceScope(root->parent->name);
     }
     ClassDefMutable *cd=getClassMutable(scope);
-    if (cd && scope+"::"==rname.left(scope.length()+2)) // found A::f inside A
+    MemberName *mn = Doxygen::functionNameLinkedMap->find(rname);
+    bool found=false;
+    if (mn) // symbol with the same name already found
     {
-      // strip scope from name
-      rname=rname.right(rname.length()-root->parent()->name.length()-2);
-    }
-    QCString rtype = root->type;
-    rtype.stripPrefix("typedef ");
-    if (!root->parent()->name.isEmpty() && root->parent()->section.isCompound() && cd)
-    {
-      AUTO_TRACE_ADD("typedef '{}' in class '{}'", rname,cd->name());
-      addVariableToClass(root,cd,MemberType::Typedef,rtype,rname,root->args,false,nullptr,
-                         root->protection,Relationship::Member);
-    }
-    else
-    {
-      MemberName *mn = Doxygen::functionNameLinkedMap->find(rname);
-      bool found=false;
-      if (mn) // symbol with the same name already found
+      for (auto &imd : *mn)
       {
-        for (auto &imd : *mn)
+        QCString rtype = root->type;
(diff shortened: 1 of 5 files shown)


doxygen:
@mosra Looks good. Thanks for contributing!

PR #11142, opened 2024-09-16
Incorrect marking of enum values in source browser

albert-github:
When having a simple example like:
```
/// \file

enum class Enum_cls
{
    All_cls
    = 0,
    Functions_cls,
    New_cls
};

enum Enum
{
    All
    = 0,
    Functions,
    New
};
```
we see in the source listing (when `EXTRACT_ALL=YES`) that the lines 3, 6, 7, 9, 11, 14, 15 and 17 are (grey) marked instead of the lines 3, 6, 7, 8, 11, 13, 15 and 16. This has been corrected.

Example: [example.tar.gz](https://github.com/user-attachments/files/17012498/example.tar.gz)

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -2732,7 +2732,14 @@ static MemberDef *addVariableToFile(
   //md->setOuterScope(fd);
   if (!root->explicitExternal)
   {
-    mmd->setBodySegment(root->startLine,root->bodyLine,root->endBodyLine);
+    if (mmd->isEnumValue())
+    {
+      mmd->setBodySegment(root->bodyLine,root->bodyLine,root->endBodyLine);
+    }
+    else
+    {
+      mmd->setBodySegment(root->startLine,root->bodyLine,root->endBodyLine);
+    }
     mmd->setBodyDef(fd);
   }
   addMemberToGroups(root,md.get());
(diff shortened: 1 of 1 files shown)


src/doxygen.cpp
```diff
@@ -2732,7 +2732,14 @@ static MemberDef *addVariableToFile(
   //md->setOuterScope(fd);
   if (!root->explicitExternal)
   {
-    mmd->setBodySegment(root->startLine,root->bodyLine,root->endBodyLine);
+    if (mmd->isEnumValue())
+    {
+      mmd->setBodySegment(root->bodyLine,root->bodyLine,root->endBodyLine);
```
doxygen, line 2737: I think it is better to make sure `startLine` is set correctly also for enum values, rather than making an exception here.

albert-github: I do agree, but was (initially) not able to find a better solution.

Did some further tests and also the following seems to work:
```
diff --git a/src/scanner.l b/src/scanner.l
index 35850404e6..2fcdf96084 100644
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -4060,8 +4060,16 @@ NONLopt [^\n]*
                                           if (!yyextra->current->name.isEmpty())
                                           {
                                             yyextra->current->fileName   = yyextra->fileName;
-                                            yyextra->current->startLine  = yyextra->yyLineNr;
-                                            yyextra->current->startColumn = yyextra->yyColNr;
+                                            if (yyextra->current_root->section.isEnum() || yyextra->current_root->spec.isEnum())
+                                              yyextra->current->startLine  = yyextra->current->bodyLine;
+                                              yyextra->current->startColumn = yyextra->current->bodyColumn;
+                                            else
+                                              yyextra->current->startLine  = yyextra->yyLineNr;
+                                              yyextra->current->startColumn = yyextra->yyColNr;
                                             if (!yyextra->current_root->spec.isEnum())
                                             {
                                               yyextra->current->type       = "@"; // enum marker
```

though this relies on the fact that somewhere a `,` is added at the end of the text for a second pass (`parseCompound`) of the scanner, though I don't see where this `,` is added. This makes me a bit reluctant to create a new PR for it.

albert-github: Found the adding of `,`:
```
yyextra->current->program << ','; // add field terminator
```
this is done twice for enum values.

I've just pushed a new proposed patch, pull request #11145 that can supersede this one.

albert-github:
Obsolete due to #11145

PR #11143, opened 2024-09-16
More flexibility showing undocumented namespaces

albert-github:
Contrary to classes (`HIDE_UNDOC_CLASSES`) there is for namespaces no possibility to flexible show / hide them (only by means of documenting the namespace or set `EXTRACT_ALL=YES`). With `HIDE_UNDOC_NAMESPACES` this possibility is created and the default behavior remains as it was (i.e. `HIDE_UNDOC_NAMESPACES=YES`).

Example: [example.tar.gz](https://github.com/user-attachments/files/17013424/example.tar.gz)

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -997,6 +997,17 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
+    <option type='bool' id='HIDE_UNDOC_NAMESPACES' defval='1'>
+      <docs>
+<![CDATA[
+ If the \c HIDE_UNDOC_NAMESPACES tag is set to \c YES, Doxygen will hide all
+ undocumented namespaces that are normally visible in the namespace hierarchy.
+ If set to \c NO, these namespaces will be included in the
+ various overviews.
+ This option has no effect if \ref cfg_extract_all "EXTRACT_ALL" is enabled.
+]]>
+      </docs>
+    </option>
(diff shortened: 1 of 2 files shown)


src/config.xml
```diff
@@ -994,6 +994,17 @@ Go to the <a href="commands.html">next</a> section or return to the
  various overviews. This option will also hide undocumented C++ concepts
  if enabled.
  This option has no effect if \ref cfg_extract_all "EXTRACT_ALL" is enabled.
+]]>
+      </docs>
+    </option>
+    <option type='bool' id='HIDE_UNDOC_NAMESPACES' defval='1'>
```
doxygen, line 1000: I think defval should be 0 for backward compatible behavior.

albert-github: I don't think so.
The key line here is:
```
(hasDocumentation() || !hideUndoc || getLanguage()==SrcLangExt::CSharp)
```
where the `|| !hideUndoc` has been added and has the value `bool hideUndoc = Config_getBool(HIDE_UNDOC_NAMESPACES);`.

So in the default value case of `1` (`YES`) the `!hideUndoc`  is `false` and does not alter the value of the `||` construct.

doxygen: Indeed, I was confused by the default of `HIDE_UNDOC_CLASSES` being `1` and the double/tripple negation logic used. i.e. `!hideUndoc`, which in my head I incorrectly interpreted as`showDoc` instead of `showUndoc`.

Issue #11147, opened 2024-09-19
Functions are incorrectly flagged as undocumented if not generating HTML or LaTeX


SRGDamia1:
**Describe the bug**
A simple function with complete documentation is flagged as not having a documented parameter or return type if HTML or LaTeX is not generated.

**To Reproduce**

Attached: 
[DoxyOutputTest.zip](https://github.com/user-attachments/files/17066956/DoxyOutputTest.zip)

Create a simple header file with a single documented function:
```cpp
/**
 * @file text.h
 */

/**
 * @brief a test function
 *
 * @param input_param an input param
 * @return The return value
 */
uint8_t aTestFxn(uint8_t input_param);

```

enable all warnings:
```
WARNINGS               = YES
WARN_IF_UNDOCUMENTED   = YES
WARN_IF_DOC_ERROR      = YES
WARN_IF_INCOMPLETE_DOC = YES
WARN_NO_PARAMDOC       = YES
WARN_IF_UNDOC_ENUM_VAL = YES
WARN_LOGFILE           = warnings.txt
WARN_AS_ERROR          = FAIL_ON_WARNINGS_PRINT
```

enable XML output and **NOT** HTML or LaTeX:
```
GENERATE_HTML          = NO
GENERATE_LATEX         = NO
GENERATE_XML           = YES
```

get a warning for undocumented parameters and returns:
```log
...text.h:8: warning: parameters of member aTestFxn are not documented
...text.h:8: warning: return type of member aTestFxn is not documented
```

re-enable either the LaTeX *or* the HTML output
```
GENERATE_HTML          = NO
GENERATE_LATEX         = YES
GENERATE_XML           = YES
```

No warnings are produced.

**Expected behavior**
There should not be errors listed for properly documented functions when only XML is generated.

I also tested a function with no documentation, with partial, and with incorrectly formatted documentation, and those same to give me the expected errors regardless of which output I chose. (See the attached zip.)

**Version**

1.12.0 (c73f5d30f9e8b1df5ba15a1d064ff2067cbb8267) on Windows 11

albert-github:
It looks like the problem was introduced in doxygen 1.9.4, more precise:
2c76059d714ea6f2a179578966021df9e365a2a6 is the first bad commit
```
commit 2c76059d714ea6f2a179578966021df9e365a2a6 (HEAD)

Date:   Wed Apr 20 19:52:43 2022 +0200

    Minor optimisation

 src/docparser_p.h  |  4 ++++
[...]

PR #11148, opened 2024-09-20
issue #11147 Functions are incorrectly flagged as undocumented if not generating HTML or LaTeX

albert-github:
Regression on:
```
commit 2c76059d714ea6f2a179578966021df9e365a2a6 (HEAD)

Date:   Wed Apr 20 19:52:43 2022 +0200

    Minor optimisation
```

The fact that XML output is a bit an odd one out was not considered (although mentioned in the comment:  "when only XML format there should be warnings as well (XML has its own write routines)").

### DIFF
--- a/src/outputlist.cpp
+++ b/src/outputlist.cpp
@@ -176,7 +176,7 @@ void OutputList::generateDoc(const QCString &fileName,int startLine,
 
   auto count=std::count_if(m_outputGenList.begin(),m_outputGenList.end(),
                            [](const auto &e) { return e.enabled; });
-  if (count>0)
+  if (count>0 || Config_getBool(GENERATE_XML))
   {
     // we want to validate irrespective of the number of output formats
     // specified as:
@@ -204,7 +204,7 @@ void OutputList::parseText(const QCString &textStr)
   auto count=std::count_if(m_outputGenList.begin(),m_outputGenList.end(),
                            [](const auto &e) { return e.enabled; });
 
-  if (count>0)
+  if (count>0 || Config_getBool(GENERATE_XML))
   {
     // we want to validate irrespective of the number of output formats
     // specified as:


albert-github:
Note:
- We also see the comment: " no formats there should be warnings as well" this would imply that the `if` condition (and thus also the `std::count_if` is not necessary.

So when not selecting any output format we ge besides the correct warning:
```
warning: No output formats selected! Set at least one of the main GENERATE_* options to YES.
```

also some, incorrect and misleading, warnings like:
```
text.h:8: warning: parameters of member aTestFxn are not documented
text.h:8: warning: return type of member aTestFxn is not documented
text.h:33: warning: parameters of member badParamFxn are not documented
text.h:33: warning: return type of member badParamFxn is not documented
text.h:26: warning: parameters of member undocParamFxn are not documented
text.h:26: warning: return type of member undocParamFxn is not documented
text.h:19: warning: parameters of member undocReturnFxn are not documented
text.h:19: warning: return type of member undocReturnFxn is not documented
```

instead of:
```
text.h:33: warning: argument 'some_other_param' of command @param is not found in the argument list of badParamFxn(uint8_t input_param)
text.h:33: warning: The following parameter of badParamFxn(uint8_t input_param) is not documented:
  parameter 'input_param'
text.h:26: warning: parameters of member undocParamFxn are not documented
text.h:19: warning: return type of member undocReturnFxn is not documented
```

@doxygen should we correct this here as well?

doxygen:
@albert-github I think the "optimization" done with https://github.com/doxygen/doxygen/commit/2c76059d714ea6f2a179578966021df9e365a2a6 is actually wrong (for the reasons stated in the comment). I think we should revert this and do the validation always and the `writeDoc()` conditionally again. The same holds for the `parseText()` function.

albert-github:
Obsolete by #11151

PR #11157, opened 2024-09-25
Warning in case of double specification of cite name

albert-github:
Give a warning in case a cite name is double used in bib files (LaTeX silently ignores subsequent entries) as this is troublesome when by accident the same name is used but to a different reference.

Example: [example.tar.gz](https://github.com/user-attachments/files/17127281/example.tar.gz)
Note the differences in the  title and url.
This now gives the warning:
```
tst.bib:11: warning: multiple use of citation name 'cgal:ol-kinetic-24b', (first occurrence: tst.bib, line 1)
```
(vite names are case insensitive).

### DIFF
--- a/src/cite.cpp
+++ b/src/cite.cpp
@@ -27,6 +27,7 @@
 #include "growbuf.h"
 #include "entry.h"
 #include "commentscan.h"
+#include "linkedmap.h"
 
 #include <map>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github Looks like the wrong example is listed for this pull request. I've made some small tweaks (using `LinkedMap` is a bit overkill, so I replaced it with `std::unordered_map`), so please check if the fix still works.

albert-github:
Oops, indeed wrong example. Correct example: [example.tar.gz](https://github.com/user-attachments/files/17178469/example.tar.gz)

Tests revealed no problems

PR #11159, opened 2024-09-27
Update available emojis

albert-github:
Updating the available emoji based on the status on GitHub .
- updating definitions
- updating retrieval procedure (some images were not present and stopped the  retrieval procedure)
- updating tests,as apparently,some emoji got on GitHub a new definition.


albert-github:
Maybe the zip file shoudl also be updated 
> For convenience a zip with the result of running the script can also be downloaded from
https://www.doxygen.nl/dl/github_emojis.zip

Issue #11172, opened 2024-10-04
Aliases with embed:rst:leading-asterisk no longer work in Doxygen 1.12


TruongMinh89:
**Describe the bug**
[example3.zip](https://github.com/user-attachments/files/17260665/example3.zip)

In Doxygen 1.9.7, I could use the following aliases to embed reStructuredText (RST) content with leading asterisks removed:
```
ALIASES += "rst=\verbatim embed:rst:leading-asterisk"
ALIASES += "endrst=\endverbatim"
```

However, in Doxygen 1.12, this no longer works as expected. The issue occurs mainly when aliases are used within this RST block. The content is not rendered correctly, and any alias inside the rst block is ignored. 

**Screenshots**
![image](https://github.com/user-attachments/assets/35302583-30fd-4684-822e-5fca4b45878a)

**To Reproduce**
Use
```
ALIASES += "rst=\verbatim embed:rst:leading-asterisk"
ALIASES += "endrst=\endverbatim"
```
and use any aliases within.

**Expected behavior**
Aliases within rst block should be rendered correctly

**Version**
Doxygen version used: 1.12.0

**Stack trace**
N/A

**Additional context**
N/A

albert-github:
In don't see any example with which we could try to reproduce the problem, so:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem? Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

TruongMinh89:
I have added the example.zip:
- Doxyfile: Contain relevant ALIASES only. I hope that you can add it to general Doxyfile to test that.
- An example .c file that used those ALIASES

TruongMinh89:
Sorry, have to update the example several times.

albert-github:
I looked at the example from https://github.com/doxygen/doxygen/issues/11172#issuecomment-2393930233 
When running this example with doxygen 1.9.7 I get:
```
error: Illegal ALIASES format 'csv_table{3::}=.. csv-table:: \1^^\3   :header: \2^^'. Use "name=value" or "name{n}=value", where n is the number of arguments
```
which is logical as the "Allow custom separator for ALIASES" were introduced in doxygen 1.9.8

- So this is not the right definition for the 1.9.7 version.

When running the current doxygen 1.12.0 version and using the debug option `commentcnv` (so `doxygen -d commentcnv`) we see:
```
...
/**
[...]

PR #11175, opened 2024-10-09
issue #11172 Aliases with embed:rst:leading-asterisk no longer work in Doxygen 1.12

albert-github:
Don't remove "common initial space" in the replacement of `^^`.


src/aliases.cpp
```diff
@@ -159,7 +159,7 @@ static void addValidAliasToMap(std::string_view alias)
 
 static std::string escapeAlias(std::string_view value)
 {
-  std::string newValue = substituteStringView(value,"^^","\\ilinebr ");
+  std::string newValue = substituteStringView(value,"^^","@ilinebr ");
```
TruongMinh89, line 162: Why we have additional space at the end of "@ilinebr "?

albert-github: - symmetry
- otherwise the command `ilinebr` cannot be correctly determined e.g. when a "normal" word is following

TruongMinh89: It create 4 spaces instead of 3 spaces in my example.

doxygen: maybe it is better to replace `^^ ` (with space) by `@ilinebr ` (with space) and then `^^` (without space) by `@ilinebr ` (with space), so we don't increase the amount of existing whitespace.

albert-github: @doxygen 
Currently the replacement of `^^` is `\ilinebr ` so with space and, as far as I remember, this was done as otherwise the command ilinebr cannot be correctly determined e.g. when a "normal" word is following.

doxygen: I understand. With my suggestion `^^nospace` would still be replaced by `@ilinebr nospace` (adding an extra space) and `^^ onespace` would be replaced by `@linebr onespace` (so no extra space).

albert-github: @doxygen I missed the subtle difference of when already a space is present after the `^^` no extra space is added otherwise one is is added. Might be a good idea (@TruongMinh89 would also mean that your workaround, 2 versus 3 spaces, is not necessary anymore).

albert-github: I've just updated this proposed pull request

TruongMinh89 (approved):

albert-github:
@sloriot 
- Which project?
- which xml generated?
So:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem? Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

albert-github:
@sloriot 

I assume you are talking about the overnight CGAL documentation run.
- What happens when you rerun the `CGAL-6.0.1-I-347` so without extra CGAL pull requests?

doxygen:
@sloriot Please check if with [this commit](https://github.com/doxygen/doxygen/commit/e3a54a27a75cb129188c2305b468b79149bc5236) the problem is fixed.

albert-github:
@doxygen 
I did some tests on my local copy of CGAL and this looks OK.  Also the overnight CGAL documentation run was OK but I'm not sure whether here a workaround was implemented for the problem(@sloriot ?).

I did see a not yet mentioned problem in HTML and one in XML which I have to investigate (but it doesn't look like they are related to the solved problem).

Issue #11184, opened 2024-10-17
\ref is unable to resolve references to fields in the documented struct


arwedus:
**Describe the bug**
Describe what you see that (you think) is wrong.

The following code:

```cpp
/// \brief Date and Time Record.
/// Represent a specific point in time.
/// \invariant \ref date and \ref day_of_the_week must not be set together.
struct DateTimeRecord
{
    /// \brief Minutes after start of the day. Valid range: 0 (00:00) to 1439 (23:59) (inclusive)
    std::optional<Int32> minutes_after_midnight;
    /// \brief This value is only set if \ref day_of_the_week has no value.
    std::optional<DateRecord> date;
    /// \brief This value is only set if \ref date has no value.
    std::optional<WeekDay> day_of_the_week;
};
```

results in the following doxygen warnings:

```
Generating Sqlite3 output for class example::sit::DateRecord
Generating Sqlite3 output for class example::sit::DateTimeRecord
/workspace/build/ref_example/example.h:55: warning: unable to resolve reference to 'date' for \ref command
/workspace/build/ref_example/example.h:55: warning: unable to resolve reference to 'day_of_the_week' for \ref command
Generating Sqlite3 output for class example::sit::TimeRestrictionRecord
```

NOTE: This warning is only issued when writing the Sqlite3 output.
When you set `GENERATE_SQLITE3 = NO` (default), the warning is not shown by doxygen build.
The references get correctly resolved in the HTML output, no matter how GENERATE_SQLITE3 is set.

NOTE: The same is true for the \copybrief or \copydetails commands:

```
Generating Sqlite3 output for class example::sit::DateTimeRecord
/workspace/build/ref_example/example.h:54: warning: @copydetails or @copydoc target 'minutes_after_midnight' not found
/workspace/build/ref_example/example.h:55: warning: unable to resolve reference to 'date' for \ref command
/workspace/build/ref_example/example.h:55: warning: unable to resolve reference to 'day_of_the_week' for \ref command
```

**To Reproduce**

See attached example:
[ref_example.zip](https://github.com/user-attachments/files/17422781/ref_example.zip)

**Expected behavior**
Describe what you would have expected or think is correct.

The warning should not be issued when `GENERATE_SQLITE3 = YES`

**Version**

doxygen: I've tested 1.10 and 1.12 (built from source)
OS: Ubuntu 20.04

albert-github:
I've just pushed a proposed patch, pull request #11185

[...]

PR #11185, opened 2024-10-18
issue #11184 \ref is unable to resolve references to fields in the documented struct

albert-github:
problem looks like to be that the mentioned fields (`date` and `day_of_the_week`) are evaluated against the outer class of the current class (so against `example::sit` instead  of `example::sit::`). Similar problems occur with the namespace (and probably also with modules and concepts) so this also has been fixed.

### DIFF
--- a/src/sqlite3gen.cpp
+++ b/src/sqlite3gen.cpp
@@ -1460,6 +1460,21 @@ static void getSQLDesc(SqlStmt &s,const char *col,const QCString &value,const De
     )
   );
 }
+
+static void getSQLDescCompound(SqlStmt &s,const char *col,const QCString &value,const Definition *def)
+{
+  bindTextParameter(
+    s,
+    col,
+    getSQLDocBlock(
+      def,
+      def,
+      value,
+      def->docFile(),
+      def->docLine()
+    )
+  );
+}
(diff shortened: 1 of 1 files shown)


albert-github:
> Do you want me to try it out on my original project?

Yes please as I've doing it a bit "in the dark".

There is a testsqlite3.py and it looks like I once tried it, but it is not integrated (as far as I know) in the test procedure, so a bit hard to create a regression test.

arwedus:
Quick update: I've run the test on the original project from which I've extracted that small example. The warnings are not issued any more.

I still get warnings of type "unable to resolve reference to * for \ref command" and "@copy* or @copydoc target * not found" in other projects, which only appear when setting GENERATE_SQLITE3=YES. They may be unreleated errors. I will provide an example with one of those in a couple days.

albert-github:
> They may be unreleated errors. I will provide an example with one of those in a couple days.

Might be that they have to do with groups / pages / .... or other compounds of which no example was present.

Yes please add an example to the original issue.

albert-github:
@arwedus  Any progress on the problems you mentioned in https://github.com/doxygen/doxygen/pull/11185#issuecomment-2422898187

PR #11187, opened 2024-10-18
Handling of `#pragma once`

albert-github:
The handling of the `#pragma once` was different from the handling of a file with the `#ifndef` guard construction, making, especially when nested include files were in e.g. a namespace, very slow.

Found by Fossies for the mysql-workbench-community 8.0.40 package.

Example: [example.tar.gz](https://github.com/user-attachments/files/17438388/example.tar.gz)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -306,6 +306,7 @@ struct preYY_state
   DefineMap                                localDefines;   // macros defined in this file
   DefineList                               macroDefinitions;
   LinkedMap<PreIncludeInfo>                includeRelations;
+  StringVector                             pragmaList;
 
   int                lastContext = 0;
   bool               lexRulesPart = false;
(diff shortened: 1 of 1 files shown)


src/pre.l
```diff
@@ -1063,6 +1065,21 @@ WSopt [ \t\r]*
                                         }
 <Command>"pragma"{B}+"once"             {
                                           yyextra->expectGuard = FALSE;
+                                          if (std::find(yyextra->pragmaList.begin(), yyextra->pragmaList.end(), yyextra->fileName.data()) != yyextra->pragmaList.end())
```
doxygen, line 1068: Using `StringUnorderedSet` is much more efficient for simple "is present" type of checks (O(1) v.s. O(N), where N is the list size)

PR #11224, opened 2024-11-08
Add support to ignore certain words when AUTOLINK_SUPPORT is enabled

diegodiasm:


diegodiasm:
The motivation for this to avoid misleading links the the [L4Re documentation](https://l4re.org/doc/). Whenever the terms L4 and L4Re are written in a page, they are automatically linked to the documentation of these namespaces, sometimes in a misleading way, like [here](https://l4re.org/doc/group__api__l4re__parent.html).

The usual approach of adding % to a word is not sufficient, because if a class is added with a name used in several places, then one has to revisit several files to add the % symbol to prevent autolinks from being generated.

This approach allows the problem to be circumvented by adding the following to the project's configuration file:


```
AUTOLINK_IGNORED_TOKENS = "L4" \
                          "L4Re"
```

That is more convenient and does not prevent intential links to be added with the `\link .. \endlink` command, or the `#` operator.

albert-github:
- Are `L4` and  `L4Re` classes / namespaces? (looks like they are namespaces)
- what happens in the source code with the links to e.g. `L4` and `L4Re`  are they linked ? (for the class / namespace name)

diegodiasm:
Hi Albert,

Both `L4` and `L4Re` are references to namespaces.

The change causes autolinks to be disabled for these two words. For example, consider the following documentation snippet:

```
/**
 * \defgroup api_l4re_parent Parent API
 * \ingroup api_l4re
 * \brief Parent interface.
 *
 * The parent interface provides means for an L4 task to signal changes in its
 * execution state. The main purpose is to signal program termination to the
 * program that started it, so that its resources can be reclaimed. In a typical
 * L4Re system, this program will be Moe or Ned.
 *
 * - Reference to L4 or L4Re does not trigger autolink to these namespaces
 * - Links to #L4 and #L4Re namespaces can be achieved using the hash syntax (#)
 * - Links to \link L4 \endlink and \link L4Re \endlink namespaces can be achieved
 *   using the link..endlink command.
 * - Occurrences already marked using the percentage (%) syntax remain rendered
 *   without the percentage symbol: %L4 and %L4Re
 *
 * \see L4Re::Parent for information about the concrete interface.
 */
```

without the proposed patch, it will render in the following way:

image

and with the proposed patch, it renders in the following way:

image

diegodiasm:
Hi Albert, do you need any additional explanation from my side? Is it possible that this pull request is accepted to the `doxygen:master` branch?

Thanks

doxygen:
@diegodiasm I made some changes, see https://github.com/doxygen/doxygen/commit/f4a9fe3904bc7c8e32969bdcaef6209f6c079549. Should not impact the functionality, let me know otherwise. 
The name of the option has changed to `AUTOLINK_IGNORE_WORDS`.

Issue #11225, opened 2024-11-10
Improve support for code blocks in included markdown files


AdarshRawat1:
## Describe the bug
When a markdown is included in other markdown using `\include{doc} "./Secondary.md"` Then, the output of code snippets breaks. 

## Screenshots

- ![image](https://github.com/user-attachments/assets/cbdea7e1-c561-4e5f-ae2e-70eb283d6775)

- ![image](https://github.com/user-attachments/assets/b9bbb603-3794-40a9-b793-f69f2061fd60)

## Observations 
This happens only if there is formatting in the primary markdown - like ( **BOLD** , _Italic_ ). **if we remove the formattings from primary markdown then everything works fine**, but we can't avoid using formatting, so I tried few workarounds one of them worked. 

i.e. There is a need to handle the formatting edge cases.

## Temporary Fix
- This version gives correct output but `\code{.c}` doesn't work on GitHub Renders, so I have to keep 2 copies of the same code in docs. (One for Doxygen and one for GitHub )
``` 
\code{.c}
    struct secondary_struct {
        int a ; 
        int b ; 
    }
\endcode

```

p.s. I've been on this for quite some time 

## To Reproduce
Use this Zip -  [includedocs.zip](https://github.com/user-attachments/files/17690916/includedocs.zip)

## Expected behavior
Correctly formatted code snippets

## Version
1.12.0 (Windows 11, WSL, Ubuntu) 

## Additional context
- [Related Stack overflow issue ](https://stackoverflow.com/questions/52120027/doxygen-how-to-include-a-markdown-page-to-document-a-group)

- https://github.com/doxygen/doxygen/issues/7430

albert-github:
I've just pushed a proposed patch, pull request #11226

albert-github:
The problem with the undesired `#!/bin/bash` in the code section cannot be solved directly by doxygen as shell scripts are not supported by doxygen.

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

AdarshRawat1:
> The problem with the undesired `#!/bin/bash` in the code section cannot be solved directly by doxygen as shell scripts are not supported by doxygen.

When the bash script is in a normal markdown( primary markdown) then it works, but breaks only for included codes i.e. `\include{doc} "./Secondary.md"`

albert-github:
@AdarshRawat1 

Can you please give an example so we are sure we use the same example / see the same problem:
[...]

PR #11226, opened 2024-11-10
issue #11225 Improve support for code blocks in included markdown files

albert-github:
Looks like the `**` in the line:
```
This is sample **content** and code for Primary file
```
are mistakenly seen as continuation comment signs of a Cpp  line also due to the initial setting of the lex state to `CComment`


albert-github:
Fix doesn't look like to be sufficient, an analogous `.h` file has the same problem.
Looking into it ...

doxygen:
@albert-github Please check the change I made in https://github.com/doxygen/doxygen/commit/723ea541fd298152fef64e0aa464ba0f8c35faee Maybe you had a reason not to use `computeIndent()`?

Issue #11240, opened 2024-11-18
[Small Feature Request] Add CSS class for `@p` command


parmi93:
Currently the `@p` and `@c` commands when rendered in HTML both generate the same output (`<code></code>`), it would be nice if a CSS class was added by default to distinguish the two commands.

Like:
- `@p` --> `<code class="param"></code>`
- `@c` --> `<code class="code"></code>`

This way the User would have the possibility to define his own CSS classes (`.param{}` and `.code{}`), to be able to visually distinguish more easily when referring to parameters and when referring to simple code.

Related issue: https://github.com/jothepro/doxygen-awesome-css/issues/162

albert-github:
I've just pushed a proposed patch, pull request #11243

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.13.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

parmi93:
I'm not sure if it's a problem related to the introduction of the `param` class.

```c
/**
 * @p key/@p value pair was added to the dictionary successfully.
 */
void my_func(int key, int value);
```
Result:
![image](https://github.com/user-attachments/assets/40f367e9-6690-4ab2-a9c4-19623ff455a6)

I was expecting this:
![image](https://github.com/user-attachments/assets/23ca7b6f-7f0b-4a24-872d-0d3df7573ee6)

albert-github:
No this does not have to do with changes for this issue, I can see it also in e.g. doxygen version1.12.0, so please file a new issue for this.
(also a line like `@b key/@p value pair was added to the dictionary successfully.` will give some unexpected results).

PR #11243, opened 2024-11-19
issue #11240 Add CSS class for `@p` command

albert-github:
Added general class attribute to style change commands. The value of the class attribute is `doxygen_` followed by the doxygen command or tag name.

### DIFF
--- a/src/htmldocvisitor.cpp
+++ b/src/htmldocvisitor.cpp
@@ -482,46 +482,55 @@ void HtmlDocVisitor::operator()(const DocHorRuler &hr)
 void HtmlDocVisitor::operator()(const DocStyleChange &s)
 {
   if (m_hide) return;
+  HtmlAttribList extraAttribs;
+  if (s.enable())
+  {
+    HtmlAttrib opt;
+    opt.name  = "class";
+    opt.value = "doxygen_" + s.tagName();
+    extraAttribs.push_back(opt);
+    mergeHtmlAttributes(s.attribs(),extraAttribs);
+  }
   switch (s.style())
   {
     case DocStyleChange::Bold:
-      if (s.enable()) m_t << "<b" << htmlAttribsToString(s.attribs()) << ">";      else m_t << "</b>";
+      if (s.enable()) m_t << "<b"  << htmlAttribsToString(extraAttribs) << ">";      else m_t << "</b>";
       break;
     case DocStyleChange::S:
-      if (s.enable()) m_t << "<s" << htmlAttribsToString(s.attribs()) << ">";      else m_t << "</s>";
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github I think it does not add much to add the class attribute to all HTML tags, and it adds a lot of characters to the output, so I limited it to `\p` and `\a` now, which could be attributed a special meaning beyond just formatting. It still fulfills the original request, I think.

doxygen:
Hi Albert, I see it differently. We have commands that express the 'look' of the text, like `\c`, `\e`,  and `\b`, and we have commands that express the 'type' of the word that follows it, such as `\p`, and `\a`. 

I agree, there need to be a way to style the 'type' commands differently from the 'look' commands, but I don't see the need to give everything a tag, i.e. have `<span class="doxygen_span>...</span>` because that can already be styled using e.g. `div.contents span` inside the CSS.

albert-github:
Indeed not everything needs a tag, so I probably overdid a little bit.
I think that when having 2 or more commands that result in the same stylechange (or command?, added this just for completeness / my small sorting - unique script gave them) at least 1 or more of them should have an extra attribute.

For completeness the overview of multiple mappings  in cmdmapper.cpp:
```
  { "c",             CommandType::CMD_CODE },
  { "p",             CommandType::CMD_CODE },

  { "javalink",      CommandType::CMD_JAVALINK },
  { "javalinkplain", CommandType::CMD_JAVALINK },

  { "arg",           CommandType::CMD_LI },
  { "li",            CommandType::CMD_LI },

  { "remark",        CommandType::CMD_REMARK },
  { "remarks",       CommandType::CMD_REMARK },

  { "sa",            CommandType::CMD_SA },
  { "see",           CommandType::CMD_SA },

  { "strong",     HtmlTagType::HTML_BOLD },
  { "b",          HtmlTagType::HTML_BOLD },

  { "code",       HtmlTagType::HTML_CODE },
  { "dfn",        HtmlTagType::HTML_CODE },

  { "tt",         HtmlTagType::XML_C },
  { "c",            HtmlTagType::XML_C },

  { "a",             CommandType::CMD_EMPHASIS },
  { "e",             CommandType::CMD_EMPHASIS },
  { "em",            CommandType::CMD_EMPHASIS },

  { "exception",     CommandType::CMD_EXCEPTION },
  { "throw",         CommandType::CMD_EXCEPTION },
  { "throws",        CommandType::CMD_EXCEPTION },

  { "result",        CommandType::CMD_RETURN },
  { "return",        CommandType::CMD_RETURN },
  { "returns",       CommandType::CMD_RETURN },

  { "var",        HtmlTagType::HTML_EMPHASIS },
  { "em",         HtmlTagType::HTML_EMPHASIS },
  { "i",          HtmlTagType::HTML_EMPHASIS },
```

Issue #11245, opened 2024-11-19
[Small Feature Request] Add `class` attribute to the `@qualifier` command in the HTML output


parmi93:
I am trying to achieve something like this:

![image](https://github.com/user-attachments/assets/890d940e-a664-44fb-867f-a927e907a872)

where the `@qualifier Thread-safe` has a different `background-color` than `inline`.
Unfortunately this is not possible because all qualifiers are rendered as `<span class="mlabel">...</span>`

I would like to purpose to change this command into something like that:

`\qualifier <name> | "(text)" [<label>]` --> `<span class="mlabel label">name | "(text)"</span>`

where `<name> | "(text)"` is displayed to the user, and `<lablel>` is used as the value for the `class` attribute.

albert-github:
The current syntax is
```
\qualifier <label> | "(text)"
```
the proposed syntax is nice but I'm thinking more in the direction of:
```
\qualifier['{'option[,option]∗'}'] <label> | "(text)"
```
where `option` can be a class label.

albert-github:
Nearly, this is indeed how you would use it but without the signal quotes (the `'{'` sifbals that the '{' is mandatory when using options:
```
\qualifier Thread-safe --> <span class="mlabel">Thread-safe</span>
\qualifier{my_class1,my_class2} Thread-safe --> <span class="mlabel my_class1 my_class2">Thread-safe</span>
```

parmi93:
Thanks for the clarification, I think this new feature might be fine =)

parmi93:
Maybe I'm asking a bit too much now =)
Would it be possible to add support for the `@ref` tag when used inside the `<lablel> | "text"` of the `@qualifier` command?

What I'm trying to achieve is to create a clickable qualifier, that takes the user to the HTML page where the meaning of Thread-safe is explained, I tried to use the `'@qualifer` command like this, but it didn't work =(

`@qualifier "@ref thread_safe \"Thread-safe\""`

albert-github:
`\ref` Interesting thought but a bit out of scope for this issue, just create another issue for it.

albert-github:
I've just pushed a proposed patch, pull request #11250

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.13.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

Issue #11247, opened 2024-11-21
[Feature Request] Add support for `@ref` command inside `@qualifier` command


parmi93:
Would it be possible to add support for the `@ref` command when used inside the <lablel> | "text" of the @qualifier command?

What I'm trying to achieve is to create a clickable qualifier, that takes the user to the HTML page where the meaning of Thread-safe is explained, I tried to use the `@qualifer` command like this, but it didn't work =(

@qualifier "@ref thread_safe \"Thread-safe\""

Issue #4854, opened 2018-07-02
Formulas in member groups ignored (Origin: bugzilla #682258)


doxygen:
status NEW severity *normal* in component *general* for *---*
Reported in version *1.8.2-SVN* on platform *Other*
Assigned to: Dimitri van Heesch

On 2012-08-20 13:05:39 +0000, Philipp Moeller wrote:

> When using a formula inside the name of a member group it is not rendered as math but printed verbatim including the delimiters into the HTML output.
> 
> This class demonstrates the problem:
> 
> /*!
>   My X.
> */
> class X
> {
> public:
>   /// \name A \f$alpha\f$ Heading
>   /// @{
> 
>   /// Nope.
>   void what();
>   /// @}
> };
> 
> The member group heading will be rendered as "A \f$alpha\f$ Heading" in the HTML output. This is not influenced by the USE_MATHJAX setting.

On 2012-11-18 11:07:17 +0000, Dimitri van Heesch wrote:

> Changed version 'latest' to '1.8.2-SVN' so I can remove 'latest' as an option as it is a moving target.

PR #11250, opened 2024-11-23
issue #11245 Add class attribute to the `@qualifier` command in the HTML output

albert-github:
Adding the possibility to have (HTML) class attribute values as options with the `@qualifier` command.

### DIFF
--- a/doc/commands.dox
+++ b/doc/commands.dox
@@ -654,13 +654,15 @@ Structural indicators
       option \ref cfg_show_enum_values "SHOW_ENUM_VALUES"
 
 <hr>
-\section cmdqualifier \\qualifier \<label\> | "(text)"
+\section cmdqualifier \\qualifier['{'[option][,option]*'}'] \<label\> | "(text)"
 
   \addindex \\qualifier
 
   With this command it is possible to add custom qualifier labels
   to members and classes. These labels will be shown in the output in the same way
   as the automatically generated labels such as "static", "inline", and "final".
+
+  The `option` is the name to be appended of the HTML class atribute for the qualifier.
 
   For instance to indicate that a function is only meant for testing purposes
   one could add `\qualifier test`
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -302,6 +302,7 @@ add_library(doxymain STATIC
     plantuml.cpp
     qcstring.cpp
     qhp.cpp
+    qualifiers.cpp
     reflist.cpp
     regex.cpp
     resourcemgr.cpp
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -271,7 +271,7 @@ class ClassDefImpl : public DefinitionMixin<ClassDefMutable>
     bool isSliceLocal() const override;
     bool hasNonReferenceSuperClass() const override;
     QCString requiresClause() const override;
-    StringVector getQualifiers() const override;
+    QualifierInfoVector getQualifiers() const override;
     bool containsOverload(const MemberDef *md) const override;
     ClassDef *insertTemplateInstance(const QCString &fileName,int startLine,int startColumn,
                                 const QCString &templSpec,bool &freshInstance) const override;
@@ -297,7 +297,7 @@ class ClassDefImpl : public DefinitionMixin<ClassDefMutable>
     void setCompoundType(CompoundType t) override;
     void setClassName(const QCString &name) override;
     void setClassSpecifier(TypeSpecifier spec) override;
-    void addQualifiers(const StringVector &qualifiers) override;
+    void addQualifiers(const QualifierInfoVector &qualifiers) override;
     void setTemplateArguments(const ArgumentList &al) override;
     void setTemplateBaseClassNames(const TemplateNameMap &templateNames) override;
     void setTemplateMaster(const ClassDef *tm) override;
@@ -586,7 +586,7 @@ class ClassDefAliasImpl : public DefinitionAliasMixin<ClassDef>
     { return getCdAlias()->hasNonReferenceSuperClass(); }
     QCString requiresClause() const override
     { return getCdAlias()->requiresClause(); }
-    StringVector getQualifiers() const override
+    QualifierInfoVector getQualifiers() const override
     { return getCdAlias()->getQualifiers(); }
     bool containsOverload(const MemberDef *md) const override
     { return getCdAlias()->containsOverload(md); }
@@ -800,7 +800,7 @@ class ClassDefImpl::IMPL
     /** C++20 requires clause */
     QCString requiresClause;
 
-    StringVector qualifiers;
+    QualifierInfoVector qualifiers;
 
     bool hasCollaborationGraph = false;
     CLASS_GRAPH_t typeInheritanceGraph = CLASS_GRAPH_t::NO;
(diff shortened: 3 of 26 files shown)


doxygen:
@albert-github I think there is a more local change possible, that doesn't require the user to specify any extra options, and also provides the possibility to style existing labels.

My idea is to derive the class from the label automatically, like so:
```c++
void HtmlGenerator::writeLabel(const QCString &label,bool /*isLast*/)
{
  DBG_HTML(m_t << "<!-- writeLabel(" << label << ") -->\n";)

  auto convertLabelToClass = [](const std::string &lab) -> QCString {
    QCString input = convertUTF8ToLower(lab);
    QCString result;
    size_t l=input.length()+1;
    result.reserve(l);
    result+=' ';
    for (size_t i=0; i<l; i++)
    {
      char c = input.at(i);
      if (isId(c) || c<0) result+=c;
      else if (c==' ' || c=='-') result+='-';
    }
    return result;
  };

  m_t << "<span class=\"mlabel" << convertLabelToClass(label.stripWhiteSpace().str()) << "\">" << label << "</span>";
}
```

This way
```c++
    /** \qualifier Thread-Safe
     *  \qualifier "Some other qualifier"
     *  More docs
     */
    inline void nvs_init() {}
```
will make Doxygen generate the labels like this:
```html
<span class="mlabels">
  <span class="mlabel inline">inline</span>
  <span class="mlabel thread-safe">Thread-Safe</span>
  <span class="mlabel some-other-qualifier">Some other qualifier</span>
</span>
```

Let me know what you think.

Note that turning specific spans into links can be done with a bit of JavaScript.

albert-github:
In principle it doesn't look bad, but it will mean that with each qualifier an extra part will be added to the class attribute (argument with PR 11243).
I think my solution is more flexible (maybe to flexible) as a user can style all qualifiers individually (e.g. when having a qualifier X with C++ it can be made green and with C# it can be made red) also one can use multiple classes added to do some other differentiating.

From the given example I understand that the "default" labels also get automatically a class attribute, that is good (was enlarging my proposed code change a bit).

The `isId` also allows the `$` but as far as I can tell this character is not allowed in a HTML class name (and maybe there are more).

The remark:
> Note that turning specific spans into links can be done with a bit of JavaScript.

what do you mean here I don't get it, the "class" are steered through the css are does the JS give some other / extra flexibility?

Furthermore there is the enhancement request #11247 ([Feature Request] Add support for `@`ref command inside `@qualifier` command) will this interfere with your proposal?

doxygen:
> In principle it doesn't look bad, but it will mean that with each qualifier an extra part will be added to the class attribute (argument with PR 11243). 

In this case it is justified, as all qualifiers have the same span and cannot be distinguished.

> I think my solution is more flexible (maybe to flexible) as a user can style all qualifiers individually (e.g. when having a qualifier X with C++ it can be made green and with C# it can be made red) also one can use multiple classes added to do some other differentiating.

I think it is indeed more flexibility, but at the cost of the user having to specify things explicitly, so a bit overly flexible in my opinion. It is always a trade-off of course.

> From the given example I understand that the "default" labels also get automatically a class attribute, that is good (was enlarging my proposed code change a bit).
> 
> The `isId` also allows the `$` but as far as I can tell this character is not allowed in a HTML class name (and maybe there are more).

Indeed. I found the spec, and can change it to

```c++
void HtmlGenerator::writeLabel(const QCString &label,bool /*isLast*/)
{
  DBG_HTML(m_t << "<!-- writeLabel(" << label << ") -->\n";)

  auto convertLabelToClass = [](const std::string &lab) {
    QCString input = convertUTF8ToLower(lab);
    QCString result;
    size_t l=input.length();
    result.reserve(l);

    // Create valid class selector, see 10.2 here https://www.w3.org/TR/selectors-3/#w3cselgrammar
    // ident     [-]?{nmstart}{nmchar}*
    // nmstart   [_a-z]|{nonascii}
    // nonascii  [^\0-\177]
    // nmchar    [_a-z0-9-]|{nonascii}

    bool nmstart=false;
    for (size_t i=0; i<l; i++)
    {
      char c = input.at(i);
      if (c<0 || (c>='a' && c<='z') || c=='_') // nmstart pattern
      {
        nmstart=true;
        result+=c;
      }
      else if (nmstart && (c<0 || (c>='a' && c<='z') || (c>='0' && c<='9') || c=='_')) // nmchar pattern
      {
        result+=c;
      }
      else if (nmstart && (c==' ' || c=='-')) // show whitespace as -
      {
        result+='-';
      }
    }
    return result;
  };

  m_t << "<span class=\"mlabel " << convertLabelToClass(label.stripWhiteSpace().str()) << "\">" << label << "</span>";
}
```

> 
> The remark:
> 
> > Note that turning specific spans into links can be done with a bit of JavaScript.
> 
> what do you mean here I don't get it, the "class" are steered through the css are does the JS give some other / extra flexibility?
> 
> Furthermore there is the enhancement request #11247 ([Feature Request] Add support for `@`ref command inside `@qualifier` command) will this interfere with your proposal?

I was indeed referring to the enhancement request. You can do this also by embedding a piece of JavaScript in the header, e.g.

```.js
$(document).ready(function() {
    $('span.thread-safe').each(function() {
        var spanContent = $(this).text();
        $(this).replaceWith('<span class="mlabel thread-safe"><a href="https://en.wikipedia.org/wiki/Thread_safety">' + spanContent + '</a></span>');
    });
});
```

albert-github:
- Definition of the class identifier.
The lexical grammar in https://www.w3.org/TR/selectors-3/#w3cselgrammar specifies that it is case insensitive anf this is respected by the call to `QCString input = convertUTF8ToLower(lab);`. A bit strange is that the example in https://www.w3schools.com/html/tryit.asp?filename=tryhtml_classes_capitals is case sensitive (I fiddled a bit with the class names) but I think this is might be a bug here. (I'll try to post in the forum there but awaiting approval)
- C++ / C# different layout of e.g. "inline" label
Would there any possibility to distinguish these?
- `\ref` support
The given JavaScript code will work for a fixed address, but I don't see show this will work when the `\ref` points to an internal doxygen page. Think here about e.g. the difference whether `CREATE_SUBDIRS` is set or not,  `SHORT_NAMES` or  slightly reorganizing the code or did I oversee some possible flexibility in the JS code?

doxygen:
> * Definition of the class identifier.
>       The lexical grammar in https://www.w3.org/TR/selectors-3/#w3cselgrammar specifies that it is case insensitive anf this is respected by the call to `QCString input = convertUTF8ToLower(lab);`. A bit strange is that the example in https://www.w3schools.com/html/tryit.asp?filename=tryhtml_classes_capitals is case sensitive (I fiddled a bit with the class names) but I think this is might be a bug here. (I'll try to post in the forum there but awaiting approval)
> 

Ok, interesting.

>   * C++ / C# different layout of e.g. "inline" label
>     Would there any possibility to distinguish these?

No, but is that really something someone wants to do? Sounds a bit farfetched to me. This particular example could also be handled in an automated way by adding a language identifier as a class specifier.

>   * `\ref` support
>     The given JavaScript code will work for a fixed address, but I don't see show this will work when the `\ref` points to an internal doxygen page. Think here about e.g. the difference whether `CREATE_SUBDIRS` is set or not,  `SHORT_NAMES` or  slightly reorganizing the code or did I oversee some possible flexibility in the JS code?

Good point. This would indeed require it be processed by Doxygen. But if `\ref` can be supported, why not other things like markup commands, images, formulas, or something crazy like a table? Then it is maybe easier/clearer to have a dedicated command like `\qualifierlink item_to_link_to "text"`. But I'm not convinced this is worth the effort.

albert-github:
- Definition of the class identifier.
login has been approved and I created: https://w3schools.invisionzone.com/topic/64106-case-insensitive-of-class-attribute-values/
- C++ / C# different layout of e.g. "inline" label
Might sound far fetched but I think that giving more general possibilities would give the users more possibilities. We don't know what they "invent" / would like to show which special concepts (not to confuse with C++ concepts) they use. 
For a language it might, probably, be reasonably easy to create, but I used the language as an example as this is easy to understand / visualize in one's mind.
- `\ref` support
I can imagine that in the text of the label there can be things like a small formula, this could / should be limited to the possibilities (ore even further) as possible with the section headers.
This possibility not only applies for the `\ref` possibility with the `\qualifier` but is actually a broader request to have the text part  of the `\ref` command be extended with layout possibilities as with section headers.
Actually this also accounts for the `\name` command, the title of a `\defgroup`, the title of a `\page`.
In the recent past I already had a look at the `\name command (Formulas in member groups ignored (Origin: bugzilla #682258), https://github.com/doxygen/doxygen/issues/4854). It is not easy, but well feasible. Hopefully I can work on it in the near future and I will keep for the request #11247 also in mind and hopefully this can automatically be part of it.

Issue #11271, opened 2024-12-13
bug - in man page, verbatim printf("string") becomes printf('string')


duncan-roe:
This has been around since doxygen 1.9.3 - I just never noticed it before.
**git bisect** identifies 6c5b9295509e as culprit. I've made a patch, will send PR soon.
Will send test case soon.

duncan-roe:
Have discovered double-quote becoming single in regular text as well. Investigating further

duncan-roe:
The damage is done in `ManDocVisitor::filter()` defined at line **987** of file **src/mandocvisitor.cpp**. The simplest fix is to delete line 999 `case '"':  c = '\''; // fall through`. You hardly need a PR for that.
But it seemed to me there must be a reason for having line 999. So originally, when I thought it was only a problem in verbatim code, I only changed line 230 `filter(s.text());` to be `manfilter(s.text());` where `manfilter()` was a clone of `filter()` with line 999 removed.
Then when I found wrong quoting was more widespread, I added a `bool fixquotes` argument to `filter()` and guessed that the `operator()` calls would want it on and the rest wouldn't. That turned out to be wrong, so I commented out
```
case '"':  if(fixquotes)
             c = '\''; // fall through
```
I'll send that as a PR in case it might save you a few minutes' work. The guess at line 230 is right and you'll have to decide about the rest.
In the meantime I'll upload the test case. I guess it's significant that it uses javadoc-style comments.

duncan-roe:
[reproducer3.tar.gz](https://github.com/user-attachments/files/18121328/reproducer3.tar.gz)

duncan-roe:
Simply **./run.sh** to run doxygen and see the man page

duncan-roe:
I tried to send a PR but one of the tests failed, when trying to download Ghostscript from chocolatey???
Nothing I can do about that so will try sending it again.

duncan-roe:
PR sent

duncan-roe:
Updated PR sent

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.13.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

duncan-roe:
I built and installed Doxygen1.13.0 and can confirm #11271 is fixed

PR #11272, opened 2024-12-13
issue #11271 man page verbatim changes \" to \'

duncan-roe:
6c5b9295509e introduced ManDocVisitor::filter() which corrects verbatim \\ but converts \" to \' which is inappropriate for case DocVerbatim::Verbatim and at least one other case.

This commit adds a bool to filter() so one can tune which calls to it do the \" conversion.

I have no idea which other calls should not do the \" conversion so for now that conversion is turned off for all calls by commenting it out.

Fixes: 6c5b9295509e: Issue with spacing around <programlisting>


duncan-roe:
The failing check on Windows is in installing Ghostscript

PR #11274, opened 2024-12-13
issue #11271 man pages incorrectly convert \" to \'

duncan-roe:
6c5b9295509e introduced `ManDocVisitor::filter()` which corrects \\ but converts \" to \' which is wrong for `case DocVerbatim::Verbatim` and the `case` of normal text.

On the assumption that there are cases where it is corect to convert \" to \', this commit adds a bool to `filter()` so one can tune which calls do the \" conversion.

I could not find which other call should not do the \" conversion, so for now that conversion is turned off for all calls by commenting it out.

Fixes: 6c5b9295509e: Issue with spacing around \<programlisting\>

### DIFF
--- a/src/mandocvisitor.cpp
+++ b/src/mandocvisitor.cpp
@@ -43,7 +43,7 @@ ManDocVisitor::ManDocVisitor(TextStream &t,OutputCodeList &ci,
 void ManDocVisitor::operator()(const DocWord &w)
 {
   if (m_hide) return;
-  filter(w.word());
+  filter(w.word(), true);
   m_firstCol=FALSE;
 }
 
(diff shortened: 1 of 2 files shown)


src/mandocvisitor.cpp
```diff
@@ -996,7 +996,8 @@ void ManDocVisitor::filter(const QCString &str)
       {
         case '.':  m_t << "\\&."; break; // see  bug652277
         case '\\': m_t << "\\\\"; break;
-        case '"':  c = '\''; // fall through
+        //case '"':  if(fixquotes)
```
doxygen, line 999: Looks like this lines is the only real change. The other changes do not have any effect as `fixquotes` doesn't appear to be used. 

Suggest to cleanup this PR.

duncan-roe:
#11271 Had an explanation for the odd-looking code. But OK - I'll send a new PR of a 1-liner

PR #11275, opened 2024-12-15
update TinyDeflate.hh to get fix clang 19.0

joequant:
update TinyDeflate to version 68ced8b which fixes
clang >= 19.0 compiles

### DIFF
--- a/deps/TinyDeflate/gunzip.hh
+++ b/deps/TinyDeflate/gunzip.hh
@@ -702,7 +702,7 @@ namespace gunzip_ns
                     // Note: Throws away progress already made traversing the tree
                     return ~std::uint_least32_t(0); // error flag
                 }
-                cur = (unsigned(cur) << 1) | unsigned(bool(p));
+                cur = (unsigned(cur) << 1) | bool(p);
             #ifdef DEFL_DO_HUFF_STATS
                 if(len > maxlen)
                 {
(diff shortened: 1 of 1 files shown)


albert-github:
@joequant Did you report is also at the developers of the original version at https://github.com/bisqwit/TinyDeflate i.e. at https://github.com/bisqwit/TinyDeflate 
- I don't think we should be out of line with the original version (when possible)
- As discussed in #10928 the `char8_t` type is a feature introduced in a version of the standard which not all compilers doxygen requires supports (doxygen needs a compiler that is C++ 2017 compliant but not necessarily C++ 2020)

albert-github:
I see you refer to https://github.com/bisqwit/TinyDeflate/commit/68ced8bd5c819264e628d4f063500753b77f613d so actually it might be better to update to that version of TinyDeflate instead of using a local update

albert-github:
I can see that the GitHub actions give quite a few warnings when using the TinyDeflate/gunzip.hh especially in respect to the removed define of CONCAT and its usage. Probably only the change at line 705 should be sufficient but I'm not sure.
I think the remark in https://github.com/doxygen/doxygen/pull/11275#issuecomment-2555398569 is still valid (regarding updating to https://github.com/bisqwit/TinyDeflate/commit/68ced8bd5c819264e628d4f063500753b77f613d) as this also should fix the CLang >= 19.0 problems.

Issue #11273, opened 2024-12-13
Using a tilde in the URL of a Markdown image tag breaks the tag if the image is in the IMAGE_PATH


azrogers:
**Describe the bug**
Using a Markdown image tag with a `~` character in the image URL (like `![](Example~/test.png)` to an image in the `IMAGE_PATH`, the following output is produced:
```html
<img src="Example" alt=""/>
<div class="caption">
    ~/test.png
</div>
```
This, of course, breaks the image, not to mention that having path of the image path in the caption is rather undesirable. The easy workaround would be to rename the folder containing images to something that doesn't contain a tilde, but we are using Doxygen to produce documentation for a Unity package, and [adding a tilde to the end of the folder name](https://docs.unity3d.com/Manual/SpecialFolders.html) allows the folder to be ignored by Unity. Otherwise, our documentation folder shows up in the Unity assets folder, which isn't intended.

This only seems to happen when the image is in the `IMAGE_PATH`.

**Screenshots**
![image](https://github.com/user-attachments/assets/888e4b1a-5f91-4c42-b137-c57db97a0bb9)

**To Reproduce**
Here is a bare-minimum example of this issue: [doxygen-bug.zip](https://github.com/user-attachments/files/18131736/doxygen-bug.zip)

**Expected behavior**
The intended output should be:
```html
<img src="test.png" alt="" />
```

**Version**
Doxygen `1.12.0 (c73f5d30f9e8b1df5ba15a1d064ff2067cbb8267)` running on Windows 11 Pro 64-bit. The same result also occurs when running the same version of Doxygen under WSL2 running Ubuntu 24.04.1 LTS using a binary obtained from the GitHub release.

albert-github:
Special characters in file / directory names are always a nuisance and especially when a specific package assigns a special meaning to it. (I'm not talking about the special meaning that e.g. *nix gives to a starting `~` or a starting `.` as these are not package specific but related to the OS).

The mentioned line
```
![](Example~/test.png)
```
translates internally in doxygen to:
```
@image html Example~/test.png \ilinebr @image latex Example~/test.png \ilinebr @image rtf Example~/test.png \ilinebr @image docbook Example~/test.png \ilinebr @image xml Example~/test.png \ilinebr
```


It looks like that in the file / directory parsing the `~` is not considered (I quickly check doctokenizer, commentscan and scanner).

@doxygen, should the `~ be added as (non first) character with a special meaning?

doxygen:
[...]

PR #11294, opened 2024-12-24
issue #11273 Using a tilde in the URL of a Markdown image tag breaks the tag if the image is in the IMAGE_PATH

albert-github:
Adding possibility to use a `~` in the file / directory name except on the first position of a file / directory name

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -600,8 +600,8 @@ ATTR      ({B}+[^>\n]*)?
 DOCNL     "\n"|"\\ilinebr"
 LC        "\\"{B}*"\n"
 NW        [^a-z_A-Z0-9]
-FILESCHAR [a-z_A-Z0-9\x80-\xFF\\:\\\/\-\+=@&#]
-FILEECHAR [a-z_A-Z0-9\x80-\xFF\-\+=@&#]
+FILESCHAR [a-z_A-Z0-9\x80-\xFF\\:\\\/\-\+=@&#~]
+FILEECHAR [a-z_A-Z0-9\x80-\xFF\-\+=@&#~]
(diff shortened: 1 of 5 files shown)


doxygen:
@albert-github I've tested on my Mac if I can also create an `~Example` directory with the test image file inside and that also seems to work fine! The directory appears as `'~Example'` when I list it (so with quotes around it), but can be accessed by doxygen without problems. This would mean that all the REJECT statements can be removed.

albert-github:
@doxygen 
That would make everything quite a bit simpler.
Did you test it on the Mac with or without the `REJECT` statements in place?

Should I change the PR or will you do it?

doxygen:
@albert-github I tested without the rejects (with the reject the image is completely ignored). I'll change it.

albert-github:
OK.

> with the reject the image is completely ignored

That was the intention of my changes, to keep it consistent with the old version when the `~` was at the first place and only allow a `~` at the second and following places, but as your tests reveal that at the first place it works as well it is good to allow it as well (less exceptions).
I assume there is no user `Example` on your Mac, what happens when you use a file at `~User` where `User` is a user of the system?

doxygen:
> I assume there is no user Example on your Mac, what happens when you use a file at ~User where User is a user of the system?

I just tested with user `~dimitri` as well (which does exist on my Mac). Also works. I guess the ~name is only something interpreted by the shell, not something special in the filesystem or the functions that work on the filesystem like `fopen()`.

albert-github:
Most likely / probably the `deps/filesystem` takes care of it.
It might give differences between *nix systems and Windows as on windows the `~` concept is not known (but it will hardly be the case that a user will reference to his home directory, will most likely be bad practice).

Issue #11315, opened 2025-01-02
Issue with @copybrief, @ref, and Markdown links in Markdown table


lcvgit:
**Describe the bug**
In a Markdown table, using `@copybrief` in a cell causes `[...](@ref ...)` in the next cell to render incorrectly.

The use of `[...](@ref ...)` in a cell that precedes `@copybrief` renders correctly; and removing `@copybrief` causes the use of `[...](@ref ...)` in the next cell to render correctly.

Possibly related to #7998.

**Screenshots**
![doxygen-bug](https://github.com/user-attachments/assets/dd7484a6-1c4c-44c0-9301-b7cfbdbb710e)

**To Reproduce**
```
/**
@page A
@brief Brief text.

| ... | ... | ... |
| --- | --- | --- |
| [`CORRECT`](@ref A) | @copybrief A | [`INCORRECT`](@ref A) |
| [`CORRECT`](@ref A) | No `@copybrief A` | [`CORRECT`](@ref A) |
*/
```

**Expected behavior**
Expect `@copybrief` in one cell of a Markdown table to not influence the next cell (or any other cell).

**Version**
1.12.0
ManjaroLinux 24.2.1
Binary from distribution

albert-github:
I've just pushed a proposed patch, pull request #11317

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.14.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #11317, opened 2025-01-03
issue #11315 Issue with `@copybrief`, `@ref`, and Markdown links in Markdown table

albert-github:
Detecting `\ilinebr` and seeing it as end of command as well.

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -551,6 +551,10 @@ size_t Markdown::Private::isSpecialCommand(std::string_view data,size_t offset)
       // skip over name (and optionally type)
       while (offset_<data_.size() && (c=data_[offset_])!='\n' && (allowSpaces || c!=' ') && c!='(')
       {
+        if (offset_<data_.size()-9 && data_[offset_]=='\\')
+        {
+          if (qstrncmp(&data_[offset_+1],"ilinebr ",8)==0) break;
+        }
         offset_++;
       }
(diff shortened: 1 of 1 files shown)


src/markdown.cpp
```diff
@@ -551,6 +551,10 @@ size_t Markdown::Private::isSpecialCommand(std::string_view data,size_t offset)
       // skip over name (and optionally type)
       while (offset_<data_.size() && (c=data_[offset_])!='\n' && (allowSpaces || c!=' ') && c!='(')
       {
+        if (offset_<data_.size()-9 && data_[offset_]=='\\')
```
doxygen, line 554: Please use `offset_+9<data_.size()` instead of `offset_<data_.size()-9`. Reason: if `offset_==7` and  `data_.size()==8` then `data.size()-9` is a very big value due to wrap around of an unsigned, and `offset_<data_.size()-9` will be true. Then `data[offset_+1]` below will do an out of bound access on a string_view.

albert-github: Corrected.
The change was, at the original writing of the PR, conform the other tests. They were changed due to:
```
Commit: b58548e0bace2a8448891f22874e6c276804a0a5 [b58548e]

Date: Thursday, January 9, 2025 3:09:34 PM

issue #11333 rebuilding fedora ignition-transport causes abort in isExplicitPage with doxygen 1.13.1
```

PR #11318, opened 2025-01-03
Adding debug and version information to doxycommentview

albert-github:
The README says:
Furthermore you can display the doxygen version used at the top of the output using `--version` and the warnings emitted by doxygen at the bottom of the output by means of `--debug`.

### DIFF
--- a/addon/doxycommentview/doxycommentview.py
+++ b/addon/doxycommentview/doxycommentview.py
@@ -63,8 +63,9 @@ def main():
                     self.wfile.write("<hr>".encode())
                 self.wfile.write(result.stdout.encode())
                 if DEBUG:
-                    self.wfile.write("<hr>".encode())
-                    self.wfile.write(result.stderr.replace('\n','<br>').encode())
+                    self.wfile.write("<hr>\n<pre>".encode())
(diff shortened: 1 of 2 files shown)


addon/doxycommentview/doxycommentview.py
```diff
@@ -50,7 +58,14 @@ def do_POST(self):
                 self.send_response(200)
                 self.send_header('Content-type', 'text/html')
                 self.end_headers()
+                if VERSION:
+                    self.wfile.write(VERSION_STR.stdout.encode())
+                    self.wfile.write("<hr>".encode())
                 self.wfile.write(result.stdout.encode())
+                if DEBUG:
+                    self.wfile.write("<hr>\n<pre>".encode())
+                    self.wfile.write(result.stderr.replace('&','&amp;').replace("'","&apos;").replace('<','&lt;').replace('>','&gt;').encode())
```
doxygen, line 67: Maybe we could use [`html.encode()`](https://docs.python.org/3/library/html.html) instead

albert-github: Did you mean `html.escape` ?

I have no objections against it, I'll leave it up to you.
(The debug info helped me quite a bit for "simple" problems).

PR #11319, opened 2025-01-04
Interaction between markdown blockquotes and sections

albert-github:
In an example (Fossies) I found the code like:
```
@page pg1 page

> text
## section
```
and this results in the warning like:
```
warning: found </blockquote> tag without matching <blockquote>
```
so it is better to terminate the block quote when a potential section command appears.

Example: [example.tar.gz](https://github.com/user-attachments/files/18306639/example.tar.gz)

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -2854,6 +2854,9 @@ size_t Markdown::Private::writeBlockQuote(std::string_view data)
       }
       else
       {
+        curLevel=level;
+        i = indent;
+        if (txt[0] == '#') break; // begin of markdown section command
         out += txt;
       }
       isGitHubFirst = false;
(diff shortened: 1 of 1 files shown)


src/markdown.cpp
```diff
@@ -2854,6 +2854,9 @@ size_t Markdown::Private::writeBlockQuote(std::string_view data)
       }
       else
       {
+        curLevel=level;
+        i = indent;
+        if (txt[0] == '#') break; // begin of markdown section command
```
doxygen, line 2859: looks like there is a risk we get here while `txt` is empty and since it is a string_view `txt[0]` can then trigger an out-of-bounds assert (on some systems).

albert-github: Indeed.

Solved some corner cases as well
```
> text
##NO header

##Second test on no header
```

and
```
> text
 ## text
```

Issue #11308, opened 2025-01-01
SVG dot graphs have added padding


meator:
> [!NOTE]
> A proposed solution is described at https://github.com/doxygen/doxygen/issues/11308#issuecomment-2573771030

**Describe the bug**
Dot graphs generated with `DOT_IMAGE_FORMAT = svg` have weird extra padding in them.

**Screenshots**
Expected result:

![image](https://github.com/user-attachments/assets/a2bbac33-56f3-4cbf-8ee2-32cd80544413)

Actual result:

![image](https://github.com/user-attachments/assets/2644387e-fd7b-4b5c-a7ba-d5610300e157)

**To Reproduce**
`main_page.dox`
```cpp
/** @mainpage
@dot
digraph {
    node [shape=plaintext]

    provider [label=<
<TABLE BORDER="1" CELLBORDER="0" CELLSPACING="0">
  <TR><TD>My table element</TD></TR>
</TABLE>>]
}
@enddot
*/
```

`Doxyfile`
```
INPUT = main_page.dox
HAVE_DOT = YES
DOT_IMAGE_FORMAT = svg
```

**Version**
Doxygen 1.12.0 with graphviz version 12.2.0 using Void Linux.

albert-github:
I ran (on windows)  with graphviz 12.2.0 and I got the expected behavior, so I don't think it is a doxygen problem.

Doxygen just copies the given dot code (to the html directory) and runs `dot` on it.
You can run doxygen with the setting `DOT_CLEANUP = NO`  so you can see what is written to the `.dot` file. When running doxygen with `-d extcmd` you can see which command doxygen invokes. 
The resulting svg file can then be inspected.

meator:
For future reference: this seems to be caused by font issues: https://forum.graphviz.org/t/unwanted-space-around-html-font-tags/460/6

A solution would be to set

```
DOT_IMAGE_FORMAT = svg:cairo
```

as described in the forum, but Doxygen unfortunately does not support the `svg:cairo` format.

albert-github:
Problem here is that `svg:cairo` is not recognized as possible format.
Reopening as enhancement.

albert-github:
@meator did you try `svg:cairo` as possible output format?

meator:
Doxygen does not support it (although it supports `png:cairo` and other `png` modifiers).

```
warning: argument 'svg:cairo' for option DOT_IMAGE_FORMAT is not a valid enum value
```

[...]

PR #11324, opened 2025-01-06
issue #11308 SVG dot graphs have added padding

albert-github:
- adding different possible image formats to `DOT_IMAGE_FORMAT` (found by trying e.g. `dot -Tsvg:xx`  and inspecting the error given)
- Format `svg:cairo` and `svg:vcairo:cairo` cannot be used with `INTERACTIVE_SVG` and also have a slightly different place of `width` and `height`)


src/dotfilepatcher.cpp
```diff
@@ -562,6 +563,10 @@ static bool readSVGSize(const QCString &fileName,int *width,int *height)
     {
       found=true;
     }
+    else if (sscanf(line.c_str(),"<svg xmlns=\"http://www.w3.org/2000/svg\" xmlns:xlink=\"http://www.w3.org/1999/xlink\" width=\"%d\" height=\"%d\"",width,height)==2)
```
doxygen, line 566: When using sscanf the parameters should be &width and &height otherwise doxygen may crash

albert-github: No not in this case, `width` and `height` are from `static bool readSVGSize(const QCString &fileName,int *width,int *height)` also check a view lines above for the old test.

doxygen: ah indeed, I didn't see this important bit of context.

PR #11338, opened 2025-01-12
Doxycommentview html corrections

albert-github:
correcting  the index.html of doxycommentview
- make `meta` tag correct
- the `<` inside the script is not HTML complient unless in a `CDATA section
- making startup tags consistent with "normal" doxygen output and xhtml

### DIFF
--- a/addon/doxycommentview/index.html
+++ b/addon/doxycommentview/index.html
@@ -1,7 +1,7 @@
-<!DOCTYPE html>
-<html lang="en">
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "https://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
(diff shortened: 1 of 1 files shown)


addon/doxycommentview/index.html
```diff
@@ -1,7 +1,7 @@
-<!DOCTYPE html>
-<html lang="en">
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "https://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US">
```
doxygen, line 2: Is there a reason to make the header use XHTML instead of HTML5? These are different standards and is probably also why the CDATA section is needed below. For XHTML also the file name should be .xhtml.

albert-github: - It is now the same as with the "normal" doxygen output (consistency)
- `CDATA` gave a small problem hence the `//` comment signs.

PR #11339, opened 2025-01-13
Add a config example to show how to display documentation of static inline functions

duncan-roe:
I thought it worth documenting how to deal with static inlines because it took me a while to figure out.
I thought the best way to do that was via a new config example.

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -208,6 +208,32 @@ Here Doxygen's preprocessor is used to substitute some
 macro names that are normally substituted by the C preprocessor,
 but without doing full macro expansion.
 
+\subsection config_examples_staticinlines Show static inline functions
+
+Supppose you have a c library project contained within a single directory with
+minimal configuration file:
+\verbatim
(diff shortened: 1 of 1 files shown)


src/config.xml
```diff
@@ -208,6 +208,32 @@ Here Doxygen's preprocessor is used to substitute some
 macro names that are normally substituted by the C preprocessor,
 but without doing full macro expansion.
 
+\subsection config_examples_staticinlines Show static inline functions
+
+Supppose you have a c library project contained within a single directory with
```
albert-github, line 213: spelling: Supppose -> Suppose

albert-github:
I'm not sure the configuration part of the manual is the correct place for this. Probably better as FAQ , with the `\static` command,an own example (in the appendices)  or ...

duncan-roe:
I sent in #11345 which is a new FAQ as @albert-github suggested, so no need to keep this one open

Issue #7522, opened 2020-01-18
Source code browser does not distinguish between same-named functions 


jrwrigh:
I have a fortran project that has same-named "twin" subroutines defined in different files (and different directories). Clicking on one of these subroutines in the source code browser will bring up the same subroutine regardless of which one is actually being used.

This has been a known issue for quite awhile, but is yet to be fixed. There was a patch submitted with #4365, but it never even got feedback (note it was submitted before the move to GitHub). I'm not sure how relevant that patch is to the current source code, but the changes are very minimal. 

Fixing this would not only resolve #4365, but also #3676, #5448, #979, and #3676. It may also fix #6251, but I'm not quite sure about that one.

doxygen version: 1.8.16

albert-github:
A bit of a late reaction.

I tried to reproduce the original problem (#5448) but failed.
To check this issue and evaluate the patch (#4365) / see whether an other implementation is necessary:
- Can you please attach a, small, self contained example (source+configuration file in a tar or zip) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.

albert-github:
I've just pushed a proposed patch, pull request #11356

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

PR #11356, opened 2025-01-19
issue #7522 Source code browser does not distinguish between same-named functions

albert-github:
Prefer the local file definition in case of multiple definitions

Example: [example.tar.gz](https://github.com/user-attachments/files/18468682/example.tar.gz)

### DIFF
--- a/src/fortrancode.l
+++ b/src/fortrancode.l
@@ -1213,7 +1213,7 @@ static MemberDef *getFortranDefs(yyscan_t yyscanner,const QCString &memberName, 
         { // found function in global scope
           if (cd == nullptr)
           { // Skip if bound to type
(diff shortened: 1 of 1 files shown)


src/fortrancode.l
```diff
@@ -1212,7 +1213,8 @@ static MemberDef *getFortranDefs(yyscan_t yyscanner,const QCString &memberName,
         { // found function in global scope
           if (cd == nullptr)
           { // Skip if bound to type
-            return md.get();
+            if (md.get()->getOutputFileBase() == yyextra->sourceFileDef->getOutputFileBase()) return md.get();
```
doxygen, line 1216: I think the check should be `md->getFileDef()==yyextra->sourceFileDef` instead of comparing the output file, which may be different depending on the `SEPARATE_MEMBER_PAGES` setting, for instance.

albert-github: Might be the case, changed as indicated.

PR #11357, opened 2025-01-19
Remove macro defines from C++ scoped enum values

gwllx:
Currently ``#define``s are listed as values of C++ scoped enums. For example:

```cpp
// config: EXTRACT_ALL=YES
// config: MACRO_EXPANSION=YES

enum A {
#define ID(X) X,
  ID(A1)
  ID(A2)
};

enum class B {
#define ID(X) X,
  ID(B1)
  ID(B2)
};
```

Produces the following output with v1.13.2:

Define_in_Scoped_Enum_Values

This is a somewhat common pattern, usually with an ``#include`` that uses some user-defined macro:

```cpp
enum class B {
#define ID(X) X,
#include "values.inc"
#undef ID
};
```

This PR is to exclude ``#define``s from the "strong" enum values in C++ only. These members would remain as members of the surrounding scope, as seen in the test case.

Note: the ``#define``s are not listed in unscoped/weak enums because they use a ``MemberNameLinkedMap`` lookup which does not include the ``#define`` names.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -7751,6 +7751,7 @@ static void addEnumValuesToEnums(const Entry *root)
               bool isJavaLike = sle==SrcLangExt::CSharp || sle==SrcLangExt::Java || sle==SrcLangExt::XML;
               if ( isJavaLike || root->spec.isStrong())
               {
+                if (sle == SrcLangExt::Cpp && e->section.isDefine()) continue;
                 // Unlike classic C/C++ enums, for C++11, C# & Java enum
                 // values are only visible inside the enum scope, so we must create
                 // them here and only add them to the enum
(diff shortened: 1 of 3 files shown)


albert-github:
@gwllx 
- See GitHub action results, looks like to be caused by the multiple definition of `ID`.
- This is caused by the fact that the define names are the same (not caused by this PR), we see the same problem in the doxygen internal documentation.


@doxygen 
- what would here be the best way to proceed?

albert-github:
The chosen solution regarding the names of the `define`s is good for this PR, but the background problem still remains, but should be handled based on a new issue / PR.

(Note: that the word `ID` is just the word used by `xmllint` as an abbreviation for "identification" and has nothing to do with the name of the `define` as this name is made into the hash value shown).

doxygen:
> @doxygen
> 
>     * what would here be the best way to proceed?

Normally `#define`'s need to be undefined using `#undef` before they can be redefined. Doxygen should already support this.

albert-github:
@doxygen
In my opinion the `#undef` have no influence on this problem as it is about the name (hash) and this is the same irrespective of the `#undef`

PR #11367, opened 2025-01-24
bug_771452 Warnings show wrong filename when same addtogroup name used in multiple files

albert-github:
Positional commands added at to of command block as otherwise when adding extra blocks the wrong line might be used


albert-github:
Looking into the problem regarding the tests 64, 74 and 81.

albert-github:
Fixed test and documentation problems

PR #11404, opened 2025-02-07
Incorrect field when ampersand in URL

albert-github:
When having an ampersand in an URL like:
```
https://en.wikipedia.org/w/index.php?title=Unsharp_masking&oldid=750486803#Photographic_unsharp_masking
```
this can lead to a warning in the xmllint checker like:
```
html\md_aa.html:108: parser error : EntityRef: expecting ';'
ck"><p><a href="https://en.wikipedia.org/w/index.php?title=Unsharp_masking&oldid
                                                                               ^
```

Note in e.g. the xml and docbook format there were already twice the filter call.

Example: [example.tar.gz](https://github.com/user-attachments/files/18707233/example.tar.gz)

(Found in the ITK package)

### DIFF
--- a/src/htmldocvisitor.cpp
+++ b/src/htmldocvisitor.cpp
@@ -401,7 +401,8 @@ void HtmlDocVisitor::operator()(const DocURL &u)
   else // web address
   {
     m_t << "<a href=\"";
-    m_t << u.url() << "\">";
+    filter(u.url());
+    m_t << "\">";
     filter(u.url());
     m_t << "</a>";
   }


src/htmldocvisitor.cpp
```diff
@@ -401,7 +401,8 @@ void HtmlDocVisitor::operator()(const DocURL &u)
   else // web address
   {
     m_t << "<a href=\"";
-    m_t << u.url() << "\">";
+    filter(u.url());
```
doxygen, line 404: The url needs URL encoding inside the href and HTML encoding outside, see https://stackoverflow.com/a/1812486/784672. Probably also holds for XML and docbook.

albert-github: Looks indeed the case that is required for all formats that potentially use the string to start a HTML browser, so probably not only XML and Docbook but also LaTeX, RTF, ...

doxygen: Hmm, I tried URL encoding, but then & is not escaped, so it still doesn't pass XML/XHTML checks. So I think your fix is correct as it is.

PR #11405, opened 2025-02-08
Not possible to place a `\cite` in a `\par` title

albert-github:
When having:
```
\par cite_based \cite lorigo2001.
```
this will lead to:
```
warning: Illegal command '@cite' found as part of a title section
```
although there is no reason prohibiting it, especially as in a `\section` it is possible to have it.

(During evaluation also the `\anchor` caused some problems with a subsequent `\par` statement (due to switching to a wrong state).

Example: [example.tar.gz](https://github.com/user-attachments/files/18719355/example.tar.gz)

(Found for the ITK project)


src/docparser.cpp
```diff
@@ -1028,24 +1028,131 @@ void DocParser::handleAnchor(DocNodeVariant *parent,DocNodeList &children)
...
+  }
+
+  tokenizer.pushState();
+  tokenizer.setStateCite();
+  tok=tokenizer.lex();
+  if (tok.is_any_of(TokenRetval::TK_NONE,TokenRetval::TK_EOF))
+  {
+    warn_doc_error(context.fileName,tokenizer.getLineNr(),"THE ONE unexpected end of comment block while parsing the "
+        "argument of command '\\{}'",context.token->name);
+    return;
```
doxygen, line 1142: @albert-github Looks like here the state pushed at line 1135 is not popped.

Maybe better to create a RAII object, i.e. 

```c++
class AutoTokenizerState
{
   public:
     AutoTokenizerState(DocTokenizer *tok) : m_tok(tok) { m_tok->pushState(); }
    ~AutoTokenizerState() { m_tok->popState(); }
   private:
     DocTokenizer *m_tok = nullptr;
}; 
```
so you can do

```c++
  { 
     AutoTokenizerState(tokenizer);
     ....
     if (cond) return;
     ....
  }
```
and have state always automatically popped when exiting the block.

albert-github: I'm not really familiar with "RAII objects" so when it is better please use this to implement it.

Issue #11423, opened 2025-02-15
Allow explicit linking inside of name objects (\throws, \retval...)


meator:
**Describe the feature request**
This is a feature request.

```cpp
/**
 * @throws MyError If not successful.
 */
```

The problem is that it's impossible to make `MyError` link to MyError when `AUTOLINK_SUPPORT` is disabled. Prepending it with `#` does nothing and using `\ref` in this context is invalid. Attempting to use `\ref` in the description leads to

```
/doxygen-repro/lib.h:45: warning: unexpected token TK_COMMAND_AT in comment block while parsing the argument of command throws
```

It is unfortunate that this feature is only accessible indirectly, there is currently no way to explicitly link the `MyError` name in `@throws` to the `MyError` class. I have tried to ask around whether there is some way I am not aware of to do this, but none of the answers I got did work: #11296

This could be useful for other commands which accept a name and a description too. Another example would be `\retval` (where it's arguably less useful).

**Screenshots**
![Documentation with autolinking enabled](https://github.com/user-attachments/assets/d9dd1114-e731-4baf-b460-287613d96a28)
autolink
![Documentation with autolinking disabled](https://github.com/user-attachments/assets/d40fdc2c-3f72-4df7-9b20-3fb417a4598c)
no autolink

**To Reproduce**
[doxygen-repro.tar.gz](https://github.com/user-attachments/files/18812637/doxygen-repro.tar.gz)

**Version**
Doxygen 1.13.2 using Void Linux.

paule32:
A possible Solution would be:

- I replace the Line: 2186 of **src/htmldocvisitor.cpp** with:  `m_t << "<a class=\"UUUel\" ";`
- then I have compile with nmake (Microsoft CL.EXE)
- then I have create a Python File with the following Content:
```
# ----------------------------------------------------------------------------
# File : filter.py
# Autor: (c) 2024 by Jens Kallup - paule32
#        all rights reserved.
#
# only for education, and non-profit !
# ----------------------------------------------------------------------------
import sys
import os
import re
import shutil
import subprocess

# ----------------------------------------------------------------------------
# cut ../ recursive from a given string "text" ...
# ----------------------------------------------------------------------------
def recursive_cut(text):
    # -------------------------------------------------
    # default: when text is empty, or dont contain ../
    # then result will be the text ...
    # -------------------------------------------------
    if not text or not text.startswith("../"):
        return text
    
    # -------------------------------------------------
    # recursive cut "../" ...
[...]

PR #11425, opened 2025-02-15
Update htmlgen.cpp 2025-02-16

paule32:
Line: 1787 - change **el** to **RENAMED_el** as workaround for a CHM Link produce Problem. When RENAMED_ is not available, doxygen tends to produce double quotes for a Link

### DIFF
--- a/src/htmlgen.cpp
+++ b/src/htmlgen.cpp
@@ -1784,7 +1784,7 @@ void HtmlGenerator::writeObjectLink(const QCString &ref,const QCString &f,
   }
   else
   {
-    m_t << "<a class=\"el\" ";
+    m_t << "<a class=\"RENAMED_el\" ";   // paule32: 16-02-2023 TODO !!!
   }
   m_t << "href=\"";
   QCString fn = f;


albert-github:
CHM is an very old and unmaintained format, it is still present in doxygen as it "works".
When changing the `el` into `RENAMED_el` this will break all other usage of the `el` class and thus is definitely not the way to go.

You wrote:
> When RENAMED_ is not available, doxygen tends to produce double quotes for a Link

Please give a small example that shows the problem:
- where is the problem with the double quotes?

paule32:
This Request is related to the Discussion: https://github.com/doxygen/doxygen/discussions/11407
I realize, when I use the original Source Code, Doxygen produce double Qoutes in  **<a class "el" href=""**

`<tr class="memitem:"><td class="memItemLeft" align="right" valign="top">struct &#160;</td><td class="memItemRight" valign="bottom"><a class="el" href=""../../dc/d9f/structstruct.html"">struct</a></td></tr>
`

with:  **<a class "RENAMED3_el" href="**

`<tr class="memitem:"><td class="memItemLeft" align="right" valign="top">struct &#160;</td><td class="memItemRight" valign="bottom"><a class="RENAMED3_el" href="../../dc/d9f/structstruct.html">struct</a></td></tr>`

I don't know these exception how it comes up.
But in a CHM Project when double Qoutes are present, the Help Linkage does a UB (Undefined Beauvoir) and the linked Page will not open and display. The resulting behauvoir is, that the actual WebPage will be open.

With the Prefix **RENAMED3_** the linked Page will be open and display.

A similar Problem is with \ref issue #11423

I attached a little bigger Test Project. so you can be reproduce this:
[packed.zip](https://github.com/user-attachments/files/18814294/packed.zip)

albert-github:
I'd like to try to reproduce but there are no sources nor Doxyfile in the provided zip file, when running the packed.zip I also don't see a href with double double quotes. Please provide a new packed.zip that reproduces your problem.

- indeed strange that set of double double quotes with the href attribute
- you mentioned `a class "el"` so a missing equal sign but in the larger code snippet I don't see this.

I don't think the `\ref`problem in #11423 is in any way related to the problem here.

paule32:
I am sorry for this. You have right - my typo mistake - I forget the equal sign.
I attached a new test project.
[packed.zip](https://github.com/user-attachments/files/18814524/packed.zip)

albert-github:
When generating the documentation with doxygen for the project provided (without the provided bat file as this is just some user code around the generation that is not part of doxygen) I cannot find the double double quotes in the output. 
Out of curiosity I browsed through the bat file and found the filter.py and I ran that on the generated documentation and the result was that the double double quotes were present. (might be line 133 of filter.py  but I didn't investigate and won't do it either).

Conclusion the problem is not a doxygen problem but a problem introduced by the user by postprocessing the results.

paule32:
Yes, right. How did you found the typo so fast ?
You must be a genius :-)

albert-github:
Just using my friends "grep" and "vi", using a sound educational guess about the string that are involved and having a bit of luck.

I first always go to the bare minimum, excluding "user" influence to see whether it is a doxygen problem or a "user" problem and from this I go some steps further.

PR #11431, opened 2025-02-17
Sort Related Pages by title

cs2350:


albert-github:
Would always be nice when there is a reason given for closing the PR.

Issue #9869, opened 2023-02-21
Change inheritance hierarchy graph orientation


groundedtako:
**Describe the bug**
Is there a way to change the orientation of a inheritance hierarchy graph from horizontal to vertical, or vise versa? If the inheritance hierarchy is too big, the graph tends to overflow the screen.

**Expected behavior**
Able to change orientation in doxyfile config settings.

**Version**
1.8.17

albert-github:
To the best of my knowledge there is no way to change the orientation of the inheritance hierarchy graph through doxygen, though it is an interesting idea.
- Which output engine are you referring to (HTML / LaTeX / ...)
- Which `DOT_IMAGE_FORMAT` are you using (in case of HTML the svg format might be an idea).

We are talking here about the `rankdir="LR"` for the graph and when setting this on a global scale this would influence all graphs.


Note the 1.8.17 version of doxygen is from December 27, 2019, the current doxygen version is 1.9.6. It is advised to update.

Issue #4453, opened 2018-07-02
would like config option to make TB rather than LR dot graphs (Origin: bugzilla #663120)


doxygen:
status NEW severity *enhancement* in component *general* for *---*
Reported in version *1.8.2-SVN* on platform *Other*
Assigned to: Dimitri van Heesch

On 2011-10-31 19:35:21 +0000, Chip Christian wrote:

> I can do this via post-processing but it would be nice to be able to specify this via a Doxyfle.

On 2011-10-31 19:38:26 +0000, Chip Christian wrote:

> This refers to rankdir which I meant to mention before...

On 2012-11-18 11:07:56 +0000, Dimitri van Heesch wrote:

> Changed version 'latest' to '1.8.2-SVN' so I can remove 'latest' as an option as it is a moving target.

PR #11440, opened 2025-02-20
Add configuration options to set rankdir for dot graphs

ulmus-scott:
These commits add configuration options to set rankdir for dot graphs.

Closes https://github.com/doxygen/doxygen/issues/4453 see DOT_GRAPH_ATTR
Closes https://github.com/doxygen/doxygen/issues/9869 see DOT_RANKDIR_CLASS

I'm not sure what config.xml is supposed to include and setting a defval for DOT_GRAPH_ATTR caused it to print it in the generated doxyfile, but without the preceding newline and "# The default value is: ".

I kept the inverse behavior of the call graphs and the strange switching to LR with DotClassGraph.

The final commit is optional and changes VHDL flow charts, applying the other global attributes to the dot graph.

For quick testing I created a simple class hierarchy: 
[doxygen-test-rankdir.tar.gz](https://github.com/user-attachments/files/18883135/doxygen-test-rankdir.tar.gz)

I started getting a nonsense warning from gcc after adding DOT_RANKDIR_DIR_DEP:
```
In file included from /home/scott/build/doxygen/src/types.h:19,
                 from /home/scott/build/doxygen/src/definition.h:21,
                 from /home/scott/build/doxygen/src/namespacedef.h:22,
                 from /home/scott/build/doxygen/src/namespacedef.cpp:18:
In function ‘int qstrcmp(const char*, const char*)’,
    inlined from ‘bool operator==(const QCString&, const char*)’ at /home/scott/build/doxygen/src/qcstring.h:611:17,
    inlined from ‘virtual bool NamespaceDefImpl::isLinkableInProject() const’ at /home/scott/build/doxygen/src/namespacedef.cpp:1537:23:
/home/scott/build/doxygen/src/qcstring.h:70:33: warning: ‘int strcmp(const char*, const char*)’ of a string of length 20 and an array of size 16 evaluates to nonzero [-Wstring-compare]
   70 | { return (str1 && str2) ? strcmp(str1,str2) :     // both non-empty
      |                           ~~~~~~^~~~~~~~~~~
/home/scott/build/doxygen/src/namespacedef.cpp: In member function ‘virtual bool NamespaceDefImpl::isLinkableInProject() const’:
/home/scott/build/doxygen/src/namespacedef.cpp:1536:21: note: in this expression
 1536 |   if (extractAnonNs &&                             // extract anonymous ns
      |       ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1537 |       name().mid(i,20)=="anonymous_namespace{"     // correct prefix
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


albert-github:
Thanks for the efforts.
The way doxygen handles PRs etc in respect to closing issues etc. is in such a way that the issues are kept open when a PR is merged until the next official release, as you used the sentence with "closes issue_number" GitHub has the (for doxygen bad) intelligence to add an automatic issue close on complete:
image

- Please remove these 2 links (hitting the "wheel" and removing the check marks will probably do).

Regarding the compilation warning, strange. I don't see at first glance what could be wrong.
- which compiler and compiler version did you use?

The code itself, I have to study it a bit and will write comments to the code later on.

Regarding:
> I have not investigated, and am not planning on it, but it would also be nice if there were doxygen commands to be able to set the rankdir on a per graph basis, e.g. some graphs in the same category may be more compact LR, others TB.

No there is no no doxygen command to set any attribute on a per graph base, would probably be a nice enhancement but quite difficult seen the different possibilities for all graphs (especially inheritance / collaboration graphs). I think the command should not be limited to just the rankdir but also have other attribute possibilities (and maybe even the possibility of setting the maximum number of nodes of a particular graph, which is now a  global settings number). I think these remarks might also be valid for the settings as done here which is limited to just the rankdir. This will need some serious thinking and definition I think.

src/config.xml
```diff
@@ -3802,6 +3802,130 @@ where `loc1` and `loc2` can be relative or absolute paths or URLs.
...
       </docs>
     </option>
+    <option type='list' id='DOT_GRAPH_ATTR' format='string' depends='HAVE_DOT'
+      defval='bgcolor=\"transparent\"'>
+      <docs>
+<![CDATA[
+ \c DOT_GRAPH_ATTR is common attributes for graphs.
+ For details please see
+ <a href="https://graphviz.org/doc/info/attrs.html">Node, Edge and Graph Attributes specification</a>.
+ <br>The default value is:<br>
```
albert-github, line 3812: Text for the default value is generated automatically, so sentence should not be here.

src/config.xml
```diff
@@ -3802,6 +3802,130 @@ where `loc1` and `loc2` can be relative or absolute paths or URLs.
 ]]>
       </docs>
     </option>
+    <option type='list' id='DOT_GRAPH_ATTR' format='string' depends='HAVE_DOT'
+      defval='bgcolor=\"transparent\"'>
```
albert-github, line 3806: I don't think the double quotes should be escaped (`\"`), it is not necssary as the string is embedded in single quotes , probably the double quotes can removed all together (see e.g. `DOT_COMMON_ATTR`)

src/config.xml
```diff
@@ -3802,6 +3802,130 @@ where `loc1` and `loc2` can be relative or absolute paths or URLs.
...
+<![CDATA[
+ \c DOT_GRAPH_ATTR is common attributes for graphs.
+ For details please see
+ <a href="https://graphviz.org/doc/info/attrs.html">Node, Edge and Graph Attributes specification</a>.
+ <br>The default value is:<br>
+]]>
+      </docs>
+      <value name='bgcolor=\"transparent\"'/>
+    </option>
+    <option type='string' id='DOT_RANKDIR_CALL' format='string' depends='HAVE_DOT' defval='LR'>
```
albert-github, line 3817: Shouldn't the default be the empty string?

src/dotcallgraph.cpp
```diff
@@ -164,11 +164,16 @@ QCString DotCallGraph::getBaseName() const
 
 void DotCallGraph::computeTheGraph()
 {
+  QCString rankdir = Config_getString(DOT_RANKDIR_CALL);
+  if (rankdir.size() >= 2 && m_inverse)
+  {
+    std::swap(rankdir[0], rankdir[1]);
+  }
```
albert-github, line 171: How would you get 2 or more values on the rankdir part?
What happens with the values 3, 4, ...

albert-github (commented):
I'm not a fan of this many extra settings. I already made some comments about this in https://github.com/doxygen/doxygen/pull/11440#issuecomment-2670863887 and the total concept of handling the attributes for `dot` should in my opinion be reconsidered.

src/config.xml
ulmus-scott: Except when I run `doxygen -g`, the sentence isn't generated but the value is inserted (with a period after and no space or new line).

src/dotcallgraph.cpp
ulmus-scott: rankdir is supposed to be only "TB", "BT", "LR", or "RL"; this is supposed to swap the characters in the string: "TB" becomes "BT", etc.

My version of `dot` appeared to reset to its default (TB) if given an invalid string, e.g. "foo", so I saw no reason to explicitly check and replace when swapping does the same thing.

src/config.xml
ulmus-scott: If DotCallGraph::m_inverse was false, rankdir was "LR".  I am trying to preserve the previous behavior.

src/config.xml
ulmus-scott: I think the behavior is different since this is `type='list'` not `type='string'`.

src/config.xml
ulmus-scott: Escaping the double quotes is necessary to create the same output as before:
```
  bgcolor="transparent";
```

I think having un-escaped `"` would confuse the parsing of the doxyfile?

You are correct that dot also accepts `bgcolor=transparent`, but I was trying to keep the output the same.

ulmus-scott:
> Thanks for the efforts. The way doxygen handles PRs etc in respect to closing issues etc. is in such a way that the issues are kept open when a PR is merged until the next official release, as you used the sentence with "closes issue_number" GitHub has the (for doxygen bad) intelligence to add an automatic issue close on complete: image
> 
>     * Please remove these 2 links (hitting the "wheel" and removing the check marks will probably do).
> 

I did not see the gear icon, but replacing "closes" with "references" in the comment appears to have removed it.

> 
> Regarding the compilation warning, strange. I don't see at first glance what could be wrong.
> 
>     * which compiler and compiler version did you use?
> 

```
gcc --version
gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
```

> 
> The code itself, I have to study it a bit and will write comments to the code later on.
> 
> Regarding:
> 
> > I have not investigated, and am not planning on it, but it would also be nice if there were doxygen commands to be able to set the rankdir on a per graph basis, e.g. some graphs in the same category may be more compact LR, others TB.
> 
> No there is no no doxygen command to set any attribute on a per graph base, would probably be a nice enhancement but quite difficult seen the different possibilities for all graphs (especially inheritance / collaboration graphs). I think the command should not be limited to just the rankdir but also have other attribute possibilities (and maybe even the possibility of setting the maximum number of nodes of a particular graph, which is now a global settings number). I think these remarks might also be valid for the settings as done here which is limited to just the rankdir. This will need some serious thinking and definition I think.

Adding commands to supply dot statements that would be inserted at the end of `DotGraph::writeGraphHeader()`, probably replacing the new rankdir parameter and similar to how DOT_GRAPH_ATTR works, would probably work.  I don't know how the maximum number of nodes is set, so that may also be possible via the same proposed commands.

Generalizing this from rankdir to any graph attribute would probably require changes to how DotClassGraph and DotCallGraph behave.

With DotClassGraph's unique dynamic selection of rankdir, it would be easiest to remove, but maybe something like:
```
auto attributes = getConfig(DOT_ATTR_CLASS);
if (!attributes.contains("rankdir") && m_lrRank)
{
    attributes.append("rankdir=\"LR\"");
}
```

DotCallGraph's inverse direction may be achieved with appending `edge [dir=back]` (https://graphviz.org/docs/attrs/dir/), but I would need to test it to be sure.

> I'm not a fan of this many extra settings. I already made some comments about this in [#11440 (comment)](https://github.com/doxygen/doxygen/pull/11440#issuecomment-2670863887) and the total concept of handling the attributes for `dot` should in my opinion be reconsidered.

I don't think this is significantly different to how the dot graphs are already handled.  See https://www.doxygen.nl/manual/commands.html#cmdgroupgraph, etc.; and the configuration options: CLASS_GRAPH, COLLABORATION_GRAPH, GROUP_GRAPHS, TEMPLATE_RELATIONS, INCLUDE_GRAPH, INCLUDED_BY_GRAPH, CALL_GRAPH, CALLER_GRAPH, GRAPHICAL_HIERARCHY, and DIRECTORY_GRAPH.

src/dotcallgraph.cpp
albert-github: Which version of `dot` do you use?

ulmus-scott:
```
dot -V
dot - graphviz version 2.43.0 (0)
```

I had tested the behavior with an invalid rankdir using a valid rankdir in DOT_GRAPH_ATTR and setting an invalid DOT_RANKDIR_INCLUDE.

ulmus-scott:
> DotCallGraph's inverse direction may be achieved with appending `edge [dir=back]` (https://graphviz.org/docs/attrs/dir/), but I would need to test it to be sure.

`edge [dir=back]` only reverses the direction of the arrows, i.e. A->B shows as A<-B, the position of the nodes in the rendered graph remain the same, so extracting rankdir from the attribute list would be required to reverse/mirror it.

ulmus-scott:
[doxygen-test-rankdir_with-call-graphs.tar.gz](https://github.com/user-attachments/files/18919693/doxygen-test-rankdir_with-call-graphs.tar.gz)
If you set `DOT_RANKDIR_CALL = foo` in Doxyfile, the call graphs will all be TB (default) and the `.dot` files will show either `rankdir="foo";` (e.g. structLevel0_a6116ba32d0d153509eb40e6b95b50345_cgraph.dot) or `rankdir="ofo";` (e.g. structLevel0_a6116ba32d0d153509eb40e6b95b50345_icgraph.dot).  Of course, it also works correctly as expected with the valid LR, RL, TB, and BT rankdirs.

src/dotcallgraph.cpp
ulmus-scott: If rankdir is empty, the graph is not reversed with m_inverse true.  We could set rankdir to "LR" if its size is not 2, mentioning this in the config.xml documentation.  Or add a separate configuration option for the inverse call graph rankdir.

PR #11448, opened 2025-02-23
Handling of `TIMESTAMP` in html footer and header

albert-github:
In ITK: https://github.com/InsightSoftwareConsortium/ITK/issues/5190  there has been a discussion about the use of `TIMESTAMP` in the html footer (and header) in respect to reproducibility (and creation of difference sets for documentation sites). When the date is replaced by the actual value all the html files are involved. I the ITK project a html string was used and its value was replaced by means of a JavaScript function during the load of the html page.

This technique has now been implemented in doxygen as well.


src/htmlgen.cpp
```diff
@@ -122,6 +122,32 @@ static void writeServerSearchBox(TextStream &t,const QCString &relPath,bool high
   }
 }
 
+const char *replaceDatetime =  "<span class=\"datetime\"></span>;";
+const char *replaceDate     =  "<span class=\"date\"></span>;";
+const char *replaceTime     =  "<span class=\"time\"></span>;";
+const char *replaceYear     =  "<span class=\"year\"></span>;";
+static void writeBuildDateJS(TextStream &t)
```
doxygen, line 129: Since running the script does add to the load time of each page, it is probably wise to only do the string replacements for the elements that are present, and even omit the script if non are present (which looks to be the default).

albert-github: It indeed adds some load time, but it will probably be possible to load only at the moment there are one of these replaceable strings in the header / footer.

I'll have a look at it.

albert-github: Done.

Is there a reason why the `selectBlocks` is done in `substituteHtmlKeywords` for each file instead of doing it directly during the `init` ob the `g_footer` / `g_header`

doxygen: I think there is an even simpler way to do this without any javascript. Here is proof of concept:

a file `date.css` with
```.css
.show-date:after { content: "21-03-2025"; }
```

used like this:
```.html
<!DOCTYPE html>
<html>
<head>
<title>Show variable defined in CSS</title>
<link href="date.css" rel="stylesheet" type="text/css"/>
</head>
<body>
This file was generated on <span class="show-date"></span> from source code.
</body>
</html>
```

Let me know what you think.

albert-github: I followed the way it is done in the ITK project.
I don't know what the time consumption differences between the javascript and css solution would be, but it looks that both approaches are valid approaches,.
I'll leave it up to you to choose the better approach.

albert-github: I've updated the PR and made a CSS version instead of a JS version out of it.

albert-github:
Unfortunately this does not work / does not solve the requirement to have minimal changes to the files in case of a regeneration.
- as soon as e.g `$datetime` / `$date` / `$time` is used in a header of footer the time of generation is propagated to all generated HTML files and thus not the minimum change.

doxygen:
> as soon as e.g $datetime / $date/$time` is used in a header of footer the time of generation is propagated to all generated HTML files and thus not the minimum change.

This should also not be used in the header or footer. Instead `<span class="timestamp"></span>` should be used at the point where a data/time is needed. The browser will do the replacement with the time stored in doxygen.css, which is the only file that changes when regenerating the output.

albert-github:
Fixed partly by [9509724](https://github.com/doxygen/doxygen/commit/950972416ec8760811d009f67429bb1a72a9c2d2)
Obsoleted by #11603 in respect to `$datetime`, `$date`, `$time` and `$year`

PR #11449, opened 2025-02-23
Don't repeat the exact same warning messages

albert-github:
Some warnings are repeated a number of times even though they refer to the same problem. By storing a md5sum of the warnings (less space than the full message) this can be prevented.

Example: [example.tar.gz](https://github.com/user-attachments/files/18931027/example.tar.gz)

### DIFF
--- a/src/message.cpp
+++ b/src/message.cpp
@@ -163,7 +163,10 @@ void warn_(WarningType type, const QCString &file, int line, fmt::string_view fm
 
 void warn_uncond_(fmt::string_view fmt, fmt::format_args args)
 {
-  if (checkWarnMessage(g_errorStr+fmt::vformat(fmt,args))) fmt::print(g_warnFile,"{}{}",g_warningStr,vformat(fmt,args));
+  {
+    std::unique_lock<std::mutex> lock(g_mutex);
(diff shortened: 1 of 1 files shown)


src/message.cpp
```diff
@@ -37,9 +39,20 @@ static QCString        g_warnlogFile;
...
 //-----------------------------------------------------------------------------------------
 
+static bool checkWarnMessage(QCString result)
+{
+  uint8_t md5_sig[16];
+  char sigStr[33];
+  MD5Buffer(result.data(),result.length(),md5_sig);
+  MD5SigToString(md5_sig,sigStr);
+
+  return g_warnHash.insert(sigStr).second;
```
doxygen, line 53: This is not thread-safe. I think it will lead to memory corruption if multiple threads are enabled. Should only be called when the `g_mutex` lock is held.

albert-github: I agree, I thought that I checked whether or not the mutex was held, will have to recheck.

albert-github: Fixed 2 place, but wouldn't the problem also not have been in the original code or is `fmt::print(g_warnFile,"{}{}",g_warningStr,vformat(fmt,args));` thread safe?

PR #11468, opened 2025-02-28
Add Linux ARM to CI

### LINKED ISSUE #11467: Request: Linux ARM Builds
Add an `ubuntu-arm` image to CI builds and distribute them as apart of Doxygen binaries, so we can use it on Linux ARM machines.

gmitch215:
This PR solves issues for #11467. Not tested.

The details for the arm image are [here](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md). There is no `ubuntu-20.04-arm` image, so I used `ubuntu-22.04-arm`.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -44,28 +44,28 @@ jobs:
 
         - {
             name: "Ubuntu Latest GCC Release (Arm)",
-            os: ubuntu-22.04-arm,
+            os: ubuntu-24.04-arm,
             build_type: "Release", cc: "gcc", cxx: "g++",
             build_gen: "Unix Makefiles",
             cmake_extra_opts: "-Dbuild_search=YES -Dbuild_app=YES -Dbuild_parse=YES"
           }
         - {
             name: "Ubuntu Latest GCC Debug (Arm)",
-            os: ubuntu-22.04-arm,
+            os: ubuntu-24.04-arm,
             build_type: "Debug", cc: "gcc", cxx: "g++",
             build_gen: "Unix Makefiles",
             cmake_extra_opts: "-Dbuild_search=YES -Dbuild_app=YES -Dbuild_parse=YES"
           }
         - {
             name: "Ubuntu Latest Clang Release (Arm)",
-            os: ubuntu-22.04-arm,
+            os: ubuntu-24.04-arm,
             build_type: "Release", cc: "clang", cxx: "clang++",
             build_gen: "Unix Makefiles",
             cmake_extra_opts: "-Duse_libclang=YES -Dstatic_libclang=YES -Duse_libc++=NO"
           }
         - {
             name: "Ubuntu Latest Clang Debug (Arm)",
-            os: ubuntu-22.04-arm,
+            os: ubuntu-24.04-arm,
             build_type: "Debug", cc: "clang", cxx: "clang++",
             build_gen: "Unix Makefiles",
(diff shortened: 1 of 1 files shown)


gmitch215:
Updated things are mostly the same, though Ubuntu 22.04 requires later versions of different packages.

Using `libxapian-dev` is the same on Intel and Arm as well.

.github/workflows/build_cmake.yml
```diff
@@ -380,15 +427,15 @@ jobs:
       with:
         name: "Html documentation artifacts"
         path: build/html/
-      if: matrix.config.name == 'Ubuntu Latest GCC Release'
+      if: matrix.config.name == 'Ubuntu Latest GCC Release (Intel)' || matrix.config.name == 'Ubuntu Latest GCC Release (Arm)'
```
albert-github, line 430: - The HTML and Latex documentation artifacts should only be done for one platform, so for `Ubuntu Latest GCC Release (Intel)` only
- The generation of the internal documentation should only be done for `Ubuntu Latest GCC Release (Intel)`

albert-github:
Looks like that for Qt also the arm version has to be loaded (when available):
```
 /bin/sh: 1: /home/runner/work/doxygen/Qt/5.15.2/gcc_64/bin/moc: Exec format error
```
(I killed rest of the jobs)

gmitch215:
According to [this comment](https://github.com/jurplel/install-qt-action/issues/248#issuecomment-2597167011) there is a fork of the `install-qt-action` that supports Linux ARM. I just switched the action.

albert-github:
In principle I'm not fond of changing things to no "official" versions, but for a test I just started it and unfortunately it looks like that for the `arm` version it doesn't work. For the other versions it worked. So I don't see a difference yet with the old version.

jdpurcell:
I'm the author of that `install-qt-action` fork. I created it so I can use https://github.com/jurplel/install-qt-action/pull/268 and https://github.com/jurplel/install-qt-action/pull/273 while we are waiting for them to be officially released. After that I don't think there will be any point to using it, but I might keep it up to date (basically just a clone of the official version) in case I need a quick fix for something and don't want to wait for official. Nonetheless, probably a good idea to use the fork just for testing and wait for the official release, or if you do merge this sooner, change it back to official after its next release.

As for the install error with ARM, Qt didn't publish pre-compiled binaries for Linux ARM64 until 6.7, so any older Qt versions will not work. In addition, I believe you'd need to stick specifically to 6.7.x if you intend to use the Ubuntu 22.04 runner. Starting with Qt 6.8, they are compiled on an Ubuntu 24.04 host; if I try to use the 6.8 binaries it seems to require the Ubuntu 24.04 runner to avoid an error related to glibc when compiling.

albert-github:
@jdpurcell 
Thank you for your clear explanations.

@gmitch215 
It looks to me that the non arm64 platforms should stay at their original Qt version and that for the arm64 either Ubuntu 24.04 with Qt 6.8.* or higher should be used (preferred) or Ubuntu 22.04 with Qt 6.7.* should be used.



(see also https://github.com/actions/runner-images/issues/10636 with the list of supported packages etc.)

gmitch215:
@albert-github @jdpurcell It appears your fork is only behind upstream (so I'm assuming that the official action supports ARM builds).

I also upgraded the ARM runners to `ubuntu-24.04-arm` and used Qt v6.8.

gmitch215:
Should be good now. The CI on my fork is passing.

PR #11470, opened 2025-03-03
bump minimal ci runner version to Ubuntu 22.04

fliiiix:
> We will soon start the deprecation process for Ubuntu 20.04.
> While the image is being deprecated, you may experience longer queue times
> during peak usage hours.
> Deprecation will begin on 2025-02-01 and
> the image will be fully unsupported by 2025-04-01.

https://github.com/actions/runner-images/issues/11101

This servers more as an issue / reminder.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -15,28 +15,28 @@ jobs:
         config:
         - {
             name: "Ubuntu Latest GCC Release",
-            os: ubuntu-20.04,
+            os: ubuntu-22.04,
             build_type: "Release", cc: "gcc", cxx: "g++",
             build_gen: "Unix Makefiles",
             cmake_extra_opts: "-Dbuild_search=YES -Dbuild_app=YES -Dbuild_parse=YES"
           }
         - {
             name: "Ubuntu Latest GCC Debug",
-            os: ubuntu-20.04,
+            os: ubuntu-22.04,
             build_type: "Debug", cc: "gcc", cxx: "g++",
             build_gen: "Unix Makefiles",
             cmake_extra_opts: "-Dbuild_search=YES -Dbuild_app=YES -Dbuild_parse=YES"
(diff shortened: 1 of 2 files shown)


albert-github:
- Looks like there are some problems installing CLang on Ubuntu, probably the versions don't exist anymore (see also https://github.com/actions/runner-images/issues/10636).
- Looks like CLang is also installed when compilation is done with GCC, this is a bit strange, so probably the, old, condition `if: matrix.config.os == 'ubuntu-20.04'` was already a bit to broad.

fliiiix:
I will dig into the current errors and update this MR again :+1:

fliiiix:
Now llvm-14 is used and i made the clang install only for cc clang on ubunut

albert-github:
Looks like the internal structure did change slightly so I now see the error:
```
/home/runner/work/doxygen/doxygen/src/clangparser.cpp:9:10: fatal error: 'clang-c/Index.h' file not found
#include <clang-c/Index.h>
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

fliiiix:
I don't fully understand why that is the case because the file itself should still be provided 

image

albert-github:
My first guess is that there is something incorrect with the include path, maybe a symbolic link is needed.

fliiiix:
maybe removing llvm 13 & 15 is more essential than i thought :shrug:  

```
/usr/lib/llvm-13/include/
/usr/lib/llvm-14/include/
/usr/lib/llvm-15/include/
```

fliiiix:
There might be potential to cleanup something with the update but i would leave that to someone more familiar with the project and don't forget that `[coverity.yml](https://github.com/doxygen/doxygen/pull/11470/files#diff-8c8c0fb40f08ea8c82334c99bd87c646b0442031a1f2c69368aca8a7a6c3050e)` did not run yet

PR #11345, opened 2025-01-17
Add an FAQ about showing static inline functions

duncan-roe:
Much as #11339  except this time it's a new FAQ as you suggested. Do you think it need an example? If so, where are the appendices?

PR #11473, opened 2025-03-04
Add an FAQ about showing static inline functions

duncan-roe:
Same as #11345 except PR is a single commit

### DIFF
--- a/doc/faq.dox
+++ b/doc/faq.dox
@@ -312,6 +312,25 @@ The, technical, reason for this is that the `stdout` can be buffered.
 It is possible to overcome this by means of the `-b` of Doxygen, like e.g `doxygen -b > out.txt 2>&1`.
 Note this might cost a little more time though.
 
+\section faq_staticinline How do I get Doxygen to show static inline functions?
+
+Ordinarily, Doxygen does not show static functions.
+But you have to declare an inline function to be static,
+otherwise it can only be used once.\n
+Meanwhile some static inline functions may be intended only for internal library use.
+The convention here is that internal functions have names starting with 2 underscores.\n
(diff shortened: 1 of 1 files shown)


doxygen:
@duncan-roe Sorry for reacting late, but I'm not really sure which problem you are addressing. Is this for C code? An inline function in C++ does not need to be static. I don't think this is a _frequently_ asked question (I don't remember it being asked before).

duncan-roe:
Yes for C. An inline function _declared in a header_ has to be static else **ld** complains (the 2nd time the header is included IIRC). `static inline` is absolutely the normal way to declare public  inline functions - see for example the Linux source tree.
I do think it's worth documenting how to deal with `static inline` functions.
In #11339 I had this documentation as a **config** example, but @albert-github suggested to make it an FAQ.

sasha2048:
For context (I'm posting this just because, as a C++ developer, I was surprised about this C behavior):
* ["C99" section in "Inline (C and C++)" Wikipedia article](https://en.wikipedia.org/wiki/Inline_(C_and_C%2B%2B)#C99)
* ["GNU inline vs. C99 inline" by Josef “Jeff” Sipek](https://blahg.josefsipek.net/?p=529)
* ["Different semantics for inline functions" subsection in "Porting to GCC 5"](https://gcc.gnu.org/gcc-5/porting_to.html#inline)
* ["inline function specifier (C documentation)" on cppreference.com](https://en.cppreference.com/w/c/language/inline.html)
* [Answer on Reddit on why adding just inline may be not enough](https://www.reddit.com/r/C_Programming/comments/lkjryl/comment/gnk8sj5/)

PR #11477, opened 2025-03-05
issue #11469 Add support for native plantuml

### LINKED ISSUE #11469: Using native-image instead of jar
Is there a way to use a native-image https://github.com/plantuml/plantuml/releases/tag/v1.2025.1-native instead of a jar file.

From digging through the code this does not seem possible.
https://github.com/doxygen/doxygen/blob/master/src/plantuml.cpp#L206

fliiiix:
I took a stab at supporting the native plantuml in doxygen.
Let me know what you think about that.

(Motivation for this is that i work in a build container where i don't have access to java and the plantuml.jar)

### DIFF
--- a/doc/commands.dox
+++ b/doc/commands.dox
@@ -3709,6 +3709,7 @@ class Receiver
   The location of the PlantUML file should be specified using
   \ref cfg_plantuml_jar_path "PLANTUML_JAR_PATH". The other jar files should also reside
   in this directory.
+  Instead of a jar file a binary capable to convert PlanUML diagrams can be provided, usually a PlantUML native image.
   \note The use of the `<engine>` `ditaa` is not possible in \LaTeX as PlantUML only
   supports the `png` format and Doxygen requires, temporary, `eps` output.
 
--- a/src/config.xml
+++ b/src/config.xml
@@ -4216,6 +4216,7 @@ to be found in the default search path.
 <![CDATA[
  When using PlantUML, the \c PLANTUML_JAR_PATH tag should be used to specify the path where
  java can find the \c plantuml.jar file or to the filename of \c jar file to be used.
+ Instead of a jar file a binary capable to convert PlanUML diagrams can be provided, usually a PlantUML native image.
  If left blank, it is assumed PlantUML is not used or
  called during a preprocessing step. Doxygen will generate a warning when it encounters a
  \ref cmdstartuml "\\startuml" command in this case and will not generate output for the diagram.
--- a/src/plantuml.cpp
+++ b/src/plantuml.cpp
@@ -253,10 +253,18 @@ static void runPlantumlContent(const PlantumlManager::FilesMap &plantumlFiles,
                test_doxygen/DOXYGEN_OUTPUT/html/A
    */
   int exitCode = 0;
-  QCString plantumlJarPath = Config_getString(PLANTUML_JAR_PATH);
+  QCString plantumlPath = Config_getString(PLANTUML_JAR_PATH);
   QCString plantumlConfigFile = Config_getString(PLANTUML_CFG_FILE);
 
-  QCString pumlExe = "java";
+  FileInfo plantumlPathFile(plantumlPath.str());
+  bool isJarPlantUML = plantumlPathFile.extension(false) == "jar";
+
+  QCString pumlExe = "";
+  if(isJarPlantUML)
+  {
+    pumlExe = "java";
+  }
+
   QCString pumlArgs = "";
   QCString pumlType = "";
   QCString pumlOutDir = "";
@@ -278,7 +286,11 @@ static void runPlantumlContent(const PlantumlManager::FilesMap &plantumlFiles,
     }
   }
   if (!pumlIncludePathList.empty()) pumlArgs += "\" ";
-  pumlArgs += "-Djava.awt.headless=true -jar \""+plantumlJarPath+"\" ";
+  pumlArgs += "-Djava.awt.headless=true ";
+  if(isJarPlantUML)
+  {
+    pumlArgs += "-jar \""+plantumlPath+"\" ";
+  }
   if (!plantumlConfigFile.isEmpty())
   {
     pumlArgs += "-config \"";
@@ -356,8 +368,14 @@ static void runPlantumlContent(const PlantumlManager::FilesMap &plantumlFiles,
 
       if ((exitCode=Portable::system(pumlExe.data(),pumlArguments.data(),true))!=0)
       {
-        err_full(nb.srcFile,nb.srcLine,"Problems running PlantUML. Verify that the command 'java -jar \"{}\" -h' works from the command line. Exit code: {}.",
-            plantumlJarPath,exitCode);
+        QCString command = plantumlPath;
+        if(isJarPlantUML)
+        {
+          command = "java -jar \"" + plantumlPath + "\" -h";
+        }
+
+        err_full(nb.srcFile,nb.srcLine,"Problems running PlantUML. Verify that the command '{}' works from the command line. Exit code: {}.",
+            command,exitCode);
       }
 
       if ( (format==PlantumlManager::PUML_EPS) && (Config_getBool(USE_PDFLATEX)) )
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -4532,7 +4532,7 @@ Token DocPara::handleCommand(char cmdChar, const QCString &cmdName)
       break;
     case CommandType::CMD_STARTUML:
       {
-        QCString jarPath = Config_getString(PLANTUML_JAR_PATH);
+        QCString plantumlPath = Config_getString(PLANTUML_JAR_PATH);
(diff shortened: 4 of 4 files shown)


doc_internal/tags_history.md
```diff
@@ -441,6 +441,8 @@ The following table gives an overview of the doxygen configuration settings and
 <dt></dt>              <dd>SQLITE3_RECREATE_DB</dd>
 <dt>New in 1.8.20</dt>
 <dt></dt>              <dd>PYTHON_DOCSTRING</dd>
+<dt>New in 1.14.0</dt>
```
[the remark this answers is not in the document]
albert-github: This should not be here. For the version 1.9.0 and newer it is done automatically!

albert-github (commented):
I didn't really review it as currently the native image does not work, on windows, so I cannot test it.
I made one remark that at least has to be corrected.

doc_internal/tags_history.md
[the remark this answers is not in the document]
fliiiix: ahh nice removed :+1:

albert-github:
I see the usage of the docker image as a workaround and I think it would be better to wait for the Plantuml people to get the Plantuml native image working (I already made a comment in https://github.com/plantuml/plantuml/issues/1344so hopefully there will be a follow up soon there).

Due to the GitHub "intelligence" on the chosen words in the PR text the issue https://github.com/doxygen/doxygen/issues/11469 will be automatically be closed as can be seen in right hand side in the browser):

image

this is not how doxygen works (doxygen will close, automatically, issues that are integrated at the moment of a release, so please remove this mentioning.

src/config.xml
```diff
@@ -4151,9 +4151,21 @@ to be found in the default search path.
...
  When using PlantUML, the \c PLANTUML_JAR_PATH tag should be used to specify the path where
  java can find the \c plantuml.jar file or to the filename of \c jar file to be used.
+ See \c PLANTUML_BIN_PATH if you use a native image.
  If left blank, it is assumed PlantUML is not used or
  called during a preprocessing step. Doxygen will generate a warning when it encounters a
  \ref cmdstartuml "\\startuml" command in this case and will not generate output for the diagram.
+]]>
+      </docs>
+    </option>
+    <option type='string' id='PLANTUML_BIN_PATH' format='file' defval=''>
```
doxygen, line 4161: Instead of a yet another option, which in this case seems to be mutual exclusive to `PLANTUML_JAR_PATH`, why not allow `PLANTUML_JAR_PATH` to optionally point to a non-jar file and then treat that as if `PLANTUML_BIN_PATH` were used. Would that work?

fliiiix: Hey i think that would absolutely work these where kind of the 2 options i saw to implement this and since the discussion in the issue did not uncover a preference i just went for one
image

The downside of reusing the `PLANTUML_JAR_PATH` in my opinion are that this will be more convoluted to configure and maybe a bit harder to figure out. But as you said an other option as implemented here is also not without its downsides.

So if you think we should go for a reuse/extend of the option `PLANTUML_JAR_PATH` i can absolutely update the MR to do that just let me know how you would like to see this work :)

fliiiix: updated & rebased to master to reuse the existing config flag instead of introducing a new flag @doxygen @albert-github

fliiiix:
So this is blocked right now because the native image does not work on windows? @albert-github @doxygen ?

albert-github:
I'm not sure the standalone image does work on other platforms as I cannot test it, but yes it blocked for the time being.

Maybe you could try to push people behind https://github.com/plantuml/plantuml/issues/1344 (or plantuml) a bit as well.

fliiiix:
@albert-github can we move forward with this MR?
Basic umls seem to work now which should be good enough to test the integration into doxygen and then hopefully at some point they get the native image working better for all diagrams.

albert-github:
That is my intention and at the moment I'm just checking out the problems in respect to missing diagram types in doxygen (<s>#12045, #12046, #12047, #12048</s> #12049) and the problems with the full and headless native executables (see https://github.com/plantuml/plantuml/issues/1344#issuecomment-4074363553 and in the mean time I found some more ....).
Furthermore I saw some inconsistencies in plantuml (https://github.com/plantuml/plantuml/issues/2597, https://github.com/plantuml/plantuml/issues/2598, https://github.com/plantuml/plantuml/issues/2599 and https://github.com/plantuml/plantuml/issues/2602).

I will have a look at this PR and see what is still needed.

doc/commands.dox
```diff
@@ -3522,6 +3522,8 @@ class Receiver
   The location of the PlantUML file should be specified using
   \ref cfg_plantuml_jar_path "PLANTUML_JAR_PATH". The other jar files should also reside
   in this directory.
+  Instead of a jar file a binary capable to convert PlanUML diagrams can be provided.
+  Usually a PlantUML native image.
```
albert-github:
```suggestion
  Instead of a jar file a binary capable to convert PlanUML diagrams can be provided, usually a PlantUML native image.
```

src/config.xml
```diff
@@ -4150,6 +4150,8 @@ to be found in the default search path.
 <![CDATA[
  When using PlantUML, the \c PLANTUML_JAR_PATH tag should be used to specify the path where
  java can find the \c plantuml.jar file or to the filename of \c jar file to be used.
+ Instead of a jar file a binary capable to convert PlanUML diagrams can be provided.
+ Usually a PlantUML native image.
```
albert-github:
```suggestion
 Instead of a jar file a binary capable to convert PlanUML diagrams can be provided, usually a PlantUML native image.
```

src/plantuml.cpp
```diff
@@ -203,7 +211,11 @@ static void runPlantumlContent(const PlantumlManager::FilesMap &plantumlFiles,
     }
   }
   if (!pumlIncludePathList.empty()) pumlArgs += "\" ";
-  pumlArgs += "-Djava.awt.headless=true -jar \""+plantumlJarPath+"\" ";
+  pumlArgs += "-Djava.awt.headless=true ";
+  if(isJarPlantUML)
+  {
+    "-jar \""+plantumlPath+"\" ";
```
albert-github:
```suggestion
    pumlArgs += "-jar \""+plantumlPath+"\" ";
```

fliiiix:
thanks for the review adapted the changes and rebased on the lastest master :sparkles:

albert-github:
Looks like that the macOS builds on GitHub have a bit of a problem (see this also with our other GitHub Actions builds).

albert-github:
I think this PR looks OK now (and we should give it a try). Plantuml still has some problems with the native image for a number of diagrams but the main diagrams seem to work.

fliiiix:
I tested this on my project with a custom build based on 564450be425b02b294aff3fc842a339762c53894 and this patch and at least for all the basic diagrams i used this worked

fliiiix:
Can i do something to make the macos actions pass?

albert-github:
> Can i do something to make the macos actions pass?

What do you mean by "macros"? (maybe a small exaple?)
There are the settings  `PLANTUML_CFG_FILE` and `PLANTUML_INCLUDE_PATH` do they give a solution?

fliiiix:
Im talking about the failed macOS jobs here in gitlab :see_no_evil: 
image

albert-github:
Oops I read "macros" instead of "macos".

With the "macos" CI builds there have been some small problems (as far as I can tell outside the scope of doxygen) last week. looks like they work again so I restarted the 2 macos builds.

fliiiix:
No worries i learned that the words macros and macos are very similar :D 

Is there anything else i need to do before this can be merged?

fliiiix:
@albert-github anything missing here before it can get merged?

albert-github:
On June 3, 2026 I got a message with https://github.com/plantuml/plantuml/issues/2610 with the request to retest the native image problems and I wrote:
> I will have a go at it, but you actually mean to retest the problems mentioned in [#](https://github.com/plantuml/plantuml/issues/)1344 I assume as I don't see essential changes in the BUILDING.md in building plantuml native myself (but when the native build 1.2026.5 is OK this is not an issue anymore).

So some progress looks like to be in the pipeline...

fliiiix:
So this can not be merged before all the plantuml bugs are fixed? (Because i think this is already useful before this)

albert-github:
No that is not the case but the most important ones should be fixed so we have a consistent usage between the native image and the jar file implementations.

fliiiix:
This is now solved by allowing to set `PLANTUML_TOOL` with https://github.com/doxygen/doxygen/pull/12301

albert-github:
Please don't close this issue, this will be automatically done when #12301 is released i.e. with the following doxygen release

fliiiix:
@albert-github so this `(see issue https://github.com/doxygen/doxygen/issues/11469)` is enough to close it on release?  Also you reopened the MR and not the issue :eyes:

albert-github:
Oops you are correct, it is OK that the PR is closed as this now is not necessary anymore the text was indeed intended for the issue #11469

PR #11482, opened 2025-03-06
Consistency of use of `HIDE_COMPOUND_REFERENCE`

albert-github:
In the question https://stackoverflow.com/questions/79483301/doxygen-hide-compound-reference-for-namespace it was noted that the `HIDE_COMPOUND_REFERENCE` does not work for e.g. namespaces. Currently it only works for classes (and also for Struct,  Union, Interface, Protocol, Category and Exception, as they all use the routine `trCompoundReference`) Besides the problem for namespaces ithe issue is present for other parts as well that call the routines:
- `trFileReference`
- `trNamespaceReference`
- `trDirReference`
- `trCompoundReferenceFortran`
- `trModuleReference`
- `trEnumReference`
- `trConstantGroupReference`
- `trServiceReference`
- `trSingletonReference`
- `trCustomReference`
- `trCompoundReferenceSlice`
- `trConceptReference`

The usage of `HIDE_COMPOUND_REFERENCE` has ben implemented for these calls as well

### DIFF
--- a/src/classdef.cpp
+++ b/src/classdef.cpp
@@ -2887,31 +2887,73 @@ QCString ClassDefImpl::title() const
 
   if (lang==SrcLangExt::Fortran)
   {
-    pageTitle = theTranslator->trCompoundReferenceFortran(displayName(),
-              m_compType,
-              !m_tempArgs.empty());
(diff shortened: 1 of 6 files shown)


IsaacBlancICL:
Just tested this and it has resolved the issue. Thank you!

PR #11494, opened 2025-03-14
Flexibility of `\cite` command

albert-github:
Adding some options to the `\cite` command to make it more flexible so that not only a reference number can be shown.

Example: [example.tar.gz](https://github.com/user-attachments/files/19249528/example.tar.gz)


doxygen:
@albert-github I added a [type-safe wrapper](https://github.com/doxygen/doxygen/commit/6de442232cd7741d38591a943004b6cef557ac57) around the bitfield, like was also done for other types in the past.

Issue #11492, opened 2025-03-13
@plantumlfile command not recognized


ferran-bofill-elausa:
**Describe the bug**
When using the `\plantumlfile` command, the HTML output is generated without errors. However, when opening the generated HTML file, which should contain the PlantUML diagram, the `\plantumlfile` command and its attributes appear as plain text. It seems that `@plantumlfile` is not being recognized and is instead treated as regular text.

**Screenshots**

![Image](https://github.com/user-attachments/assets/ecde630a-5922-4b29-83aa-6653a5970868)

**To Reproduce**
Attached an example project that reproduces this issue. The project contains only three files:

-**diagram.puml** – Contains the PlantUML diagram referenced by the `@plantumlfile` command.
-**Doxyfile** – The Doxygen configuration file.
-**example.dox** – Contains the `@plantumlfile` command, which is not working as expected, along with an embedded PlantUML diagram using `@startuml` and `@enduml` to verify that PlantUML is functioning correctly.

To test the issue, open the provided example project, adjust the Doxyfile configuration as needed based on your system’s file paths, and then run Doxygen.

![Image](https://github.com/user-attachments/assets/a8eb7538-a514-439a-a375-c8b5afca461f)

Also attached the generated HTML output, which demonstrates the issue.

[example.zip](https://github.com/user-attachments/files/19232047/example.zip)

**Version**
Doxygen version used: 1.9.1

ferran-bofill-elausa:
I also tested the same example project with Doxygen version 1.13.2, built from source. The @plantumlfile command no longer appears as plain text, but the diagram still doesn't appear.

albert-github:
In the puml file we see:
```
@startuml example_diagram
Bob -> Alice : hello
@enduml
```
and apparently this translates to the filename `inline_umlgraph_1.svg example_diagram.svg` although on the HTML code we see:  `inline_umlgraph_1.svg` so looks like a mix-up of the output filename. Quick solution would be to remove the `example_diagram` from the puml file.

Note that it doesn't work in doxygen version 1.9.1 is logical as the `\plantumlfile` command was introduced in doxygen version 1.13.0

albert-github:
I've just pushed a proposed patch, pull request #11500

albert-github:
Adjusted code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
[...]

PR #11500, opened 2025-03-17
issue #11492 `@plantumlfile` command not recognized

albert-github:
- The original problem is about the fact that fact that the `@startuml` command had a filename and this was not expected / anticipated, the result was a concatenation of the name with and "inline" name (i.e. an generated dummy name) and the given name
- Potential problem with handing out the same number for an "inline" file name (multithreading, now protected by means of a mutex.
- With a `\plantumlfile` it is possible to have multiple diagrams in one file, only one was shown, now all are shown.

### DIFF
--- a/src/plantuml.cpp
+++ b/src/plantuml.cpp
@@ -13,6 +13,7 @@
  *
  */
 
+#include <mutex>
 #include "plantuml.h"
 #include "util.h"
 #include "portable.h"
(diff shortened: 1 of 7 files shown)


src/plantuml.cpp
```diff
@@ -95,34 +73,153 @@ QCString PlantumlManager::writePlantUMLSource(const QCString &outDirArg,const QC
         case '\t': break;
         case ' ':  break;
         case '@':
-          if (initial && literal_at(p,"start")) // @start...
+          if (!insideComment && literal_at(p,"start")) // @start...
```
doxygen, line 76: I don't really understand the use-case of this inline parser stuff. Do you have an example?

src/plantuml.cpp
```diff
@@ -95,34 +73,153 @@ QCString PlantumlManager::writePlantUMLSource(const QCString &outDirArg,const QC
...
-  Debug::print(Debug::Plantuml,0,"*** writePlantUMLSource generateType: {}\n",generateType);
+void PlantumlManager::generatePlantUmlFileNames(const QCString &fileName,OutputFormat format,const QCString &outDir,
+                                                QCString &baseName,QCString &puName,QCString &imgName)
+{
+  static int umlindex=1;
+
+  if (fileName.isEmpty()) // generate name
+  {
+    std::lock_guard<std::mutex> lock(g_PlantUmlMutex);
+    puName = "inline_umlgraph_"+QCString().setNum(umlindex);
```
doxygen, line 199: Using numbers like this will give instable output, i.e. when using multiple threads every run the numbers can be assigned to different graphs, and thus the output will change even if the input did not. A typical solution is to use a hash of the filename + line number, representing where the graph is found in the input, and use that to create the output name.

src/plantuml.cpp
albert-github: The example: [example.tar.gz](https://github.com/user-attachments/files/19525459/example.tar.gz)

src/plantuml.cpp
albert-github: The original was already using the non stable numbers and this could lead to multiple use of the same number (it is unlikely that this happens, but better to be safe hence the mutex). It indeed would be better to use a hash.

albert-github:
At first glance I don't see any problems.

Issue #11001, opened 2024-07-15
\copydoc does not respect section hierarchy in table of contents


DCDEIE:
**Describe the bug**

H1/H2/H3 (section/subsection/subsubsection) headings in text that are copied to another file by means of a `\copydoc` tag do not port over the section headings hierarchy in the copydoc text to the table of contents.

**Screenshots**

<img width="828" alt="dox_issue_copydoc" src="https://github.com/user-attachments/assets/5b276e75-fc6e-4b3a-85c4-95b25e8cd58f">

**To Reproduce**

[test_copydoc_toc.zip](https://github.com/user-attachments/files/16233506/test_copydoc_toc.zip)

**Expected behavior**

While all text using `\copydoc` is correctly inserted from the copydoc source into the target text inline, it would be nice if the table of contents also recognised all sections in the source copydoc text and correctly listed them in the table of contents hierarchy in the target for the copydoc text. As you can see on the page **Target 1 heading 1**, the headings **Copydoc H1/H2/H3** correctly appear inline with text on the page, but are not listed in the table of contents hierarchy to the left under **Copydoc target page**.

**Version**

Doxygen v1.10
Windows 10, 64-bit

albert-github:
The problem here is the moment evaluation of the `\copydoc`, the interpretation of `\tableofcontents` and  the determination of when the available sections. The determination of the available sections is done before the evaluation of the `\copydoc` and hence sections are not known to the `\tableofcontents`.
I think you should use the `\snippet{doc}` command here and also use the options `prefix` to prevent double label names and maybe `raise` to get everything at the right level .
(don't forget to set the `EXAMPLE_PATH`).

I did a small test with: [example.tar.gz](https://github.com/user-attachments/files/16236371/example.tar.gz)
which gives:
![image](https://github.com/user-attachments/assets/f810c62e-ed8b-421b-813d-2ccb26cd44e6)

So I think the problem is usage.

DCDEIE:
I thought it might be helpful to add there that the syntax in your sample files for the snippet does not work in v1.10 (the version I use). I needed to modify it from...

\snippet{doc,raise=1,prefix=fn} copydoc_source_common_doc_for_reuse.dox COPYDOC_SOURCE

...to...

\snippetdoc {raise=1,prefix=fn} copydoc_source_common_doc_for_reuse.dox COPYDOC_SOURCE

And many thanks for the test files.

albert-github:
@DCDEIE 
[...]

PR #11529, opened 2025-04-03
issue #11001 \copydoc does not respect section hierarchy in table of contents

albert-github:
When linking to pages / sections prefer the version with the "prefix" (i.e. local version) but when no match is found also try the one without "prefix" (i.e. global version).


DCDEIE:
I tested out this fix and can confirm that it resolved my issue. Is there any estimate when this version will be officially released? Many thanks for the fast turnaround 🥇

PR #11533, opened 2025-04-06
fix: fix regression of #11390

e-kwsm:


albert-github:
The file caseconvert.py is nowhere used and has only be used to generate caseconvert.h in the past, so that is why it was not detected.

Issue #11528, opened 2025-04-02
\example{lineno} does not generate the example


nekenyu:
**Describe the bug**
Per: https://www.doxygen.nl/manual/commands.html#cmdexample

The following comment does not work: \example{lineno} exampleFile.cpp 
The following comment does: \example exampleFile.cpp

**To Reproduce**
Given an existing exampleFile.cpp available from the EXAMPLE_PATH:

I should be able to write: \example{lineno} exampleFile.cpp
1. I expect the Examples page to include: exampleFile.cpp ; it does not.
2. I expect a generated html file for exampleFile.cpp with line numbers; it is not.
3. I expect the following to work: \ref exampleFile.cpp ; I get warning: unable to resolve reference to 'exampleFile.cpp' for \ref command.

When I write: \example exampleFile.cpp
1. I expect the Examples page to include: exampleFile.cpp ; it does.
2. I expect a generated html file for exampleFile.cpp without line numbers; it is generated without line numbers.
3. I expect the following to work: \ref exampleFile.cpp ; it does.

Side note: Both use cases above exist between two other "/example ...cpp" commands without the lineno, which work both times.

**Expected behavior**
See: To Reproduce.

**Version**
`doxygen -v` reports: `1.9.8` on Windows11 WSL: running Ubuntu 24.04.2 LTS (installed via `su apt install doxygen`.

**Stack trace**
n/a

**Additional context**
n/a

albert-github:
Doxygen 1.9.8 is from August 25 2023, the current doxygen version is 1.13.2

To be able to reproduce your problem:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem?
   Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

nekenyu:
Presently I can only recreate it in the repository where I found it, not a test repository.

I generated a new Doxyfile, put sample code and doxygen documentation around it. The command worked. in the sample test case, it worked. I reduced the differences to my makefile to PROJECT_NAME and PROJECT_BREIF.

albert-github:
- best it to try to reduce the number of INPUT files and see which file(s) cause(s) the problem. (Note also pay attention to the emitted warnings)
- is it a proprietary repository / code or can we find it somewhere

nekenyu:
I have:
    WARN_AS_ERROR          = FAIL_ON_WARNINGS_PRINT
[...]

PR #11534, opened 2025-04-06
Warning ` mismatch `#if` / `#endif` from preprocessor

albert-github:
When an included file is not terminated by means of a `\n` it is possible that we get a warning like:
```
warning: More #if's than #endif's found (might be in an included file).
```
adding the `\n`, as already done with files that are "normally" preprocessed , solves the problem

Example: [example.tar.gz](https://github.com/user-attachments/files/19622658/example.tar.gz)

(Found as a side effect of #11528)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -2259,6 +2259,7 @@ static std::unique_ptr<FileState> checkAndOpenFile(yyscan_t yyscanner,const QCSt
     }
     else
     {
+      addTerminalCharIfMissing(fs->fileBuf,'\n');
       fs->oldFileBuf    = state->inputBuf;
       fs->oldFileBufPos = state->inputBufPos;
     }


nekenyu:
TLDR: I would recommend the current proposed change **and** issue an optional (default on) warning when the last line trailing linefeed is not detected. This, of course, involves another option to override this

Sorry, I wanted to think-out this issue.... And, and allow the conclusion to be argued....

Long Answer:

I understand that POSIX requires then trailing newline. Clearly, at least one compiler does not. Clearly my IDE (which shall remain nameless) -- from which this issue was first noticed in https://github.com/doxygen/doxygen/issues/11528 -- does not do that.

Given the unique nature of doxygen (processing multiple language source codes for documentation comments), one would expect doxygen to work if the compiler does. Reality is, it might not always.

Also, this raises the question of any other tests having this issue.

There are a few obvious solutions:
- back out the change that required the trailing new line 
- internally, doxygen can add the new line if needed -- that is the proposed solution
- update the error message message to suggest this most obvious problem
- add a separate warning checker that explicitly states the lack of trailing new line

Obviously, it would be good not to back it out. The proposed change solves the immediate problem, and should solve the larger issue of other testers. Updating all error checkers message for the new line is too much work. But, if doxygen is going to enter into the mess of the last line trailing newline -- which some original change did -- then it is probably best to add a separate warning that can indicates features and other tools may ignore such a last line.

I could also see a case where someone wants you to not silently add it... Maybe to emulate how another tool would behave? But, I would suggest they are better off with the warning to know other toosl might have a problem... It should not harm to add the trailing newline only within doxygen processing.

I would recommend the current proposed change **and** issue an optional (default on) warning when the last line trailing linefeed is not detected. This, of course, involves another option to override this

albert-github:
Nice thoughts.

The original file is, apparently, a valid file when the trailing `\n` is missing.
Doxygen does not change the original file, it just adds it when missing in the internal buffer, so doxygen can process the file (it might be that e.g. compilers do the same as there is no message when you compile your code) In case of doxygen it was just missing for the included files.

Adding another setting might confuse the user more (and doxygen will still need to add the `\n` internally for proper processing)

PR #11537, opened 2025-04-07
use c++20 as the standard for the project

dpronin:


albert-github:
At the moment doxygen does not use C++ 20 features, so I don't think there is any need to change the minimum version (and it might result in that some people cannot build doxygen anymore without changes to their system).
- What would be the advantage?

dpronin:
The idea is to use char8_t that is a part of c++20. Code, being compiled with modern clang and libc++, stops compiling due to not supporting std::basic_string<unsigned char> any longer at least in libc++. If you want to go along the time you will definitely sooner or later come to this decision. My business has merely been to offer
I am requesting to merge into main branch, I am not insisting to propagate this change to older versions, so why do we care about those people who might like to compile modern doxygen with gcc-3?

kaorihinata:
This is required (in some form) to actually finish fixing #10928 and fix compilation for newer versions of macOS and the Command Line Tools package. The change in #10928 updated the code to fix a missing usage of `JAVACC_CHAR_TYPE`, but didn't add the plumbing in `CMakeLists.txt` to change it when necessary for Clang/libc++.

If we don't want to change it universally, we can always just make it a Clang dependency:

```patch
--- CMakeLists.txt.orig 2025-04-09 05:34:53
+++ CMakeLists.txt      2025-04-09 05:37:18
@@ -82,7 +82,11 @@
 endif()

 # use C++17 standard for compiling
-set(CMAKE_CXX_STANDARD 17)
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    set(CMAKE_CXX_STANDARD 20)
+else()
+    set(CMAKE_CXX_STANDARD 17)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS ON)

@@ -170,7 +174,11 @@
 endif()

 # needed for JavaCC
-set(JAVA_CC_EXTRA_FLAGS "-DJAVACC_CHAR_TYPE=\"unsigned char\"")
+if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    set(JAVA_CC_EXTRA_FLAGS "-DJAVACC_CHAR_TYPE=\"char8_t\"")
+else()
+    set(JAVA_CC_EXTRA_FLAGS "-DJAVACC_CHAR_TYPE=\"unsigned char\"")
+endif()
 set(CMAKE_CXX_FLAGS       "${CMAKE_CXX_FLAGS}       ${JAVA_CC_EXTRA_FLAGS}")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${JAVA_CC_EXTRA_FLAGS}")

```

dpronin:
OK, I would agree if we arranged about this trade-off

doxygen:
@kaorihinata it is not just a clang version, it is clang 19 and later that has the problem. If we make a cmake switch it should still use C++17 for older clang versions, to avoid issues such as https://github.com/doxygen/doxygen/actions/runs/14307727795/job/40843054842?pr=11537

kaorihinata:
@doxygen Unfortunately, it's not just Clang 19 as far as Apple Clang goes. Xcode 6.3 and the current release of Command Line Tools currently ship a build of 17 which exhibits the behavior described:

```
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: arm64-apple-darwin24.5.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

There are still some supported versions of macOS though that use Xcode 6.2 and below and ship Apple Clang 16 which does not require a patch, so I guess the border for a system that identifies its compiler as AppleClang would be 17. Does a check for Clang 19 and AppleClang 17 sound like a more preferable solution?

kaorihinata:
This seems to work, though it uses `VERSION_GREATER_EQUAL` which requires CMake 3.7+. Could always use `VERSION_GREATER` with a silly high minor/patch version if you'd prefer.

```patch
@@ -22,7 +22,7 @@
 option(build_search    "Build external search tools (doxysearch and doxyindexer)" OFF)
 option(build_doc       "Build user manual (HTML and PDF)" OFF)
 option(build_doc_chm   "Build user manual (CHM)" OFF)
-if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     option(use_libc++  "Use libc++ as C++ standard library." ON)
 endif()
 option(use_libclang    "Add support for libclang parsing." OFF)
@@ -81,8 +81,19 @@
     endif()
 endif()

-# use C++17 standard for compiling
-set(CMAKE_CXX_STANDARD 17)
+# use C++17 standard for compiling (unless using >= AppleClang 17 or >=
+# Clang 19)
+if (
+    (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
+     CMAKE_CXX_COMPILER_VERSION VERSION_LESS 17)
+    OR
+    (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
+     CMAKE_CXX_COMPILER_VERSION VERSION_LESS 19)
+)
+    set(CMAKE_CXX_STANDARD 17)
+else()
+    set(CMAKE_CXX_STANDARD 20)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS ON)

@@ -170,7 +181,11 @@
 endif()

 # needed for JavaCC
-set(JAVA_CC_EXTRA_FLAGS "-DJAVACC_CHAR_TYPE=\"unsigned char\"")
+if (CMAKE_CXX_STANDARD EQUAL 20)
+    set(JAVA_CC_EXTRA_FLAGS "-DJAVACC_CHAR_TYPE=\"char8_t\"")
+else()
+    set(JAVA_CC_EXTRA_FLAGS "-DJAVACC_CHAR_TYPE=\"unsigned char\"")
+endif()
 set(CMAKE_CXX_FLAGS       "${CMAKE_CXX_FLAGS}       ${JAVA_CC_EXTRA_FLAGS}")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${JAVA_CC_EXTRA_FLAGS}")

```

doxygen:
@kaorihinata looks like a good solution. Doxygen already requires cmake 3.14+.
@dpronin can you include these changes in the PR?

dpronin:
@doxygen done

PR #11546, opened 2025-04-18
Update doxygen.css to fix validation issue (II)

KaruroChori:
It is unclear what `text-wrap: unrestricted` is supposed to do, as it is not standard and I cannot find any documentation but this page: http://www.css-infos.neolao.com/property/text-wrap

I replaced it with inherit to hopefully neuter it out.

https://jigsaw.w3.org/css-validator/validator?uri=https%3A%2F%2Fdocs.opencv.org%2F4.10.0%2Fd4%2Fd17%2Fnamespacecv_1_1aruco.html&profile=css3svg&usermedium=all&warning=1&vextwarning=&lang=en

### DIFF
--- a/doc/doxygen_manual_chm.css
+++ b/doc/doxygen_manual_chm.css
@@ -155,10 +155,11 @@ dl.el {
 
 div.line {
 	font-family: 'JetBrains Mono', Consolas, Monaco, 'Andale Mono', 'Ubuntu Mono', monospace, fixed;
-        font-size: 13px;
+    font-size: 13px;
 	min-height: 13px;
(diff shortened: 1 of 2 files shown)


doxygen:
Seems like this was once in the [CSS3 standard](https://www.w3.org/TR/2011/WD-css3-text-20110215/#text-wrap) but has been dropped in the meantime.


Maybe 
```css
text-wrap: wrap; 
word-break: break-all;
``` 
is a good replacement.

KaruroChori:
I think those failures are unrelated. I got several failing workflows over the past few hours in unrelated projects.

Issue #11549, opened 2025-04-21
Leftover variable in CSS (validation and possible logic issue)


KaruroChori:
https://github.com/doxygen/doxygen/blob/c941da77e4ca6e2ba932d7736b5ef0d9c21aeeab/templates/html/navtree.css#L89

This is probably a leftover from some templating system which was in place before CSS variables.  
I cannot tell for sure what this width should be, as there is not much context there for me to say.

KaruroChori:
I isolated the point it was introduced and where it was being replaced:
https://github.com/doxygen/doxygen/blob/ceb4115c7b941039411e1793e01239610ff112a2/src/ftvhelp.cpp#L1299

albert-github:
I think the variable is indeed a leftover and the `width` is now handled by means of the `navtree.js`, so the entry can be removed from the `navtree.css`.

albert-github:
I've just pushed a proposed patch, pull request #11550

KaruroChori:
Too fast :D. I was worried about support without JS. I would have kept `width` with the default value of 250px.

albert-github:
I don't think it is possible to have doxygen documentation without JS, so should not be a problem.

KaruroChori:
Is it so? I am often accessing doxygen documentation on terminal browsers or dillo which have no js support. 
From what I can tell, it can be navigated quite well. Was this no-js support removed recently or never there and just a happy accident?

albert-github:
Probably a happy accident.

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.14.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #11550, opened 2025-04-23
issue #11549 Leftover variable in CSS (validation and possible logic issue)

albert-github:
- `width: $width` is handled in navtree.js
- css files are not handled by means of `copyResource...` anymore
- handling of the contents of Luminance files is done solely in css files now, so no separate Luminance template files anymore

### DIFF
--- a/src/resourcemgr.cpp
+++ b/src/resourcemgr.cpp
@@ -99,62 +99,6 @@ bool ResourceMgr::copyResourceAs(const QCString &name,const QCString &targetDir,
           }
         }
         break;
-      case Resource::Luminance:
-        {
-          QCString n = name;
-          n = n.left(n.length()-4)+".png"; // replace .lum by .png
-          const uint8_t *data = res->data;
-          uint16_t width   = (data[0]<<8)+data[1];
-          uint16_t height  = (data[2]<<8)+data[3];
-          ColoredImgDataItem images[2];
-          images[0].name    = n.data();
-          images[0].width   = width;
-          images[0].height  = height;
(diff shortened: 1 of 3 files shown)


src/resourcemgr.cpp
```diff
@@ -99,62 +99,6 @@ bool ResourceMgr::copyResourceAs(const QCString &name,const QCString &targetDir,
...
-          images[0].width   = width;
-          images[0].height  = height;
-          images[0].content = &data[4];
-          images[0].alpha   = &data[4+width*height];
-          images[1].name    = nullptr; // terminator
-          writeColoredImgData(targetDir,images);
-          return TRUE;
-        }
-        break;
-      case Resource::CSS:
```
doxygen, line 138: This case was still unused (got a compiler warning for it), so I reintroduced it (without the `navtree.css` specifics).

albert-github: I see in the GitHub Actions indeed a warning:
```
home/runner/work/doxygen/doxygen/src/resourcemgr.cpp:85:12: warning: enumeration value ‘CSS’ not handled in switch [-Wswitch]
   85 |     switch (res->type)
      |       
```
Though, as far as I can tell, the relevant code is never called. The `CSS` is used in the setting `initResources` but in the routine `getAsString` it is not used at all. 
We probably need another follow up PR to handle this.

albert-github: > We probably need another follow up PR to handle this.

Will be hard as the `type` is used at the above mentioned line 85 in the switch statement.

Issue #8731, opened 2021-08-21
grouping variables requires an empty line before closing @}


thom-sch:
**Describe the bug**
grouping variables requires an empty line before closing @} or an additional \brief
If not the description of last variable erroneously appears as detailed description.

**To Reproduce**
See attached zip-file:
[Test16_grouping_variables.zip](https://github.com/doxygen/doxygen/files/7026014/Test16_grouping_variables.zip)

**Version**
Windows 10/64
Doxygen version used: 1.9.2 (caa4e3de211fbbef2c3adf58a6bd4c86d0eb7cb8)
but also observed in V.1.9.1 and earlier Versions

**Additional context**

Regards from Meerbusch
Thomas Schneider

albert-github:
I've just pushed a proposed patch, pull request #11559

thom-sch:
Just done a quick test: looks fine :)

Regards from Meerbusch
Thomas

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.14.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #11559, opened 2025-05-02
issue #8731 grouping variables requires an empty line before closing `@}`

albert-github:
See to it that a comment block with `@}` is handled better after an inline comment.

### DIFF
--- a/src/commentcnv.l
+++ b/src/commentcnv.l
@@ -388,6 +388,14 @@ SLASHopt [/]*
 <Scan>\n                           { /* new line */
                                      copyToOutput(yyscanner,yytext,yyleng);
                                    }
+<Scan>{CPPC}[!/]/.*\n[ \t]*{CPPC}[!/][ \t]*{CMD}"}" {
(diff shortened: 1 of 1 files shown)


src/commentcnv.l
```diff
@@ -436,14 +444,6 @@ SLASHopt [/]*
 				     yyextra->readLineCtx=YY_START;
 				     BEGIN(ReadLine);
                                    }
-<Scan>{CPPC}[!/]/.*\n              { /* one line special C++ comment */
```
doxygen, line 439: @albert-github can this rule really be removed? Seems like any case that is not a `/*! @}` will now match the rule below, treating it as a normal comment and missing any special commands inside it? or is this then still handled elsewhere?

albert-github: @doxygen 
Not sure about it. I expected to get a warning about something like a double rule, but that is not the case so probably it doesn't hurt to leave those lines.

Should I update the PR or will you do it upon integration?

PR #11561, opened 2025-05-03
Fix links not being generated

### LINKED ISSUE #11560: Autolinks to functions don't work without @file command
**Describe the bug**

Autolinks to functions aren't generated if the function's documentation is in a file without a @file command.

**To Reproduce**

```
$ ls
foo.c foo.h

$ cat foo.h
/** @file */
void foo1(void);
void foo2(void);

$ cat foo.c
/**
 * See foo2()
 */
void foo1(void) {}

/**
 * Lorem ipsum
 */
void foo2(void) {}

$ (doxygen -g && doxygen) >/dev/null

$ grep -r 'See foo2' html
html/foo_8h.html:<p>See foo2() </p>
```

**Expected behavior**

I'd expect `foo2()` to be autolinked.

**Version**

1.13.2 on macOS (homebrew)

**Additional information**

This only seems to happen when linking to documentation in files without a `@file` command.

nwellnhof:
Ignore non-linkable definitions when resolving links. This avoids links not being generated if a definition in a non-linkable file has higher priority than the one in the linkable file.

Add a flag `onlyLinkable` to `getResolvedSymbolRec` which makes it skip over unlinkable definitions. This flag is only set in `getDefsNew`. The result of this function is ultimately used in

- DocParser::handleLinkedWord via resolveRef
- DocRef::DocRef via resolveLink
- DocLink::DocLink via resolveLink
- linkifyText via getDefs

All these users ignore unlinkable definitions.

Fixes #11560.

### DIFF
--- a/src/symbolresolver.cpp
+++ b/src/symbolresolver.cpp
@@ -144,6 +144,7 @@ struct SymbolResolver::Private
                            const QCString &args,        // in
                            bool checkCV,                // in
                            bool insideCode,             // in
+                           bool onlyLinkable,           // in
                            const MemberDef **pTypeDef,  // out
                            QCString *pTemplSpec,        // out
                            QCString *pResolvedType);    // out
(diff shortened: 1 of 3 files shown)


nwellnhof:
Force-pushed a documentation fix.

PR #11570, opened 2025-05-12
Translation german update

rstephan:
Add missing translations from 1.8.19 to fill the gap to the current version.

### DIFF
--- a/src/translator_de.h
+++ b/src/translator_de.h
@@ -146,7 +146,7 @@
 #ifndef TRANSLATOR_DE_H
 #define TRANSLATOR_DE_H
 
-class TranslatorGerman : public TranslatorAdapter_1_8_15
+class TranslatorGerman : public Translator
 {
   public:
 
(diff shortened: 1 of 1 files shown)


src/translator_de.h
```diff
@@ -2248,6 +2248,108 @@ class TranslatorGerman : public TranslatorAdapter_1_8_15
...
+    { return "Liste aller Schnittstellen"; }
+
+    QCString trInterfaceListDescription() override
+    { return "Hier folgen die Schnittstellen mit Kurzbeschreibung:"; }
+
+    QCString trInterfaceHierarchy() override
+    { return "Schnittstellenhierarchie"; }
+
+    QCString trInterfaceHierarchyDescription() override
+    { return "Die Liste aller Schnittstellen ist -mit Einschränkungen- alphabetisch sortiert:"; }
```
albert-github, line 2285: Wouldn't it be better to use `,`instead of `-`

rstephan: I have replicated the "style" from the previous translation of descriptions, like in `trClassHierarchyDescription()`. But I agree with you, that a `,` looks way better that the `-`.
If you like, I can update all the relevant places with the comma-synax.

albert-github: I'm not 100% sure what is normally used in German speaking countries, but I think there also the `,` is used. I think it would be best to replace it but this should be done everywhere at that moment in the ´translator_de.h´

PR #11584, opened 2025-05-23
Closes #8362

### LINKED ISSUE #8362: Images with same name overwrite each other in output directory
Given the following structure:

src/component1/doc/component1.md
src/component1/doc/img/structure.svg
src/component2/doc/component2.md
src/component2/doc/img/structure.svg

where component1.md and component2.md both have the following content:

```
...
![Structure](img/structure.svg)
...
```

the output will show the component2/doc/img/structure.svg for both component1 and component2. There is no warning or error message during doxygen generation.

**Expected behavior**
The current behavior is not safe.

Chose between one of the following design options:

1. Print an error message when copying an image from somewhere below IMAGE_PATH to OUTPUT_DIRECTORY would result in overwriting that image because an image with the same name already exists.
2. Prepend the path to the base name of the image in the OUTPUT_DIRECTORY, e.g. OUTPUT_DIRECTORY/component1/doc/img/structure.svg - and modify the image HTML tag.
3. Prepend the path to the file name of the image in the OUTPUT_DIRECTORY. e.g. OUTPUT_DIRECTORY/component1_doc_img_structure.svg - and modify the image HTML tag.
4. Append a counter to the images e.g. structure.svg structure_1.svg, structure_2.svg and adapt the references.

I think that the current behavior is by design, as Doxygen is probably supposed to be used on single component level of a software project only.

see also: issue #1822 ( https://github.com/doxygen/doxygen/issues/1822 )

doxygen version: 1.8.20
Ubuntu 18.04

0xspringtime:
Closes #8362
Refs #140

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -289,7 +289,7 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
-    <option type='bool' id='CREATE_SUBDIRS' defval='0'>
+    <option type='bool' id='CREATE_SUBDIRS' defval='1'>
       <docs>
 <![CDATA[
  If the \c CREATE_SUBDIRS tag is set to \c YES then Doxygen will create


albert-github:
- Why would this solve issue #8362?

I don't see this as a bug fix / improvement, there are numerous packages that rely on the default `CREATE_SUBDIRS=NO` and they would get now, unexpected, sub directories and maybe even their processing might fail as they didn't expect these sub directories.

The setting is present for large projects so that the directories would not get overcrowded (especially a speed problem in the past).

PR #11588, opened 2025-05-25
Some Fixes for MinGW-w64

MehdiChinoune:


CMakeLists.txt
```diff
@@ -177,7 +177,7 @@ if (WIN32)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc")
     endif()
 endif()
-if (CMAKE_SYSTEM_NAME MATCHES "CYGWIN")
+if (CYGWIN OR MINGW)
```
albert-github, line 180: isn't in this test the `CMAKE_SYSTEM_NAME MATCHES` part missing twice?

addon/doxysearch/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ find_package(xapian REQUIRED)
 find_package(ZLIB REQUIRED)
 
 if (WIN32)
-  set(WIN_EXTRA_LIBS uuid.lib rpcrt4.lib ws2_32.lib)
+  set(WIN_EXTRA_LIBS uuid rpcrt4 ws2_32)
```
albert-github, line 5: Probably MinGW specific, so it probably will need a test around it.

src/portable.cpp
```diff
@@ -667,7 +667,7 @@ std::ofstream Portable::openOutputStream(const QCString &fileName,bool append)
   std::ios_base::openmode mode = std::ofstream::out | std::ofstream::binary;
   if (append) mode |= std::ofstream::app;
 #if defined(__clang__) && defined(__MINGW32__)
-  return std::ofstream(fs::path(fileName.str()).wstring(), mode);
+  return std::ofstream(fs::path(fileName.str()).string(), mode);
```
albert-github, line 670: As of which version of CLang?

CMakeLists.txt
MehdiChinoune: I don't understand your question

albert-github: Shouldn't the test be something like `if (CMAKE_SYSTEM_NAME MATCHES "CYGWIN" OR CMAKE_SYSTEM_NAME MATCHES "MINGW")`?

doxygen:
@albert-github I need your feedback/approval for this PR as I cannot really test it myself. I'm a bit worried about `wstring()` to `string()` change in `portable.cpp`. Seems like this code was introduced in the past by [this commit](https://github.com/doxygen/doxygen/commit/3317465fe3de7c5f5b362e7d45f4df0c45f95a3a). I wonder if the string() conversion is needed at all (since it isn't done for all other platforms/compilers).

albert-github:
@doxygen
- `wstring()` issue now has been removed from the patch, but I think there were good reasons to introduce it originally.  I cannot test it as I don't have either the clang compiler nor MinGW installed. Might be that the original poster of the `wstring()`  patch could shine a light whether or not it is still necessary.
- The other fixes, look OK to me (tested now https://github.com/doxygen/doxygen/pull/11588#discussion_r2106795051, did test before https://github.com/doxygen/doxygen/pull/11588#discussion_r2107018526).

Issue #11607, opened 2025-06-08
Broken HTML tables


wismill:
**Describe the bug**
Doxygen fails to compile table with `<thead>`

**Screenshots**

**To Reproduce**

```html
<table>
<thead>
<tr><th>xxx</th></tr>
</thead>
</table>
```

**Expected behavior**
Doxygen compile the HTML correctly.

**Version**
1.14.0, Linux (openSUSE Tumbleweed)

**Stack trace**

**Additional context**
It used to work with the previous 1.13 series.

wismill:
To be honest, I am really concerned by the poor QA of this project. Breaking such a **_basic_ feature** gives me little faith in the new releases. Is there no testing at all?

I am sorry for the rant, but please do not resort to the “too few resources” excuse. *Testing* (e.g. golden tests) *is* was makes an *efficient use of resources*. I should know, I am myself a FLOSS dev. I know very well how thankless the job is, so I greatly value the automated tests to reduce my workload.

But please, *help yourself* and write tests!!

albert-github:
I've just pushed a proposed patch, pull request #11608

<strike>Edit: PR is incomplete.</strike>

doxygen:
> To be honest, I am really concerned by the poor QA of this project. Breaking such a **_basic_ feature** gives me little faith in the new releases. Is there no testing at all?
> 
> I am sorry for the rant, but please do not resort to the “too few resources” excuse. _Testing_ (e.g. golden tests) _is_ was makes an _efficient use of resources_. I should know, I am myself a FLOSS dev. I know very well how thankless the job is, so I greatly value the automated tests to reduce my workload.
> 
> But please, _help yourself_ and write tests!!

We do have tests that compare the output against expectations. This is mostly limited to XML output for automation reasons. While there are currently 100+ tests, it would require millions more to cover all possible scenarios, so that is not very feasible. 

I do agree that we can do a better job to more systematically add test-cases when fixing bugs in order to prevent future regressions.

In this particular case we did have a table as part of the manual, so there was some test coverage, but that table didn't include the table head and body tags, which are supposed to be ignored by Doxygen anyway.

@albert-github Could you add a test case for this, as part of your fix?

wismill:
> We do have tests that compare the output against expectations. This is mostly limited to XML output for automation reasons. While there are currently 100+ tests, it would require millions more to cover all possible scenarios, so that is not very feasible.

[...]

PR #11608, opened 2025-06-08
Issue #11607 Broken HTML tables

albert-github:
Catching end tags for `<thead>`, `<tfoot>` and `<tbody>` as well so they won't throw a warning

### DIFF
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -2214,6 +2214,18 @@ getrow:
       {
         retval = Token::make_RetVal_EndTable();
       }
+      else if (tagId==HtmlTagType::HTML_TBODY && parser()->context.token->endTag)
+      {
+        // for time being ignore </t....> tag
+      }
+      else if (tagId==HtmlTagType::HTML_THEAD && parser()->context.token->endTag)
+      {
+        // for time being ignore </t....> tag
+      }
+      else if (tagId==HtmlTagType::HTML_TFOOT && parser()->context.token->endTag)
+      {
+        // for time being ignore </t....> tag
+      }
       else  // found some other tag
       {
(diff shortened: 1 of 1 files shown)


src/docnode.cpp
```diff
@@ -1946,8 +1946,20 @@ Token DocHtmlRow::parse()
...
   Token tok=parser()->tokenizer.lex();
-  // skip whitespace
-  while (tok.is_any_of(TokenRetval::TK_WHITESPACE,TokenRetval::TK_NEWPARA)) tok=parser()->tokenizer.lex();
+  // skip whitespace and tbody, thead and tfoot tags
+  while (tok.is_any_of(TokenRetval::TK_WHITESPACE,TokenRetval::TK_NEWPARA,TokenRetval::TK_HTMLTAG))
+  {
+    if (tok.is(TokenRetval::TK_HTMLTAG))
+    {
+      HtmlTagType tagId=Mappers::htmlTagMapper->map(parser()->context.token->name);
+      if ((tagId==HtmlTagType::HTML_TBODY) || (tagId==HtmlTagType::HTML_THEAD) || (tagId==HtmlTagType::HTML_TFOOT)) tok=parser()->tokenizer.lex();
```
sethrj, line 1955: Maybe define this as a helper function so you can reuse it on line 2017? Or in fact isn't this whole block identical to the one below, in which case shouldn't a larger helper be used?

EDIT: looks like this code block is used in four places: *definitely* define a helper function, otherwise future fixes/changes will likely miss some of the duplicate cases.

doxygen: I fully agree. I was already working on that when I read your comment.

Issue #11625, opened 2025-06-19
Autolinking does not work with USE_MDFILE_AS_MAINPAGE


bkloster-sma:
**Describe the bug**
When declaring a markdown file as the mainpage with either USE_MDFILE_AS_MAINPAGE or `{#mainpage}`, no automatic links are generated for function names or classes mentioned in that markdown file.

**To Reproduce**
The [attached zip](https://github.com/user-attachments/files/20817117/doxygen_autolink_on_mainpage.zip) contains two doxyfiles, one with `USE_MDFILE_AS_MAINPAGE` set, and the other without. When running doxygen on `NoMainPage.doxyfile`, autolinking works as expected. When running it on `WithMainpage.doxyfile`, links are only generated with explicit `@ref`.

**Expected behavior**
Links should be automatically generated for markdown documents set as mainpage.

**Version**
This only happens since 1.14.0. With 1.13.2, the mainpage contains auto-generated links as expected.

albert-github:
It looks like the mentioned problem is a side effect of:
```
Commit: 3c5255e0ad809fd86543744195df8b7a9814f785 [3c5255e]
Date: Friday, May 9, 2025 8:57:44 PM

issue #6998 Markdown: Links in ATX headings processed incorrectly?
```

albert-github:
I've just pushed a proposed patch, pull request #11627

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.15.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #11627, opened 2025-06-20
Issue #11625 Autolinking does not work with USE_MDFILE_AS_MAINPAGE

albert-github:
Also the main page should follow the rules regarding autolinking.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -4935,7 +4935,7 @@ static void writeIndex(OutputList &ol)
       ol.startTitleHead(QCString());
       ol.generateDoc(Doxygen::mainPage->docFile(),Doxygen::mainPage->getStartBodyLine(),
                   Doxygen::mainPage.get(),nullptr,Doxygen::mainPage->title(),false,false,
-                  QCString(),true,false,Config_getBool(MARKDOWN_SUPPORT));
+                  QCString(),true,false);
       headerWritten = TRUE;
     }
   }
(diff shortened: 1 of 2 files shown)


src/index.cpp
```diff
@@ -4935,7 +4935,7 @@ static void writeIndex(OutputList &ol)
       ol.startTitleHead(QCString());
       ol.generateDoc(Doxygen::mainPage->docFile(),Doxygen::mainPage->getStartBodyLine(),
                   Doxygen::mainPage.get(),nullptr,Doxygen::mainPage->title(),false,false,
-                  QCString(),true,false,Config_getBool(MARKDOWN_SUPPORT),false);
+                  QCString(),true,false,Config_getBool(MARKDOWN_SUPPORT));
```
doxygen: the `false` parameter is needed for titles, so this one should stay, the change below is correct.

src/index.cpp
```diff
@@ -4972,7 +4972,7 @@ static void writeIndex(OutputList &ol)
     ol.generateDoc(defFileName,defLine,Doxygen::mainPage.get(),nullptr,
                 Doxygen::mainPage->documentation(),true,false,
                 QCString(),false,false,
-                Config_getBool(MARKDOWN_SUPPORT),false);
+                Config_getBool(MARKDOWN_SUPPORT));
```
doxygen: Here also the Config_getBool(MARKDOWN_SUPPORT) parameter can be removed as it is already the default.

albert-github: In most cases I'm personally not a fan of default arguments as it is less good visible from the call what is happening and one has to either look in the `.h` file where the function is declared.

Found another superfluous `Config_getBool(MARKDOWN_SUPPORT)` as well.

Issue #11647, opened 2025-07-02
Backslashes not properly escaped in generated search files


Ssamm45:
**Describe the bug**
Php uses backslash as namespace separators so generated names can contain them. In the generated html/search/*.js files these backslashes aren't properly escaped which can cause unexpected output, or in the case of '\u' can break the search entirely

**To Reproduce**

Running doxygen with the default configuration on a file like this:
```php
<?php

/**
* @file
* example file
*/

/**
* This is a freestanding url function
*/
function url()
{
	return "https://example.com";
}

/**
* Example Class
*/
class MyClass {

	/**
	* Generates a url
	*/
	public function url():string
	{
		return "https://example.com";
	}
}

```

will generate an 'html/search/functions_0.js' that look like this

```JavaScript
var searchData=
[
  ['url_0',['url',['../classMyClass.html#abe51b72ce9e4d53d8fdd4c53bcdf3916',1,'MyClass\url()'],['../test_8php.html#abc5ed2ad3f9365fb31cc1eb625b974d9',1,'url():&#160;test.php']]]
];
```
Then if you open the documentation in the browser and search 'url' you get 'No Matches' and the browser console contains the error 

'Uncaught SyntaxError: malformed Unicode character escape sequence functions_0.js:3:87'

**Expected behavior**

The expected output would be `'MyClass\\url()'` instead of `'MyClass\url()'`

```JavaScript
var searchData=
[
  ['url_0',['url',['../classMyClass.html#abe51b72ce9e4d53d8fdd4c53bcdf3916',1,'MyClass\\url()'],['../test_8php.html#abc5ed2ad3f9365fb31cc1eb625b974d9',1,'url():&#160;test.php']]]
];
```

**Version**
doxygen version 1.14.0
installed from Arch Linux package manager
LSB Version:	n/a
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a

albert-github:
I've just pushed a proposed patch, pull request #11648

albert-github:
Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.15.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #11648, opened 2025-07-03
issue #11647 Backslashes not properly escaped in generated search files

albert-github:
Escape the backslash in the separator


Ssamm45:
I tested it and it fixes the issue I was having.

Issue #11634, opened 2025-06-24
Doxygen Latex Tables not working correctly


ConradMDoosan:
**Describe the bug**
I am using Doxygen to make a table in latex. I want this to be reproducible without having to manually change it. In the LaTeX setting if I set them to the the setting I below I get a funky looking table as seen in the picture below. 

GENERATE_LATEX         = YES
LATEX_OUTPUT           = latex
LATEX_CMD_NAME         =
MAKEINDEX_CMD_NAME     = makeindex
LATEX_MAKEINDEX_CMD    = makeindex
COMPACT_LATEX          = YES
PAPER_TYPE             = a4
EXTRA_PACKAGES         = array \
                         longtable \
                         booktabs \
                         tabularx \
                         doxygen
LATEX_HEADER           =
LATEX_FOOTER           =
LATEX_EXTRA_STYLESHEET =
LATEX_EXTRA_FILES      =
PDF_HYPERLINKS         = YES
USE_PDFLATEX           = YES
LATEX_BATCHMODE        = BATCH
LATEX_HIDE_INDICES     = YES
LATEX_BIB_STYLE        = plain
LATEX_EMOJI_DIRECTORY  =

![Image](https://github.com/user-attachments/assets/f65642aa-cdd5-49dd-94ef-f8a4812a2797)

The code that I am using is:

/*
* @page numbers Numbers
*
* *<table>
 *<tr><th>File Name</th><th>Start Position</th><th>Length</th><th>Number</th></tr>
 *<tr><td>File 1</td><td>3.1</td><td>1 Byte</td><td>513</td></tr>
 *<tr><td>File 2</td><td>4.1</td><td>2 Byte</td><td>190</td></tr>
 *</table>
*
*/

I have also used this format to with no avail.
| File Name | Start Pos | Len | Number |
| ----------- | ---------- | ---- | --------- |

When I turn Latex_BatchMode = NO, I get the following error:
(C:\Users\conradmercer\AppData\Local\Programs\MiKTeX\tex/latex/wasysym\uwasy.fd
)
! Undefined control sequence.
<argument> ...rtpbox {\@nextchar }\insert@pcolumn
                                                  \@endpbox \do@row@strut \o...
l.16 \begin{longtabu}spread 0pt [c]{*{4}{|X[-1]}|}

?

I am using the latest version. My html tables look good, I just need it to work for Latex. I would love any help thanks!

dPAutomation:
Same problem on my side:

[...]

PR #11651, opened 2025-07-04
Removing explicit tables code from source code

albert-github:
Removing explicit tables code from source code so it might be easier maintainable / there might be easier to create workarounds.


albert-github:
Thanks to @davidcarlise for the help on some problems (see https://tex.stackexchange.com/questions/747234/file-ended-while-scanning-use-of-txgetbody)

Note this does not solve the problems as reported in #11634

PR #11653, opened 2025-07-08
Imp/dot n plantuml improvements

verheand:
There are some improvements regarding speed especially for systems with expensive file i/o or calls to executables:
- option DOT_DRY_RUN does not call dot from doxygen. Generation of files can be done afterwards e.g. using xargs dot.exe.
- there's a look-up table that checks if the dot-graph has been already written to a file. This avoids file i/o checking if the file already exists.
- the dot-files do not start with getOutputFileBase(). This reduces the number of files with identical content and also the time to render the graphs.
- there's a look-up table that checks if the uml-graph has been already processed. If yes, it is skipped. 
- PlantUML file inline_umlgraph_* is now split up into multiple files. PlantUML supports only multi-threading for separated files. Thus, writing all graphs to one file uses only one thread in plantuml. 
- The inline_umlgraph files contain now a reference to the origin. This helps to fix syntax issues in the graphs. 
- Plantuml syntax errors are now reported as warnings.

This is my first pull request, so consider it as a draft proposal far away from been perfect and fully tested.

### DIFF
--- a/doc_internal/Doxyfile.in
+++ b/doc_internal/Doxyfile.in
@@ -350,6 +350,7 @@ DIR_GRAPH_MAX_DEPTH    = 1
 DOT_IMAGE_FORMAT       = svg
 INTERACTIVE_SVG        = YES
 DOT_PATH               =
+DOT_DRY_RUN            = NO
 DOTFILE_DIRS           =
 DIA_PATH               =
 DIAFILE_DIRS           =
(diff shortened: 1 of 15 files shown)


doxygen:
@verheand Thanks for contributing. For a pull request to be integrated easily, it is important for it to only fix a single bug or add a single feature. Your pull request seems to combine a number of changes into one large PR which is typically not desired.

verheand:
Thank you for your answer. I will close the commit now and will try to split up the changes.

PR #11703, opened 2025-08-05
Replace ShellExecuteExW with simple CreateProcessW #11699

### LINKED ISSUE #11699: Portable::system gets stuck in spinlock on Windows with high CPU core counts
**Describe the bug**
The use of `ShellExecuteExW` with unbounded (or high) parallelism is problematic, as by an implementation detail of `ShellExecuteExW` itself, it's trying to cache all sorts of details regarding the configuration of the shell in a shared cache. As far as I can tell, this internal cache is as badly implemented as it can only get, and results in a soft-lock with zero progression if too many threads attempt to initialize it simultaneously.

Specifying `COINIT_APARTMENTTHREADED` has no effect whatsoever on this cache.

**Screenshots**
What it looks like with 64+ threads:
Image

None of those threads gets to the point where they ever manage to start a process, they all get stuck looping.

When viewed with a profiler, the entirety of the CPU time is spent in `windows.storage.dll!CShellExecute::_DoExecute`.

Only once one of them (by chance) breaks out of the loop, all of them are then able resume.

**To Reproduce**
Can only be reproduced on high-core-count systems (>= 60 threads), with `DOT_NUM_THREADS == 0` and a number of (`dot`) jobs sufficient to saturate all cores. Did not trigger on WIindows 10 / Server 2022, but does trigger on Windows 11 / Server 2025.

You won't be able to trigger the contention/loop with just using a lower number of real cores.

There may be an additional dependency on used antivirus solutions on the host, as any OS level interception of registry accesses etc. makes it more likely to enter the bad case.

**Expected behavior**
`ShellExecuteExW` is far from threadsafe. Completely getting stuck is a quite obviously a bug in the `windows.storage.dll`, but even in the best case `ShellExecuteExW` has a massive overhead.

Not getting stuck would be minimum - bust mostly not using `ShellExecuteExW` in the first place would be expected. Most of the commands passed in don't even require a fully fleshed shell. Just a dumb `CreateProcess` with the command line passed in the 2nd argument `lpCommandLine` would have done the job correctly for almost all of the invocations in this codebase.

The only exception in that entire codebase might be `gswin32`.

**Version**
Doxygen 1.13.2 on Windows 11 / Windows Server 2025 with 64 threads AMD Threadripper and `DOT_NUM_THREADS == 0`.

Ext3h:
Fix #11699 by replacing the call to `ShellExecuteExW` with the fragile `CoInitializeEx` setup by a plain `CreateProcessW`.

Not only should this avoid the bug itself, but also speed up the invocation of the `dot.exe` etc. significantly.

### DIFF
--- a/src/portable.cpp
+++ b/src/portable.cpp
@@ -200,59 +200,43 @@ int Portable::system(const QCString &command,const QCString &args,bool commandHa
   }
   else
   {
-    // Because ShellExecuteEx can delegate execution to Shell extensions
-    // (data sources, context menu handlers, verb implementations) that
-    // are activated using Component Object Model (COM), COM should be
-    // initialized before ShellExecuteEx is called. Some Shell extensions
-    // require the COM single-threaded apartment (STA) type.
-    // For that case COM is initialized as follows
-    CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE);
+    uint16_t* fullCmdW = nullptr;
+    recodeUtf8StringToW(fullCmd, &fullCmdW);
 
-    uint16_t *commandw = nullptr;
-    recodeUtf8StringToW( commandCorrectedPath, &commandw );
-    uint16_t *argsw = nullptr;
-    recodeUtf8StringToW( args, &argsw );
+    STARTUPINFOW sStartupInfo;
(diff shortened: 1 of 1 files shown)


albert-github:
I just did  small test on my Windows (10) system with an older compiler and I didn't see any problems.

PR #11705, opened 2025-08-08
Support Mathjax version 4

albert-github:
Added support for Mathjax version 4 and update documentation

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -2800,14 +2800,16 @@ obfuscate email addresses.
       </docs>
       <value name="MathJax_2"/>
       <value name="MathJax_3"/>
+      <value name="MathJax_4"/>
     </option>
     <option type='enum' id='MATHJAX_FORMAT' defval='HTML-CSS' depends='USE_MATHJAX'>
       <docs>
 <![CDATA[
  When MathJax is enabled you can set the default output format to be used for
  the MathJax output.
- For more details about the output format see <a href="http://docs.mathjax.org/en/v2.7-latest/output.html">MathJax version 2</a>
- and <a href="http://docs.mathjax.org/en/latest/web/components/output.html">MathJax version 3</a>.
+ For more details about the output format see <a href="https://docs.mathjax.org/en/v2.7/output.html">MathJax version 2</a>,
+ <a href="https://docs.mathjax.org/en/v3.2/output/index.html">MathJax version 3</a>
+ and <a href="https://docs.mathjax.org/en/v4.0/output/index.htm">MathJax version 4</a>.
 ]]>
       </docs>
       <value name="HTML-CSS" desc="(which is slower, but has the best compatibility. This is the name for Mathjax version 2, for MathJax version 3 this will be translated into \c chtml)"/>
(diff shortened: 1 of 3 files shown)


src/config.xml
```diff
@@ -2819,33 +2821,42 @@ obfuscate email addresses.
       <docs>
 <![CDATA[
  When MathJax is enabled you need to specify the location relative to the
- HTML output directory using the \c MATHJAX_RELPATH option. The destination
+ HTML output directory using the \c MATHJAX_RELPATH option. For Mathjax version 2 the destination
  directory should contain the `MathJax.js` script. For instance, if the \c mathjax
  directory is located at the same level as the HTML output directory, then
- \c MATHJAX_RELPATH should be <code>../mathjax</code>. The default value points to
+ \c MATHJAX_RELPATH should be <code>../mathjax</code>.s
+ For Mathjax versions 3 and 3 the destination directory should contain the
```
doxygen: '3 and 3' -> '3 and 4'?

src/config.xml
```diff
@@ -2819,33 +2821,42 @@ obfuscate email addresses.
...
-For example for MathJax version 3 (see http://docs.mathjax.org/en/latest/input/tex/extensions/index.html):
+For example for MathJax version 3 (see https://docs.mathjax.org/en/v3.2/input/tex/extensions/):
 \verbatim
 MATHJAX_EXTENSIONS     = ams
 \endverbatim
+For example for MathJax version 4 (see https://docs.mathjax.org/en/v4.0/input/tex/extensions/):
+\verbatim
+MATHJAX_EXTENSIONS     = units
+\endverbatim
+Note that for Matjax version 4 quite a few extensions are already automatically loaded.
```
doxygen: Matjax -> Mathjax

src/htmlgen.cpp
```diff
@@ -464,6 +464,69 @@ static QCString substituteHtmlKeywords(const QCString &file,
 
     switch (mathJaxVersion)
     {
+      case MATHJAX_VERSION_t::MathJax_4:
```
doxygen, line 467: Looks like a lot of code duplication between this and the MathJax_3 case. Maybe one or more helper functions could be used.

src/config.xml
```diff
@@ -2800,14 +2800,16 @@ obfuscate email addresses.
       </docs>
       <value name="MathJax_2"/>
       <value name="MathJax_3"/>
+      <value name="MathJax_4"/>
```
doxygen, line 2803: should this also already be the default?

albert-github: Due to compatibility I don't think so for compatibility reasons.

src/htmlgen.cpp
albert-github: Might be possible but when implementing support for Mathjax 4 it was unclear what exactly would be needed, hence some duplication.

src/config.xml
doxygen: but existing Doxyfiles that have already selected Mathjax 2 or 3 explicitly will not be affected by a new default, only config files that did not yet specify a version explicitly, but these then would most likely also not specify `USE_MATHJAX` and then the version setting doesn't matter as Mathjax support is disabled by default.

albert-github: Most packages will have set the Mathjax version but it is not necessay, one can just set `USE_MATHJAX` and in that case the default Mathjax version will be used.
Setting Mathjax version 4 as default sounds nice, though the version 4 is new (even though it has  been quite long in alpha / beta  state) one advantage might be that it fixes some discrepancies between LaTeX and Mathjax (a known problem is the usage of `\text` with underscores.
To get in text mode inside a formula something like:
image

- In Latex one has to write:  `\text(src\_dst}` as `\text(src_dst}`  will give an error
- In Mathjax 2 and 3 one has to write `\text(src_dst}` as with  `\text(src\_dst}`  one gets:
  image
- In Mathjax 4 one can follow the (standard) LateX way: `\text(src\_dst}` as \text(src_dst}` will give an error.

(By chance I was looking into this problem again for Fossies).

albert-github: Note that `MATHJAX_VERSION` was introduced in doxygen version 1.9.2 (August 28, 2021), so "relatively" recent.

doxygen:
@albert-github I've reduced the code duplication a bit. See [this commit](https://github.com/doxygen/doxygen/commit/07f0177d715120f4634311c6eadd2c21ce318c70). Let me know if I've missed anything.

albert-github:
At first glance I don't see a difference.

albert-github:
As some extra information, especially regarding the problem of the `_` in `text` (https://github.com/doxygen/doxygen/pull/11705#discussion_r2265208555) a proposed pull request #11713 has been made

Issue #11708, opened 2025-08-11
[BUG] Variadic macro expansion doesn't work when the arguments contain comments starting with `//`


parmi93:
```c
/**
 * @file temp.h
 */
#include "config.h"

/**
 * @brief My struct A.
 */
typedef PACK(struct {
    int struct_a_1; ///< Description of struct_a_1
    int struct_a_2; // This comment causes the problem 
}) my_structA_t;

/**
 * @brief My struct B.
 */
PACK(struct my_structB_t {
    int struct_b_1; ///< Description of struct_b_1
    int struct_b_2; // This comment causes the problem 
});

/**
 * @brief My Enum A.
 */
typedef PACK(enum {
    ENUM_A_1 = 1, ///< My enum A1
    _ENUM_A_MAX   // This comment causes the problem 
}) my_enumA_t;

/**
 * @brief My Enum B.
 */
PACK(enum my_enumB_t{
    ENUM_B_1,    ///< My enum B1
    _ENUM_B_MAX, // This comment causes the problem 
});
```

This file is not relevant to the issue, but I’m adding it for completeness
`config.h`:
```c
#ifndef CONFIG_H
#define CONFIG_H

#ifndef PACK
#if defined(__GNUC__) || defined(__clang__)
// It works with structs and enums
#define PACK(...) __VA_ARGS__ __attribute__((__packed__))
#endif

#if _MSC_VER
/* It works only with structs, MSVC does not support packing of enums; in any case, this macro does
 * nothing when used with enums. */
#define PACK(...) __pragma(pack(push, 1)) __VA_ARGS__ __pragma(pack(pop))
#endif
#endif // PACK

#endif
```

`Doxyfile`:
```
MACRO_EXPANSION        = YES
PREDEFINED             = PACK(...)=__VA_ARGS__
```

Output:
<img width="665" height="1058" alt="Image" src="https://github.com/user-attachments/assets/a05af8bf-05ce-4955-8fc2-6517ebf9620c" />

The problem disappears if I replace `// This comment causes the problem` with one of the following:
- `/// This comment is ok`
- `///< This comment is ok`
- `/* This comment is ok */`
- `/** This comment is ok */`
- `/**< This comment is ok */`

Doxygen v1.14.0
[doxygen_bug.zip](https://github.com/user-attachments/files/21718625/doxygen_bug.zip)

albert-github:
I've just pushed a proposed patch, pull request #11712

albert-github:
[...]

PR #11715, opened 2025-08-13
Spurious asterisk after preprocessing when using macros

albert-github:
When having something like:
```
typedef PACK(enum {
    ENUMA, /**< @brief The asterisk character at the beginning of the next line is not removed, but
            * becomes part of the documentation. */
}) my_enum_t;
```
and a `PREDEFINED `PACK` this leads to an `*` in the output before the word "becomes. This is due to the, incorrect, joining of lines in case of a multi line block of comment.

This is based on the comment: https://github.com/doxygen/doxygen/issues/11708#issuecomment-3183502931

Example: [example.tar.gz](https://github.com/user-attachments/files/21758855/example.tar.gz)

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -928,9 +928,8 @@ WSopt [ \t\r]*
                                           BEGIN(FindDefineArgs);
                                         }
 <ArgCopyCComment>\n                     {
-                                          yyextra->defArgsStr+=' ';
+                                          yyextra->defArgsStr+=yytext;
                                           yyextra->yyLineNr++;
-                                          outputChar(yyscanner,'\n');
                                         }
 <ArgCopyCComment>.                      {
                                           yyextra->defArgsStr+=yytext;


src/pre.l
```diff
@@ -928,9 +928,8 @@ WSopt [ \t\r]*
                                           BEGIN(FindDefineArgs);
                                         }
 <ArgCopyCComment>\n                     {
-                                          yyextra->defArgsStr+=' ';
+                                          yyextra->defArgsStr+=yytext;
                                           yyextra->yyLineNr++;
-                                          outputChar(yyscanner,'\n');
```
doxygen, line 933: Will this not cause the line numbers to get out of sync? (i.e. increasing `yyLineNr` without outputting a `\n`)

albert-github: I checked the linenumbers of the resulting preprocessing and they looked OK to me.

Issue #11727, opened 2025-08-15
DOT_NUM_THREADS should have a bigger maximum limit


peterurbanec:
The maximum number of threads that can be specified for `DOT_NUM_THREADS` is 32.

This is not enough on modern machines. On a Threadripper Pro with 64 cores and 128 threads, the Doxygen run takes longer than it needs to, while the machine is underutilised.

Please remove the upper bound on this configuration variable, or at least increase it to 512.

Threadripper Pro machines with 192 threads are available today, so a concurrency level of 200 dot threads is a reasonable setting with only minimal over-commit to account for I/O.

albert-github:
I think this issue is related to the comments https://github.com/doxygen/doxygen/issues/11699#issuecomment-3155180448 and https://github.com/doxygen/doxygen/issues/11699#issuecomment-3158330640

@doxygen which actions should we take.

albert-github:
• Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).
• (Issue was closed automatically by GitHub due to the used terminology in the pull request)

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in doxygen version 1.15.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #11728, opened 2025-08-15
Increase maximum number of threads fix doxygen/doxygen#11727

peterurbanec:
This addresses doxygen/doxygen#11727

### DIFF
--- a/src/config.xml
+++ b/src/config.xml
@@ -881,7 +881,7 @@ Go to the <a href="commands.html">next</a> section or return to the
 ]]>
       </docs>
     </option>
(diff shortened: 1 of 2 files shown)


albert-github:
Thanks for the patch
As Indicated in the issue comment https://github.com/doxygen/doxygen/issues/11727#issuecomment-3191387029probably we should with this fix also
- synchronize the `NUM_PROC_THREADS` and `DOT_NUM_THREADS`
- maybe remove the upper limit all together


Just starting the build, but remaining are the above points.

Issue #11740, opened 2025-08-19
Regression: Doxygen >= 1.11.0 no longer able to produce .chm files that link to each other using TAGFILES


puetzk:
**Describe the bug**
PR #10860 introduced a restriction that .chm files should only be able to link to http/https URLs, in an attempt to fix #6217 

I think this check is too strong. In particular, it bans the "ms-its:" URL scheme IE uses for linking into .chm files (and since .chm files use the MSHTML render, this is the means by which a set of separate-but-related related .chm files would link into each other, which is what `TAGFILES` is all about. 

See https://learn.microsoft.com/en-us/previous-versions/windows/desktop/htmlhelp/about-html-help-urls

**To Reproduce**

[Doxygen-TAGFILES-ms_its.zip](https://github.com/user-attachments/files/21868585/Doxygen-TAGFILES-ms_its.zip)

`doxygen Doxyfile.foo.doxygen.chm` -> generates foo.doxygen.chm and foo.doxygen.tags
`doxygen Doxyfile.bar.doxygen.chm` -> generates bar.doxygen.chm

doxygen 1.10 works as intended, producing a bar.doxygen.chm in which both of those mentions of `Foo` (in the signature and in the description) are links to `ms-its:foo.doxygen.chm::/class_foo.html`, which will open the corresponding page in foo.doxygen.chm in the same browser frame/instence of hh.exe. 
> void  bar ([Foo](ms-its:foo.doxygen.chm::/class_foo.html) *foo) 
>  Oh the things you could do, if you had a [Foo](ms-its:foo.doxygen.chm::/class_foo.html).

Caveat: there's some further interactions here between `ms-its:`, IE's Local Zone lockdown  and the "Mark of the web" (Zone.Identifier proprety).  If the "bar.chm" file has a "Zone.Identifier" identifying it as downloaded from some website, then the browser isn't allowed follow these links to other local files, even if those other files bear the same "mark of the web" domain. So admittedly this is only usable between .chm files generated locally, or shipped as part of an installer, etc - loose files downloaded by a browser often end up with their links broken unless you do the File : Properties : General : Unblock in explorer to release them from quarantine. But I think caring about that kind of browser security feature is beyond the doxygen's scope - it generated a file that works fine, unless blocked by a cross-domain security measure.

Doxygen 1.14 instead produces a message

[...]

PR #11741, opened 2025-08-20
issue #11740 Regression: Doxygen >= 1.11.0 no longer able to produce chm files that link to each other using TAGFILES

albert-github:
Added `ms-its` protocol for HTMLHELP files

### DIFF
--- a/src/configimpl.l
+++ b/src/configimpl.l
@@ -2006,7 +2006,8 @@ void Config::checkAndCorrect(bool quiet, const bool check)
       if (eqPos!=std::string::npos) // tag command contains a destination
       {
         QCString url = QCString(s.substr(eqPos+1)).stripWhiteSpace().lower();
-        validUrl = url.startsWith("http:") || url.startsWith("https:");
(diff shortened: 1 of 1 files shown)


puetzk (approved):
Thanks for the quick turnaround! That's certainly all it looked like it would take, but I hadn't set up the dependencies to build Doxygen from source and try it yet... and I didn't want to send a PR without actually trying my changes at all :-)

If this made the foo/bar example attached to #11740 work, looks good to me...

albert-github:
@puetzk 
You can find, till it is automatically removed by GitHub, executables in the list of artifacts at https://github.com/doxygen/doxygen/actions/runs/17094432325/attempts/1?pr=11741

puetzk:
Ah. I looked, but only found the 
"macOS Latest Debug (Apple Silicon)" artifacts from attempt 2. I hadn't noticed that it was "attempt 2" and the other artifacts were in the canceled "attempt 1" 

Anyway, yes, fix confirmed.

Issue #11683, opened 2025-07-28
duplicates in searchdata.xml "text" field


MatthiasIsele:
**Describe the bug**
Brief comments in a base class result in searchdata.xml containing the field "text" N times where N is the number of classes derived from the base class.
This doesn't seem like a big issue in this small example. However in my huge project with a large class hierarchy this results in getting huge fields which is also visible in the search results on the website using the external search engine.

**To Reproduce**
There are three c++ classes defined:
ClassA, ClassB and ClassC where ClassB and ClassC are both derived from ClassA.
ClassA has a method myMethod with a brief comment containing the text "This is a brief comment".
Generate using the following settings:
```
JAVADOC_AUTOBRIEF = YES
EXTERNAL_SEARCH = YES
```
the attached zip contains a minimal example to replicate.

As a result I get an entry in searchdata:
```
<doc>
  <field name="type">function</field>
  <field name="name">ClassA::myMethod</field>
  <field name="args">(int arg)</field>
  <field name="url">class_class_a.html#a464f8f980ac6ef91714a60d2bd906203</field>
  <field name="keywords">myMethod ClassA::myMethod ClassA</field>
  <field name="text">This is a brief comment This is a brief comment This is a brief comment</field>
</doc>
```

**Expected behavior**
As a result I should get an entry in searchdata like:
```
<doc>
  <field name="type">function</field>
  <field name="name">ClassA::myMethod</field>
  <field name="args">(int arg)</field>
  <field name="url">class_class_a.html#a464f8f980ac6ef91714a60d2bd906203</field>
  <field name="keywords">myMethod ClassA::myMethod ClassA</field>
  <field name="text">This is a brief comment</field>
</doc>
```

**Version**
Tested with current master but got the same behaviour with version 1.13.x and 1.14.x
Running on linux or windows results in the same behaviour.

[searchdata.zip](https://github.com/user-attachments/files/21470853/searchdata.zip)

albert-github:
This looks a bit overdone to have 3 times the same text. Has to be investigated.
Looks like it is independent of `JAVEDOC_AUTOBRIEF`

MatthiasIsele:
You are right. In my initial test I had a comment without `@brief` tag. So the issue only occured when setting `JAVADOC_AUTOBRIEF` on.
I can confirm that the issue persists when `JAVADOC_AUTOBRIEF` is off.

albert-github:
I've just pushed a proposed patch, pull request #11744

[...]

PR #11744, opened 2025-08-23
issue #11683 duplicates in searchdata.xml "text" field

albert-github:
In case of an inherited member the, brief, text of the inherited member should not be added to the search text field.

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -2614,11 +2614,13 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
      )
   {
     auto parser { createDocParser() };
+    Doxygen::searchIndex.setInherited(inheritedFrom!=nullptr);
     auto ast    { validatingParseDoc(*parser.get(),
                                      briefFile(),briefLine(),
                                      getOuterScope()?getOuterScope():d,
                                      this,briefDescription(),TRUE,FALSE,
                                      QCString(),TRUE,FALSE) };
+    Doxygen::searchIndex.setInherited(false);
     if (!ast->isEmpty())
     {
       ol.startMemberDescription(anchor(),inheritId);
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -180,7 +180,7 @@ DocWord::DocWord(DocParser *parser,DocNodeVariant *parent,const QCString &word) 
       DocNode(parser,parent), m_word(word)
 {
   //printf("new word %s url=%s\n",qPrint(word),qPrint(parser->context.searchUrl));
-  if (Doxygen::searchIndex.enabled() && !parser->context.searchUrl.isEmpty())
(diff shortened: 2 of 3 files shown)


src/memberdef.cpp
```diff
@@ -2614,11 +2614,13 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
      )
   {
     auto parser { createDocParser() };
+    Doxygen::searchIndex.setInherited(inheritedFrom!=nullptr);
```
doxygen, line 2617: I don't think this will work correctly when multiple threads are enabled. This code can be run from multiple threads in parallel, but if one thread manipulates the shared object `Doxygen::searchIndex` this will also impact the other threads. 

A possible solution is to give the `m_inherited` variable thread-local storage (see `THREAD_LOCAL` define in doxygen.h).

doxygen:
@albert-github Implemented a solution using thread local storage.

src/memberdef.cpp
albert-github: I think a valid point.

I tried to make the `m_inherited`:
```
static THREAD_LOCAL bool m_inherited;
```
but this gave problems (even when including `doxygen.h`) like:
```
error C2144: syntax error: 'bool' should be preceded by ';'
```
but this was due to the fact that `doxygen.h` included `searchindex.h` (so a "chicken - egg problem"), placing the defines in `doxygen.h` before the includes helped a bit here but not everywhere so still needed to change the place of `doxygen.h` the respective cpp files, but gave linking problems like (depending on the compiler):
```
.../ld: ../lib/libdoxymain.a(memberdef.cpp.o):memberdef.cpp:(.rdata$.refptr.__emutls_v._ZN15SearchIndexIntf11m_inheritedE[.refptr.__emutls_v._ZN15SearchIndexIntf11m_inheritedE]+0x0): undefined reference to `__emutls_v._ZN15SearchIndexIntf11m_inheritedE'
```

```
error LNK2001: unresolved external symbol "private: static bool SearchIndexIntf::m_inherited" (?m_inherited@SearchIndexIntf@@0_NA)
```

I also have a doubt how to initialized `m_inherited`

So still a valid point, but probably a more sophisticated solution is needed (the call in `memberdef.cpp` could be embedded in a mutex, but I don't think this is the right way to go, although it will be just a small string, brief string, that is normally handled here).

When finishing this comment I saw the implementation, most likely the sophisticated way to go.

doxygen: While looking a bit more into the code, I noticed `validatingParseDoc()` already has a `indexWords` parameter to tell the function if words need to be added to the index. So setting this to false should have already been enough. Did you try this already? If not, I'll test this and make a simplified solution.

albert-github: I saw this flag as well, but I didn't trust to use it.

PR #11745, opened 2025-08-24
Handling storage class specifier `thread_local`

albert-github:
The `thread_local` storage class specifier was handled differently from the other storage class specifiers (like `mutable`) this has been corrected


src/xmlgen.cpp
```diff
@@ -904,6 +904,10 @@ static void generateXMLForMember(const MemberDef *md,TextStream &ti,TextStream &
     if (md->isMutable()) t << "yes"; else t << "no";
     t << "\"";
 
+    t << " thread_local=\"";
```
doxygen: Since `thread_local` is rarely used, maybe better to only write something if it is `yes` like is done for the attributes below. That way, none of the examples should have to be updated and the size of the XML output doesn't increase with a lot of default text.

albert-github: Valid point, I'll adjust the proposed PR.

Wouldn't this be valid for `mutable` as well?

doxygen: > Wouldn't this be valid for mutable as well?

Yes, I guess so, but maybe some existing tools that parse the XML by now depend on it being always present, so probably safer to keep it as-is.

albert-github: I think those tools are wrong / its is a bug in those existing tools, as in the xsd file we see:
```
<xsd:attribute name="mutable" type="DoxBool" use="optional"/>
```

Issue #11748, opened 2025-08-25
ALIASES expansion to form @section key causes : warning: Invalid section id ' '; ignoring section


r-downham:
I would like to use ALIASES to create section/subsection keys for use with @section and @ref.
Generating via ALIASES for @ref works without issue
Same ALIASES for @section fails with : warning: Invalid section id '...'; ignoring section
Creating the @section manually with same key that would have been generated by the ALIASES works ok.

Z:\>doxygen --version
1.14.0 (cbe58f6237b2238c9af7f51c6b7afb8bbf52c866)

New/Clean project.

Only changes to Doxyfile as follows:-
```
INPUT = D:\testDoxygen
INPUT += PageOne.dox
...
ALIASES += "ezGetKey{3}=\3\1{name}\2"
ALIASES += "ezSection{2}=@ezGetKey{\1,\2,section}"
ALIASES += "ezSubSection{2}=@ezGetKey{\1,\2,subsection}"
...
```

PageOne.dox
```
/*! @page PageOne
@fileinfo  <!-- PageOne.dox : as expected  -->
@fileinfo{name} <!-- PageOne  : as expected -->

@ezGetKey{@fileinfo,TopicOne,section} <!-- sectionPageOneTopicOne : as expected -->
@ezSection{@fileinfo,TopicOne} <!-- sectionPageOneTopicOne : as expected -->

@ref @ezSection{@fileinfo,TopicOne} <!-- Works fine : as expected -->
@section @ezSection{@fileinfo,TopicOne} My Topic One <!-- (warning: Invalid section id 'sectionPageOneTopicOne'; ignoring section) -->

<!-- @section sectionPageOneTopicOne My Topic One --> <!-- Create manually works as expected -->

*/
```

albert-github:
The problem looks a bit like when are the `@fileinfo` and the `ALIASES` expanded also in related to the detection of section ids

I made a bit a smaller example:
```
/*! @page PageOne

@section @ezSection{@fileinfo,TopicOne} My Topic One

@section sectionPageOneTopicManual My Topic Manual
*/
```

and when running `doxygen -d commentcnv` we get:
```
/*! @page PageOne

@section section@fileinfo{name}TopicOne My Topic One

@section sectionPageOneTopicManual My Topic Manual
*/
```

so we see here that `ALIASES` are expanded.
When running `doxygen -d commentscan`) we get:
```
@section sectionPageOneTopicOne My Topic One

[...]

PR #11749, opened 2025-08-26
issue #11748 ALIASES expansion to form `@section` key causes : warning: Invalid section id ' '; ignoring section

albert-github:
Enabled possibility to have `@fileinfo` and `@lineinfo` in section labels.

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -177,6 +177,8 @@ static bool handleIPrefix(yyscan_t yyscanner,const QCString &, const StringVecto
 
 [[maybe_unused]] static const char *stateToString(int state);
 
+static QCString fileInfoVal(QCString name, FileInfo fi);
+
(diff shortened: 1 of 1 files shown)


src/commentscan.l
```diff
@@ -1918,25 +1920,90 @@ STopt  [^\n@\\]*
...
+                                                }
+                                                first = false;
+                                              }
+                                              else
+                                              {
+                                                warn(yyextra->fileName,yyextra->lineNr,"Unknown option specified with \\fileinfo: '{}'", optStripped);
+                                              }
+                                            }
+                                          }
+                                          if (first) // no options specified
```
doxygen, line 1961: if `@fileinfo` produces a file name with a (partial) path i.e `dir/file`, then the label will be invalid, so shouldn't the separators be escaped, for instance via `escapeCharsInString(name,true,true)`?

albert-github: It might indeed create an invalid label, I have no objections against the fact that this PR wil be changed in that way.

Issue #11751, opened 2025-08-28
Latex documentation with many functions causes pdflatex to hit memory limit


valentingehrke:
**Describe the bug**
Generating PDF/Latex documentation from sources with e.g. many functions/classes causes pdflatex to hit its resource limit (in this instance 9000 out of 9000 fonts max, see refman.log below). This is caused by a (for the lack of a better word) memory leak that seems to stem from `tabularry`'s  use of `\intarray_new:cn` which creates "arrays" that are allocated in font memory and afaict cannot be freed.

The pdflatex command terminates with the error `! Font \g__tblr_row_8920_intarray=cmr10 at 0.1365pt not loaded: Not enough room left.` where 8920 seems to refer to the 8920th table in the document.

output (cropped)
```
[891]
[892]
[893]
[894]
! Font \g__tblr_row_8920_intarray=cmr10 at 0.1365pt not loaded: Not enough room
 left.
\__intarray_new:N ...font_int \c__intarray_sp_dim 
                                                  \scan_stop: \int_step_inli...
l.51704 \end
            {DoxyParams}
? 
! Emergency stop.
\__intarray_new:N ...font_int \c__intarray_sp_dim 
                                                  \scan_stop: \int_step_inli...
l.51704 \end
            {DoxyParams}
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on refman.log.
Please consult refman.log to see the error messages
```

tail of refman.log:
```
Here is how much of TeX's memory you used:
 40361 strings out of 467890
 966063 string characters out of 5410143
 830644 words of memory out of 5000000
 57461 multiletter control sequences out of 15000+600000
 4297965 words of font info for 9000 fonts, out of 8000000 for 9000
 1141 hyphenation exceptions out of 8191
 103i,15n,123p,1499b,1419s stack positions out of 10000i,1000n,20000p,200000b,200000s
!  ==> Fatal error occurred, no output PDF file produced!
```

**Screenshots**
(no screenshots)

**To Reproduce**

1. Get [poc-memory-leak.zip](https://github.com/user-attachments/files/22028080/poc-memory-leak.zip)
2. Unpack
3. `doxygen`
4. `make -C latex`

**Expected behavior**
Good question. Technically the memory leak is in tabularry, but maybe tabularry is not the right choice for doxygen with its heavy use of tables.

**Version**
doxygen commit 0829265de0ea47bbe364525defc6a3612338663c
`doxygen -v`: `1.15.0`

[...]

PR #11753, opened 2025-08-29
issue #11751 Latex documentation with many functions causes pdflatex to hit memory limit

albert-github:
Added correction code for the tabularray version 2025A (see also discussion: https://tex.stackexchange.com/questions/750342/check-for-exact-version-of-loaded-package)


valentingehrke:
Thank you!

Issue #11760, opened 2025-09-05
\doxyconfig does not work as an if-argument


thom-sch:
**Describe the bug**
Trying to use `\doxyconfig` to circumvent the restriction of `\internal` command (which doesn't span over sections):

doxyfile:
```
INTERNAL_DOCS = YES
ENABLED_SECTIONS = INTERNAL_DOCS_YES
```
and my mainpage.md:
```
\if INTERNAL_DOCS_\doxyconfig INTERNAL_DOCS

Some internal documentation

- should be shown if INTERNAL_DOCS == YES
- but always not shown because \doxyconfig seems not to work when used as an if-argument

\endif
```

**To Reproduce**
s. attached zip file:

[Test35_doxyconfig.zip](https://github.com/user-attachments/files/22168629/Test35_doxyconfig.zip)

**Expected behavior**
May be this is the intended behavior; but I think it should work as an if-argument, too

**Version**
Version 1.14.0

**Additional context**
- `\doxyconfig` expands successfully in various use-cases (s. example)
- The documentation says: "...as used in Doxygen's configuration file...": Because first I run in little trouble it may be helpful to clarify something like: "...i.e. `\doxyconfig` doesn't expand to the respective default value but to an empty string if `<config_option>` not used in the configuration file"

Regards from Meerbusch
Thomas Schneider

albert-github:
Interesting use case, has not been considered yet. Setting issue to enhancement.
Note that not all types of configuration setting are suitable in the `\if` context.

You mentioned in the `mainpage.md`:
```
Also try ALIASES Internal1 and Internal2 instead of the if-statement or
ALIAS DoxyCfgSection{INTERNAL_DOCS} as the if-argument
```

would be nice to have these test cases in the example as well (so we know 100% sure we are talking about the same).

thom-sch:
Hello Albert,

thanks for quick response; here is the complete/enhanced example, incl. html output:

[Test35_doxyconfig_complete.zip](https://github.com/user-attachments/files/22171421/Test35_doxyconfig_complete.zip)

Regards from Meerbusch
Thomas

albert-github:
I've just pushed a proposed patch, pull request #11767

thom-sch:
Just done a little test: works fine for me.

Thanks for enhancement.

Thomas

albert-github:
I've just pushed a proposed patch, pull request #12277 taking also into consideration the `\cond` command, this proposed PR makes #11767 obsolete.

[...]

PR #11767, opened 2025-09-09
issue #11760 \doxyconfig does not work as an if-argument

albert-github:
Adding the possibility to use the `\doxyconfig` command in the section label of `\if`

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -66,6 +66,9 @@ typedef yyguts_t *yyscan_t;
 #include "trace.h"
 #include "debug.h"
 #include "stringutil.h"
+#include "config.h"
+#include "configimpl.h"
+#include "configoptions.h"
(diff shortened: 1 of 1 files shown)


doxygen:
@albert-github This makes it work for `\if` but not for `\cond`. I think it would be better to do the `\doxyconfig` handling inside the `CondParser`.

albert-github:
That is indeed the case I created a "proof of concept" working for the `\if` command and thought about adding it somewhere else in a generic way as well. I thought I would have to change the `constexpr` parser for this (and was not fond of it to do that, also seen the limited time I had at that moment).
Feel free to to create an alternative implementation.

albert-github:
A new proposed PR has been mad #12277 taking also the `\cond` command into consideration making this PR obsolete

PR #11775, opened 2025-09-20
perf(dir): avoid unnecessary string copy in isEmpty by using const reference

artyom-fedosov:
Changes:
- Avoids copying the string, reducing memory overhead and improving performance slightly.
- Behavior of the method remains unchanged.

### DIFF
--- a/src/dir.cpp
+++ b/src/dir.cpp
@@ -260,7 +260,7 @@ bool Dir::exists() const
   return fi.exists() && fi.isDir();
 }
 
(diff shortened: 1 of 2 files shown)


artyom-fedosov:
Hello @doxygen, could you check this PR? Two checks failed, but the error says "You are using macOS 13".

doxygen:
@artyom-fedosov moved the Github action to use macOS 14.

PR #11778, opened 2025-09-25
Refactor: minor code cleanups

artyom-fedosov:


src/cite.cpp
```diff
@@ -101,11 +101,8 @@ void CitationManager::insert(const QCString &label)
 
 const CiteInfo *CitationManager::find(const QCString &label) const
 {
-  auto it = p->entries.find(label.lower().str());
-  if (it!=p->entries.end())
-  {
+  if (auto it = p->entries.find(label.lower().str()); it != p->entries.end())
     return it->second.get();
```
doxygen, line 106: For readability, I prefer keeping the `{ .. }` in case the conditional statement is placed on a line by itself. Only if it naturally fit directly after the `if` the `{ .. }` can be omitted.

PR #11779, opened 2025-09-27
Improvement/code simplification

artyom-fedosov:


src/debug.cpp
```diff
@@ -153,22 +153,18 @@ DebugLex::~DebugLex()
...
-      fprintf(stderr,"%s lexical analyzer: %s\n",state, lexName);
-    }
-  }
+  if (!Debug::isFlagSet(mask)) return;
+
+  fprintf(stderr, "%s lexical analyzer: %s%s%s\n",
+          state,
+          lexName,
+          (fileName && *fileName) ? " (for: " : "",
+          (fileName && *fileName) ? fileName : "");
```
doxygen: I don't think this is quite right, the `)` is missing after the change. Probably not worth it to combine everything in one `printf`. Only moving the `isFlagSet()` check out would be enough.

src/definition.cpp
```diff
@@ -1572,13 +1572,9 @@ static QCString abbreviate(const QCString &s,const QCString &name)
...
 
-  // capitalize first word
-  if (!result.isEmpty())
-  {
-    char c=result[0];
-    if (c>='a' && c<='z') c+='A'-'a';
-    result[0]=c;
-  }
+  // capitalize first character
+  if (!result.isEmpty()) result[0] = static_cast<char>(std::toupper(result[0]));
```
doxygen: `std::toupper()` is not the same as it depends on the C locale setting, whereas the string is UTF-8 encoded.

src/doxygen.cpp
```diff
@@ -9460,15 +9460,11 @@ static void flushUnresolvedRelations()
...
 {
-  const FileDef *fd = md->getFileDef();
-  if (!fd)
-  {
-    return FALSE;
-  }
-
-  return fd->absFilePath() == root->fileName;
+  if (const FileDef *fd = md->getFileDef())
+    return fd->absFilePath() == root->fileName;
```
doxygen, line 9467: missing `{...}` here

src/util.cpp
```diff
@@ -111,35 +111,19 @@ void TextGeneratorOLImpl::writeString(std::string_view s,bool keepSpaces) const
...
-      if (c==' ')
-      {
-        m_ol.writeNonBreakableSpace(1);
-      }
-      else
-      {
-        cs[0]=c;
-        m_ol.docify(cs);
-      }
+      if (c == ' ') m_ol.writeNonBreakableSpace(1);
```
doxygen: for `if`...`else` constructs always use `{...}` for both the `if` and the `else` cases

doxygen (approved):

PR #11788, opened 2025-10-04
Improvement/remove dead code

artyom-fedosov:


src/image.h
```diff
@@ -42,7 +42,6 @@ class Image
     void fillRect(uint32_t x,uint32_t y,uint32_t width,uint32_t height,uint8_t colIndex,uint32_t mask);
     bool save(const QCString &fileName);
     friend uint32_t stringLength(const QCString &s);
-    uint32_t width() const;
     uint32_t height() const;
```
doxygen, line 46: I would think it's better here to keep `width()`, even if it is not used, in order to provide a consistent API, i.e. an image naturally has a width and a height.

PR #11796, opened 2025-10-08
Improvement/formatting

artyom-fedosov:
The `.clang-format` file usage is optional; it doesn’t enforce formatting, but it saves time and keeps code style consistent.


doxygen:
@artyom-fedosov I added some additional settings to `.clang-format` to better match doxygen's coding style (by letting copilot analyze existing files). It is still not perfect, though. So apply it to new code only.

PR #11804, opened 2025-10-11
Improvement/dependabot

artyom-fedosov:
Run nightly to avoid disturbing other PRs during the day.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,6 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
+      time: "03:00"
+    open-pull-requests-limit: 3


.github/dependabot.yml
```diff
@@ -3,4 +3,6 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
+      time: "03:00"
+    open-pull-requests-limit: 3
```
albert-github, line 8: - How does the interaction manifest itself during the day?
- Wouldn't it be possible to limit it to once a week, say Sunday at 03:00?
- Why is this limit there for 3 PR's?

artyom-fedosov: - As I understand, Dependabot runs once at 03:00 UTC and creates PRs if there are any, but does nothing during the day.
- It is possible, but I think that would be too infrequent. Note that `daily` means every weekday.
- It prevents flooding with too many PRs. Three seems like a reasonable limit.

PR #11805, opened 2025-10-12
Improvement/cmake

artyom-fedosov:
- Improved formatting
- Added extra clang-tidy checks
- Simplified the C++ standard selection logic

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,14 +88,19 @@ if(build_wizard)
   endif()
 endif()
 
-set(CMAKE_CXX_STANDARD 17)
+# use C++17 standard for compiling (unless very new Clang is present)
+if(
+  (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
+  CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 17) OR
+  (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
+  CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 19)
+)
+  set(CMAKE_CXX_STANDARD 20)
+else()
+  set(CMAKE_CXX_STANDARD 17)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS ON)
-
-check_cxx_compiler_flag("-std=c++20" COMPILER_SUPPORTS_CXX20)
-if(COMPILER_SUPPORTS_CXX20)
-  set(CMAKE_CXX_STANDARD 20)
-endif()
 
 if(ENABLE_CLANG_TIDY)
   find_program(CLANG_TIDY clang-tidy)


CMakeLists.txt
```diff
@@ -55,76 +55,68 @@ set(TOP "${PROJECT_SOURCE_DIR}")
...
-    (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
-     CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 17) OR
-    (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
-     CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 19)
-)
-    set(CMAKE_CXX_STANDARD 20)
-else()
-    set(CMAKE_CXX_STANDARD 17)
-endif()
+set(CMAKE_CXX_STANDARD 17)
```
doxygen: At this moment it is better to keep C++17 as the standard (so we also test with it and no C++20 only features creep in) and only use C++20 for the cases where it is needed (clang 19+ based dev envs). I expect your change will basically enable C++20 on all platforms we use for testing.

CMakeLists.txt
```diff
@@ -55,76 +55,68 @@ set(TOP "${PROJECT_SOURCE_DIR}")
...
-    endif()
+check_cxx_compiler_flag("-std=c++20" COMPILER_SUPPORTS_CXX20)
+if(COMPILER_SUPPORTS_CXX20)
+  set(CMAKE_CXX_STANDARD 20)
+endif()
+
+if(ENABLE_CLANG_TIDY)
+  find_program(CLANG_TIDY clang-tidy)
+  if(CLANG_TIDY)
+    set(CLANG_TIDY_CHECKS "-*,bugprone-*,clang-analyzer-*,cppcoreguidelines-*,performance-*,portability-*")
```
doxygen: I suggest keeping the commented out checks, as these are checks I tried and fixed explicitly. Enabling all checks gives too much noise in my opinion.

CMakeLists.txt
artyom-fedosov: Ok, now it seems I understand. But if we use only C++17 features in the code, we can compile the program using the C++17 or a greater version if it is supported by a compiler because C++ is backward compatible. There are some optimizations in the standard library in newer versions, for example, a lot more noexcept and constexpr functions and methods.
So I would propose this:
```cmake
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS ON)

check_cxx_compiler_flag("-std=c++26" COMPILER_SUPPORTS_CXX26)
check_cxx_compiler_flag("-std=c++23" COMPILER_SUPPORTS_CXX23)
check_cxx_compiler_flag("-std=c++20" COMPILER_SUPPORTS_CXX20)

if(COMPILER_SUPPORTS_CXX26)
    set(CMAKE_CXX_STANDARD 26)
elseif(COMPILER_SUPPORTS_CXX23)
    set(CMAKE_CXX_STANDARD 23)
elseif(COMPILER_SUPPORTS_CXX20)
    set(CMAKE_CXX_STANDARD 20)
endif()
```

albert-github: I think in this way it is still possible that non C++17 creep in when the developer has a compiler that supports a newer standard version. This might still lead to problems.

artyom-fedosov: But currently the C++ version already can be 20, so C++20 features can also creep in.
```cpp
if (
    (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
     CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 17) OR
    (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
     CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 19)
)
    set(CMAKE_CXX_STANDARD 20)
```

CMakeLists.txt
artyom-fedosov: I restricted the checks to these:
```cmake
    set(CLANG_TIDY_CHECKS "-*"
      "bugprone-use-after-move"
      "bugprone-infinite-loop"
      "bugprone-narrowing-conversions"
      "clang-analyzer-deadcode.DeadStores"
      "cppcoreguidelines-pro-type-cstyle-cast"
      "performance-unnecessary-copy-initialization"
      "performance-for-range-copy"
      "performance-move-const-arg"
      "portability-*"
    )
```
There are some warnings in the code, but not too many. These checks help to prevent silly mistakes. Can I propose them?

doxygen: yes please do

PR #11806, opened 2025-10-12
Fix/compiler warnings

artyom-fedosov:
- Fixed some compiler warnings

### DIFF
--- a/addon/doxywizard/doxywizard.cpp
+++ b/addon/doxywizard/doxywizard.cpp
@@ -807,6 +807,16 @@ void MainWindow::outputLogFinish()
 //-----------------------------------------------------------------------
 int main(int argc,char **argv)
 {
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+  static const char ENV_VAR_QT_DEVICE_PIXEL_RATIO[] = "QT_DEVICE_PIXEL_RATIO";
+  if (!qEnvironmentVariableIsSet(ENV_VAR_QT_DEVICE_PIXEL_RATIO)
+      && !qEnvironmentVariableIsSet("QT_AUTO_SCREEN_SCALE_FACTOR")
+      && !qEnvironmentVariableIsSet("QT_SCALE_FACTOR")
+      && !qEnvironmentVariableIsSet("QT_SCREEN_SCALE_FACTORS")) {
+    QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+  }
+#endif
+
   QApplication a(argc,argv);
   int locArgc = argc;
 


addon/doxywizard/doxywizard.cpp
```diff
@@ -807,14 +807,6 @@ void MainWindow::outputLogFinish()
...
 int main(int argc,char **argv)
 {
-  static const char ENV_VAR_QT_DEVICE_PIXEL_RATIO[] = "QT_DEVICE_PIXEL_RATIO";
-  if (!qEnvironmentVariableIsSet(ENV_VAR_QT_DEVICE_PIXEL_RATIO)
-        && !qEnvironmentVariableIsSet("QT_AUTO_SCREEN_SCALE_FACTOR")
-        && !qEnvironmentVariableIsSet("QT_SCALE_FACTOR")
-        && !qEnvironmentVariableIsSet("QT_SCREEN_SCALE_FACTORS")) {
-    QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
-  }
-
```
albert-github, line 817: Not sure this should / can be  removed seen the comment with the introduction:
```
Commit: e7f41104e8e3df75cfd0e654481755f8412dfaea [e7f4110]

Date: Monday, June 3, 2024 5:49:06 PM

issue #10922 Doxywizard Font Size Rendering Problem on 4k HiRes Monitors

Based on the comment https://stackoverflow.com/a/48582594/1657886 the code has been adjusted.
```

addon/doxywizard/doxywizard.cpp
```diff
@@ -63,14 +63,14 @@ MainWindow::MainWindow()
 {
   QMenu *file = menuBar()->addMenu(tr("File"));
   file->addAction(tr("Open..."),
-                  this, SLOT(openConfig()), Qt::CTRL|Qt::Key_O);
+                  this, SLOT(openConfig()), QKeySequence{ Qt::CTRL | Qt::Key_O });
```
albert-github, line 66: When has `QKeySequence` been introduced in Qt?

addon/doxywizard/wizard.cpp
```diff
@@ -533,7 +533,8 @@ Step1::Step1(Wizard *wizard,const QHash<QString,Input*> &modelData) : m_wizard(w
 
   //---------------------------------------------------
   QFrame *f = new QFrame( this );
-  f->setFrameStyle( QFrame::HLine | QFrame::Sunken );
+  f->setFrameShape(QFrame::HLine);
+  f->setFrameShadow(QFrame::Sunken);
```
albert-github, line 537: Why not use the old code here?

addon/doxywizard/doxywizard.cpp
artyom-fedosov: Sorry, I will fix that, it is necessary for Qt5. The citation from [the Qt6 documentation](https://doc.qt.io/qt-6/highdpi.html#configuring-x11): Note that **the behavior depends on the Qt version in use. Qt 5 behavior assumes that AA_EnableHighDpiScaling has been set (this flag is not needed on Qt 6).**

addon/doxywizard/doxywizard.cpp
artyom-fedosov: I think it was introduced in [Qt 3](https://documentation.help/Qt-3.0.5/qkeysequence.html#details).

addon/doxywizard/wizard.cpp
artyom-fedosov: Because compiler warns that bitwise operation between different enumeration types ('QFrame::Shape' and 'QFrame::Shadow') is deprecated [-Wdeprecated-enum-enum-conversion]

artyom-fedosov:
> In general what are the warnings? Which version(s) of Qt did you test against?

All warnings are in the commit messages. I tested the code on Qt6.

PR #11809, opened 2025-10-18
Fix/clang-tidy-warnings

artyom-fedosov:
- Removed dead code
- Improved performance
- Replaced C-style cast

### DIFF
--- a/src/dotnode.cpp
+++ b/src/dotnode.cpp
@@ -197,12 +197,11 @@ QCString DotNode::convertLabel(const QCString &l, LabelStyle style)
 {
   QCString bBefore("\\_/<({[: =-+@%#~?$"); // break before character set
   QCString bAfter(">]),:;|");              // break after  character set
-  const QCString &p(l);
-  if (p.isEmpty()) return QCString();
+  if (l.isEmpty()) return QCString();
   QCString result;
   char pc=0;
   uint32_t idx = 0;
-  int charsLeft=static_cast<int>(p.length());
+  int charsLeft=static_cast<int>(l.length());
   int sinceLast=0;
   int foldLen = Config_getInt(DOT_WRAP_THRESHOLD); // ideal text length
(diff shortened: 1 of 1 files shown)


src/debug.cpp
```diff
@@ -87,9 +87,9 @@ void Debug::print_(DebugMask mask, int prio, fmt::string_view fmt, fmt::format_a
 }
 
 static char asciiToLower(char in) {
-    if (in <= 'Z' && in >= 'A')
-        return in - ('Z' - 'z');
-    return in;
+  if (in <= 'Z' && in >= 'A')
+    return in + ' ';
```
doxygen: It is not obvious that `'z'-'Z'==' '`, so I don't think this is an improvement. Maybe `in - 'A' + 'a'` could be used instead.

src/definition.cpp
```diff
@@ -120,7 +120,7 @@ void DefinitionImpl::Private::setDefFileName(const QCString &df)
 void DefinitionImpl::Private::init(const QCString &df, const QCString &n)
 {
   setDefFileName(df);
-  QCString lname = n;
+  const QCString &lname = n;
```
doxygen: Looks like there is no reason (anymore) to have `lname`, so best use `n` everywhere instead..

src/dotnode.cpp
```diff
@@ -197,7 +197,7 @@ QCString DotNode::convertLabel(const QCString &l, LabelStyle style)
 {
   QCString bBefore("\\_/<({[: =-+@%#~?$"); // break before character set
   QCString bAfter(">]),:;|");              // break after  character set
-  QCString p(l);
+  const QCString &p(l);
```
doxygen: Seems like there is no reason to have `p` anymore, best use `l` instead everywhere.

doxygen (changes requested):

PR #11813, opened 2025-10-20
Fix/clang-tidy-warnings

artyom-fedosov:
- Improved performance
- Replaced C-style cast

### DIFF
--- a/src/searchindex_js.cpp
+++ b/src/searchindex_js.cpp
@@ -224,7 +224,7 @@ static void addMemberToSearchIndex(const MemberDef *md)
       )
      )
   {
-    QCString n = md->name();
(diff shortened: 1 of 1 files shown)


src/searchindex_js.cpp
```diff
@@ -284,42 +283,41 @@ static void addMemberToSearchIndex(const MemberDef *md)
       )
      )
   {
-    QCString n = md->name();
-    if (!n.isEmpty())
+    if (!md->name().isEmpty())
```
doxygen: I don't think this is necessarily an improvement. `n` was computed only once before and then reused, and now `md->name()` is called multiple times, which leads to code duplication, and is probably less efficient as well. Same for the case above.

PR #11825, opened 2025-10-24
Add -Wcomma option for Clang

artyom-fedosov:
- Removed duplicate option
- Added `-Wcomma` option
- Fixed some warnings

### DIFF
--- a/src/markdown.cpp
+++ b/src/markdown.cpp
@@ -768,7 +768,11 @@ size_t Markdown::Private::findEmphasisChar(std::string_view data, char c, size_t
     if (data[i]=='`')
     {
       int snb=0;
-      while (i<size && data[i]=='`') snb++,i++;
+      while (i < size && data[i] == '`')
+      {
+        snb++;
+        i++;
+      }
 
       // find same pattern to end the span
       int enb=0;
@@ -967,11 +971,13 @@ int Markdown::Private::processNmdash(std::string_view data,size_t offset)
   int count=1;
   if (i<size && data[i]=='-') // found --
   {
-    count++,i++;
+    count++;
+    i++;
   }
   if (i<size && data[i]=='-') // found ---
   {
-    count++,i++;
+    count++;
+    i++;
   }
   if (i<size && data[i]=='-') // found ----
   {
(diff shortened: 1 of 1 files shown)


cmake/CompilerWarnings.cmake
```diff
@@ -60,7 +60,7 @@ function(set_project_warnings project_name)
       -Wdouble-promotion # warn if float is implicit promoted to double
       -Wformat=2 # warn on security issues around functions that format output
                  # (ie printf)
-      -Wpedantic # warn if non-standard C++ is used
+      $<$<COMPILE_LANGUAGE:CXX>:-Wcomma>
```
albert-github, line 63: The `-Wpedantic` was added to signal a superfluous `;` after the closing `}` of a class, not sure that it is OK to remove it.

src/definition.cpp
```diff
@@ -790,13 +797,21 @@ bool readCodeFragment(const QCString &fileName,bool isMacro,
         else if (pc=='/' && c=='/') // skip single line comment
         {
           while ((c=*p++)!='\n' && c!=0);
-          if (c=='\n') lineNr++,col=0;
+          if (c == '\n')
```
albert-github, line 800: Not sure whether or not these constructs would be better readable when using:
```
if (c=='\n') { lineNr++; col=0; }
```
also in respect of using 1 versus 5 lines.

src/markdown.cpp
```diff
@@ -768,7 +768,11 @@ size_t Markdown::Private::findEmphasisChar(std::string_view data, char c, size_t
     if (data[i]=='`')
     {
       int snb=0;
-      while (i<size && data[i]=='`') snb++,i++;
+      while (i < size && data[i] == '`')
```
albert-github, line 771: see comment with a previous if statement.

cmake/CompilerWarnings.cmake
artyom-fedosov: I wrote in the commit message that it is the duplicate option (there is the same option 5 lines above)

albert-github: Sorry didn't see that, only saw a few lines, removing the duplicate is good.

PR #11835, opened 2025-10-28
issue #11831 @ref not supported in <summary> tag

albert-github:
Added possibility to use the `\ref` command in the `<summary>` tag


src/doctokenizer.h
```diff
@@ -194,6 +194,9 @@ class DocTokenizer
     void pushState();
     void popState();
 
+    int  getState();
```
doxygen: I don't think these are needed, one can use `pushState()` and `popState()` instead, which also avoids "leaking" the state information to the caller.

src/docnode.cpp
```diff
@@ -1436,6 +1436,12 @@ void DocHtmlSummary::parse()
     {
       break;
     }
+    else if (((tok.value()==TokenRetval::TK_COMMAND_AT) || (tok.value()==TokenRetval::TK_COMMAND_BS)) &&
```
doxygen: suggest to use `tok.is_any_of(TokenRetval::TK_COMMAND_AT, TokenRetval::TK_COMMAND_BS)`

src/docnode.cpp
```diff
@@ -1436,6 +1436,12 @@ void DocHtmlSummary::parse()
...
+             (Mappers::cmdMapper->map(parser()->context.token->name)== CommandType::CMD_REF))
+    {
+      parser()->handleRef(thisVariant(),children(),
+                          tok.value()==TokenRetval::TK_COMMAND_AT ? '@' : '\\',parser()->context.token->name);
```
doxygen: suggest to use `tok.command_to_char()`

albert-github:
Resurfaced again in https://github.com/CGAL/cgal/pull/6914

PR #11843, opened 2025-10-30
Replace PROJECT_SOURCE_DIR to TOP in CMake

artyom-fedosov:
There was this variable in the main CMakeLists.txt file: ```set(TOP "${PROJECT_SOURCE_DIR}")```.
This variable is shorter, so I replaced all ```PROJECT_SOURCE_DIR``` to ```TOP```.


albert-github:
Wouldn't it be better to abandon `TOP` completely in favor of `PROJECT_SOURCE_DIR` ?
(don't forget the `cmake` directory as here currently both terms are used.)

artyom-fedosov:
@albert-github, I replaced in the cmake/ directory variables also.

doxygen:
I'm also not sure about this change. `PROJECT_SOURCE_DIR` has a clearly defined meaning that people understand when reading a `CMakeLists.txt` file. For `TOP` one has to first search where it is defined to understand what it means.

artyom-fedosov:
Ok, I removed TOP at all.

PR #11851, opened 2025-11-02
Removed some dead code

artyom-fedosov:
- These functions used nowhere

### DIFF
--- a/src/growbuf.h
+++ b/src/growbuf.h
@@ -64,6 +64,7 @@ class GrowBuf
       m_str = std::exchange(other.m_str,static_cast<char*>(nullptr));
       return *this;
     }
+    void reserve(size_t size) { if (m_len<size) { m_len = size; m_str = static_cast<char*>(realloc(m_str,m_len)); } }
     void clear()      { m_pos=0; }
     void addChar(char c)  { if (m_pos>=m_len) { m_len+=GROW_AMOUNT; m_str = static_cast<char*>(realloc(m_str,m_len)); }
                         m_str[m_pos++]=c;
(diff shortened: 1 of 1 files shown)


src/arguments.h
```diff
@@ -102,10 +102,8 @@ class ArgumentList
     void push_back(const Argument &a)     { m_args.push_back(a); }
     Argument &back()                      { return m_args.back(); }
     const Argument &back() const          { return m_args.back(); }
-    Argument &front()                     { return m_args.front(); }
```
albert-github, line 105: May be leave it for consistency / symmetry with e.g. `back`

src/arguments.h
```diff
@@ -102,10 +102,8 @@ class ArgumentList
...
     const Argument &front() const         { return m_args.front(); }
     Argument &at(size_t i)                { return m_args.at(i); }
-    const Argument &at(size_t i) const    { return m_args.at(i); }
```
albert-github, line 108: May be leave it for consistency / symmetry with e.g. back

src/growbuf.h
```diff
@@ -112,11 +111,9 @@ class GrowBuf
                       addStr(tmp);
     }
     char *get() { return m_str; }
-    const char *get() const { return m_str; }
```
albert-github, line 115: Maybe leave it for consistency / symmetry.

src/xmlgen.cpp
```diff
@@ -352,12 +352,6 @@ void XMLCodeGenerator::writeLineNumber(const QCString &extRef,const QCString &co
   }
 }
 
-void XMLCodeGenerator::finish()
```
albert-github, line 355: Shouldn't also the commented line using `finish` (line 502) be removed)?

artyom-fedosov: Seems yes. I removed.

src/arguments.h
doxygen: Indeed. It is typically desirable to define both the const and non-const flavors, so the compiler can pick the right one back on the context. So treat them as "reserved for future use".

artyom-fedosov: Okay.

PR #11853, opened 2025-11-05
style: applied the clang-format to the whole code base

dropbearrob:
```
find . -type f \( -name "*.cpp" -o -name "*.h" \) -exec clang-format -i
{} +
```
.clang-format was updated in Commit 8d66827
Applying this bulk update to the style, separates the concerns and ensures that future updates which may apply formatting automatically are reviewed on technical changes only and automatic style corrections are not re-reviewed


albert-github:
Always having the same code layout is good but it is hard to enforce (as mentioned before with another PR) as not all developers have / use the clang formatter.

There are a number of places where this definitely should not be done as:
- it is external code like the things in the deps directory 
- it is example / test code like in the example and testing directories
- generated code in the vhdlparser directory there are quite a few generated (by javaCC) files

dropbearrob:
> Always having the same code layout is good but it is hard to enforce (as mentioned before with another PR) as not all developers have / use the clang formatter.
> 
> There are a number of places where this definitely should not be done as:
> 
> * it is external code like the things in the deps directory
> * it is example / test code like in the example and testing directories
> * generated code in the vhdlparser directory there are quite a few generated (by javaCC) files

If I remove the deps, test and generated files from the formatted paths would that be acceptable.

and what if I added a github action to apply the clang-format on files which aren't in this exclusion criteria.
would that address the issues you have and allow the PR to be successful?

albert-github:
> If I remove the deps, test and generated files from the formatted paths would that be acceptable.

Would certainly be necessary make it a bit acceptable

> and what if I added a github action to apply the clang-format on files which aren't in this exclusion criteria.
would that address the issues you have and allow the PR to be successful?

Would be a nice extension but I see some artifacts in the updated code that are in my opinion less desirable e.g. addon/doxyparse/doxyparse.cpp the removing of the indentation of the `#include` inside the `#if` construct. Another example is the amount of white space added in addon/doxyapp/doxyapp.cpp on the lines 55 and 57 - 75 because of the long data type with the `clone` function. When going through the code there will probably be more.

Btw / off topic do you have access to the coverity warnings: https://scan.coverity.com/projects/2860

dropbearrob:
> > If I remove the deps, test and generated files from the formatted paths would that be acceptable.
> 
> Would certainly be necessary make it a bit acceptable

ok will do. 
 
> > and what if I added a github action to apply the clang-format on files which aren't in this exclusion criteria.
> > would that address the issues you have and allow the PR to be successful?
> 
> Would be a nice extension but I see some artifacts in the updated code that are in my opinion less desirable e.g. addon/doxyparse/doxyparse.cpp the removing of the indentation of the `#include` inside the `#if` construct. Another example is the amount of white space added in addon/doxyapp/doxyapp.cpp on the lines 55 and 57 - 75 because of the long data type with the `clone` function. When going through the code there will probably be more.

are you requesting that the current clang-format be changed to address these concerns?  because this was just appllying the existing clang-format such that when I save and auto-apply the clang-format I dont re-format the file and future changes are only the functional changes for review.
> 
> Btw / off topic do you have access to the coverity warnings: https://scan.coverity.com/projects/2860

I wasn't aware of this, I will look into the coverity warnings.  To be honest I was just trying to give a little back as I was extending the use of doxygen for my own safety critical needs in my own fork (not even originally on github), so I just jumped in and hacked away which must be very frustrating for the community to teach me to suck eggs, but I'll aspire to be better.

albert-github:
> are you requesting that the current clang-format be changed to address these concerns? because this was just applying the existing clang-format such that when I save and auto-apply the clang-format I don't re-format the file and future changes are only the functional changes for review.

I just gave my opinion ad of course the source form is quite  individual.


> I wasn't aware of this, I will look into the coverity warnings. To be honest I was just trying to give a little back as I was extending the use of doxygen for my own safety critical needs in my own fork (not even originally on github), so I just jumped in and hacked away which must be very frustrating for the community to teach me to suck eggs, but I'll aspire to be better.

I thought you might not be aware of it. Any input is welcome and with this comes some effort from our side as well, but in the long run everybody will benefit from it (so no problem).

dropbearrob:
Actually I'll rename the pull request and reduce its scope.

PR #11861, opened 2025-11-08
bug 512677 New command specifying an empty comment

albert-github:
A new command `\trivial` is introduced signaling that the entity is "trivial" and does not need to be documented. Advantage is that the entity will appear in the documentation without using a setting like `EXTRACT_ALL`


doxygen:
@albert-github If you have a number of methods that are trivial, I think it rather tedious, and somewhat error prone, and ugly to have to specify 

```c++
   //! \trivial
   void foo()
   //! \trivial
   void bar()
   //! \trival
   void baz()
```
(note the typo).

So I was thinking about a simpler, more visually pleasing way. How about a single `-` to indicate no docs?

```c++
   //! -
   void foo()
   //! -
   void bar()
   //! -
   void baz()
```

Let me know what you think

albert-github:
@doxygen 
I'm not in favor of using a "shortcut" like `-`.
- it is in my opinion less clear and for the not so experienced code reader is is less clear. I prefer a bit more explicit (`\trivial`) versus implicit (`-`).
- it might lead to conflicts when parsing lists (it would be a useless list though)
-  I've also been thinking about the ellipse (`...`)  but this might be confused with other meanings of it (i.e. still to be filled in).

furthermore I think the tediousness is limited (a good editor will do the `\\! \trivial` with the same number of keystrokes as `\\! -` when adding the text (just the definition of the macro might be 7 more characters).

thom-sch:
Why to integrate this in doxygen?

A simple ALIAS like

> ALIASES = "notdocumented=`&nbsp;`"

does the job and everyone was free to choose an alias name according his requirements.

Regards from Meerbusch
Thomas

albert-github:
That came to my mind as well, but that would result in the fact that there is some empty space in output as shown in:
image

instead of:
image


(in HTML it might be possible to overcome this problem in a css file, but in other output formats this might not be possible).

dkaszews:
For skipping multiple, I think a pair of tags like `\starttrivial` and `\endtrivial` would be better instead of some shortcut syntax. Using commands is explicit and unambiguous.

dkaszews:
Also, "trivial" has special meaning in C++, so I would suggest naming the command `\empty` or `\undoc(umented)`

addon/doxmlparser/doxmlparser/compound.py
```diff
@@ -1785,44 +1791,44 @@ def validate_DoxCompoundKind(self, value):
                 lineno = self.gds_get_node_lineno_()
                 self.gds_collector_.add_message('Value "%(value)s"%(lineno)s does not match xsd enumeration restriction on DoxCompoundKind' % {"value" : encode_str_2_3(value), "lineno": lineno} )
                 result = False
-    def validate_DoxLanguage(self, value):
```
dkaszews, line 1788: Really weird diff, looks like the functions got reordered for no good reason?

dkaszews (commented):
Small notes regarding the diff

addon/doxmlparser/doxmlparser/compound.py
albert-github: `compound.py` is generated code and only present for convenience (i.e. people who do not have `generateDS` installed).

Issue #11828, opened 2025-10-24
Inconsistent vertical spacing of ordered list items in generated HTML


ipilcher:
**Describe the bug**

I have an ordered list.  Some of the list items consist of multiple paragraphs and some do not.  The vertical spacing of these items is inconsistent.  Whitespace is added below the multi-paragraph items, but it is not added below the single-paragraph items in the same list.  This makes the output less attractive and more difficult to read.  Adding additional blank lines between the single-paragraph list items has no effect on the generated HTML.

**Screenshots**

<img width="893" height="682" alt="Image" src="https://github.com/user-attachments/assets/ba1f2be3-33ab-46da-b9db-670fea7cdbd7" />

**To Reproduce**

[spacing.zip](https://github.com/user-attachments/files/23129576/spacing.zip)

```c
/**
 * @file
 *
 * The insertion algorithm operates as follows.
 *
 * 1. The ideal position of the key to be inserted (the "candidate key") is
 *    computed, and the search for a position at which the key can be stored
 *    begins at that position.  At its ideal position, the candidate's PSL is 0.
 *
 * 2. If the position being examined (the current position) is empty, the key is
 *    stored at that position and the algorithm is complete.
 *
 * 3. If the PSL of the entry at the current position is equal to the
 *    candidate's PSL, check whether the entry's key is equal to the candidate's
 *    key.
 *
 *    This check can be skipped if it is known that the candidate key is not
 *    present in the table.
 *
 *    If the keys are equal, then the candidate key is already present in the
 *    hash table at the current position.  The value associated with the key (if
 *    any) can be updated or not, depending on the needs of the program, at
 *    which point the algorithm is complete.
 *
 * 4. If the PSL of the entry at the current position is less than the
 *    candidate's PSL, the candidate is stored at the current position,
 *    displacing its current occupant.
 *
 *    a. The occupant of the current position is copied to a temporary location.
 *
 *    b. The candidate is copied to the current position.
 *
 *    c. The former occupant of the current position (in its temporary location)
 *       becomes the new candidate.
 *
 *    d. The new candidate PSL is simply the PSL of the new candidate.
 *
 *    Continue with the next step, but skip the check for duplicate keys (step
 *    3) until the algorithm is complete.
 *
 * 5. Advance to the next position in the table and increment the candidate PSL.
 *
 * 6. Return to step 2.
 */
```

**Expected behavior**

Vertical spacing between list items should be consistent.

**Version**

[...]

PR #11862, opened 2025-11-10
issue #11828 Inconsistent vertical spacing of ordered list items in generated HTML

albert-github:
The current situation is a bit inconsistent as depending on the number of paragraphs in an item the `<p class="startli">` is added or nor.

In case one of the items consists of multiple paragraphs the `<p class="startli">` is added for all paragraphs in the items in the current item block. When having maximal 1 paragraph for each item the  `<p class="startlisingle">` is added so the user can (through a css file) choose a different layout for both situations.

### DIFF
--- a/src/docnode.cpp
+++ b/src/docnode.cpp
@@ -3017,6 +3017,7 @@ Token DocAutoList::parse()
 
     children().append<DocAutoListItem>(parser(),thisVariant(),m_indent,num++);
     retval = children().get_last<DocAutoListItem>()->parse();
+    m_singleParagraph = m_singleParagraph &&  children().get_last<DocAutoListItem>()->children().size()<=1 ;
     //printf("DocAutoList::parse(): retval=0x%x parser()->context.token->indent=%d m_indent=%d "
     //       "m_isEnumList=%d parser()->context.token->isEnumList=%d parser()->context.token->name=%s\n",
     //       retval,parser()->context.token->indent,m_indent,m_isEnumList,parser()->context.token->isEnumList,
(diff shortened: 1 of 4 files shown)


doxygen:
@albert-github Looks like a bit of a complicated solution. I think the new `startlisingle` class is not really needed.

One can do something like:

```css
li {
  background: yellow;
}

li > p.startli {
  background: green;
}
```

to style the single and multi paragraph cases differently already.

albert-github:
@doxygen 

When applying the example you propose in https://github.com/doxygen/doxygen/pull/11862#issuecomment-3518514245 to the "original example"  and to situations with just single paragraphs situations and multi paragraph situations  (see example: [example.tar.gz](https://github.com/user-attachments/files/23495268/example.tar.gz)) we would get:
image

Note also the coloring in the treeview and on other places.

When using the proposed patch and changing:


```
li {
  background: yellow;
}
```
into
```
li > p.startlisingle {
  background: yellow;
}
```

We get:
image


So in my opinion:
- the usage of the `<p>` tag should be used with all text `<li>` in the auto lists (consistency as is the topic of the undelying issue #11828)
- just redefining the `<li>` does not work as this also influences other usages of the HTML `<li>` tag.
- the usages of the class name `startlisingle` is not 100% necessary and could be replaced by `startli`  but for extra flexibility I think it is better to have `startlisingle` as well

doxygen:
@albert-github I kept the consistent use of `<p>` but replaced the `startlisingle` by a CSS construct. See https://github.com/doxygen/doxygen/commit/3447a2ed8ef64e03cc7615f2c7e97c67485106f5

albert-github:
Looks like the css solution it is working as it should.

Issue #11702, opened 2025-08-05
LaTeX output does not properly break lines with long indexed parameters


JonathanBittner:
**Describe the bug**
I have several C/C++ functions with multiple parameters some of which are arrays/matrices with text string indices.  If they come near the end of the line they might not wrap properly and extend past the margin and even be cut off.

example:
```
constexpr int number_of_strings = whatever;
constexpr max_string_length = something_else;
 int function_name(int parameter1, int parameter2, std::string i_am_an_array[number_of_strings][max_string_length], ...)
```
if the line gets too long and `i_am_an_array[][]` gets placed at the end, it might not wrap.
The LaTeX output is (ignoring hyperlinks)
`i\+_\++am\+_\+\+an\+_\++array\mbox{[}\mbox{number\+_\+\+of\+_\+\+strings}\mbox{]}\mbox{[}\mbox{max\+_\+\+string\+_\+\+length}\mbox{]}`

**To Reproduce**
I can't share an example from our code base and I wanted to post this bug before moving on to other projects so I don't have a specific example. I imagine it also depends on things like paper size, font size, etc.

**Expected behavior**
There's no easy way to fix this in LaTeX since things like \sloppy, etc. can be inconsistent.  By definition, LaTeX does not allow breaking inside an mbox.  With the current code, LaTeX does not allow breaking between the  mboxes and there is opportunity to at least improve by some strategic placement of the \+ custom command already in use.  HTML (at least in Firefox) seems to allow wrapping between `][`

recommendation: include a \+ after some of the mbox commands like the ones for the opening bracket.  Taking the index name out of the mbox would probably break something else

example
`i\+_\++am\+_\+\+an\+_\++array\mbox{[}\+\mbox{number\+_\+\+of\+_\+\+strings}\mbox{]}\+\mbox{[}\mbox{max\+_\+\+string\+_\+\+length}\mbox{]}`

**Version**
1.14.0 (https://github.com/doxygen/doxygen/commit/cbe58f6237b2238c9af7f51c6b7afb8bbf52c866) Win10/x64, via download

JonathanBittner:
This only happens when PDF_HYPERLINKS = YES is on, as the mbox'es are not generated otherwise and wrapping appears to be better

albert-github:
In the example given it is not clear where the long part is (the example given is not correct either as it is missing a `;` at the end of the last line).
When adding a parameter `int parameter0` as first argument it is more clear how the problem manifests itself.
[...]

PR #11866, opened 2025-11-14
issue #11702 LaTeX output does not properly break lines with long indexed parameters

albert-github:
making the `\mbox` usage a bit more sophisticated (checking the length so long links can be wrapped) using `\doxymbox`. Based on
- idea from https://tex.stackexchange.com/a/18577/44119 , there were problems with a.o. `\hbox` in hyperlinks, so using:
  - strategy / code used for length determination of tables

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -2437,7 +2437,7 @@ void filterLatexString(TextStream &t,const QCString &str,
           else
             t << static_cast<char>(c);
           break;
-        case '#':  t << "\\#";           break;
+        case '#':  t << "\\+\\#";        break;
         case '$':  t << "\\$";           break;
         case '%':  t << "\\%";           break;
         case '^':  processEntity(t,pdfHyperlinks,"$^\\wedge$","\\string^");    break;


templates/latex/doxygen.sty
```diff
@@ -657,9 +657,21 @@
   \end{list}%
 }
 
+\newsavebox{\doxymboxbox}
+\newlength{\doxymboxlength}
+\newcommand{\doxymbox}[1]{%
+  \sbox{\doxymboxbox}{#1}%
+  \setlength{\doxymboxlength}{\wd\doxymboxbox}%
+  \ifthenelse{\doxymboxlength>\linewidth}{%
```
doxygen, line 665: Seems like this does not take into account the case where the link is not at the start of the line. 
See for example this case where there is text in front of the link:
```c++
namespace org::somenamedorg::somepackage::subpackage {
  /** my class */
  class theinterface {
    public:
      /** my function */
      void normalfunctionname(int x,short s,String s,BigInteger b);
  };
}

/** \mainpage
 *  See
 * <ul>
 * <li>This is a bit too long or not {@link org::somenamedorg::somepackage::subpackage::theinterface#normalfunctionname(int, short, String, BigInteger)} yes it might be a bit long.</li>
 * </ul>
 */
```
without the mbox the link is ok, with the doxymbox it still runs off of the page.

albert-github: I have no problem with the example when I apply the patch to the current master.

I do see a small problem with the `#` in the link, looks like this should also be `\+\#` here in the case without `\item`

Example: [example.tar.gz](https://github.com/user-attachments/files/23650741/example.tar.gz)

doxygen: I changed the example to use `::` instead of `.` last minute without retesting (was originally a Java example).
This has the problem for both links:

```c++
namespace org::somenamedorg::somepackage::subpackage {
  /** my class */
  class theinterface {
    public:
      /** my function */
      void normalfunctionname(int x,short s,String s,BigInteger b);
  };
}

/** \mainpage
 *  See
 * <ul>
 * <li>This is a bit too long or not {@link org.somenamedorg.somepackage.subpackage.theinterface#normalfunctionname(int, short, String, BigInteger)} yes it might be a bit long.</li>
 * </ul>
 *
 * This is a bit too long or not {@link org::somenamedorg::somepackage::subpackage::theinterface#normalfunctionname(int, short, String, BigInteger)} yes it might be a bit long.
 */
```
Adding a `\+` before the `#` does indeed seem to fix the problem. Probably a `\+` should also be placed before the `(` and `,` of the argument list.

albert-github: I already thought it was originally a java example (the `org::` part), looks like now they all break conveniently.

Indeed probably before the `(`, but I think that after a `)` would also be a suitable place for a `\+`.
Regarding the `,` I think the `\+` should be after the comma.

Feel free to add them.

Issue #9603, opened 2022-09-14
Unexpected warning "no matching file member found" for C macro depending on include structure


rstahn:
**Describe the bug**
We have  a simple header file `DataBase2_macros_params.h` with a simple C macro definition:
````
#ifndef DATABASE2_MACROS_PARAMS_H
#define DATABASE2_MACROS_PARAMS_H

//! \brief Macro definition
#define DB2_PARAM_DEF_SYS_VOL(name)      name

#endif
````
This header file is included in another header file `DB2_Data_Types.h`
````
#ifndef DB2_DATA_TYPES_H
#define DB2_DATA_TYPES_H

#include "DataBase2_macros_params.h"

typedef ATTRIBUTE_PACKED struct
{
  DB2_PARAM_DEF_SYS_VOL(int iTest);
} ParamSystemVOL_t;

#endif
````
The header `DB2_Data_Types.h` is included in an application header `APPL_Logging.h`:
````
#ifndef APPL_LOG_H
#define APPL_LOG_H

#include "../Extern/DataBa/DB2_Data_Types.h"

#endif
````
Finally, both `APPL_Logging.h` and `DB2_Data_Types.h`are included in an application module `APPL_Evaluation.c`:
````
#include "APPL_Logging.h"
#include "../Extern/Databa/DB2_Data_Types.h"
````
Depending on the order of the includes in `APPL_Evaluation.c`, Doxygen issues an unjustified warning:
````
DB2_Data_Types.h:8: warning: no matching file member found for DB2_PARAM_DEF_SYS_VOL(int iTest)
Possible candidates:
 '#define DB2_PARAM_DEF_SYS_VOL(name)' at line 5 of file DataBase2_macros_params.h
````

**Expected behavior**
Doxygen should be able to resolve the macro regardless of the order of the include files.

**To Reproduce**
Attached is a minimum example with a doxygen project to reproduce the problem: [C_Macro_Includes_Warning.zip](https://github.com/doxygen/doxygen/files/9567502/C_Macro_Includes_Warning.zip)
Switching the order of the includes in `APPL_Evaluation.c` removes the warning.

**Version**
We reproduced the problem with versions 1.8.20, 1.9.4 and 1.9.5 (latest). Other versions have not been tested.

rstahn:
[...]

PR #11875, opened 2025-11-24
issue #9603 Unexpected warning "no matching file member found" for C macro depending on include structure

albert-github:
Making that anonymous structs / classes etc. are recognized as a typedef and not only when there is some comment between  the words `typedef` and `struct`.)
- `scanner.l` handle `typedef` always  as a separate word (an not only when ere is some comment between  the words `typedef` and `struct`)
- `memberdef.cpp` see to that for anonymous items also the word `typedef` is shown in the output
- `111 _autolink_8cpp.xml` consequence of better handling of `typedef`
- `htmlgen.cpp` consistent handling / alignment of items


albert-github:
Note also the changes made in #11877 in respect to `memItemLeft`

Issue #11891, opened 2025-12-08
noway to control newline generation in html Variable Documentation


mlshort:
I am seeing unexpected newlines in my (html) Variable Documentation.

For Example, see the following:
![Image](https://github.com/user-attachments/assets/8ed8b531-e71e-41dd-b471-46a47604ab6a)

In my source code, I have a newline in the variable declaration for asthetics and other display limitations when viewing / editing the raw C++ code.

For Example, see the following:
![Image](https://github.com/user-attachments/assets/47686963-1b19-471f-b6b6-a2afe3ddc65e)

**To Reproduce**
Easy to reproduce and appears to be the default behavior.  However I don't remember this being the case in an older version several years back.

**Expected behavior**
I would like to see the Variable Documentation not carry forward the newline from the source code.
In the generated html documentation, I would like to see the **Variable Initial:** value be displayed as:

`= ( SPI_CR_MASTER | SPI_CR_MODE3 | SPI_CR_BAUDDIV_4 | SPI_CR_SS_NONE | SPI_CR_EN_MANUAL_START_ENABLE )`

NOT

```
= ( SPI_CR_MASTER | SPI_CR_MODE3 |
                                                       SPI_CR_BAUDDIV_4 | SPI_CR_SS_NONE | SPI_CR_EN_MANUAL_START_ENABLE )
```

**Version**
Windows 11, 64 bit).
1.14.0 (cbe58f6237b2238c9af7f51c6b7afb8bbf52c866)

**Stack trace**
not needed

**Additional context**

albert-github:
> Easy to reproduce and appears to be the default behavior. However I don't remember this being the case in an older version several years back.

So to be sure that we are talking about the same problem:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem?
   Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

mlshort:
Seriously?

Fine.  Please find the following attachment of c++ source file containing the variable declaration + Doxygen configuration file.

[Lab5.zip](https://github.com/user-attachments/files/24031248/Lab5.zip)

albert-github:
> Seriously?

Yes, In a lot of cases the used settings have an influence and also it is not possible to cut and paste from an image so `constexpr uint32_t SPI_CFG_AutoCS_ManualSTART_NoSS` would have to been typed by hand.
[...]

PR #11894, opened 2025-12-09
issue #11891 noway to control newline generation in html Variable Documentation

albert-github:
The initial part of the initializing of a variable i.e. the part before the `=` sign was omitted from the initializer and hence a bit a strange indentation.
- `scanner.l` add the initial part as white space
- `memberdef.cpp` remove the initial common whitespace of all initializer lines
- `memberdef.cpp` in `setAnchor` the extra white space has to be added to remain compatible (otherwise the anchor will change, tests 59, 72, 104, 105)

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -4548,30 +4548,7 @@ void MemberDefImpl::setInitializer(const QCString &initializer)
   while (p>=0 && isspace(static_cast<uint8_t>(m_initializer.at(p)))) p--;
   m_initializer=m_initializer.left(p+1);
   m_initLines=m_initializer.contains('\n');
-  QCString s(m_initializer);
-  const char *data = s.data();
-  size_t i = 0;
-  size_t col = 0;
-  size_t size = s.length();
-  m_initializer.clear();
-  while (i<size)
-  {
-    char c = data[i++];
-    switch(c)
-    {
-      case '\n':
-        m_initializer+=c;
-        col=0;
-        break;
-      case ' ':
-        if (col < indent) col++;
-        else m_initializer+=c;
-        break;
-      default:
-        m_initializer+=c;
-        break;
-    }
-  }
+  stripIndentationVerbatim(m_initializer,indent, false);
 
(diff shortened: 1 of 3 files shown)


src/memberdef.cpp
```diff
@@ -4395,13 +4395,12 @@ void MemberDefImpl::setAnchor()
   if (!m_args.isEmpty()) memAnchor+=m_args;
   if (m_memSpec.isAlias()) // this is for backward compatibility
   {
-    memAnchor.prepend(" = "+m_initializer);
+    memAnchor.prepend(" =  "+m_initializer);
```
doxygen, line 4398: Is this extra space intentional?

src/memberdef.cpp
```diff
@@ -4549,6 +4548,31 @@ void MemberDefImpl::setInitializer(const QCString &initializer)
   while (p>=0 && isspace(static_cast<uint8_t>(m_initializer.at(p)))) p--;
   m_initializer=m_initializer.left(p+1);
   m_initLines=m_initializer.contains('\n');
+  QCString s(m_initializer);
```
doxygen: Doesn't this do the same as `stripIndentationVerbatim()` in `util.cpp`?

src/memberdef.cpp
albert-github: Yes, see comment with PR:
> `memberdef.cpp` in `setAnchor` the extra white space has to be added to remain compatible (otherwise the anchor will change, tests 59, 72, 104, 105)

src/memberdef.cpp
albert-github: The comment with the function says:
> (excluding the first line)

and later on:
> skip the initial line from stripping


So I don't think this is the case, the function could be enhanced in a way that also the first line is considered (e.g by an argument `bool stripFirstLine = false` that in this case can be set to `true`)

Feel free to adjust this.

albert-github: Used extended `stripIndentationVerbatim` based on review

PR #11899, opened 2025-12-15
Correct line numbers for warning in perlmod output

albert-github:
The line numbers in the perlmod output were based on the start line of the definition although this should have been to to the start of the brief / detailed description (as done at other places).

When having (and also the appropriate images):
```
/** \file
  *
  * the base first graph
  * \dotfile base.ex
  *
  * the spec1 first graph
  * \dotfile spec1/base.ex
  *
  * the spec2 first graph
  * \dotfile spec2/base.ex
  *
  * do an inline as well
  */
```

we get as warnings:
```
.../aa.c:4: warning: included dot file name 'base.ex' is ambiguous.
Possible candidates:
  .../base.ex
  .../spec1/base.ex
  .../spec2/base.ex
.../aa.c:2: warning: included dot file name 'base.ex' is ambiguous.
Possible candidates:
  .../base.ex
  .../spec1/base.ex
  .../spec2/base.ex
```
instead of just the warning for line 4.

Example: [example.tar.gz](https://github.com/user-attachments/files/24164058/example.tar.gz)

### DIFF
--- a/src/perlmodgen.cpp
+++ b/src/perlmodgen.cpp
@@ -1520,8 +1520,8 @@ void PerlModGenerator::generatePerlModForMember(const MemberDef *md,const Defini
     .addFieldQuotedString("protection", getProtectionName(md->protection()))
     .addFieldBoolean("static", md->isStatic());
 
-  addPerlModDocBlock(m_output,"brief",md->getDefFileName(),md->briefLine(),md->getOuterScope(),md,md->briefDescription());
-  addPerlModDocBlock(m_output,"detailed",md->getDefFileName(),md->docLine(),md->getOuterScope(),md,md->documentation());
+  addPerlModDocBlock(m_output,"brief",md->briefFile(),md->briefLine(),md->getOuterScope(),md,md->briefDescription());
+  addPerlModDocBlock(m_output,"detailed",md->docFile(),md->docLine(),md->getOuterScope(),md,md->documentation());
   if (md->memberType()!=MemberType::Define &&
       md->memberType()!=MemberType::Enumeration)
     m_output.addFieldQuotedString("type", md->typeString());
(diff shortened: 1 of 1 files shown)


src/perlmodgen.cpp
```diff
@@ -1520,8 +1520,8 @@ void PerlModGenerator::generatePerlModForMember(const MemberDef *md,const Defini
     .addFieldQuotedString("protection", getProtectionName(md->protection()))
     .addFieldBoolean("static", md->isStatic());
 
-  addPerlModDocBlock(m_output,"brief",md->getDefFileName(),md->getDefLine(),md->getOuterScope(),md,md->briefDescription());
-  addPerlModDocBlock(m_output,"detailed",md->getDefFileName(),md->getDefLine(),md->getOuterScope(),md,md->documentation());
+  addPerlModDocBlock(m_output,"brief",md->getDefFileName(),md->briefLine(),md->getOuterScope(),md,md->briefDescription());
```
doxygen: If we change this to use `md->briefLine()` then also `md->getDefFileName()` should be changed to  `md->briefFile()` (brief description can be in a header and `getDefFileName()` can point to the source file).

src/perlmodgen.cpp
```diff
@@ -1520,8 +1520,8 @@ void PerlModGenerator::generatePerlModForMember(const MemberDef *md,const Defini
...
+  addPerlModDocBlock(m_output,"detailed",md->getDefFileName(),md->docLine(),md->getOuterScope(),md,md->documentation());
```
doxygen: If we change this to use `md->docLine()` then also `md->getDefFileName()` should be changed to  `md->docFile()` (detailed description can be in a header and `getDefFileName()` can point to the source file).

src/perlmodgen.cpp
albert-github: Indeed

PR #11901, opened 2025-12-15
GroupDef::sortSubGroups() should not be gated by SORT_BRIEF_DOCS

kartben:
Fixes a regression introduced with commit 9cb220d whereby subgroup sorting was gated using `SORT_BRIEF_DOCS`. This is incorrect as sortSubGroups() is called in the context of sorting member groups, which is gated by `SORT_GROUP_NAMES` already, and there should not be an additional/different option being introduced.

### DIFF
--- a/src/groupdef.cpp
+++ b/src/groupdef.cpp
@@ -1992,13 +1992,10 @@ void GroupDefImpl::removeMemberFromList(MemberListType lt,MemberDef *md)
 
 void GroupDefImpl::sortSubGroups()
 {
-  if (Config_getBool(SORT_BRIEF_DOCS))
-  {
-    std::stable_sort(m_groups.begin(),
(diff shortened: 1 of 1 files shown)


kartben:
@albert-github any chance this can make it in a 1.15.1? This is IMO an important enough regression that it would make sense to address it in a maintenance release? Or maybe 1.16.0 is already around the corner?
Thanks!

albert-github:
The merged fix is in the 1.16.0 master version and will be in the future 1.16.0 release.
I'm not the one that decides when a new release will be made. I know it always takes a bit of time to create the release and especially building the executables for the different platforms. For the different platform builds some initiatives are going on regarding building the different required libraries in a static way (see https://github.com/doxygen/doxygen-release).

PR #11902, opened 2025-12-17
Protect counter with mutex

albert-github:
Protect counter of the inline defined images (`\dot` and \msc` commands) with a mutex.

### DIFF
--- a/src/dot.cpp
+++ b/src/dot.cpp
@@ -279,6 +279,7 @@ void writeDotGraphFromFile(const QCString &inFile,const QCString &outDir,
  *  \param graphId a unique id for this graph, use for dynamic sections
  *  \param srcFile the source file
  *  \param srcLine the line number in the source file
+ *  \param newFile signal whether or not the file has been generated before (value `false`) or not.
  */
 void writeDotImageMapFromFile(TextStream &t,
                             const QCString &inFile, const QCString &outDir,


src/docbookvisitor.cpp
```diff
@@ -378,16 +383,18 @@ DB_VIS_C
...
-        static int dotindex = 1;
         QCString baseName(4096, QCString::ExplicitSize);
         QCString name;
         QCString stext = s.text();
         m_t << "<para>\n";
-        name.sprintf("%s%d", "dot_inline_dotgraph_", dotindex);
-        baseName.sprintf("%s%d",
+        {
+          std::lock_guard<std::mutex> lock(dotindex_mutex);
+          name.sprintf("%s%d", "dot_inline_dotgraph_", dotindex);
```
doxygen: Protecting the counter is a good idea, but the use of a counter itself is already problematic, as it can result in different output when running Doxygen multiple times on the same input. The same count value can be assigned to a different graph if the order in which the graphs are processed changes due to a different thread scheduling order.

albert-github: I agree but the only solution I see here is to use a file name including a md5 hash and a "list" so that files with the same md5 hash (i.e. the same content) will only be rendered once. But with the md5 has we won;t need a counter anymore but should place the mutex over the list.
In principle I don't think it is very problematic that there are different numbers although some test strategies between different builds think otherwise.

@doxygen What is your suggestion to do?

doxygen:
@albert-github I noticed quite some code duplication (8x more or less the same code), so I moved it to a separate function. Please check if I didn't break anything in the process.

albert-github:
I checked and found a small problem when `DOT_CLEANUP=YES` as I get the warnings:
```
.../aa.c:17: error: Problems running dot: exit code=2, command='dot.exe', arguments='".../doc/html/inline_dotgraph_4b689ea5cd4847e4f5ac9ff55d29fff0.dot" -Tcmapx -o ".../doc/html/dot_inline_dotgraph_4b689ea5cd4847e4f5ac9ff55d29fff0.map"'
Failed to open input file '.../doc/html/inline_mscgraph_7c6951245abdf3122b13597d42cd76cb.msc'
.../bb.c:15: error: Problems generating msc output (error=FILE ERROR). Look for typos in you msc file '.../doc/html/inline_mscgraph_7c6951245abdf3122b13597d42cd76cb.msc'
```
Probably the line like:
```
if (Config_getBool(DOT_CLEANUP)) Dir().remove(fileName.str());`
```
should not be in the `*docvisitor.cpp` files but should be in a cleanup phase after `g_outputList->cleanup();` in `doxygen.cpp` by running over the `static StringUnorderedSet writeFileContents_set;`  when the `DOT_CLEANUP` is set.

@doxygen 
- what do you think?
- will you implement this or should I give it a try?


Example: [example.tar.gz](https://github.com/user-attachments/files/24271079/example.tar.gz)

albert-github:
In the mean time I created a proposed pull request #11906

albert-github:
- Code of #11906 has been integrated in master on GitHub.

PR #11909, opened 2025-12-24
Special one line CPP comment in string

albert-github:
When having  special one line comment (`//!` or `///`) inside a string the rule `<*>{CPPC}[!/][^\n]*` would kick in and due to its greediness an error like:
```
input buffer overflow, can't enlarge buffer because scanner uses REJECT
    lexical analyzer: .../doxygen/src/code.l (for: Chart_bundle_js.cpp)
```
can appear when the, remainder, of the string is to large to fit in the lex buffer.

By specifying all states except those that were already rejected this problem can be overcome.

Example: [example.tar.gz](https://github.com/user-attachments/files/24332199/example.tar.gz)

(Found by Fossies for the proxysql package)

### DIFF
--- a/src/code.l
+++ b/src/code.l
@@ -2063,8 +2063,7 @@ ENDQopt ("const"|"volatile"|"sealed"|"override")({BN}+("const"|"volatile"|"seale
                                           codifyLines(yyscanner,yytext);
                                           endFontClass(yyscanner,true);
                                         }
-<*>{CPPC}[!/][^\n]*                       { // strip special one-line comment
-                                          if (YY_START==SkipComment || YY_START==SkipString) REJECT;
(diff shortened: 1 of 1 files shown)


src/code.l
```diff
@@ -2063,8 +2063,7 @@ ENDQopt ("const"|"volatile"|"sealed"|"override")({BN}+("const"|"volatile"|"seale
                                           codifyLines(yyscanner,yytext);
                                           endFontClass(yyscanner,true);
                                         }
-<*>{CPPC}[!/][^\n]*                       { // strip special one-line comment
-                                          if (YY_START==SkipComment || YY_START==SkipString) REJECT;
+<SkipStringS,SkipVerbString,SkipCPP,SkipCxxComment,Body,BodyVar,FuncCall,MemberCall,MemberCall2,SkipInits,ClassName,AlignAs,AlignAsEnd,PackageName,ClassVar,CppCliTypeModifierFollowup,Bases,SkipSharp,ReadInclude,TemplDecl,TemplCast,CallEnd,ObjCMethod,ObjCParams,ObjCParamType,ObjCCall,ObjCMName,ObjCSkipStr,ObjCCallComment,OldStyleArgs,ConceptName,UsingName,RawString,InlineInit,ModuleName,ModuleImport>{CPPC}[!/][^\n]*                       { // strip special one-line comment, all states with the exception of SkipComment and SkipString which are handled separately
```
doxygen, line 2066: I'm afraid this will make things hard to maintain (easy to forget adding a new state here as well). Next to that, it doesn't really fix the problem if the issue happens for any of these states.

albert-github: That is true, though I only saw this in these cases (so far). I can imagine that there might be other states as well but due to the lack of an example it is hard to fix.

Unfortunately lex doesn't have a possibility of "all states except".

Fixing the real problem would be quite a big job (and at the moment probably not worth the effort).

PR #11910, opened 2025-12-25
issue #11828 Inconsistent vertical spacing of ordered list items in generated HTML

albert-github:
The current situation is a bit inconsistent as depending on the number of paragraphs in an item the `<p class="startli">` is added or not.

In case one of the items consists of multiple paragraphs the `<p class="startli">` is added for all paragraphs in the items in the current item block. Now the `<p class="startli">` is always added, though this gave some problems with checked / unchecked lists, this has been overcome by moving the "check" marker to the `<ul>`, so it is now the list marker and not the item marker, hence also for each item an own list and a correction regarding the margins.

### DIFF
--- a/src/htmldocvisitor.cpp
+++ b/src/htmldocvisitor.cpp
@@ -975,11 +975,7 @@ void HtmlDocVisitor::operator()(const DocAutoList &l)
   }
   else
   {
-    if (l.isCheckedList())
-    {
-      m_t << "<ul class=\"check\">";
-    }
-    else
+    if (!l.isCheckedList())
     {
       m_t << "<ul>";
     }
@@ -992,7 +988,10 @@ void HtmlDocVisitor::operator()(const DocAutoList &l)
   }
   else
   {
-    m_t << "</ul>";
+    if (!l.isCheckedList())
+    {
+      m_t << "</ul>";
+    }
   }
   if (!l.isPreformatted()) m_t << "\n";
   forceStartParagraph(l);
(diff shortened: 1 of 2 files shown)


src/htmldocvisitor.cpp
```diff
@@ -1001,21 +1000,29 @@ void HtmlDocVisitor::operator()(const DocAutoList &l)
 void HtmlDocVisitor::operator()(const DocAutoListItem &li)
 {
   if (m_hide) return;
-  switch (li.itemNumber())
+  const auto docAutoList= std::get_if<DocAutoList>(li.parent());
+
+  if (docAutoList && docAutoList->isCheckedList())
```
doxygen, line 1005: doesn't this wrap each list item inside its own list of one? And if not, then how to deal with some list items being checked and others not, if the "checked" property is moved to the list level?

albert-github: Yes it does wrap each check list item in its own list and in this way it is possible to set for each item its own "character" (i.e `\2611\A0` etc.)
In a check list items are either "checked" or "unchecked" and both are handled.

doxygen: @albert-github I think this is still a bit of a hack. The output of a list of items should also be a single list in my opinion, otherwise styling it by a custom CSS would become very difficult. I think it should be possible to style a list properly. I made this small prototype to test it out: https://gist.github.com/doxygen/5043aac1cd0dfd834eb39dba21527d0a. I'll see if I can apply this to Doxygen's output.

albert-github: @doxygen 

It is indeed a bit of a hack and a single list would definitely be preferred.
The method should also be tested properly with long list items (a quick test didn't show problems though).
Did you find this method somewhere on the net?

doxygen: @albert-github I used [copilot](https://github.com/copilot) to generate the initial example. Was not 100% perfect but got me in the right direction.

albert-github: OK I see, never used copilot, maybe I should read-up a bit on it.

doxygen: Can you check if https://github.com/doxygen/doxygen/commit/bca0c6ef0cc4e42f6b5e79dad52767323acb1b93 already sufficiently solves the problem?

albert-github: I tested it yesterday already with https://github.com/doxygen/doxygen/commit/bca0c6ef0cc4e42f6b5e79dad52767323acb1b93 and the presented results (https://github.com/doxygen/doxygen/issues/11828#issuecomment-3702255340) were from that version, I now retested again with 1.16.0 (da1b04381f4d7f3831d5a1d8c5600e8e6dbfffc2) and the results are still the same.

doxygen: I still fail to see the problem.

albert-github: I'll see if I can make it a bit clearer in a small comprehensive example (will take a little while).

albert-github: When we have (test.h):
```
/**
 * @file
 *
 * Example
 *
 * 1. Single paragraph
 *
 * 2. multi paragraph
 *
 *    multi paragraph
 *
 *    multi paragraph
 *
 */
```
and for convenience the coloring css settings:
```
li {
  background: red;
}

li > p.startli {
  background: green;
}
```
this results in:
image

in other words the list item with the single paragraph is colored (red) like the default `<li>` tag  and the multi paragraphs are colored (green) like the specified class for the <`>` tag with the `<li> tag.

when looking in the html/test_86.html we see:
```
<ol type="1">
<li>Single paragraph</li>
<li><p class="startli">multi paragraph</p>
<p class="startli">multi paragraph</p>
<p class="startli">multi paragraph </p>
</li>
</ol>
```

when we would change:
```
<li>Single paragraph</li>
```
into
```
<li><p class="startli">Single paragraph</p></li>
```

the problem would be solved and we would have a more consistent output.
Just setting the color for the default `<li>`  to `green` would not solve the issue as this would also influence other `<li>` tags.

Example: [example.tar.gz](https://github.com/user-attachments/files/24452966/example.tar.gz)


I hope this shines a bit light on the problem I see.

Issue #11624, opened 2025-06-18
Classes always ordered alphabetically


isaac-blanc:
**Describe the bug**
Python classes are ordered alphabetically, rather than by declaration order, even when all SORT configs are turned off.

**Screenshots**

The following module renders as shown below.

```
"""Demo module"""

class C():
    pass

class B():
    pass

class A():
    pass
```

![Image](https://github.com/user-attachments/assets/16a2e4f8-ec58-4f46-8d1c-426537f6e664)

**To Reproduce**
Demo attached.

**Version**
Doxygen 1.14.0 (e0653197fd6bf9800f708ef02e09094d91778445*)
Windows 11 Pro 64-bit

[demo.zip](https://github.com/user-attachments/files/20793511/demo.zip)

albert-github:
As far as I know this is by design. 
By coincidence a nice related discussion about the sorting is https://github.com/doxygen/doxygen/discussions/11616 where the opposite is requested (though for concepts).

isaac-blanc:
Ah ok, well not to worry then. Thanks all the same for the reply!

doxygen:
@albert-github The sorting of members is controlled by `SORT_MEMBER_DOCS`. One could argue that the list of classes as shown within a module, namespace, group, or file page is also a list of 'members' of that symbol's page. It is a bit of an implementation detail that it is now always sorted because internally it is not a "member list" type.

This is different from #11616 where the global list of concepts is not sorted as it should. 

So in my view global lists should always be sorted, whereas the sorting of lists inside the overview of the symbol pages should always be controlled by `SORT_MEMBER_DOCS`.

albert-github:
Indeed the problem is "what is a member" though in the mentioned concept case (#11616) the concepts are, as far as I understand it, also not global as they are part of a namespace so comparable (in my opinion) to classes that are part of a namespace. Both should be handled in a consistent way.

One thing one should not forget is the fact that when using `NUM_PROC_THREADS` with a value not equal to `1` the result might not be deterministic anymore.

doxygen:
@isaac-blanc Please verify if the referenced commit fixes the problem for you. Do not close the issue, this will be done automatically when the next official release becomes available. 

Note that the toggle to use is `SORT_BRIEF_DOCS` and not `SORT_MEMBER_DOCS`, which is for the detailed section only.

isaac-blanc:
Just checked and can confirm this commit fixes the problem for me. Thank you!

[...]

Issue #8425, opened 2021-03-12
Would like a way to sort class list but not brief member list


kmccarty-dgi:
**Describe the bug**
In Doxygen 1.6.1, the default observed behavior was to sort the list of classes alphabetically on the class list page; and on each individual class doc page, to leave the brief member descriptions in the same order as they were in the header file, while alphabetically sorting the long member descriptions.

In Doxygen 1.9.1, to which I'm trying to have my workplace upgrade, the class list is no longer sorted alphabetically by default.  Now it seems to be in a more or less random order (maybe this is the order in which the source files are listed, but that isn't at all obvious on the generated HTML).  In a project with many different classes, where multiple classes are defined in many of the headers (some of which have names unrelated to the header file names), this makes the class listing very confusing.

***This seems to be a problem only for classes inside an enclosing C++ namespace.***

Setting SORT_BRIEF_DOCS = YES causes the class list to be sorted alphabetically as before.  But this also has a side effect of alphabetically sorting the brief member list on each class's individual page.  This is not what I want, since the class members are declared in a logical order in our source header files and I'd like them to remain in that order in the HTML for the sake of clarity.

Also setting SORT_MEMBER_DOCS = NO causes the *full* member descriptions to no longer be alphabetically sorted -- but the brief member descriptions are still alphabetically sorted, so this doesn't really help me.

The setting of SORT_BY_SCOPE_NAME doesn't seem to have any bearing on this behavior.

**Expected behavior**
What I'd really like is a way to have the class list be alphabetically sorted, while the brief member descriptions appear in the same order as in the source header file.  (I don't care too much what order the full member descriptions are in, although I'd have a mild preference for alphabetical.)

Is there a way to accomplish this now (without making major changes to our source code)?  If not, could a new setting that would allow this be added?

Thanks so much!
Kevin

kmccarty-dgi:
N.B.  The random ordering seems to happen only for classes enclosed in a C++ namespace.  I'll attach an example shortly.

kmccarty-dgi:
Sample project exhibiting the problem:

[libfoobar.tar.gz](https://github.com/doxygen/doxygen/files/6131803/libfoobar.tar.gz)

albert-github:
What is the difference between First.h and Last.h except for the name of the struct?

kmccarty-dgi:
There is no difference in the struct contents -- the point of my example was to illustrate that an untoward sort ordering is obtained.

[...]

PR #11930, opened 2026-01-07
Add option to control list sorting in treeview

tyler-yankee:
Commit 9cb220d forced `SORT_BRIEF_DOCS` to control sorting of both the briefs as well as the lists in the treeview (when `GENERATE_TREEVIEW` is enabled). This commit makes those two behaviors distinct by adding a new option.


tyler-yankee:
Hello! I help maintain the [Drake](https://drake.mit.edu/) project, and we use Doxygen for our [C++ documentation](https://drake.mit.edu/doxygen_cxx/index.html).

On our website, we've always had the navigation and members sorted, but not the briefs (i.e. `SORT_MEMBER_DOCS = YES` `SORT_BRIEF_DOCS = NO` ). For our use case, this is the best combination of navigability+usability of the page, while also maintaining somewhere a logical ordering of the code authors' intention when declaring members of their class. When upgrading to Doxygen 1.15, we noticed via the linked commit (and subsequently linked issue) that `SORT_BRIEF_DOCS` now also controls sorting in the treeview/navigation, which for us represents two distinct behaviors, only one of which we want.

I didn't update any of the test cases in this PR as I'd first like guidance as to whether this is a welcome change. Thanks!

doxygen:
> I didn't update any of the test cases in this PR as I'd first like guidance as to whether this is a welcome change. 

@tyler-yankee Although I try to limit the number of options, I think this may have some value.

tom-osika:
Hello all. I work alongside @tyler-yankee on Drake. To add on to what Tyler has mentioned, I noticed the following behavior when testing out these changes (although it was observed on master as well).

With myclasses.h:

```cpp
class C {};
class B {};
class A {};
```

Generating the documentation with all sorting options turned **off**, I see that A, B, and C are listed in alphabetical order instead of declaration order (see the attached screenshot).

Screenshot from 2026-01-05 17-01-47


The sorting behaves as expected when moving these classes into a namespace.

Is this a case of "[global lists should always be sorted](https://github.com/doxygen/doxygen/issues/11624#issuecomment-2992614856)", or is there something else going on here? Thank you.

doxygen:
> @tom-osika Is this a case of "https://github.com/doxygen/doxygen/issues/11624#issuecomment-2992614856", or is there something else going on here?

Yes, this list is created from all classes found in the input. There is no clear order in that case (these classes can come from any file), and then sorting them in alphabetical order makes it easier to find a particular class when the list becomes longer.

tom-osika:
> Yes, this list is created from all classes found in the input. There is no clear order in that case (these classes can come from any file), and then sorting them in alphabetical order makes it easier to find a particular class when the list becomes longer.

Great, thanks for the clarification!

tyler-yankee:
Hi @doxygen, I believe the one CI failure is unrelated to my change, but please correct me if I'm wrong. Otherwise, I guess it looks like this behavioral change doesn't fail any existing tests, but let me know if you would like any new tests added to cover this. Thanks.

src/config.xml
```diff
@@ -1162,8 +1162,18 @@ Go to the <a href="commands.html">next</a> section or return to the
...
+]]>
+      </docs>
+    </option>
+    <option type='bool' id='SORT_NAVIGATION_LISTS' defval='0'>
+      <docs>
+<![CDATA[
+ If the \c SORT_NAVIGATION_LISTS tag is set to \c YES then Doxygen will sort the
+ file, namespace and class lists in the navigation tree alphabetically by
+ member name. If set to \c NO, the members will appear in declaration order.
+ \note If \ref cfg_generate_treeview "GENERATE_TREEVIEW" is set to \c NO
```
doxygen, line 1175: This statement caught my eye. And given the way it is now implemented, this doesn't seem to be the case.

Take for example a `test.cpp` file with:
```c++
/** @file */

/** class Z */
class Z {};
/** class A */
class A {};
/** class M */
class M {};
/** funcZ */
void funcZ();
/** funcA */
void funcA();
/** funcM */
void funcM();
```
If you look at the file documentation. The order of the list of classes on the page is now controlled via `SORT_NAVIGATION_LISTS` whereas the function order is controlled via `SORT_BRIEF_DOCS`.
So the effect is not limited to the navigation lists as it should, but also to the page contents.

doxygen (changes requested):

src/config.xml
tyler-yankee: Sorry for delayed response. I see what you're saying, and can reproduce.

I'm struggling to determine how this PR should be adapted accordingly. I don't see where the treeview is actually constructed from the list of members, classes, namespaces etc. Do we need to keep multiple copies of the lists to control the sorting in a more granular way?

I've looked at backing out my changes in `filedef.cpp` (likewise for others except `namespacedef.cpp`), and that resolves the issue you've pointed out here, but I'm not sure if it maintains the intention of keeping the nav list, briefs, and members sorted with distinct controls.

tyler-yankee: Consider the following `test.cpp` as a more expansive test example:

```
/** @file */

/** class Z */
class Z {};
/** class A */
class A {};
/** class M */
class M {};
/** funcZ */
void funcZ();
/** funcA */
void funcA();
/** funcM */
void funcM();

/** namespace alpha */
namespace alpha {
    /** class B */
    class B {};
    /** class C */
    class C {};
    /** class D */
    class D {};
}

/** namespace beta */
namespace beta {
    /** class E */
    class E {};
    /** class F */
    class F {};
    /** class G */
    class G {};
}

```

tyler-yankee: @doxygen I've pivoted the implementation to address this issue, PTAL. thanks!

tyler-yankee:
Hi @doxygen, have you had a chance to look at the updated implementation on this? Happy to discuss further as needed, thanks.

tyler-yankee:
Hi @doxygen, just checking in to see the status of the latest changes, thanks.

tyler-yankee:
Actually, is this PR a duplicate of https://github.com/doxygen/doxygen/commit/75ac8423bad7b0a6ff6d1262afc578e1b0ead1a1 (#8425)?

albert-github:
@tyler-yankee 

Could you please test the current doxygen version (1.17.0) that was released on April 30, 2026 as here the alternative mentioned in https://github.com/doxygen/doxygen/issues/8425#issuecomment-3910367781

tyler-yankee:
Yes, I believe this fixes our issue, so I'll close this PR. thank you!

PR #11944, opened 2026-01-12
Improve filename matching

sebhub:
This patch set improves to items:

1. In the generated tagfile, the files of a group are fully specified. This helps in case filenames alone are ambiguous.
2. In `findFileDef()`, if multiple files have the same name, then return a unique match in case the stripped file paths are equal (perfect match). This helps if a project uses includes like this `#include <s/h.h>` and `#include <t/s/h.h>` where the `h.h` files are distinct files.

### DIFF
--- a/src/groupdef.cpp
+++ b/src/groupdef.cpp
@@ -943,11 +943,11 @@ void GroupDefImpl::writeFiles(OutputList &ol,const QCString &title)
       if (!fd->hasDocumentation()) continue;
       ol.startMemberDeclaration();
       QCString anc = fd->anchor();
-      if (anc.isEmpty()) anc=fd->displayName(); else anc.prepend(fd->displayName()+"_");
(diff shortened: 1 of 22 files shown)


vtjnash:
Would this fix LLVM's documentation, where many of the links on https://llvm.org/doxygen/classllvm_1_1Instruction.html now are getting overwritten to SandboxIR/Instruction.cpp instead of IR/Instruction.cpp?

sebhub:
> Would this fix LLVM's documentation, where many of the links on https://llvm.org/doxygen/classllvm_1_1Instruction.html now are getting overwritten to SandboxIR/Instruction.cpp instead of IR/Instruction.cpp?

There is a chance, that this issue is addressed by the patch set. I would rebuild Doxygen with the patch set and try it out.

vtjnash:
Yes, I confirmed that it looks like the output is fixed with your patch

Issue #11939, opened 2026-01-08
LaTeX output does not properly break lines with long doxylink


hachso:
**Describe the bug**
LaTeX output does not properly break lines with long doxylink

**Screenshots**

<img width="794" height="208" alt="Image" src="https://github.com/user-attachments/assets/028a0de7-fc3e-4da7-9dce-854882644bc2" />

<img width="1259" height="127" alt="Image" src="https://github.com/user-attachments/assets/3aecf0ad-c517-472a-aad5-e97e1a024678" />

**To Reproduce**
To reproduce the bug, please compile the refman.tex file using pdflatex. This document was originally generated by Doxygen and has been modified to isolate and highlight the specific issue. The bug can be seen on page 3.

[doxylink_issue.zip](https://github.com/user-attachments/files/24501630/doxylink_issue.zip)

**Expected behavior**
break lines with long doxylink.

**Version**
version : 1.16.0
OS: Windows 10 

**Stack trace**

**Additional context**

albert-github:
In your output I can see the problem. Unfortunately just some output files won't help us to solve the issue as we should be able to build so:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem?
   Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).


(Yes on the Tex exchange site always a minimal LaTeX example is requested, but that is about specific LaTeX problems, here we need to surrounding circumstances as well, so we ca check the doxygen code)

hachso:
Hello @albert-github, I have attached in my initial post the "doxylingk_isssue.zip" file in the "To reproduce" section. Are you able to download it and compile it ? Thanks again for your help !

albert-github:
Yes I'm able to download it and I see the files:
```
doxygen.sty
etoc_doxygen.sty
refman.tex
```
I'm able to compile (`pdflatex refman.tex`) and I see in the result the long line, though to reproduce it from doxygen I need an example that I can run with doxyen and after that run `make` in the latex directory, so that is why wrote before:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem?
   Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

hachso:
Hello @albert-github, yes sure here it is: 

[...]

PR #11950, opened 2026-01-14
issue #11939 LaTeX output does not properly break lines with long doxylink

albert-github:
There following parts are fixed:
- 5 Index Introducing the LaTeX environment `doxyIndexDescription` and setting an extra flag `escapeUnderscore` for the function `latexEscapeIndexChars` to false.
- 4 Section title Using the `escapeUnderscore` flag to determine whether or not the "word breaking" doxygen LaTeX command `\+` can be used (was by default set to `true` in the `filterLatexString`, when always using `false` would lead to TeX capacity exceeded or to `\+` in the index output).
- 1 File Index Solved due to point 4.

### DIFF
--- a/src/latexgen.cpp
+++ b/src/latexgen.cpp
@@ -2405,7 +2405,9 @@ void filterLatexString(TextStream &t,const QCString &str,
           else
             t << static_cast<char>(c);
           break;
-        case '#':  t << "\\+\\#";        break;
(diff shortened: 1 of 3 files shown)


albert-github:
I'll look into the problem regarding creating the doxygen manual (probably due to a formula `$<$` etc. in an index item.

albert-github:
As written withdrawn as major problems with a.o. `{` and `}`

Issue #7449, opened 2019-12-16
C#: doxygen does not understand namespace aliases


exyi:
C# supports alias a namespace in the `using` directive. For example:

```C#
using Alias1 = Ns1;

/// <see cref="Alias1.A" />
public class Class1
{
}
```

However, Doxygen does not understand that when it's used in a `<see cref="..." />` tag:

![image](https://user-images.githubusercontent.com/7894687/70942193-04e2e680-2046-11ea-8b76-c1f68f4337be.png)

I'd expect it to generate a link to `Ns1.A`.

When a full name of the class is specified, it works correctly.

FYI: C# compiler understands this as I'd expect and translates the comment to:

```xml
        <member name="T:Class1">
            <see cref="T:Ns1.A" />
        </member>
```

Small repro: [doxygen-repro-using-alias.zip](https://github.com/doxygen/doxygen/files/3970085/doxygen-repro-using-alias.zip)

exyi:
Maybe I was too specific. The behavior is actually quite strange, it seems that the `using` directives are ignored completely.

albert-github:
When testing with doxygen 1.8.16 I can see that the link is not present.
When running with doxygen 1.16.1 the problem does not occur.

I think the problem has been solved in the mean time.

PR #11967, opened 2026-02-07
Extend C++ namespace alias syntax to C# + Various fixes

nopeless:
There are various bugs for C++ namespace alias linking due to the fact that

1. C# uses `.` for namespace member access than `::` as it is for C++
2. C# also allows `::` for `global::` and namespace access within namespace
3. When resolving competing identifiers, it fails to prioritize aliasing than namespace inclusion
4. Template aliasing resolution fails due to the differing tag scheme from C++ (C# makes it so that generics (templates) of differing arity refer to different references)

1 has been fixed by simply extending `scanner.l` to accept `.` but transform it into `::`

2 will be addressed in another feature PR

3 has been fixed by making `accessibleViaUsingNamespace` return `result = 1` instead to lower priority

4 has been fixed by calculating arity on parse time and adding a `templateArity` in `resolveClass` function call

---

<details>
<summary>old description</summary>

Cause:

In C#, aliasing with `using` statement is possible

`using X = Y.Z`;

However, under current parser, `Y.Z` will be noted as `typedef Y.Z`;

When `.` is not converted to `::`, symbol resolution at `getTypeDefRec(...)` fails as lookup for `FooNamespace.BarMember` cannot be processed because it doesn't have `::`

Thus, at `scanner.l`, a conversion is needed to fit the `::` representation convention

All tests pass

If you need a testing entry let me know
</details>

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -4833,10 +4833,28 @@ static void resolveTemplateInstanceInType(const Entry *root,const Definition *sc
   int ti=ttype.find('<');
   if (ti!=-1)
   {
+    int templateArity=-1;
     QCString templateClassName = ttype.left(ti);
+    if (root->lang==SrcLangExt::CSharp)
+    {
+       templateArity=1;
+       int level=0;
+       const char *p = ttype.data() + ti + 1;
+       char c;
+       while ((c=*p++))
+       {
+         if (c=='<') level++;
+         else if (c=='>')
+         {
+           if (level==0) break;
+           level--;
+         }
+         else if (c==',' && level==0) templateArity++;
+       }
+    }
     SymbolResolver resolver(root->fileDef());
     ClassDefMutable *baseClass = resolver.resolveClassMutable(scope ? scope : Doxygen::globalScope,
-                                           templateClassName, true, true);
+                                           templateClassName, true, true, templateArity);
     AUTO_TRACE_ADD("templateClassName={} baseClass={}",templateClassName,baseClass?baseClass->name():"<none>");
     if (baseClass)
     {
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -4833,25 +4833,9 @@ static void resolveTemplateInstanceInType(const Entry *root,const Definition *sc
   int ti=ttype.find('<');
   if (ti!=-1)
   {
-    int templateArity=-1;
     QCString templateClassName = ttype.left(ti);
-    if (root->lang==SrcLangExt::CSharp)
-    {
-       templateArity=1;
-       int level=0;
-       const char *p = ttype.data() + ti + 1;
-       char c;
-       while ((c=*p++))
-       {
-         if (c=='<') level++;
-         else if (c=='>')
-         {
-           if (level==0) break;
-           level--;
-         }
-         else if (c==',' && level==0) templateArity++;
-       }
-    }
+
+    int templateArity=(root->lang==SrcLangExt::CSharp) ? resolveTemplateArity(ttype) : 0;
     SymbolResolver resolver(root->fileDef());
     ClassDefMutable *baseClass = resolver.resolveClassMutable(scope ? scope : Doxygen::globalScope,
                                            templateClassName, true, true, templateArity);
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -2089,6 +2089,13 @@ NONLopt [^\n]*
                                           init.stripPrefix("class ");
                                           init.stripPrefix("struct ");
                                           init = init.stripWhiteSpace();
+
+                                          // Use same alias system early on for C#
+                                          if (yyextra->insideCS)
+                                          {
+                                            init = substitute(init,".","::");
+                                          }
+
                                           yyextra->current->type = "typedef "+init;
                                           yyextra->current->args.clear();
                                           yyextra->current->spec.setAlias(true);
--- a/src/symbolresolver.cpp
+++ b/src/symbolresolver.cpp
@@ -1634,10 +1634,11 @@ SymbolResolver::~SymbolResolver()
 const ClassDef *SymbolResolver::resolveClass(const Definition *scope,
                                              const QCString &name,
                                              bool mayBeUnlinkable,
-                                             bool mayBeHidden)
+                                             bool mayBeHidden,
+                                             int templateArity)
 {
-  AUTO_TRACE("scope={} name={} mayBeUnlinkable={} mayBeHidden={}",
-      scope?scope->name():QCString(), name, mayBeUnlinkable, mayBeHidden);
+  AUTO_TRACE("scope={} name={} mayBeUnlinkable={} mayBeHidden={} templateArity={}",
+      scope?scope->name():QCString(), name, mayBeUnlinkable, mayBeHidden, templateArity);
   p->reset();
 
   auto lang = scope ? scope->getLanguage() :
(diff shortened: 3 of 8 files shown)


albert-github:
Thanks for the effort, however:
When running #7449 with the current doxygen version 1.16.1 I already see, with the example provided with #7449, that the missing link is present, but in that case we have a "simple" using statement and it might be that more complex cases still have a problem, so:
> If you need a testing entry let me know

Yes please, 
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the original problem?


Note: GitHub has the habit that when using words like `fixes` it adds a link to automatically close an issue when the proposed PR is integrated, this is not how doxygen likes it, at a moment of a release the, fixed, issue will automatically be closed so please remove the link  (normally on the right side of the window):
image

nopeless:
@albert-github thanks for the swift reply. This issue was actually brought up to me by a friend who is working on a much larger codebase.

The reproducible example that I use is the following:

```txt
root@c1f21a662f97:/workspaces/doxygen/issue-1# ./print_files.sh 
// GlobalImpleFile.cs ------------
using GlobalInterfaAlias = GlobalInterfa;

class GlobalImple : GlobalInterfaAlias {}
// ------------------------------


// GlobalInterfaFile.cs ------------
interface GlobalInterfa {}
// ------------------------------


// NamespaImpleFile.cs ------------
using NamespaInterfaAlias = Namespa.Interfa;

class AliasImple : NamespaInterfaAlias {}
// ------------------------------


// NamespaInterfaFile.cs ------------
namespace Namespa {
  interface Interfa {}
}
// ------------------------------
```
Also a tar gz file: [csfiles.tar.gz](https://github.com/user-attachments/files/25157594/csfiles.tar.gz)

One thing that I do note is that `scanner.l` and `code.l` has different parsing semantics for `using X = <...>` depending on whether the RHS is a compound or a single identifier. You can confirm the difference with `doxygen -d lex:code`

I am not sure if a fix for `code.l` is needed as I do not have a complete view of this project

I'll work on adding a few test cases

nopeless:
I don't think I am able to remove the linked github issue unfortuately

image

nopeless:
Leaving references here

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/namespace-alias-qualifier

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/namespaces#1452-using-alias-directives

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/namespaces#148-qualified-alias-member


Well, @albert-github , it turns out you were very right, because as I was reading the `using` directive from the language spec and turns out `::` is also valid in C#, but reserved for a different purpose than `.` access.

nopeless:
additionally, the semantics for aliasing seems to have expanded over the years, with `using static` and `global using` introduced in C# 6.0 and being able to alias any kind of type in C# 12.0

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-12.0/using-alias-types

```cs
using Point = (int x, int y);
```

Let me know how you are planning to handle this

nopeless:
image

Lot of test cases fail, we can work them out

albert-github:
- Regarding: https://github.com/doxygen/doxygen/pull/11967#issuecomment-3866033546
  - example, thanks though a doxygen settings file is missing
  - using a basic settings file does show some differences:
    -  in the inheritance graphs
    - in the hierarchy overview 
  
    are there any other things to be noted? 
  - yes the `scanner.l` and `code.l` use some different parsing as they are used for different purposes (bluntly said `scanner.l` for syntactical parsing and getting initial documentation comments, `code.l` for ode representation and coloring as well as an attempt to get a call / caller hierarchy).
-  Regarding https://github.com/doxygen/doxygen/pull/11967#issuecomment-3866036702 (link)
   - looks like you were successful after all.
- Regarding https://github.com/doxygen/doxygen/pull/11967#issuecomment-3866064782 (difference `.` and `::`)
   - doesn't make things any simpler, or better said even more complex for the user. Best is at first to concentrate on the `.` version, the integration of the `::` might be quite a bit more difficult (especially as nowadays only `.` is supported (see a.o. in `util.cpp`: `getLanguageSpecificSeparator`.
-  Regarding https://github.com/doxygen/doxygen/pull/11967#issuecomment-3866065560 (`global` and  `static`)
   -  could be handled as well, no plans yet but should probably be done in 2 enhancement PRs
-  Regarding https://github.com/doxygen/doxygen/pull/11967#issuecomment-3866096152 (test case 116)
   -  always nice to have
   - as you already wrote the test case is work in progress, looks like the case with multiple levels (like `N.Collections.Dictionary`) is not handled properly.

nopeless:
@albert-github I am not sure how to apply a fix because for a lack  of a better word the codebase simply doesn't seem to handle c# generics well?

well, it does handle c# generics, but because of that we run into an issue where generics of different arity has to be looked up

essentially I originally tried to just copy c++ generic handling logic to c# but in c# generics of different arity can mean different things so doxygen simply suffixes it with (`-<N>-g`), but this information is not available? unless I'm missing something

I'd like a better guide on how to fix this issue, currently my best "fix" is to just simply detect arity at `symbolresolver.cpp > resolveClass` but that doesn't seem like a good idea

nopeless:
I just implemented it how I think is good enough. Let me know what you think

nopeless:
`global::` support does not exist and it probably should be another feature branch like your other suggestion

nopeless:
ok, turns out that there is yet ANOTHER issue where symbol resolution succeeds, but fights when there are competing resolutions

```c#
using Alpha; // Alpha.Interface exists
using Interface = Beta.Interface;

class Implementation : Interface {} // Gets resolved into Alpha.Interface
```

Previously, `using Interface = Beta.Interface` didn't work at all, and therefore I assumed fixing that would fix this issue by proxy, but this is another issue

edit: might be a doxygen.cpp bug while I was trying to fix util.cpp, I'm looking into it
edit2: was not the case. but I extracted common logic into function so that code desync doesn't happen again

nopeless:
code has been adjusted to use a function for both `doxygen.cpp` and `util.cpp`

nopeless:
currently this is the minimal reproduction I am using

```cs
namespace N {
  interface A {}
}

namespace M {
  interface A {}
}

using N;
using A = M.A;

interface I : A {}

```

nopeless:
The issue is indeed because of "result" having ambiguous ranking

nopeless:
@albert-github this was my fix for the namespace issue, it is a 5 line fix that simply makes `result = 1` every time it is possible to just forcefully lower the ranking so that an explicit alias is always picked. Not sure if this is proper so you need to check it. The fix works, though, as `A` now correctly refers to `M.A`

doxygen:
The example 

```c#
namespace N {
  interface A {}
}

namespace M {
  interface A {}
}

using N;
using A = M.A;

interface I : A {}
```

Is interesting as the same example in C++

```c++
namespace N {
  class A {};
}

namespace M {
  class A {};
}

using namespace N;
using A = M::A;

class I : public A {};
```
doesn't compile as the definition of `A` is ambiguous.

The C# standard says in the chapter about namespaces:
> Unlike a using_alias_directive, a using_namespace_directive may import types whose identifiers are already
defined within the enclosing compilation unit or namespace body. In effect, names imported by a
using_namespace_directive are hidden by similarly named members in the enclosing compilation unit or
namespace body.

So unlike in C++, in C# the `using N` will not import `A` as it is already defined directly.

nopeless:
@doxygen Thanks for clarifying the behavior, but I believe checking if the code is correct is not the scope of Doxygen, rather it is producing correct documentation on the premise that is correct code? If so, the case of C# being the default behavior (as it is implemented now) seems to be the most straightforward implementation.

Let me know if you think otherwise and would like to suggest a different resolution

nopeless:
@doxygen other than the merge conflict, is there any blocking issue for this pr? would like to see this merged as a best effort solution

nopeless:
> @albert-github this was my fix for the namespace issue, it is a 5 line fix that simply makes `result = 1` every time it is possible to just forcefully lower the ranking so that an explicit alias is always picked. Not sure if this is proper so you need to check it. The fix works, though, as `A` now correctly refers to `M.A`

I may revert this change however as it seems a little irrelevant

PR #11983, opened 2026-02-16
Update translator_gr.h

nickreserved:


albert-github (commented):
Thanks.

As the Greek translation should be up to date again now the line:
```
class TranslatorGreek : public TranslatorAdapter_1_16_0
```
should also be replaced by:
```
class TranslatorGreek : public Translator
```

Issue #11984, opened 2026-02-16
C++20 concepts and doxygen 1.16.1 parsing problems


nickreserved:
I take
`warning: documented symbol 'concept restrict more the &&std::random_access_iterator< typename T::diagonal_iterator > && output_iterator< typename T::diagonal_iterator, typename T::value_type >' was not declared or defined.`

on this code
```
/** A writable triangular-storage matrix whose rows from first element to main diagonal element can be accessed.
A triangular-storage band matrix with mutable elements, must fulfill:
- It must fulfill a BandMatrixWithLowerTriangularStorage.
- It must fulfill a MutableSkylineMatrixWithLowerTriangularStorage.
- It must fulfill a MutableMatrixWithAccessedDiagonal.
- Must have a member type \c diagonal_vector which is a MutableDenseVector.
- Must have a member type \c diagonal_iterator which is a <tt>std::random_access_iterator</tt> and
<tt>std::output_iterator</tt> iterator on each main diagonal's elements. */
template<typename T>
concept MutableBandMatrixWithLowerTriangularStorage = BandMatrixWithLowerTriangularStorage<T>
&& MutableSkylineMatrixWithLowerTriangularStorage<T> && MutableMatrixWithAccessedDiagonal<T>
&& MutableDenseVector<typename T::diagonal_vector> // this and 2 next lines, restrict more the
&& std::random_access_iterator<typename T::diagonal_iterator> // MutableMatrixWithAccessedDiagonal
&& std::output_iterator<typename T::diagonal_iterator, typename T::value_type>;
```

This does not happen on previous non-C++20-aware doxygen versions.

nickreserved:
Removing the comments
// this and 2 next lines, restrict more the
// MutableMatrixWithAccessedDiagonal
fixes the warning. Parsing problem?

albert-github:
Yes this definitely looks like a parsing problem.

When looking a the output I see:

<img width="749" height="147" alt="Image" src="https://github.com/user-attachments/assets/14f1bf86-c5fd-42d1-bd28-335a09ec6662" />


instead of the expected:

<img width="848" height="215" alt="Image" src="https://github.com/user-attachments/assets/4b37cd68-c4c0-4430-b3f7-180abf5dd639" />

Looks like that the scanner stops after the first comment.


Example: 

[example.tar.gz](https://github.com/user-attachments/files/25335893/example.tar.gz)

albert-github:
Most likely:
[...]

Issue #9569, opened 2022-09-03
Improve support for C++ concepts


milasudril:
**Describe the bug**
As of v1.9.4, doxygen has basic support for concepts, however, it is not possible to document individual "members". Documentation comments on individual members is simply stripped from the documentation.

**Expected behavior**
Concepts should be presented in a similar way as abstract base classes.

**To Reproduce**
```c++
/**
 * \brief Defines the requirements of a "source"
 *
 * A source is something that can be read from, for example, an input buffer or a file.
 *
 * \ingroup de-serialization
 *
 */
template<class T>
concept source = requires(T a)
{
	/**
	 * \brief Shall try to consume next byte and return it. If it possible to consume another
	 *  byte, it shall return std::nullopt, which implies an End-Of-File condition.
	 */
	{ read_byte(a) } -> std::same_as<std::optional<char>>;
};
```
Becomes no more than

```c++
template<class T>
concept anon::source =  requires(T a)
    {
        
        { read_byte(a) } -> std::same_as<std::optional<char>>;
    }
```

**Version**
Version 1.9.4 build with libclang 14 on Ubuntu 22.04.

**Additional context**
The configuration file enables clang assisted parsing with -std=c++20 as option to clang.

albert-github:
- What are you missing?
- What is in your doxygen settings file: so Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file) that allows us to reproduce the problem? Please don't add external links as they might not be persistent.
- I only see a `\ingroup` nut no definition of the group


@doxygen is it by design that in the code fragment the comment is missing even when `STRIP_CODE_COMMENTS = NO` ?

milasudril:
I mean that it should write a more complete documentation page. Currently, it picks up the documentation of the concept itself, but it cannot produce a page which contains documentation about individual requirements. I think it can use the same format as a class, and document produce documentation for each requirement as if they were some kind of member function. Here I need to document that `read_byte` should consume next byte, and also, that std::nullopt is return in case of EOF, etc.

h8672:
What happens if you surround it with **addtogroup** and surround with ``` @{ @} ```?
I cant add classes to a group with **ingroup** if it's not part of a group itself, because it would just create a module link to that group.
If the group has no name, I don't think it will be added.

I changed the code slightly, change the marking
```cpp
/**
 * \brief Defines the requirements of a "source"
 *
 * A source is something that can be read from, for example, an input buffer or a file.
 *
[...]

PR #11988, opened 2026-02-17
issue #11984 C++20 concepts and doxygen 1.16.1 parsing problems

albert-github:
Analogous to the  the problem with the `//` comments also with the `/*` type of comments the problem was present.

Example: [example.tar.gz](https://github.com/user-attachments/files/25364580/example.tar.gz)

### DIFF
--- a/src/scanner.l
+++ b/src/scanner.l
@@ -3290,7 +3290,8 @@ NONLopt [^\n]*
 <ReadInitializer>{CCS}                  {
                                           if (yyextra->keepComment)
                                           {
-                                            yyextra->current->initializer << yytext;
+                                            yyextra->lastCContext = YY_START ;
+                                            BEGIN( SkipComment ) ;
                                           }
                                           else
                                           {


src/scanner.l
```diff
@@ -3290,7 +3290,8 @@ NONLopt [^\n]*
 <ReadInitializer>{CCS}                  {
                                           if (yyextra->keepComment)
                                           {
-                                            yyextra->current->initializer << yytext;
+                                            yyextra->lastCContext = YY_START ;
+                                            BEGIN( SkipComment ) ;
```
doxygen, line 3294: This doesn't look like the right solution. The idea of `keepComment` is, as the name suggests, to keep the comments as part of the initializer, so they can be rendered as part of the concept definition.

See also https://github.com/doxygen/doxygen/commit/89346424ddafc095dd9abf880586240d72b8dbed and issue #9569

albert-github: With the PR #11986 the comment has also been removed, so actually we should distinguish between comments like: `//` versus `///` / `///<` and `/*` versus `/**` /  `/**<`

doxygen:
@albert-github I've changed it such that the normal C/C++ comments are kept so they remain visible in the concept definition.

albert-github:
In principle does not look bad, though I don understand the differences between `Con3a` versus `Con5` as well as `Con4a` versus `Con6`.

```
/// \file

/** Third a fie */
template<typename T>
concept Con3a=
/// A,
CC<typename T::DV> &&
/// B
DD<typename T::DI>
;

/**  Fourth a fie */
template<typename T>
concept Con4a =
/// C
CC<typename T::DV> &&
/// D
DD<typename T::DI>
;
```
and
```
/// \file

/** Fifth fie */
template<typename T>
concept Con5= CC<typename T::DV> ///< A,
&& DD<typename T::DI> ///< B
;

/**  Sixth fie */
template<typename T>
concept Con6 = CC<typename T::DV> ///< C
&& DD<typename T::DI> ///< D
;
```


Example: [example1.tar.gz](https://github.com/user-attachments/files/25415983/example1.tar.gz)

doxygen:
@albert-github The problem is that the comments are *inside* the concept (much like comments inside the body of a function). So there is not really a concept of before or after. When putting comments at the end there is not really a good way to render them in line, so I chose to show them as normal comments. Only special comments that are meant to be placed before an item are used to split the concept definition into parts.

That said the `///<` in Con6 should also be rendered as a normal comment. I'll look into that.

albert-github:
@doxygen 

in the commit message it now stays:
```
Commit: fddfea11b223e4aada871ad7c3148400b48698be [fddfea1]

Date: Friday, February 20, 2026 2:01:21 PM

PR #11988 C++20 concepts and doxygen 1.16.1 parsing problems

Keep ///< style comments in concept definition as-is.
```
That makes it clear, but I'm not sure that the user will understand it as the user might think that `///<` comments should be rendered as well `con5a` and `con6a`. Probably a difficult problem to solve.

PR #12002, opened 2026-02-21
Discussion #12000 CHM Viewer JavaScript Problem with svg.min.js

albert-github:
Disable `INTERACTIVE_SVG` when `GENERATE_HTMLHELP=YES`

Example: [example.tar.gz](https://github.com/user-attachments/files/25457693/example.tar.gz)


paule32:
Did you use a splitted Version of **svg.min.js** ?
So, you have the new one for modern Browsers, and a Version for old Browser's like CHM Viewer ?
For me, I download the Version 2.7 of this File from **CDN**, and put it to the DoxyGen Root documentation path.
And I set the Option:
```
HTML_EXTRA_FILES  += ./dynsections.js svg.min.js
```
If you do so, the Script Error fly away with this Tasks.
**Note:**
The CDN Version comes with named: **svg-min.js**, and DoxyGen use **svg.min.js**.
So, don't forget to rename the File.

albert-github:
@paule32 
Which version of doxygen did you test?

PR #12011, opened 2026-02-24
Fix warning

sloriot:
```
vhdlparser/tokenmanager.h:25:44: warning: extra ‘;’ after in-class function definition [-Wextra-semi]
   25 |   virtual void   setParser(void* parser) {};
      |                                            ^
      |****
```


albert-github:
This is known but this is generated code and the change is outside the control of doxygen.

Issue #12009, opened 2026-02-22
Doctest and requirements mapping links and layout


maidenone:
**Describe the bug**
Thank you for adding the requirement mapping!
A few years ago i made my own "hack" to implement requirement mapping in doxygen based on alieses, xlist and post-processing. I was very happy to see that Doxygen got native requirement handling.

My issues:
1. When using Doctest, the name is not correctly resolved from the test macro, i guess Catch2 have the same issue.
TEST_CASE("my test") This is resolved to TEST_CASE(), that is not very helpful.

2. Another issue is that i do not get any clickable links to \satisfies

I think some small changes to the layout would make the html view much clearer, especially with lots of requirements.

**Screenshots**
Proposal for improved layout:
<img width="900" height="713" alt="Image" src="https://github.com/user-attachments/assets/8772ae16-28e1-45b2-8c23-da506810ea34" />

<br>
Vs the layout we currently have in our own doxygen requirement mapping tool:
(but we resolve the proper test names and get clickable links on satisfies).
<br>

<img width="634" height="618" alt="Image" src="https://github.com/user-attachments/assets/b763accc-7a93-4e0c-a6dc-174ba10298ae" />

Also, it would be great to be able to put requirements in sections and have a section description:

SECTION NAME
- Description of section
- - Requirements for section

SECTION NAME2
- Description of section
- - Requirements for section

**To Reproduce**
I have set up an example project for requirement mapping here: https://github.com/maidenone/doxreqcpp

**Version**
1.16.1 Linux

albert-github:
It is mostly very hard to work through other code especially when there are things like cmake involved as well.
Furthermore (even though it is a dedicated repository) a GitHub repository can change and or be removed ad thus making the issue incomplete therefore always:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem?
   Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

Furthermore it looks like you missed (at least) 2 settings:
```
EXTRACT_STATIC         = YES
MACRO_EXPANSION        = YES
```
I also had to add `/** \file */` to the file `src/math_function.cpp`.

At that moment I get links to the `math_function`.

Regarding the `TEST_CASE` this is probably due as doxygen cannot find the `doctest.h`  which is the `lib` directory is this directory is nowhere specified.

when adding:
```
INCLUDE_PATH = lib
```
we will get a warning:
```
[...]

PR #12014, opened 2026-02-24
issue #12009 Doctest/Catch2, requirements links and layout

albert-github:
In the preprocessor the standard defined macros are not expanded by doxygen, resulting in names like `DOCTEST_ANON_FUNC___LINE__()` . 
The standard macro `__LINE__` will now be expanded in the doxygen preprocessor.

### DIFF
--- a/src/pre.l
+++ b/src/pre.l
@@ -3078,7 +3078,13 @@ static bool expandExpression(yyscan_t yyscanner,QCString &expr,QCString *rest,in
     bool replaced=FALSE;
     macroName=expr.mid(p,l);
     //printf(" p=%d macroName=%s\n",p,qPrint(macroName));
-    if (p<2 || !(expr.at(p-2)=='@' && expr.at(p-1)=='-')) // no-rescan marker?
+    if (macroName == "__LINE__")
+    {
+      QCString resultExpr = QCString().setNum(yyextra->yyLineNr);
+      QCString restExpr=expr.right(expr.length()-8-p);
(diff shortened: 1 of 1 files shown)


src/pre.l
```diff
@@ -3078,7 +3078,13 @@ static bool expandExpression(yyscan_t yyscanner,QCString &expr,QCString *rest,in
     bool replaced=FALSE;
     macroName=expr.mid(p,l);
     //printf(" p=%d macroName=%s\n",p,qPrint(macroName));
-    if (p<2 || !(expr.at(p-2)=='@' && expr.at(p-1)=='-')) // no-rescan marker?
+    if (macroName == "__LINE__")
```
doxygen: @albert-github Probably good to also consider the other predefined macros such as `__FILE__`, `__DATE__`, and `__TIME__`
See also https://en.cppreference.com/w/c/preprocessor/replace.html#Predefined_macros

albert-github: I agree, though I first wanted to know whether or not this `__LINE__` would be acceptable.

In the "Working Draft, Programming Languages — C++" number N5014 of 2025-08-05 I see in paragraph  "15.12 Predefined macro names" quite a few names.
- Should they all be implemented?

doxygen: The ones starting with `__STDC_` are probably not so useful, they can be used to query the capabilities/properties of the compiler.

albert-github: Extended example: [example3.tar.gz](https://github.com/user-attachments/files/25621578/example3.tar.gz)

Issue #12008, opened 2026-02-22
"namespacemembers.html" and "functions.html" only show items for one specific initial letter


N-Dekker:
**Describe the bug**
It appears that for a large project, Doxygen switches from showing all items (namespace members or class members) in "namespacemembers.html" and "functions.html" to only showing the items for one specific initial letter.

**Screenshots**

"namespacemembers.html" showing only the letter C. The web page has no link to the pages with the other initial letters.
<img width="709" height="1392" alt="image" src="https://github.com/user-attachments/assets/07ca70d6-7877-4898-b63a-77022ebf5641" />

**To Reproduce**

The problem occurs when running Doxygen over the following source tree: https://github.com/SuperElastix/elastix/tree/a1643518ef0037fc6c3c736d1624de25dac798b7

**Expected behavior**
I would prefer to have all items from A to Z on one page, as it used to be. It is still like that at https://elastix.dev/doxygen/namespacemembers.html

However, if it's necessary to have each initial letter on a separate page, I would expect a link to each page on the header of "namespacemembers.html" and "functions.html"

**Version**
```
> doxygen --version
1.16.1 (669aeeefca743c148e2d935b3d3c69535c7491e6)
```
Windows 11 (24H2)

albert-github:
I'm always a bit reluctant to just having a repository especially without proper instructions on how to build JUST the documentation.

I think I get the problem, lets get to some recent doxygen, internal, documentation at https://doxygen.github.io/doxygen-docs/index.html
When clicking here in the "namespaces" in the treeview (on the left) and subsequently on "Namespace Members" we get:

<img width="1150" height="593" alt="Image" src="https://github.com/user-attachments/assets/4b793814-92a8-4ce9-8e69-725a3cf586cb" />

but above the line "Here is a list of all namespace members with links to the namespace documentation for each member:" you would like to see a line analogous to the line at e.g. file:///D:/Programs/Doxygen/fork/doxygen/build_windows_nmake_v2017/doxygen_docs/html/classes.html i.e.

<img width="1291" height="340" alt="Image" src="https://github.com/user-attachments/assets/5633234f-c858-4cfe-881c-734d2f7c345e" />

(reachable through "Classes" "Class Index" in the treeview)

- Is this correct?

N-Dekker:
[...]

PR #12016, opened 2026-02-25
issue #12008 "namespacemembers.html" and "functions.html" only show items for one specific initial letter

albert-github:
Analogous to the case for the "Class index" also create an "alphabetical list" in the page (taking the multi page possibility into consideration) but do this only in case `HTML_DYNAMIC_MENUS=YES` as otherwise the list would be double.

### DIFF
--- a/src/index.cpp
+++ b/src/index.cpp
@@ -3125,7 +3125,6 @@ static void writeMemberIndex(OutputList &ol,
     alphaLinks += "<a class=\"qindex\" href=\"" + anchor +
                   li + "\">" +
                   letter + "</a>";
-    first=FALSE;
   }
   alphaLinks += "</div>\n";
   ol.writeString(alphaLinks);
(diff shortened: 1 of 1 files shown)


src/index.cpp
```diff
@@ -3100,6 +3100,37 @@ static void writeQuickMemberIndex(OutputList &ol,
...
+    else
+      anchor=fullName+"_"+is+extension+"#index_";
+
+    if (!first) alphaLinks += "&#160;|&#160;";
+    first=false;
+    QCString li = letterToLabel(letter);
+    alphaLinks += "<a class=\"qindex\" href=\"" + anchor +
+                  li + "\">" +
+                  letter + "</a>";
+    first=FALSE;
```
N-Dekker: Nitpick: `FALSE` should be `false`, I guess  😺

albert-github: In principle correct, but part of the code is copied from the routine above where also `FALSE` is used.
There are still quite a few places in doxygen where `FALSE` (quick count on the src director around 1738 times. `TRUE` around 1463 times. Lowercase versions 1568 / 9719 times) is used.

N-Dekker: OK, thanks, I did not realize that doxygen uses both `false` and `FALSE`! However, your pull request now says:

```c++
    first=false;
    QCString li = letterToLabel(letter);
    alphaLinks += "<a class=\"qindex\" href=\"" + anchor +
                  li + "\">" +
                  letter + "</a>";
    first=FALSE;
```

So it assigns `first` twice! One of the two assignments looks redundant to me  🤷

albert-github: That is not the intention and is redundant (a good optimizing compiler would remove it anyway)

PR #12027, opened 2026-03-04
change cache mutexes to thread locals

vtjnash:
These changes seemed to help llvm's doxygen build scale well with threads, instead of getting slower. With a mutex, all of the threads spend all of their time waiting, burning a lot more CPU time, but not getting any faster. With a local cache, all of the benefits of caching are achieved, with none of the concurrency contention.

Written by Claude, so this is certainly not the most clean implementation, but was intended to test the theory that thread performance could be much better by fixing this contention. That test seemed a success. Now, I can improve upon this if this is something of interest, particularly if there is anything particular you know is an issue or style that you would like to see.

### DIFF
--- a/src/definition.cpp
+++ b/src/definition.cpp
@@ -15,6 +15,7 @@
 
 #include <algorithm>
 #include <iterator>
+#include <mutex>
 #include <unordered_map>
 #include <string>
 #include <optional>
@@ -53,6 +54,16 @@
 
 //-----------------------------------------------------------------------------------------
 
+/** once_flag wrapper that is copyable (copy default-initializes the flag) and resettable. */
+struct ResettableOnce
+{
+  mutable std::once_flag flag;
+  ResettableOnce() = default;
+  ResettableOnce(const ResettableOnce &) {} // copy: leave flag in not-yet-called state
+  ResettableOnce &operator=(const ResettableOnce &) { return *this; }
+  void reset() { flag.~once_flag(); new (&flag) std::once_flag{}; }
+};
+
 /** Private data associated with a Symbol DefinitionImpl object. */
 class DefinitionImpl::Private
 {
@@ -81,6 +92,7 @@ class DefinitionImpl::Private
     QCString localName;      // local (unqualified) name of the definition
                              // in the future m_name should become m_localName
     QCString qualifiedName;
+    ResettableOnce qualifiedNameOnce;
     QCString ref;   // reference to external documentation
 
     bool hidden = FALSE;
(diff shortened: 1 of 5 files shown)


doxygen:
@vtjnash Interesting idea. Because the cache is no longer shared between threads, I guess this results in more CPU and memory use, as the same symbol lookup now needs to be done and remembered for each thread separately. But at the same time there is also no thread contention anymore, so it indeed scales better. I like it.

vtjnash:
I think the main decision I'm seeing is that it looks like the thread pool is destroyed between each phase, which entirely frees the caches in between each phase. But right now there's also code to explicitly free some subsections of it instead (e.g. flushCachedTemplateRelations). Would it be better to explicitly preserve the thread local caches across thread pools or better to delete the partial clearing, and instead just drop them fully?

(some of this also depends on whether it's eventually worthwhile to make the ThreadPool a global static, so that it doesn't get entirely dropped between each phase, but rather is a shared resource representing each CPU)

vtjnash:
I'm seeing it actually results in substantially less CPU usage too, since copying this data between threads costs a lot more CPU effort than recomputing it. Memory usage is definitely going to be higher, but only a (random) subset of the cache will be present on each CPU depending on which files it visits, so it isn't linear. Most large core machines have large memory too for exactly this reason (e.g. the one I have access to has 128 cores and 756 GB). I didn't test any intermediate size machines to see how intermediate thread scaling would work out.

doxygen:
I see a two of issues with the proposal.
1. The clean of actions of the typeLookupCache in `doxygen.cpp` are now only done on the main thread.
2. The cache totals are also only calculated for the main thread:

Latest on Doxygen's own code:
<pre>
type lookup cache used 3153/131072 hits=135397 misses=4609
</pre>
and the same for version 1.16.1:
<pre>
type lookup cache used 37414/131072 hits=350336 misses=38873
symbol lookup cache used 91045/131072 hits=808173 misses=91062
</pre>

Maybe 1. is not a big issue (what is released by the worker thread doesn't need to be cleaned). For 2. we can probably keep the maximum per thread and update a global maximum at the end of each thread and the main thread.

vtjnash:
> The clean of actions of the typeLookupCache in doxygen.cpp are now only done on the main thread.

The other likely alternatives are either to explicitly drop all caches (including main thread) or to explicitly preserve the cache between thread pools. Both are fairly straightforward to implement, and the choice probably most depends upon what is expected in future doxygen development.

> For 2. we can probably keep the maximum per thread and update a global maximum at the end of each thread and the main thread.

We can also sum them together, or just drop them (do stats really matter, as long as performance is fine). This can easily be part of the ThreadPool teardown logic.

PR #12028, opened 2026-03-04
add persistent per-thread dot processes via stdin pipeline

vtjnash:
Keep one 'dot -T<format>' child process alive per worker thread per format, feeding each graph via stdin and reading rendered output from stdout. This eliminates fork overhead.

For llvm docs, this takes the overhead of this step on my machine from dozens of minutes, to milliseconds. Particularly on linux, fork of a multigig process is painfully slow, and serializes everything to update page tables.

Written by Claude, so this is certainly not the most clean implementation, and some comments look entirely wrong, but it was intended to test the theory that dot performance could be much better by fixing this fork cost. That test seemed a success. Now, I can improve upon this if this is something of interest, particularly if there is anything particular you know is an issue or style that you would like to see

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -258,6 +258,7 @@ add_library(doxymain STATIC
     dotincldepgraph.cpp
     dotlegendgraph.cpp
     dotnode.cpp
+    dotprocess.cpp
     dotrunner.cpp
     doxygen.cpp
     eclipsehelp.cpp
--- a/src/dotprocess.cpp
+++ b/src/dotprocess.cpp
@@ -1,0 +1,584 @@
+/******************************************************************************
+*
+* Copyright (C) 1997-2019 by Dimitri van Heesch.
+*
+* Permission to use, copy, modify, and distribute this software and its
+* documentation under the terms of the GNU General Public License is hereby
+* granted. No representations are made about the suitability of this software
+* for any purpose. It is provided "as is" without express or implied warranty.
+* See the GNU General Public License for more details.
+*
+* Documents produced by Doxygen are derivative works derived from the
+* input used in their production; they are not affected by this license.
+*
+*/
+
+#include "dotprocess.h"
+#include "message.h"
+#include "portable.h"
+
+#include <cerrno>
+#include <cstring>
+#include <map>
+
+#ifndef _WIN32
+#  include <unistd.h>
+#  include <sys/wait.h>
+#  include <fcntl.h>
+#  include <signal.h>  // strsignal
+#endif
+
+//-----------------------------------------------------------------------------------------
+// End-marker tables
+//-----------------------------------------------------------------------------------------
+
+struct EndMarker
+{
+  const char *bytes;
+  size_t      len;
+};
+
+static EndMarker getEndMarker(const QCString &format)
+{
+  if (format == "svg")   return { "</svg>\n",  7 };
+  if (format == "cmapx") return { "</map>\n",  7 };
+  if (format == "png")
+  {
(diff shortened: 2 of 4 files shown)


doxygen:
@vtjnash Another interesting idea. I think fork performance of Linux is quite good compared to Windows, so likely there is even more to gain on that platform. 

Would be nice if the platform specifics could be hidden in portable.cpp.

vtjnash:
> I think fork performance of Linux is quite good compared to Windows, so likely there is even more to gain on that platform.

In the LLVM case, linux fork is quite a lot worse compared to Windows, because of the multi-GB heap, particular when using threads (there's a lot of locks to acquire, and a lot of memory to copy). I thought about using posix_spawn here, which would have improved that for macOS, but done nothing for Windows. I also tried to do this in-process, but `dot` turns out to be not safe to use with threads. Though by removing the spawn overhead, a single-threaded in-process `dot` would actually potentially be faster than the one existing fork call, for generating all of the diagrams for llvm. So that could be another option instead here.

doxygen:
@vtjnash Looks like there are some issues with the CI pipeline. I see
```
error: dot process (format 'png'): read from dot stdout failed â process exited with code 127; will restart.
error: dot process (format 'cmapx'): read from dot stdout failed â process exited with code 127; will restart.
error: dot process (format 'png'): read from dot stdout failed; will restart.
error: dot process (format 'cmapx'): read from dot stdout failed; will restart.
```
Not sure if this can be fixed reliably (note that the dot format can be user configured). 

Listening to the "end" of the dot output via some assumed end marker is a bit fiddly I think. Maybe an idea to make a "dot-like" tool using the GraphViz libraries (released under a GraphViz compatible license) that is able to do the IO streaming in a more robust way, and then Doxygen can use this dot-like tool for each worker thread...

vtjnash:
It could be fixed for those (the output format is obviously known). Code 127 just means the binary launch failed, and that is because claude used `execv` instead of `execvp`. 

I think there's actually quite a few options here:
1. The classic solution was to pre-fork a daemon that ran all later `system` calls, as this avoids a lot of the common problems with `fork` (bad performance, racy with pipe syscall, bad interactions with threads and locks, etc.). The downside is simply that this is annoying to implement, and gives no advantage over 2.
2. The historical answer in unix was the hack of `vfork`. The downside is that this doesn't exist in macOS, and is also apparently considered to be rather buggy by libc developers (there is some race conditions and stack corruption that happens, which can't be fixed, but also usually isn't observed).
3. The modern solution is almost the same, but uses `posix_spawn` to avoid the extra worker process. The downside is simply that this doesn't help on Windows at all.
4. The `DOT_MULTI_TARGETS` could be expanded to handle everything in the existing interface. This is probably fairly clean, and the main downside is simply that scheduling becomes statically decided, since the work has to be partitioned up front. The slightly nice part is that threads are pointless here, since it can instead just launch N processes in a loop, then waitpid for N process in a loop. Where each process gets `TOTAL / N` jobs assigned to it.
5. The end marker could be "machine learned" since `dot` is a deterministic function of the input. Between each real job, the program could emit a "separator" job (just `graph{}`). The doxygen would run this separator job first, to learn the expected output, then would know exactly what to look for, for any format, to delineate real output (as long as the real input is never equivalent to `graph{}`).
6. Doxygen could make an almost-like-dot tool that it uses, though that may be an increasingly legal gray area, as far as the compatibility of the copyleft licenses for each project.

doxygen:
@vtjnash Now that the alternative implementation passing multiple dot files per dot invocation is merged, this can be closed.

Issue #11923, opened 2026-01-06
1.16 GLIBC not found errors in GitHub Actions workflow using Ubuntu 22.04


mcarans:
**Describe the bug**
In a GitHub Actions workflow that uses Doxygen, I see:
```
Downloaded: /home/runner/work/oolite/oolite/doxygen/doxygen-1.16.0.linux.bin.tar.gz
../doxygen/doxygen-1.16.0/bin/doxygen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.36' not found (required by ../doxygen/doxygen-1.16.0/bin/doxygen)
../doxygen/doxygen-1.16.0/bin/doxygen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ../doxygen/doxygen-1.16.0/bin/doxygen)
../doxygen/doxygen-1.16.0/bin/doxygen: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by ../doxygen/doxygen-1.16.0/bin/doxygen)

```

**To Reproduce**
Doxygen is called via coverxygen (1.8.1) like this: `python3 -m coverxygen` in the workflow

**Expected behavior**
Doxygen runs

**Version**
1.16 in Ubuntu 22.04 GitHub Actions workflow

albert-github:
Looks a bit like the old problem https://github.com/doxygen/doxygen/issues/6393 around static linking and older versions of gcc and the OS.

doxygen:
The binary version of Doxygen is now compiled on Ubuntu 24.04 as I had problems automating the build with 22.04 (libclang didn't build properly). Ubuntu 24.04 uses newer versions of glibc and glibc++. Unfortunately, this means the binaries no longer work on older distributions. So I suggest to reach out the `coverxygen` to stick to using 1.15.0 or upgrade your workflow to Ubuntu 24.04.

vtjnash:
Just a thought: `zig cc` can be used as a linker wrapper to avoid glibc versioning bugs like this: https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html

Otherwise, you're usually stuck maintaining an ancient docker image of some sort just to have an old enough glibc version in the build environment to avoid this sort of problem

doxygen:
[...]

PR #12032, opened 2026-03-05
[RFC] move dot inprocess

vtjnash:
Remove thread support, since dot is not thread-safe. But without fork overhead on linux, this processes all llvm docs in tens of seconds instead of tens of minutes (for me). Because of fork overheads, the threads didn't help previously anyways.

This is an alternative to #12028. That PR scales very well with threads, so it can process llvm in tenths of seconds instead of tens of seconds on huge machines. But huge core machines aren't that common, so I don't know it is worth that complexity.


src/config.xml
```diff
@@ -3854,17 +3854,6 @@ where `loc1` and `loc2` can be relative or absolute paths or URLs.
...
-      </docs>
-    </option>
-    <option type='int' id='DOT_NUM_THREADS' defval='0' minval='0' maxval='512' depends='HAVE_DOT'>
-      <docs>
-<![CDATA[
- The \c DOT_NUM_THREADS specifies the number of \c dot invocations Doxygen is
- allowed to run in parallel. When set to \c 0 Doxygen will
- base this on the number of processors available in the system. You can set it
- explicitly to a value larger than 0 to get control over the balance
- between CPU load and processing speed.
```
albert-github, line 3867: The `DOT_NUM_THREADS` has to be added to the "obsolete" part at the end of config.xml

CMakeLists.txt
```diff
@@ -245,6 +245,29 @@ else()
...
+        set(LIBGVC_INCLUDE_DIRS ${GVC_INCLUDE_DIR})
+        set(LIBGVC_LIBRARIES ${GVC_LIB} ${CGRAPH_LIB})
+    endif()
+endif()
+if(LIBGVC_FOUND)
+    message(STATUS "Found libgvc — dot graphs will be rendered in-process")
+else()
+    message(STATUS "libgvc not found — dot graphs will use subprocess")
+endif()
+
```
albert-github, line 270: This idea as might seem to be good but:
- one needs to have dot installed on compile time and thus bring some extra dependencies
- loosing flexibility for the end user as the end user cannot choose the dot version anymore (trying newer versions when e.g. certain artifacts are fixed).

or did I miss something

doc_internal/tags_history.md
```diff
@@ -441,6 +441,8 @@ The following table gives an overview of the doxygen configuration settings and
 <dt></dt>              <dd>SQLITE3_RECREATE_DB</dd>
 <dt>New in 1.8.20</dt>
 <dt></dt>              <dd>PYTHON_DOCSTRING</dd>
+<dt>Obsolete in 1.16.1</dt>
```
albert-github, line 444: Should be in config.xml and it will be automatically added to this file (see remark with config.xml)

src/configimpl.l
```diff
@@ -2036,14 +2036,6 @@ void Config::checkAndCorrect(bool quiet, const bool check)
...
   }
 
-  //------------------------
-  int dotNumThreads = Config_getInt(DOT_NUM_THREADS);
-  if (dotNumThreads<=0)
-  {
-    dotNumThreads=std::max(2u,std::thread::hardware_concurrency()+1);
-    Config_updateInt(DOT_NUM_THREADS,dotNumThreads);
-  }
-
```
albert-github, line 2046: Don't we loose the parallel possibility / flexibility in case the executable is / has to be used.

CMakeLists.txt
vtjnash: Yep, that is essentially the trade off here. Though Graphviz follows semantic versioning, so the user can drop in a new shared library just as easily / just the same as changing the binary (breaking changes are hopefully unlikely to happen to the minimal API used here? Recent breaking changes usually remove functions: https://gitlab.com/graphviz/graphviz/-/blob/main/CHANGELOG.md)

src/configimpl.l
vtjnash: Yes, that could be feature gated on the compile flag. Or I'd prefer just entirely deleting this code path so it doesn't have two different ways that things can get built leading to bug report confusion when things get configured subtly differently.

Note that there is potentially not much that runs parallel in `system` itself, since the kernel needs to hold the page table lock during the fork, so the dot process appears to need to be pretty slow and few for this option to be meaningful (and not actually be counter-productive). But in LLVM, it seems the dot processes are many and fast, both of which make fork parallelism quite limited.

CMakeLists.txt
vtjnash: It could also be implemented to use dlopen, so it doesn't actually have to be at compile time and it can still be optional. Is it common to expect users run graphviz without dot? The Debian-based distros make it more annoying to install the libgraphviz-dev package to have dlopen support (vs graphviz to have both the binaries and the shared libraries), so for that specific package manager, having it at compile time is easier.

(with dlopen, the ABI could even be dynamically changed by calling `gvcVersion`, equivalent to calling `dot -V` if the command line API ever changed)

src/configimpl.l
albert-github: The 2 different ways shouldn't be a problem as when the bug report gives the right versions of doxygen and of the settings used it will be quite traceable.
> Yes, that could be feature gated on the compile flag

so both possibilities would be possible for the end user, with the appropriate doxygen setting.
Having a 2 way possibilities is (upon user request!) already possible with the MSC tool (compiled inside, see deps directory, and optional through `MSC_TOOL` setting an executable)

CMakeLists.txt
albert-github: There are some other remarks to be made here
- is it from the license of graphviz allowed to integrate it in this way in a distribution?
- problems might occur when the graphviz library and doxygen are compiled with different compilers which can result in problems with shared libraries (see  https://github.com/doxygen/doxygen/issues/11923 and similar). To avoid such problems with dll's and attempt is made / ongoing to create static libraries fro a number of relevant parts (see https://github.com/doxygen/doxygen-release).
- are there versions available that work correct with doxygen debug executables ?

vtjnash: Dynamic linking vs calling a binary typically does not change the licensing restrictions that must be followed (IANAL, just parroting the FSF from the link below). That said, graphviz is Eclipse Public License which is indeed a licensing violation to use from a GPL2 program according to FSF (https://directory.fsf.org/wiki/License:EPL-2.0)

https://www.gnu.org/licenses/gpl-faq.html#GPLInProprietarySystem

vtjnash:
@albert-github noticed an important issue above, that the license chosen for doxygen likely intended to explicitly prohibit linking to graphviz in the way that this PR would do, so I will formally retract this PR on potentially uncertain legal grounds. I doubt that either project fully intended to cause this licensing conflict, but I also doubt that either project is in a position to dual-license their code now to resolve the conflict.

PR #12040, opened 2026-03-10
bug 129682 Wrong function dependency is made

albert-github:
The dummy argument names (parameters) in the brief description should not be linked to an entity (function, variable, ...).

(this fixes part 2 of the mentioned bug)


src/memberdef.cpp
```diff
@@ -4082,7 +4089,7 @@ void MemberDefImpl::writeMemberDocSimple(OutputList &ol, const Definition *conta
   ol.docify(doxyName);
   if (isVariable() && !argsString().isEmpty() && !isObjCMethod() && !isFunctionPtr())
   {
-    linkifyText(TextGeneratorOLImpl(ol),getOuterScope(),getBodyDef(),this,argsString());
+    linkifyText(TextGeneratorOLImpl(ol),getOuterScope(),getBodyDef(),this,argsString(),&(argumentList()));
```
doxygen: `&argumentList()` is a potentially dangerous construct (now ok-ish, but will break if this method would ever be changed return a value instead of a reference). Better use the underlying `&m_defArgList`

src/memberdef.cpp
```diff
@@ -2517,6 +2523,7 @@ void MemberDefImpl::writeDeclaration(OutputList &ol,
                 isTypedef() ?
                    substitute(argsString(),")(",") (") :
                    combineArgsAndException(argsString(),excpString()), // text
+                &(argumentList()),
```
doxygen: `&argumentList()` is a potentially dangerous construct (now ok-ish, but will break if this method would ever be changed return a value instead of a reference). Better use the underlying `&m_defArgList`

src/memberdef.cpp
albert-github: What is the suggestion? Copy, the name, to a local variable and hand this variable to function?

albert-github: OK I see now the suggestion (use `&m_defArgList` instead) but what to do when it has to be used in another file?

PR #12058, opened 2026-03-25
be slightly more compatible parsing fenced blocks for markdown

nolange:
ignore whitespaces after the fence, allow this common style:

~~~ markdown
``` plantuml
@startuml
class CBase
CBase <|-- CDerived
@enduml
```
~~~

see https://spec.commonmark.org/0.12/#fenced-code-blocks

Btw. is there a reason the code eschews regex for parsing?


albert-github:
The question:
> Btw. is there a reason the code eschews regex for parsing?

should actually not be posed here but in a separate discussion, but in general for cases that are not really complex a regular expression is a but slow and often less intuitive.

Issue #5368, opened 2018-07-02
localizing Doxywizard (Origin: bugzilla #719632)


doxygen:
status NEW severity *enhancement* in component *doxywizard* for *---*
Reported in version *1.8.6-GIT* on platform *Other*
Assigned to: Dimitri van Heesch

Original attachment names and IDs:
- _[20131201_localize_doxywizard.diff](http://ftp.stack.nl/pub/users/dimitri/bugzilla/attachments//bug719632_263230_20131201_localize_doxywizard.diff)_ (ID 263230)
- _[20131226_new_patch_for_issue719632.diff](http://ftp.stack.nl/pub/users/dimitri/bugzilla/attachments//bug719632_264900_20131226_new_patch_for_issue719632.diff)_ (ID 264900)

On 2013-12-01 11:22:06 +0000, Suzumizaki-Kimitaka wrote:

> Created attachment 263230
> Localize doxywizard. GUI and the descriptions of the options.
> 
> I tried to translate the doxywizard GUI and the detail descriptions of the options.
> Unlike the doxygen itself, I use Qt linguist toolkit.
> 
> 1. Build doxygen and doxywizard.
> 2. Execute lupdate (one of Qt linguist tool) against addon/doxywizard/for_lupdate_doxywizard.pro.
> 3. Run Qt linguist and load doxywizard_*.ts file.
> 4. Translate and save.
> 5. Release from menu, to make doxywizard_*.qm file.
> 6. Run doxywizard with option --ld:<qm_dir> and --l:<locale>.
> 
> Wildcard part of .ts and .qm file above is determined at the 2nd step, see help of lupdate.
> The default string depends on the locale of the environment.
> 
> Note:
> 1. I have not made the installer of *.qm file, I don't know to where we should place them.
> 2. Only Japanese translation (doxywizard_ja_JP.ts) is given with the patch.
> 
> Regards,
> Suzumizaki-Kimitaka

On 2013-12-26 11:59:11 +0000, Suzumizaki-Kimitaka wrote:

> Created attachment 264900
> New patch for 1.8.6+
> 
> I made new patch to catch up with doxygen(doxywizard) 1.8.6.
> 
> Regards,
> Suzumizaki-Kimitaka

albert-github:
See also proposed patch #12079

albert-github:
Seen the mentioned pull request and subsequent pull request I think this issue can be considered as fixed but not released.
[...]

PR #12079, opened 2026-04-04
feat: Add multi-language support for doxywizard

StackAndPointer:
- Add TranslationManager singleton for managing translations
- Add OptionTranslations class for config option translations
- Add 8 language translation files (zh_CN, zh_TW, de, fr, ja, ko, es, ru)
- Add Language menu in UI for manual language switching
- Support system language auto-detection on startup
- English fallback for unsupported languages
- All 456 UI text elements now support i18n
- Update CMakeLists.txt for translation file processing


albert-github:
Impressive piece of work.

- Compilation error:
  I tried to compile it but I got:
  ```
  ...\addon\doxywizard\translationmanager.cpp(239): error C2039: 'path': is not a member of 'QLibraryInfo'
  d:\qt\qt-everywhere-src-5.14.0\install_nmake_2017_32\include\qtcore\qlibraryinfo.h(51): note: see declaration of 'QLibraryInfo'
  ...\addon\doxywizard\translationmanager.cpp(239): error C3861: 'path': identifier not found
  ```
  According to https://doc.qt.io/qt-6/qlibraryinfo.html this function is present since Qt 6.0 might be that in version 5 it is called `location` (see https://doc.qt.io/archives/qt-5.15/qlibraryinfo.html)

  When locally changing it translates. So we need probably a conditional for this (see `wizard.cpp` and search for `QT_VERSION`).

- Looking at the language drop-down menu I get:

  image

  Doesn't look 100% correct

- Question how to switch between languages? (Just selection another language dos nothing, as far as I can see)
- Documentation
  I think it would be good to have a separate paragraph in doxygen wizard part of the manual or in the internationalization part of the doxygen manual to specify what has to be done
  - adding a language
  - changing a string (probably just add the string)
  - adding a new string to the doxywizard. In doxygen itself there is the `translator_adapter` etc. to handle this or does the current implementation automatically fall back to the English version of the string?

StackAndPointer:
@albert-github Currently, switching the language requires restarting the application. I need to improve other content based on your suggestions. Thank you.

albert-github:
> Currently, switching the language requires restarting the application

I set the language to `Deutsch` and restarted the doxywizard, but still everything is in English. Also using "Use current settings at startup" does not help.

StackAndPointer:
> > Currently, switching the language requires restarting the application
> 
> I set the language to and restarted the doxywizard, but still everything is in English. Also using "Use current settings at startup" does not help.`Deutsch`

Are the .qm files inside the compiled output folder?

StackAndPointer:
> > Currently, switching the language requires restarting the application
> 
> I set the language to and restarted the doxywizard, but still everything is in English. Also using "Use current settings at startup" does not help.`Deutsch`

I haven't embedded the .qm file into the application yet.

albert-github:
> I haven't embedded the .qm file into the application yet.

OK, would indeed be better. I temporarily placed the `*.qm` besides the executable and after restarting this the texts are in the chosen language though the documentation remains in English, for me that is quite confusing especially as the names of the settings are now written in a "more readable" version (through a.o. `s_optionTranslations` in the new `optiontranslations.cpp`) and these terms are also translated (e.g:
- ALWAYS_DETAILED_SEC
  - Always Detailed Section
  - Immer detaillierter Abschnitt 

When a new setting is added, recently the `MERMAID_...` settings, this also influences the doxywizard and things have to be added.

StackAndPointer:
> > I haven't embedded the .qm file into the application yet.
> 
> OK, would indeed be better. I temporarily placed the besides the executable and after restarting this the texts are in the chosen language though the documentation remains in English, for me that is quite confusing especially as the names of the settings are now written in a "more readable" version (through a.o. in the new ) and these terms are also translated (e.g:`*.qm``s_optionTranslations``optiontranslations.cpp`
> 
> * ALWAYS_DETAILED_SEC
>   
>   * Always Detailed Section
>   * Immer detaillierter Abschnitt
> 
> When a new setting is added, recently the settings, this also influences the doxywizard and things have to be added.`MERMAID_...`

I will solve it soon.

StackAndPointer:
> > 我还没把.qm文件嵌入到应用里。
> 
> 好的，确实会更好。我临时把 放在可执行文件旁边，重启后文本是选定的语言，但文档仍然是英文，这对我来说相当混乱，尤其是设置名称现在用“更易读”的版本写成（通过 a.o.在新版本中），这些术语也被翻译了（例如：`*.qm``s_optionTranslations``optiontranslations.cpp`
> 
> * ALWAYS_DETAILED_SEC
>   * 始终详尽的部分
>   * Immer detaillierter Abschnitt
> 
> 当新增设定，最近设置也会影响多克西巫师，必须添加新内容。`MERMAID_...`

The volume of documentation translation is enormous. I'm currently working on implementing dynamic language setting refresh.

albert-github:
I was just thinking about the "problem" of "ALWAYS_DETAILED_SEC", that would be probably better to have something like `<brief>..</brief>` in the `config.xml` and this can be used (through the translation `configgen.py`) in the doxywizard.

StackAndPointer:
> 我刚才在想“ALWAYS_DETAILED_SEC”这个“问题，如果能在 doxywizard 里用类似的东西，可以通过翻译来使用，可能会更好。`<brief>..</brief>``config.xml``configgen.py`

So, should we use "ALWAYS_DETAILED_SEC" or "Always Detailed Section"
”？

albert-github:
My personal preference would be `ALWAYS_DETAILED_SEC` but I don't know @doxygen his opinion.
I could also think of something where a tooltip is displayed (either way around), but when a descriptive text is used I think it certainly should land in the `config.xml` (like the proposed `<brief>` (or depending on the implementation `<tooltip>`)

StackAndPointer:
>> - Add retranslate() virtual method to Input base class and all derived classes
>> - Implement retranslateUi() for MainWindow, Wizard, Expert, Step1-4
>> - Add OptionTranslations::retranslate() to refresh config option translations
>> - Add translations.qrc for embedding .qm files in executable
>> - Add I18N_GUIDE.md documentation for internationalization development
>> - Update CMakeLists.txt for translation resource compilation
@albert-github

albert-github:
The documentation is now as a stand alone `.md` file but would be better of in the `doc` directory and added to the 
documentation.
Some remarks that quickly caught my eye
- I think there should be for the building process only references to `CMake` and `make` (`nmake`) and not to qt tools (like the QT Creator / lupdate) unless there is no way around.
- "Language Code Standards" is there an authoritative reference where these codes can be found
-  `.ts` should also be in backticks

StackAndPointer:
> The documentation is now as a stand alone `.md` file but would be better of in the `doc` directory and added to the documentation. Some remarks that quickly caught my eye
> 
> * I think there should be for the building process only references to `CMake` and `make` (`nmake`) and not to qt tools (like the QT Creator / lupdate) unless there is no way around.
> * "Language Code Standards" is there an authoritative reference where these codes can be found
> * `.ts` should also be in backticks

Now it has been added to the document, and building with cmake is normal.

StackAndPointer:
@albert-github How about now?

albert-github:
When switching to the German language and looking at the "about" box I still see some English texts, same accounts for the "run" page (and "wizard" page).
When building the documentation I get the error:
```
warning: tag INPUT: input source 'i18n_guide.md' does not exist
```

This is due to the fact that `i18n_guide.md` is not in the `DOC_FILES` list of the `CMakeLists.txt`.

When building the documentation (`make docs`) I see in the HTML result:
image


image

In the pdf it is completely missing as `doc\doxygen_manual.tex` has not been updated.

albert-github:
I also see some strange things on the "wizard" page, this is not correctly updated, as can be seen with the text "No Project logo selected" that is only switched after switching to another page and switching back to the "wizard" page.

Also in the help part nearly the complete text is gone.
I see e.g.:
image


instead of:

image

Though the "help window" should give the full information.

**Edit**
Happens after switching between languages.

StackAndPointer:
> I also see some strange things on the "wizard" page, this is not correctly updated, as can be seen with the text "No Project logo selected" that is only switched after switching to another page and switching back to the "wizard" page.
> 
> Also in the help part nearly the complete text is gone. I see e.g.: image
> 
> instead of:
> 
> image
> Though the "help window" should give the full information.
> 
> **Edit** Happens after switching between languages.

I will fix the issue you mentioned right away.

StackAndPointer:
@albert-github I have fixed the issue you mentioned

albert-github:
When compiling I get the error:
```
addon\doxywizard\doxywizard.cpp(383): error C2248: 'QString::QString': cannot access private member declared in class 'QString'
```

Building the latex documentation gives also an error:
```
! LaTeX Error: Unicode character ├ (U+251C)
               not set up for use with LaTeX.
```
and subsequent. Best to remove the "Appendix: File structure". 

The page should probably be in the Development part of the latex manual:
image

so underneath "internationalization"

In the HTML I still see the strange `i18n_guide` paragraph this is due to the `/*!` and `*/` that are used in the `.md` file and shouldn't be there (makes the difference) and in "Doxywizard Internationalization (i18n) Development Guide" I see:
image

StackAndPointer:
@albert-github Now QT5 should be able to be built normally.
 I'm really not familiar with building LaTeX.

albert-github:
Problem with the:
> I also see some strange things on the "wizard" page, this is not correctly updated, as can be seen with the text "No Project logo selected" that is only switched after switching to another page and switching back to the "wizard" page.

is still present and also the "about" box is in some parts still English (differs a bit per language):
image

Documentation, I'll have a look later today (or tomorrow).

albert-github:
(Made a small mistake before.)
I've been looking at the problems in the documentation and the results of this are in (you can use either):
- patch file [diff.patch](https://github.com/user-attachments/files/26482299/diff.patch)
- compressed file with the changed files [save_doc.tar.gz](https://github.com/user-attachments/files/26482365/save_doc.tar.gz)

StackAndPointer:
@albert-github There should be no problem now

addon/doxywizard/doxywizard.cpp
```diff
@@ -850,10 +850,6 @@ void MainWindow::updateTitle()
 {
   QString title = tr("Doxygen GUI frontend") ;
   m_resetDefault->setEnabled(m_modified);
-  if (m_modified)
-  {
-    title+=QString::fromLatin1(" +");
-  }
```
albert-github, line 746: I think this should remain, it signals that the used doxywizard has been made based on a version that is not yet completely in git.

StackAndPointer: > I think this should remain, it signals that the used doxywizard has been made based on a version that is not yet completely in git.

Do I need to add this code now?

albert-github:
I still see some problems (reported before):
- drop down language menu:
    image

- text that is updated (in wizard tab) only when going to another tab and back
     image
   after going to another tab and back:
    image

- `MERMAID_...` setings in "expert" tab sub part "Dot":
    image



Not mentioned before:
- "Topics" in "wizard" tab not translated
    image

- expert  tab index not translated:
    image

addon/doxywizard/doxywizard.cpp
albert-github: I thought it might be about the git version, but that is not true it is about the fact that, compared to what is stored is default,   a setting has been changed. So this code has to remain.

albert-github: https://github.com/doxygen/doxygen/pull/12079#discussion_r3036544750: Yes, see correct reasoning https://github.com/doxygen/doxygen/pull/12079#discussion_r3036587780

StackAndPointer:
> * expert  tab index

I think it's unnecessary to translate the "expert tab index"

StackAndPointer:
I didn't see the issue with the dropdown language menu. Could you please take a larger screenshot?

albert-github:
Translating the "expert tab index" is a bit questionable but when I see in the index "Project" and in the settings there is "Projektlogo" this feels quite strange to me.

albert-github:
The requested larger screenshot:

image


The problem might be in my installation though, so I I started the "Approve workflows to run" and we will have some Artifacts available from the GitHub Actions

StackAndPointer:
> The requested larger screenshot:
> 
> image
> The problem might be in my installation though, so I I started the "Approve workflows to run" and we will have some Artifacts available from the GitHub Actions

I've found the problem and will solve it immediately

StackAndPointer:
@albert-github The issue you mentioned has been fixed

albert-github:
Looks like I get a compilation error:
```
...\doxygen\addon\doxywizard\wizard.cpp(779): error C2228: left of '.isNull' must have class/struct/union
...\doxygen\addon\doxywizard\wizard.cpp(779): note: type is 'const QPixmap *'
...\doxygen\addon\doxywizard\wizard.cpp(779): note: did you intend to use '->' instead?
```
When changing it to `->isNull()` the application even crashes.

`if (!m_projIconLab->pixmap())` looks like to work.

The problems:
- drop down menu still present at my place
- "No project logo" problem OK
- `MERMAID...` problem OK
- "Topics"  in "wizard" tab problem is translated, OK
- "Topics" in "expert" index tab also translated, rest still old.

New:
- "+" problem, code restored, but I don't think it should be changed when just the language has been changed.

StackAndPointer:
> Looks like I get a compilation error:
> 
> ```
> ...\doxygen\addon\doxywizard\wizard.cpp(779): error C2228: left of '.isNull' must have class/struct/union
> ...\doxygen\addon\doxywizard\wizard.cpp(779): note: type is 'const QPixmap *'
> ...\doxygen\addon\doxywizard\wizard.cpp(779): note: did you intend to use '->' instead?
> ```
> 
> When changing it to the application even crashes.`->isNull()`
> 
> `if (!m_projIconLab->pixmap())` looks like to work.
> 
> The problems:
> 
> * drop down menu still present at my place
> * "No project logo" problem OK
> * `MERMAID...` problem OK
> * "Topics"  in "wizard" tab problem is translated, OK
> * "Topics" in "expert" index tab also translated, rest still old.
> 
> New:
> 
> * "+" problem, code restored, but I don't think it should be changed when just the language has been changed.

What fatal problem does the drop-down menu have?

StackAndPointer:
@albert-github All issues except for the "drop-down menu" have been fixed.

albert-github:
> What fatal problem does the drop-down menu have?

Application just terminates without any warning / message.
Probably a null pointer exception, nothing is shown at my place.

See also https://github.com/doxygen/doxygen/actions/runs/23999231901?pr=12079

The 6.2 builds and the 6.8 builds pass, not the 5.15 builds.

Saw your update, started GitHub Actions run. Created a local run, looks OK.
I'll await the GitHub Actions runs so I can see what happens there with the dropdown menu (and with an other artifact that I didn't report yet).

StackAndPointer:
> > What fatal problem does the drop-down menu have?
> 
> Application just terminates without any warning / message. Probably a null pointer exception, nothing is shown at my place.
> 
> See also https://github.com/doxygen/doxygen/actions/runs/23999231901?pr=12079
> 
> The 6.2 builds and the 6.8 builds pass, not the 5.15 builds.
> 
> Saw your update, started GitHub Actions run. Created a local run, looks OK. I'll await the GitHub Actions runs so I can see what happens there with the dropdown menu (and with an other artifact that I didn't report yet).

I haven't triggered this issue. Could you please perform additional debugging to see where the crash occurs?

albert-github:
> I haven't triggered this issue. Could you please perform additional debugging to see where the crash occurs?

With the fix it doesn't crash anymore (probably an incorrect intermediate test on my side).

albert-github:
I did now a small test with the GitHub Artifact executable
- the problem with the dropdown menu occurs there as well, looks like a it might have to do with the special characters used in the dropdown menu.
- the "an other artifact that I didn't report yet" problem I don't see with the GitHub Actions build, so must be something local with my build / Qt version / Qt Settings.
- Regarding the "+" problem I now see it even when just starting the doxywizard application (so even when no changes have been made!).


I noticed a new problem:
- when being in the "expert" tab and switching to a translated part in the index the right side with the settings doesn't change, looks like to me that the translation back doesn't work properly (in the English version I don't see a problem).

StackAndPointer:
> I did now a small test with the GitHub Artifact executable
> 
> * the problem with the dropdown menu occurs there as well, looks like a it might nave to do with the special characters used in the dropdown menu.
> * the "an other artifact that I didn't report yet" problem I don't see with the GitHub Actions build, so must be something local with my build / Qt version / Qt Settings.
> * Regarding the "+" problem I now see it even when just starting the doxywizard application (so even whe no changes have been made!).
> 
> I noticed a new problem:
> 
> * when being in the "expert" tab and switching to a translated part in the index the right side with the settings doesn't change, looks like to me that the translation back doesn't work properly (in the English version I don't see a problem).

I misunderstood your meaning. I thought that to make the "+" persist, I needed to completely remove it?

albert-github:
> I misunderstood your meaning. I thought that to make the "+" persist, I needed to completely remove it?

No it should only be shown when a setting / option has changed and the "Doxyfile" has not yet been written to disk with the changed settings.
When changing the language this is not a settings change (in the doxygen context).

StackAndPointer:
> I did now a small test with the GitHub Artifact executable
> 
> * the problem with the dropdown menu occurs there as well, looks like a it might nave to do with the special characters used in the dropdown menu.
> * the "an other artifact that I didn't report yet" problem I don't see with the GitHub Actions build, so must be something local with my build / Qt version / Qt Settings.
> * Regarding the "+" problem I now see it even when just starting the doxywizard application (so even whe no changes have been made!).
> 
> I noticed a new problem:
> 
> * when being in the "expert" tab and switching to a translated part in the index the right side with the settings doesn't change, looks like to me that the translation back doesn't work properly (in the English version I don't see a problem).

Is there a specific crash report for the dropdown menu issue?

albert-github:
- "expert" tab switching looks OK now.

albert-github:
The dropdown menu problem, it looks like that in the file translationmanager.cpp the line:
```
fr.nativeName = QString::fromLatin1("Français");
```
should be replaced by
```
fr.nativeName = QString::fromUtf8("Français");
```

analogous for the Spanish language.

StackAndPointer:
Should there be no issues now, right?

albert-github:
- There is still the "+" problem when switching between languages.

Further I think the major problems have been solved, though there are still some items around the representation / what to show
- s_optionTranslations in the new optiontranslations.cpp
   I.e. here the doxygen settings are now represented by a more human readable text instead of the settings mnemonic. In https://github.com/doxygen/doxygen/pull/12079#issuecomment-4186913525 it was already suggested to move the string to the config.xml as a `<brief>` tag which automatically means the the implementer of a new setting has to make the initial, English, text. In this respect also a suggestion was done about a tooltip (https://github.com/doxygen/doxygen/pull/12079#issuecomment-4186924638).
   would be good when @doxygen could shine his light on this
- updating translations, when this would land in the doxygen master version we should also thinking to make a sort of report analogous to https://doxygen.github.io/doxygen-docs/d2/d78/pg_trans.html (and an overview like https://www.doxygen.nl/manual/langhowto.html).

StackAndPointer:
> * There is still the "+" problem when switching between languages.
> 
> Further I think the major problems have been solved, though there are still some items around the representation / what to show
> 
> * s_optionTranslations in the new optiontranslations.cpp
>   I.e. here the doxygen settings are now represented by a more human readable text instead of the settings mnemonic. In [feat: Add multi-language support for doxywizard #12079 (comment)](https://github.com/doxygen/doxygen/pull/12079#issuecomment-4186913525) it was already suggested to move the string to the config.xml as a `<brief>` tag which automatically means the the implementer of a new setting has to make the initial, English, text. In this respect also a suggestion was done about a tooltip ([feat: Add multi-language support for doxywizard #12079 (comment)](https://github.com/doxygen/doxygen/pull/12079#issuecomment-4186924638)).
>   would be good when @doxygen could shine his light on this
> * updating translations, when this would land in the doxygen master version we should also thinking to make a sort of report analogous to https://doxygen.github.io/doxygen-docs/d2/d78/pg_trans.html (and an overview like https://www.doxygen.nl/manual/langhowto.html).

I don't understand in what form the '+' should be retained, or whether it should be removed.

albert-github:
It should only be visible in case a setting is changed an not upon a change of display language.
probably the `SIGNAL` `languageChanged` triggers an update of the "expert" tab ans subsequent a `configChanged` hence setting the "+".
Maybe in the function `languageChanged` the status of `m_modified` should be saved and at the end be reset (and a call be made to `updateTitle`)

StackAndPointer:
I think there shouldn't be any problems now.

StackAndPointer:
@albert-github @doxygen Please review the latest commits.

albert-github:
The latest fix is incorrect, see e.g.  "TAB_SIZE"  when changing this the field becomes red but the "+" does not appear (as it should).
When changing te language (without changing any setting) the "+" still appears.

StackAndPointer:
> The latest fix is incorrect, see e.g. "TAB_SIZE" when changing this the field becomes red but the "+" does not appear (as it should). When changing te language (without changing any setting) the "+" still appears.

I understand where the problem lies

StackAndPointer:
> The latest fix is incorrect, see e.g. "TAB_SIZE" when changing this the field becomes red but the "+" does not appear (as it should). When changing te language (without changing any setting) the "+" still appears.

Now I have solved the problem

albert-github:
Oops I missed the indeed incorrect call to `configChanged` in `onLanguageChanged` although I was looking there with my "solution" but this is better.

StackAndPointer:
@doxygen @albert-github I believe it's ready for merge now. Could you please conduct the final review?

albert-github:
I found an old issue: "localizing Doxywizard (Origin: bugzilla #719632)" i.e. https://github.com/doxygen/doxygen/issues/5368
The attachment will normally be hard to get so: [bug719632_264900_20131226_new_patch_for_issue719632.diff.txt](https://github.com/user-attachments/files/26504656/bug719632_264900_20131226_new_patch_for_issue719632.diff.txt)

maybe some inspirations...

I looked in doxygenwizard.cpp and saw there in the function `saveConfig`:
```
   QMessageBox::warning(this,
        tr("Error saving"),
        QString(tr("Error: cannot open the file "))+fileName+tr(" for writing!\n")+
        tr("Reason given: ")+QString::number(f.error()));
```
Looking at the line:
```
QString(tr("Error: cannot open the file "))+fileName+tr(" for writing!\n")+
```
and the translation into German this would become:
```
QString(tr("Fehler: Datei kann nicht geöffnet werden "))+fileName+tr(" zum Schreiben!\n")+
```
Though I think that the German sentence construction is not correct and the english sentence should be:
```
QString(tr("Error: cannot open the file %1 for writing!").arg(fileName))+
```
and the German translation would be:
```
QString(tr("Fehler: Datei %1 kann nicht zum Schreiben geöffnet werden!\n").arg(fileName))+
```
(not sure about the syntax of the "arg" part, but I think the idea should be clear).

The example give is just one example and there will probably be more places where similar problems occur, in general the construction of sentences in different languages can differ and therefor sentences should not be translated in parts but as a whole sentence.

StackAndPointer:
> I found an old issue: "localizing Doxywizard (Origin: bugzilla #719632)" i.e. #5368 The attachment will normally be hard to get so: [bug719632_264900_20131226_new_patch_for_issue719632.diff.txt](https://github.com/user-attachments/files/26504656/bug719632_264900_20131226_new_patch_for_issue719632.diff.txt)
> 
> maybe some inspirations...
> 
> I looked in doxygenwizard.cpp and saw there in the function `saveConfig`:
> 
> ```
>    QMessageBox::warning(this,
>         tr("Error saving"),
>         QString(tr("Error: cannot open the file "))+fileName+tr(" for writing!\n")+
>         tr("Reason given: ")+QString::number(f.error()));
> ```
> 
> Looking at the line:
> 
> ```
> QString(tr("Error: cannot open the file "))+fileName+tr(" for writing!\n")+
> ```
> 
> and the translation into German this would become:
> 
> ```
> QString(tr("Fehler: Datei kann nicht geöffnet werden "))+fileName+tr(" zum Schreiben!\n")+
> ```
> 
> Though I think that the German sentence construction is not correct and the english sentence should be:
> 
> ```
> QString(tr("Error: cannot open the file %1 for writing!").arg(fileName))+
> ```
> 
> and the German translation would be:
> 
> ```
> QString(tr("Fehler: Datei %1 kann nicht zum Schreiben geöffnet werden!\n").arg(fileName))+
> ```
> 
> (not sure about the syntax of the "arg" part, but I think the idea should be clear).
> 
> The example give is just one example and there will probably be more places where similar problems occur, in general the construction of sentences in different languages can differ and therefor sentences should not be translated in parts but as a whole sentence.

I will address the issue you mentioned

StackAndPointer:
@albert-github I have discovered a new issue unrelated to multiple languages. Doxywizard can open any file and make incorrect modifications. I believe Doxygen should have its own configuration file format?

StackAndPointer:
@albert-github The issue you mentioned above should have been resolved now.

addon/doxywizard/doxywizard_de.ts
```diff
@@ -143,12 +143,12 @@
         <translation>Kein Projekt-Logo ausgewählt.</translation>
     </message>
     <message>
-        <source>Sorry, cannot find file(</source>
-        <translation>Entschuldigung, Datei nicht gefunden(</translation>
+        <source>Sorry, cannot find file(%1);</source>
+        <translation>Entschuldigung, Datei nicht gefunden(%1);</translation>
```
albert-github: The (%1) should in the German translation direct after the word Datei, in general the specific name should directly follow the file / Datei ...

addon/doxywizard/doxywizard_de.ts
```diff
@@ -143,12 +143,12 @@
         <translation>Kein Projekt-Logo ausgewählt.</translation>
     </message>
     <message>
-        <source>Sorry, cannot find file(</source>
-        <translation>Entschuldigung, Datei nicht gefunden(</translation>
+        <source>Sorry, cannot find file(%1);</source>
```
albert-github: I'm mentioning this here but in general in the English text I see `cannot` and `can&apos;t`, should be consistent

albert-github:
Regarding https://github.com/doxygen/doxygen/pull/12079#issuecomment-4192987728
indeed this is unfortunately the case, but it historically grown. When normally using doxygen and updating the settings file (e.g. `doxygen -u Doxyfile`) a `.bak` file is generated.

@doxygen should we use in the doxywizard a similar strategy, so when saving and the file already exists create a `.bak` file?

addon/doxywizard/doxywizard_de.ts
paule32: Hello,

I try this reply from Thunderbird, not WebIF on github.com, so good best try...
For an advise from a German native speaker:

- the part of:
```
<source>Sorry, cannot find file(%1);</source>
<translation>Entschuldigung, Datei nicht gefunden(%1);</translation>
```

- could be:
```
<source>Sorry, cannot find file(%1);</source>
<translation>Entschuldigung, die Datei (%1) konnte nicht gefunden werden.</translation>
```

- or simply:
```
<source>Sorry, cannot find file(%1);</source>
<translation>Datei (%1) nicht gefunden.</translation>
```

Best wishes to Eastern Egg searcher .)
paule32

StackAndPointer:
@albert-github @paule32 I have optimized it based on your suggestions

albert-github:
Small question: why are some sentences multiple times to be translated e.g.
```
<translation>Entschuldigung, die Datei (%1) konnte nicht gefunden werden.</translation>
```
can't this be done with one translation?

StackAndPointer:
> Small question: why are some sentences multiple times to be translated e.g.
> 
> ```
> <translation>Entschuldigung, die Datei (%1) konnte nicht gefunden werden.</translation>
> ```
> 
> can't this be done with one translation?

Given the different contexts, I believe they should be separated.

albert-github:
They might be in different contexts but they are both related to a similar event type, so when going a step further one might also define a function for this and in that case there would only be one string.

StackAndPointer:
@albert-github I adopted your suggestion and fixed some other issues.

albert-github:
Looks like the German translation has not been updated (I still see both text 3 times)

albert-github:
I see also multiple occurences (i.e. twice, hence the 2in the next lines):
```
      2         <source>Browse to a file</source>
      2         <source>Browse to a folder</source>
      2         <source>Docbook</source>
      2         <source>HTML</source>
      2         <source>LaTeX</source>
      2         <source>Next</source>
      2         <source>No Project logo selected.</source>
      2         <source>Previous</source>
      2         <source>Project</source>
      2         <source>Select...</source>
      2         <source>Topics</source>
      2         <source>XML</source>
```
(accounts for all languages, used e.g. `grep "<source>" doxywizard_de.ts | sort | uniq -c | sort -n` to find them)

StackAndPointer:
> I see also multiple occurences (i.e. twice, hence the 2in the next lines):
> 
> ```
>       2         <source>Browse to a file</source>
>       2         <source>Browse to a folder</source>
>       2         <source>Docbook</source>
>       2         <source>HTML</source>
>       2         <source>LaTeX</source>
>       2         <source>Next</source>
>       2         <source>No Project logo selected.</source>
>       2         <source>Previous</source>
>       2         <source>Project</source>
>       2         <source>Select...</source>
>       2         <source>Topics</source>
>       2         <source>XML</source>
> ```
> 
> (accounts for all languages, used e.g. to find them)`grep "<source>" doxywizard_de.ts | sort | uniq -c | sort -n`

They are in different contexts

albert-github:
They might have different contexts but e.g. `Previous` and `Next` have the same purpose. same accounts for probably other texts (e.g."Browse to a file").

StackAndPointer:
@albert-github How about now?

albert-github:
Looks like a lot of new, unwanted, files went into the patch....

Looks like there are still occurrences with 2 and 3 times (hard to tell due to the extra files).

StackAndPointer:
I'll add these to the Refactor translation system later
- Docbook
- HTML
- Input
- LaTeX
- Project
- Source Browser

albert-github:
I don't know whether or not you can run the `grep "<source>" doxywizard_*.ts | sort | uniq -c | sort -n` but this give a nie insight also the e.g in the Russian translation version the text "No Project logo selected." is still 3 times and with the other just once.
The same accounts also for some other texts (for different languages).

StackAndPointer:
@albert-github There are no repetitions now

albert-github:
Indeed I don't see any double occurrences anymore.

I see in the source addon/doxywizard directoty the generated files:
```
        doxywizard_de.qm
        doxywizard_es.qm
        doxywizard_fr.qm
        doxywizard_ja.qm
        doxywizard_ko.qm
        doxywizard_ru.qm
        doxywizard_zh_CN.qm
        doxywizard_zh_TW.qm
```
as they are derived files I'm wondering why they are copied to this directory, any special reason.

paule32:
> Indeed I don't see any double occurrences anymore.
> 
> I see in the source addon/doxywizard directoty the generated files:
> 
> ```
>         doxywizard_de.qm
>         doxywizard_es.qm
>         doxywizard_fr.qm
>         doxywizard_ja.qm
>         doxywizard_ko.qm
>         doxywizard_ru.qm
>         doxywizard_zh_CN.qm
>         doxywizard_zh_TW.qm
> ```
> 
> as they are derived files I'm wondering why they are copied to this directory, any special reason.

isnit doxywizard programmed with Qt5 Framework ?
As such thing, you can use resource file to collect all these files into a one qt resource file, which can be zipped and stored into the executable image

StackAndPointer:
> Indeed I don't see any double occurrences anymore.
> 
> I see in the source addon/doxywizard directoty the generated files:
> 
> ```
>         doxywizard_de.qm
>         doxywizard_es.qm
>         doxywizard_fr.qm
>         doxywizard_ja.qm
>         doxywizard_ko.qm
>         doxywizard_ru.qm
>         doxywizard_zh_CN.qm
>         doxywizard_zh_TW.qm
> ```
> 
> as they are derived files I'm wondering why they are copied to this directory, any special reason.

My oversight. The .qm files will now be automatically cleaned up.

StackAndPointer:
> > Indeed I don't see any double occurrences anymore.
> > I see in the source addon/doxywizard directoty the generated files:
> > ```
> >         doxywizard_de.qm
> >         doxywizard_es.qm
> >         doxywizard_fr.qm
> >         doxywizard_ja.qm
> >         doxywizard_ko.qm
> >         doxywizard_ru.qm
> >         doxywizard_zh_CN.qm
> >         doxywizard_zh_TW.qm
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > as they are derived files I'm wondering why they are copied to this directory, any special reason.
> 
> isnit doxywizard programmed with Qt5 Framework ? As such thing, you can use resource file to collect all these files into a one qt resource file, which can be zipped and stored into the executable image

I had already integrated it into the exe file before. However, I didn't add the function of automatically cleaning up the generated .qm files.

albert-github:
This is better, but in my opinion not 100% OK, that they are removed from the source directory, but in the build directory they can remain just kine object files remain. The build system will otherwise think they have to be regenerated (each time), when a `ts` file changes the build system should and will take care of the rebuild.

I think the lines:
```
add_custom_command(TARGET doxywizard POST_BUILD
  COMMAND ${CMAKE_COMMAND} -E echo "Cleaning up .qm files (embedded in executable)..."
)
foreach(qm_file ${doxywizard_QM_FILES})
  add_custom_command(TARGET doxywizard POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E remove -f ${qm_file}
  )
endforeach()
```
should not be present.

albert-github:
@paule32 
Regarding https://github.com/doxygen/doxygen/pull/12079#issuecomment-4225404101

The doxygenwizard can be build with Qt5 and Qt6 and a resource file is used, check the CMakeLists.txt on the addon/doxywizard directory.

StackAndPointer:
> This is better, but in my opinion not 100% OK, that they are removed from the source directory, but in the build directory they can remain just kine object files remain. The build system will otherwise think they have to be regenerated (each time), when a file changes the build system should and will take care of the rebuild.`ts`
> 
> I think the lines:
> 
> ```
> add_custom_command(TARGET doxywizard POST_BUILD
>   COMMAND ${CMAKE_COMMAND} -E echo "Cleaning up .qm files (embedded in executable)..."
> )
> foreach(qm_file ${doxywizard_QM_FILES})
>   add_custom_command(TARGET doxywizard POST_BUILD
>     COMMAND ${CMAKE_COMMAND} -E remove -f ${qm_file}
>   )
> endforeach()
> ```
> 
> should not be present.

I have corrected

doxygen:
> Regarding [#12079 (comment)](https://github.com/doxygen/doxygen/pull/12079#issuecomment-4192987728) indeed this is unfortunately the case, but it historically grown. When normally using doxygen and updating the settings file (e.g. `doxygen -u Doxyfile`) a `.bak` file is generated.
> 
> @doxygen should we use in the doxywizard a similar strategy, so when saving and the file already exists create a `.bak` file?

Yes I think this makes sense, or maybe even better to ask the user if overwriting is desired with a dialog.

doxygen:
Regarding this PR: I'm a bit worried about the maintenance. 

Especially the OptionTranslations section will often change I'd expect. How to keep it up to date / in-sync with config.xml? The description of the option itself is not translated, so seems a bit limited.

StackAndPointer:
> Regarding this PR: I'm a bit worried about the maintenance.
> 
> Especially the OptionTranslations section will often change I'd expect. How to keep it up to date / in-sync with config.xml? The description of the option itself is not translated, so seems a bit limited.

Does OptionTranslations also adopt the strategy of generating translations using Python scripts? Describing translation workload is quite substantial, and even with AI translation, it requires a considerable amount of time.

StackAndPointer:
I will add the internationalization of configuration option descriptions later.

StackAndPointer:
@albert-github @doxygen Please review it.

albert-github:
@doxygen 

- The translation of the `config.xml` is one of the biggest worries for me as well, especially seen the, incomplete, translations in as added by 9d6be99cd8cecd1d3cd0c46570ccf571c5bad4fc (In a later comment I will give a, small, review of this implementation)
- Another problem I addressed is the translation of the mnemonic in the expert tab (see: https://github.com/doxygen/doxygen/pull/12079#issuecomment-4186924638)
- The problem regarding the writing of the "Doxyfile" you addressed in https://github.com/doxygen/doxygen/pull/12079#issuecomment-4226319656, should this be part of this PR or a separate PR? (I think it should be a separate PR)

albert-github:
@StackAndPointer 

- what is the purpose of the file `sync_option_translations.py`?
- It looks like that not all strings are in the different translations (looks like the translation for `<source>group_graphs</source>` is missing for `fr`, `jp` and `ko`)
- The `*.ts` files get through the configuration options a bit bloated, I think it would be better to have the `OptionDocs` context part in a separate file (the both files for each language could be concatenated in `CMakeLists.txt` and used by the translator.

As mentioned before the translation is a bit incomplete when looking at e.g. the `CALLER_GRAPH` we see
image

and in the German translation:
image

The original text in config.xml is:
```xml
    <option type='bool' id='CALLER_GRAPH' defval='0' depends='HAVE_DOT'>
      <docs>
<![CDATA[
 If the \c CALLER_GRAPH tag is set to \c YES then Doxygen will
 generate a caller dependency graph for every global function or class method.
 <br>Note that enabling this option will significantly increase the time of a run.
 So in most cases it will be better to enable caller graphs for selected
 functions only using the \ref cmdcallergraph "\\callergraph" command.
 Disabling a caller graph can be accomplished by means of the command
 \ref cmdhidecallergraph "\\hidecallergraph".
]]>
      </docs>
    </option>
```
We see al least:
- missing text (i.e. the Note part)
- the occurrence of `\c` in the translated text, normally this is translated in the `getDocsForNode` function in `expert,cpp` also is valid for other doxygen commands in the text
- the "default" and "requires" clauses are missing (generated by `configgen.py` with the help of the section `<generator>` in the config.xml`)

StackAndPointer:
> @StackAndPointer
> 
> * what is the purpose of the file ?`sync_option_translations.py`
> * It looks like that not all strings are in the different translations (looks like the translation for  is missing for ,  and `<source>group_graphs</source>``fr``jp``ko`)
> * The  files get through the configuration options a bit bloated, I think it would be better to have the  context part in a separate file (the both files for each language could be concatenated in  and used by the translator.`*.ts``OptionDocs``CMakeLists.txt`
> 
> As mentioned before the translation is a bit incomplete when looking at e.g. the we see`CALLER_GRAPH` image
> 
> and in the German translation: image
> 
> The original text in config.xml is:
> 
> ```
>     <option type='bool' id='CALLER_GRAPH' defval='0' depends='HAVE_DOT'>
>       <docs>
> <![CDATA[
>  If the \c CALLER_GRAPH tag is set to \c YES then Doxygen will
>  generate a caller dependency graph for every global function or class method.
>  <br>Note that enabling this option will significantly increase the time of a run.
>  So in most cases it will be better to enable caller graphs for selected
>  functions only using the \ref cmdcallergraph "\\callergraph" command.
>  Disabling a caller graph can be accomplished by means of the command
>  \ref cmdhidecallergraph "\\hidecallergraph".
> ]]>
>       </docs>
>     </option>
> ```
> 
> We see al least:
> 
> * missing text (i.e. the Note part)
> * the occurrence of  in the translated text, normally this is translated in the  function in  also is valid for other doxygen commands in the text`\c``getDocsForNode``expert,cpp`
> * the "default" and "requires" clauses are missing (generated by  with the help of the section  in the config.xml`)`configgen.py``<generator>`

sync_option_translations.py is used to ensure synchronization with config.xml.

albert-github:
> sync_option_translations.py is used to ensure synchronization with config.xml.

It is nowhere called.
Probably should  be integrated in `configgen.py` in one way or another.

StackAndPointer:
@albert-github @doxygen Please review it.

albert-github:
- translation directory is a good idea
- I see an inconsistency in the number of translations
    ```
    doxywizard_de.qm 797 translations
    doxywizard_es.qm 797 translations
    doxywizard_fr.qm 482 translations
    doxywizard_ja.qm 482 translations
    doxywizard_ko.qm 482 translations
    doxywizard_ru.qm 483 translations
    doxywizard_zh_CN.qm 482 translations
    doxywizard_zh_TW.qm 482 translations

    doxywizard_options_de.qm 314 translations
    doxywizard_options_es.qm 314 translations
    doxywizard_options_fr.qm 315 translations
    doxywizard_options_ja.qm 315 translations
    doxywizard_options_ko.qm 315 translations
    doxywizard_options_ru.qm 315 translations
    doxywizard_options_zh_TW.qm 315 translations
    ```
- I just looked at the `CALLER_GRAPH` of the German translation and saw:
   image
  Compared to   
   image
  so still quite a bit missing, also the "default" value and incorrect open / close of the `\c` comamnds
  
  I also looked  at the French treanslation
   image
  here there is some more text though the `\hidecallergraph` etc. shouldn't be links

StackAndPointer:
> * translation directory is a good idea
> * I see an inconsistency in the number of translations
>   ```
>   doxywizard_de.qm 797 translations
>   doxywizard_es.qm 797 translations
>   doxywizard_fr.qm 482 translations
>   doxywizard_ja.qm 482 translations
>   doxywizard_ko.qm 482 translations
>   doxywizard_ru.qm 483 translations
>   doxywizard_zh_CN.qm 482 translations
>   doxywizard_zh_TW.qm 482 translations
>   
>   doxywizard_options_de.qm 314 translations
>   doxywizard_options_es.qm 314 translations
>   doxywizard_options_fr.qm 315 translations
>   doxywizard_options_ja.qm 315 translations
>   doxywizard_options_ko.qm 315 translations
>   doxywizard_options_ru.qm 315 translations
>   doxywizard_options_zh_TW.qm 315 translations
>   ```
> * I just looked at the  of the German translation and saw:
>   图片
>   Compared to
>   图片
>   so still quite a bit missing, also the "default" value and incorrect open / close of the  comamnds`CALLER_GRAPH``\c`
>   I also looked  at the French treanslation
>   图片
>   here there is some more text though the  etc. shouldn't be links`\hidecallergraph`

I don't think it's necessary to translate the default value information

albert-github:
> I don't think it's necessary to translate the default value information

I think it is necessary (all translations should have the same information)  and it is missing now so the user won't always know what to expect.

When there are more then one possibility like with `CLASS_GRAPH`:
    image
    image
    image

Here the defaults are show, but not the "Possible values" ( also from the `<generator>` part).

StackAndPointer:
I might not be able to solve the problem until tomorrow

albert-github:
By chance I saw a minor problem in the about box :

image


there is a `)` to many due to:
```
tr("(Created with Qt version %1)").arg(QString::fromLatin1(QT_VERSION_STR));
```
in `doxywizard.cpp` (`)` after `%1` should not be present.

Also in the about box the translation is split in 2 `tr` functions whilst in the display of the version (`else if (!qstrcmp(argv[1],"--version"))` part) it is nicely 1.

albert-github:
Looks like the Russian translation has but no other language has:
```
<source>That default image has to be specified with full path.</source>
```
I also saw in the Russian translation (it just caught my eye):
```
<message>
        <source>is set to</source>
        <translation>establecido en</translation>
    </message>
```
looks to me Spanish translation.

I see some improvements in the layout, thought still see some problems
- with `\c`
- not always empty lines in respect to "default", "possible values" etc.

probably this is still ongoing work

StackAndPointer:
Purpose of sync_option_translations.py
Option                Description
--config              config.xml path (default: ../../src/config.xml)
--options             optiontranslations.cpp path (default: ./optiontranslations.cpp)
--update              Generate entries for optiontranslations.cpp
--docs                Analyze documentation translation entries in .ts files
--apply               Actually modify .ts files (must be used with --docs)
--ts-dir              .ts files directory (default: ./translations)
--ts-file             Check a specific .ts file
--options-only        Only process doxywizard_options_*.ts files
--verbose             Verbose output
--limit               Limit the number of entries displayed (default: 10)

StackAndPointer:
After everything is confirmed to be fine, I think we need to update the content of i18n_guide.md.

StackAndPointer:
@albert-github  Please review it.

albert-github:
I had a quick look and found in the German and Russian translations:
image

image

Looks like to occur in other languages as well (didn't check all though) I didn't look further for other tags with similar behavior.


Looking at the `ALIASES` tag in the released doxywizard (versiion 1.16.1) we see:

image


in the new version we now see:

image

so something wen wrong with the `\n` in the `\ref` command 


> After everything is confirmed to be fine, I think we need to update the content of i18n_guide.md.

An update would never hurt.

addon/doxywizard/expert.cpp
```diff
@@ -53,6 +55,69 @@ static QString convertToComment(const QString &s)
   }
 }
 
+static QString convertDoxyCmdToHtml(const QString &s)
```
albert-github: I understand that it is more convenient / better to have this routine, but why is the slightly different but working correctly code still present in `getDocsForNode`, I would have expected that the code from `getDocsForNode` would have been taken and placed in the new routine.
This might also resulted in the fact that separate handling of the `<br>` tag would not be necessary and that the handling of `\n` as mentioned in https://github.com/doxygen/doxygen/pull/12079#issuecomment-4237844942 would work.

albert-github:
Just caught my eye regarding the `<generator>` part of the `config.xml` that e.g. 
```
<source>The default directory is:</source>
```
is translated in the "regular" doxywizard translation file and not in the "options" file where I think it might belong.
(will probably be valid for all `<generator>` settings.)

I also see that not the arguments are not added in the translation ad that the translation of e.g. `minmaxdef` is missing.

StackAndPointer:
> Purpose of sync_option_translations.py Option Description --config config.xml path (default: ../../src/config.xml) --options optiontranslations.cpp path (default: ./optiontranslations.cpp) --update Generate entries for optiontranslations.cpp --docs Analyze documentation translation entries in .ts files --apply Actually modify .ts files (must be used with --docs) --ts-dir .ts files directory (default: ./translations) --ts-file Check a specific .ts file --options-only Only process doxywizard_options_*.ts files --verbose Verbose output --limit Limit the number of entries displayed (default: 10)

@albert-github I think this Python script, functionality-wise, can exist independently. It doesn't need to be integrated into configgen.py?

albert-github:
The way everything is updated when it would have been integrated can be decided at that moment / at a later moment.

StackAndPointer:
The abnormal issue with minmaxdef still needs further confirmation.

StackAndPointer:
@albert-github Please review it.

albert-github:
Some initial commment

Indeed I'm still missing the min / max  line in the German translation (didn't check other).

in the code I see:
```
    docs+=SA("%{tr:Expert:Minimum value:}")+SA(" ")+child.attribute(SA("minval"))+SA(", ");
    docs+=SA("%{tr:Expert:maximum value:}")+SA(" ")+child.attribute(SA("maxval"))+SA(", ");
    docs+=SA("%{tr:Expert:default value:}")+SA(" ")+child.attribute(SA("defval"))+SA(".");
```
I think this should be one line and not 3 separate parts and furthermore the `.` can have a different representation in different languages (when not mistaken this is a.o. the case for the Chinese language).

For the `ALIASES` I now see `<br/>` 's  but this should be `\n`'s.

albert-github:
Just caught my eye I'm missing in the output also the output for "The default directory is:" (see e.g. the different `HTML_OUTPUT`.
Also I think the line:
```
docs += SA(" ")+SA("%{tr:Expert:The default directory is:}")+SA(" <code>") + defval + SA("</code>.");`
```
should be in one part (in case the order in other languages uis a but different and for the reason of the `.` (as already mentioned in https://github.com/doxygen/doxygen/pull/12079#issuecomment-4245598720)

StackAndPointer:
There are still many issues that need to be resolved, such as incorrect link parsing in Simplified Chinese, Traditional Chinese, and Japanese in some cases.

StackAndPointer:
@albert-github You can first review the existing issues.

albert-github:
The minimum / maximum / default value is now shown, though without the extra newline, the line `docs+=SA("<br/>");` has apparently accidentally been removed.
The split sentence has not been made into one sentence yet, I think that each sentences when in one way or another possible should not be split in parts (for the reasons mentioned before and as there are the easy possibilities `%1` (`%2` ....) and `arg()` this has to be done).

The newline is also missing for the "default" line though not in all cases (and probably more similar lines).

StackAndPointer:
An unrelated issue to this PR: Wizard.cpp seems to be large enough to be split into multiple. cpp files

StackAndPointer:
175b1942d2a844ed4c72e476a5ac53d8
@albert-github It seems that there is a bug that already exists in the official version

StackAndPointer:
@albert-github Please review the latest commits. I believe the issues have been largely resolved.

StackAndPointer:
@doxygen Please review it.

doxygen:
@StackAndPointer How did you create the translations? Did you do it manually, or use some automated translation? I assume it is the latter, in which case I think we should try to bring this kind of automation into the update loop (at least for the options), otherwise I don't see a way to keep things up to date. Any thoughts on this?

StackAndPointer:
> Purpose of sync_option_translations.py
> Option                Description
> --config              config.xml path (default: ../../src/config.xml)
> --options             optiontranslations.cpp path (default: ./optiontranslations.cpp)
> --update              Generate entries for optiontranslations.cpp
> --docs                Analyze documentation translation entries in .ts files
> --apply               Actually modify .ts files (must be used with --docs)
> --ts-dir              .ts files directory (default: ./translations)
> --ts-file             Check a specific .ts file
> --options-only        Only process doxywizard_options_*.ts files
> --verbose             Verbose output
> --limit               Limit the number of entries displayed (default: 10)

@doxygen I tried using this Python script to solve the synchronization issue. At the time, when I was handling the translations, I simply had the AI extract and translate the text directly.

StackAndPointer:
I'm not familiar with LaTeX builds. Prior to this, @albert-github helped me with document processing to make it compatible with LaTeX.

doxygen:
@StackAndPointer Instead of introducing a new script, I would prefer to let the AI (or human) generate translations of config.xml (i.e. `config_de.xml`, `config_fr.xml` etc) and then have an option in the existing `configgen.py` to produce the option translation mappings for Qt. This way we could also let Doxygen generate a default config file with translated texts (`doxygen -g` for english and `doxygen -g:de` for German or something like that).

Maybe an idea to split this PR into a part without translating the options (easier to accept for me), and then add the option part later in a more maintainable/generic way.

doc/i18n_guide.md
```diff
@@ -0,0 +1,632 @@
...
+<context>
+    <name>OptionDocs</name>
+    <message>
+        <source>NEW_OPTION</source>
+        <translation>New Option Description</translation>
+    </message>
+</context>
+```
+
+\subsubsection i18n_config_dynamic Add Dynamic Content Translations
```
doxygen, line 511: This has the same label as already used on line 226, causing the tests to fail.

addon/doxywizard/doxywizard.cpp
```diff
@@ -48,6 +52,164 @@
...
+{
+  return QCoreApplication::translate("Messages", "External");
+}
+
+QString DoxygenWizard::msgDotTopic()
+{
+  return QCoreApplication::translate("Messages", "Dot");
+}
+
+QString DoxygenWizard::translateExpertTopic(const QString &name)
```
doxygen, line 190: Suggest to use a map, the returns the translated string given the name...

doxygen: also all the static functions can be done in a lambda, so you get something like

```c++
#define TR_MSG_ENTRY(name) { QLatin1String(name), []() { return QCoreApplication::translate("Message", name); } }

static QMap<QString,std::function<QString()>> map = { 
TR_MSG_ENTRY("Dot"),
TR_MSG_ENTRY("External"),
}
```

and then when using it `return map[name]();`

addon/doxywizard/translationmanager.cpp
```diff
@@ -0,0 +1,410 @@
...
+    LanguageInfo en;
+    en.code = QString::fromLatin1("en");
+    en.nativeName = QString::fromLatin1("English");
+    en.englishName = QString::fromLatin1("English");
+    en.tsFile = QString();
+    en.qmFile = QString();
+    en.optionsQmFile = QString();
+    m_languages.insert(en.code, en);
+
+    LanguageInfo zhCN;
```
doxygen, line 63: Filling the entries has quite some code duplication. Suggest to put the data in a table, loop over the rows, and derive the `tsFile`, `qmFile`, and `optionsQmFile` fields from the `code` field.

addon/doxywizard/wizard.cpp
```diff
@@ -758,28 +758,55 @@ void Step1::init()
   m_destDir->setText(getStringOption(m_modelData,STR_OUTPUT_DIRECTORY));
 }
 
+void Step1::retranslateUi()
+{
+  m_introLabel->setText(tr("Provide some information "
```
doxygen, line 763: The labels are also specified in the step's constructor. This duplication of string literals should be avoided as it is error-prone.

paule32:
Hello,
just as a Note: I write a similar Tool:
- on left Side you can see Projects for DoxyGen
- on right Side you can see the options, and help for German Users

image

with my own in-built locales Tool, I can provide different Languages for the Application:

image

This notes are only Ideas for the Work for doxywizard

StackAndPointer:
> @StackAndPointer Instead of introducing a new script, I would prefer to let the AI (or human) generate translations of config.xml (i.e. `config_de.xml`, `config_fr.xml` etc) and then have an option in the existing `configgen.py` to produce the option translation mappings for Qt. This way we could also let Doxygen generate a default config file with translated texts (`doxygen -g` for english and `doxygen -g:de` for German or something like that).
> 
> Maybe an idea to split this PR into a part without translating the options (easier to accept for me), and then add the option part later in a more maintainable/generic way.

I considered that approach at the time, but felt it would have too large an impact on the project structure, so I didn't adopt it. If you agree with this implementation approach, that would certainly be even better.

addon/doxywizard/doxywizard.cpp
StackAndPointer: I'll be quite busy next week and probably won't have time to address the issues you mentioned until next weekend.

StackAndPointer:
@doxygen Please review it.

albert-github:
No approval of everything but I have locally problems compiling, so want to see how it goes in the CI.

StackAndPointer:
Perhaps I overlooked something, but my local compilation is working properly.

StackAndPointer:
Minor issue, I will fix it immediately.

StackAndPointer:
@albert-github There shouldn't be any problem now

albert-github:
> Perhaps I overlooked something, but my local compilation is working properly.

Probably working by accident as an old file was present somewhere.

Compiles now for me.


Compilation warning:
```
addon\doxywizard\expert.cpp(177): warning C4010: single-line comment contains line-continuation character
```
for this reason the original comment was reading: `// 2xbackslash -> backslash`

StackAndPointer:
@albert-github There shouldn't be any problem now

albert-github:
`MATHJAX_RELPATH` problem (https://github.com/doxygen/doxygen/pull/12079#issuecomment-4267577103) nice  catch, thanks.
For this I've just pushed a proposed patch, pull request #12097

albert-github:
> An unrelated issue to this PR: Wizard.cpp seems to be large enough to be split into multiple. cpp files

It is a bit large and could be split, but I don't think it is worth to do it (at the moment).


Some remarks:
- I saw that the files:
  ```
   src/config_de.xml
   src/config_es.xml
   src/config_fr.xml
   src/config_ja.xml
  src/config_ko.xml
  src/config_ru.xml
  src/config_zh_CN.xml
  src/config_zh_TW.xml
  ```
  now appeared in favor of `translations/doxywizard_options_....ts` I think this is a good idea but I'm not that happy with the place, I think that it would be better to have a subdirectory under `src` that has these generated files (@doxygen what do you think?)

- In the `src/config_....xml` files I still see some English texts (especially in the examples section and the introduction text. When translating the file I think the entire file should be translated.

- When we look at the documentation of `FILTER_SOURCE_PATTERNS` we see in the English version:

  image

  In this it is not easy to find that the setting mentioned in the documentation `FILTER_SOURCE_FILES` is in the descriptive text is `Filter source files` (one has to do the reading and not just a "pattern matching" in one's memory).

  When looking at e.g. the German translation we see:

  image

  and thus that the descriptive text is `Quelldateien filtern` so here we even have to make much more effort (and even when one speaks reasonably good German it still is very hard to find the right setting).

  In the example the settings are quite close to each other other but this is not necessary always the case.

  In the past I already to use the original settings name and as tooltip the explanation in the selected language, I think that would be much clearer (or maybe some other solution, but important for me is to have the official setting also directly visible on the right hand site. @doxygen what do you think?)

albert-github:
- Code regarding `MATHJAX_RELPATH` (#12097, https://github.com/doxygen/doxygen/pull/12079#issuecomment-4321896402) has been integrated in master on GitHub

StackAndPointer:
> > An unrelated issue to this PR: Wizard.cpp seems to be large enough to be split into multiple. cpp files
> 
> It is a bit large and could be split, but I don't think it is worth to do it (at the moment).
> 
> 
> Some remarks:
> - I saw that the files:
>   ```
>    src/config_de.xml
>    src/config_es.xml
>    src/config_fr.xml
>    src/config_ja.xml
>   src/config_ko.xml
>   src/config_ru.xml
>   src/config_zh_CN.xml
>   src/config_zh_TW.xml
>   ```
>   now appeared in favor of `translations/doxywizard_options_....ts` I think this is a good idea but I'm not that happy with the place, I think that it would be better to have a subdirectory under `src` that has these generated files (@doxygen what do you think?)
> 
> - In the `src/config_....xml` files I still see some English texts (especially in the examples section and the introduction text. When translating the file I think the entire file should be translated.
> 
> - When we look at the documentation of `FILTER_SOURCE_PATTERNS` we see in the English version:
> 
>   image
> 
>   In this it is not easy to find that the setting mentioned in the documentation `FILTER_SOURCE_FILES` is in the descriptive text is `Filter source files` (one has to do the reading and not just a "pattern matching" in one's memory).
> 
>   When looking at e.g. the German translation we see:
> 
>   image
> 
>   and thus that the descriptive text is `Quelldateien filtern` so here we even have to make much more effort (and even when one speaks reasonably good German it still is very hard to find the right setting).
> 
>   In the example the settings are quite close to each other other but this is not necessary always the case.
> 
>   In the past I already to use the original settings name and as tooltip the explanation in the selected language, I think that would be much clearer (or maybe some other solution, but important for me is to have the official setting also directly visible on the right hand site. @doxygen what do you think?)
> 

Is it possible to add a search function?

albert-github:
@StackAndPointer 
> Is it possible to add a search function?

What do you mean? Can you give an example?

doxygen:
> in the context of **DoxyWizard**: what stand **INPUT_FILE_ENCODING** for ?
> 
> I see **INPUT_ENCODING** and below **INPUT_FILE_ENCODING** ... Is **INPUT_FILE_ENCODING** for to specified single Files Encoding ?
> 

@paule32 Don't hijack this PR for unrelated questions. See the documentation for these option in the manual for the difference.

StackAndPointer:
> @StackAndPointer
> 
> > Is it possible to add a search function?
> 
> What do you mean? Can you give an example?

It was my misunderstanding

StackAndPointer:
@albert-github It is strange that texts like 'See also' are not in the .cpp file;I think they should either all be in the .cpp file like 'Minimum value', or all be in the .xml file.

doxygen:
@StackAndPointer I'll start integrating this PR. I plan to do some further improvements. No need to add more commits until further notice.

albert-github:
I see the text "See also" in the original expert.cpp on line 471, but not as a `tr` but `SA` or did I misunderstand something?

albert-github:
@doxygen is it already up to date with the latest changes in `config.xml`  like `DOT_BATCH_SIZE` and the text move in `MERMAID_JS_URL`?

StackAndPointer:
> I see the text "See also" in the original expert.cpp on line 471, but not as a `tr` but `SA` or did I misunderstand something?

I'm sorry, I didn't perform a global search just now.

StackAndPointer:
> @StackAndPointer I'll start integrating this PR. I plan to do some further improvements. No need to add more commits until further notice.

OK

StackAndPointer:
> @doxygen is it already up to date with the latest changes in like and the text move in ?`config.xml``DOT_BATCH_SIZE``MERMAID_JS_URL`

I've just resolved these issues, but I'm not sure if I should commit the changes right now.

StackAndPointer:
> > @doxygen is it already up to date with the latest changes in like and the text move in ?`` config.xml``DOT_BATCH_SIZE``MERMAID_JS_URL ``
> 
> I've just resolved these issues, but I'm not sure if I should commit the changes right now.

@doxygen Do I need to submit these changes?

doxygen:
@StackAndPointer I suggest to wait until I merged this PR and then make a new PR.

StackAndPointer:
> @StackAndPointer I suggest to wait until I merged this PR and then make a new PR.

OK

doxygen:
@StackAndPointer I've merged this PR, please check if the result is ok and submit any additional changes via a new PR.

PR #12085, opened 2026-04-08
Support SVG for different formats in Mermaid diagrams

albert-github:
- support SVG for different formats in Mermaid diagrams
- synchronize handling of output format and output extension for Mermaid

### DIFF
--- a/src/mermaid.cpp
+++ b/src/mermaid.cpp
@@ -44,12 +44,32 @@ QCString MermaidManager::imageExtension(OutputFormat format)
 {
   switch (format)
   {
-    case OutputFormat::Bitmap: return ".png";
-    case OutputFormat::SVG:    return ".svg";
-    case OutputFormat::PDF:    return ".pdf";
+    case OutputFormat::Bitmap: return "png";
+    case OutputFormat::SVG:    return "svg";
+    case OutputFormat::PDF:    return "pdf";
   }
-  return ".png";
+  return "png";
 }
+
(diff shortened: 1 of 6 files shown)


src/mermaid.cpp
```diff
@@ -44,13 +44,33 @@ QCString MermaidManager::imageExtension(OutputFormat format)
...
 }
 
+MermaidManager::OutputFormat MermaidManager::imageFormat(OutputFormat format)
+{
+  if (format == MermaidManager::OutputFormat::Bitmap)
+  {
+    QCString imgExt = getDotImageExtension();
+    if (imgExt=="svg")
+    {
+      return MermaidManager::OutputFormat::SVG;
```
doxygen, line 61: If we set the dot image extension to SVG, it looks like the mermaid images will also be written in SVG format for RTF and Docbook. Do these formats support this?

albert-github: With the small test that I did I can see that there are SVG images in docbook (for XMLMind through batik plugin). In RTF there is also the possibility to use SVG images through a plugin.

doxygen:
@albert-github Check 39f39b0adfcdf7a6887b870490941b0abffa78b6. It should not change the behavior, but I found `imageFormat` a confusing function as it had the same output as input...

albert-github:
Yes the `imageFormat` had a bit a "strange" input. as I didn't think deep enough apparently and staed to close to the old version, using the new implementation is cleaner.

What to do with the coverity warnings?

Issue #12092, opened 2026-04-15
Regression in the parsing of titles of custom navbar tabs


Joseph-Edwards:
**Describe the bug**
When specifying a custom `title` in a custom navbar tab, the text is sometimes not rendered properly if the symbols `<` or `>` are included. Instead, the text between `<` and `>` is being parsed into html. This happens both with the symbol `<` and the code `&lt;`

**Screenshots**

With 1.13.2:

<img width="331" height="191" alt="Image" src="https://github.com/user-attachments/assets/944650d2-a186-4a08-8d8c-a09ba5f2754f" />

With 1.14.0:

<img width="642" height="905" alt="Image" src="https://github.com/user-attachments/assets/a33bb25a-4546-4530-bfe2-9c0e0df06963" />

**To Reproduce**

Use v1.14 or later
[repro.tar.gz](https://github.com/user-attachments/files/26755876/repro.tar.gz)

**Expected behavior**
The navbar title should appear as `outside<inside>`, rather than `outside` with the inclusion of the html tag `<inside></inside>`.

**Version**
```
> doxygen --version
1.14.0 (cbe58f6237b2238c9af7f51c6b7afb8bbf52c866)

> lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04.4 LTS
Release:	24.04
Codename:	noble
```

I compiled doxygen from the source.

**Additional context**
This bug did not exist in v1.13.2

albert-github:
- The `<` and `>` are not valid in the xml file except for tags (and in CDATA sections), so this anyway has to be `&lt;` and &gt;` though just changing this does not help.


I've just pushed a proposed patch, pull request #12093

Joseph-Edwards:
I can confirm that your proposed patch resolves my issue. Thank you very much @albert-github!

albert-github:
- Modified code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in Doxygen version 1.17.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed, and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #12093, opened 2026-04-15
issue #12092 Regression in the parsing of titles of custom navbar tabs

albert-github:
There should be the `&lt;` and `&gt;` in the output file, but it looks like that the `std::ostream` replaces these again with `<` and `>` so best is to escape the `&` as well again (so it will be `&amp;lt;` and `&amp;gt;` to the `std::ostream` resulting in the required text in the javascript file).

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -4052,6 +4052,10 @@ QCString convertToJSString(const QCString &s,bool keepEntities,bool singleQuotes
   {
     switch (c)
     {
+      case '<':  result+="&amp;lt;";
+                 break;
+      case '>':  result+="&amp;gt;";
+                 break;
       case '"':  if (!singleQuotes) result+="\\\""; else result+=c;
                  break;
       case '\'': if (singleQuotes) result+="\\\'"; else result+=c;


src/util.cpp
```diff
@@ -4052,6 +4052,10 @@ QCString convertToJSString(const QCString &s,bool keepEntities,bool singleQuotes
   {
     switch (c)
     {
+      case '<':  result+="&amp;lt;";
+                 break;
+      case '>':  result+="&amp;gt;";
+                 break;
```
doxygen, line 4058: It is not `std::ostream` that does the conversion, it's done by `convertCharEntitiesToUTF8()`.
I think the following fix would be better (less side effects, no double conversion):

```.diff
diff --git a/src/ftvhelp.cpp b/src/ftvhelp.cpp
index b80e16e826..d163f3a8f8 100644
--- a/src/ftvhelp.cpp
+++ b/src/ftvhelp.cpp
@@ -556,8 +556,8 @@ static bool dupOfParent(const FTVNodePtr &n)
 static void generateJSLink(TextStream &t,const FTVNodePtr &n)
 {
   bool nameAsHtml = !n->nameAsHtml.isEmpty();
-  QCString result = nameAsHtml ? n->nameAsHtml : n->name;
-  QCString link = convertToJSString(result,nameAsHtml);
+  QCString result = nameAsHtml ? n->nameAsHtml : convertToHtml(n->name);
+  QCString link = convertToJSString(result,true);
   if (n->file.isEmpty()) // no link
   {
     t << "\"" << link << "\", null, ";
```

albert-github: Might indeed be better, as a similar construct is used in `FTVHelp::Private::generateLink`
Maybe even better to have:
```
   if (n->nameAsHtml.isEmpty()) n->nameAsHtml = convertToHtml(n->name);
  QCString link = convertToJSString(n->nameAsHtml ,true);
```
and in `FTVHelp::Private::generateLink`:
```
if  (n->nameAsHtml.isEmpty()) n->nameAsHtml = convertToHtml(n->name);
```
and replace `text` with `n->nameAsHtml`

@doxygen what do you think? Change as what you think will be the best way.

albert-github:
Looks like the line:
```
    if (Config_getBool(HIDE_SCOPE_NAMES)) result=stripScope(result);
```
has bee removed (probably by accident) in:
```
Commit: 92c33c43583086649af80ca6f57be7a0bafae489 [92c33c4]

Date: Monday, April 27, 2026 1:26:19 PM

PR #12093 Minor refactoring to reduce impact
```

doxygen:
Looks like it indeed.

PR #12101, opened 2026-04-28
i18n: Add translations to some hard coded strings and fix localization configurations files

StackAndPointer:
- Wrap hardcoded strings in expert.cpp with Expert::tr() for translation:
  - 'Note:', 'See also:'
  - 'Doxygen usage', 'External Indexing and Searching'
  - 'Linking to external documentation', 'Including formulas'
- Add translations for these strings in all language .ts files
- Translate `DOT_BATCH_SIZE` option docs in all localized config_xx.xml files
- Fix `MATHJAX_RELPATH` formatting in localized config files to match the original English config.xml (multi-line format with proper list items)


StackAndPointer:
2d9eb898d0ac66291f6a0c2c5bbd08da
By the way, there is a link parsing error in MATHJAX_RELPATHfor Simplified Chinese, Traditional Chinese, and Japanese.

addon/doxywizard/expert.cpp
```diff
@@ -119,21 +119,21 @@ QString convertDoxyCmdToHtml(const QString &s)
 
   // \ref specials
   regexp.setPattern(SA("\\\\ref[ ]+doxygen_usage"));
-  result.replace(regexp,SA("\"Doxygen usage\""));
+  result.replace(regexp,SA("\"") + Expert::tr("Doxygen usage") + SA("\""));
```
albert-github, line 122: I see  now that you used quite often `Expert::tr` but I think it would be better to use `QObject::tr` as tr is defined there (there are no problems now as `QObject` is inherited by `Expert` but I think it is cleaner.

src/translations/config_de.xml
```diff
@@ -3,57 +3,57 @@
   <group name="Project" docs="Project related configuration options">
     <option type="string" id="DOXYFILE_ENCODING" format="string" defval="UTF-8">
       <docs>
-<![CDATA[Dieses Tag gibt die Codierung fur alle Zeichen in der Konfigurationsdatei an, die folgen. Der Standardwert ist UTF-8, was auch die Codierung fur den gesamten Text vor dem ersten Auftreten dieses Tags ist. Doxygen verwendet \c libiconv (oder das in \c libc eingebaute iconv) fur die Transkodierung. Siehe https://www.gnu.org/software/libiconv/ fur die Liste moglicher Codierungen.]]>        
+<![CDATA[Dieses Tag gibt die Codierung fur alle Zeichen in der Konfigurationsdatei an, die folgen. Der Standardwert ist UTF-8, was auch die Codierung fur den gesamten Text vor dem ersten Auftreten dieses Tags ist. Doxygen verwendet \c libiconv (oder das in \c libc eingebaute iconv) fur die Transkodierung. Siehe https://www.gnu.org/software/libiconv/ fur die Liste moglicher Codierungen.]]>
```
albert-github, line 6: I see quite a bit of white space added outside the tags, this should not be the case (as it will later on making it harder to follow the history), it is better to remove the extra white space after the tags.
I don't quickly see any other changes in these translation file (this remark is valid for all the translation files in this directory)

albert-github:
The problem noted with the translations is probably due to a faulty pattern matching where there is no simple space in front of the minus sign or the minus sign is not a simple minus sign

addon/doxywizard/expert.cpp
doxygen: If it `tr` is a member function, just `tr(...)` would have been fine as well I'd expect.

albert-github: The `Expert::tr` is used in non `Expert` member functions e.g. `static QString getDocsForNode(const QDomElement &child)` but still in the file `expert.cpp` though.

PR #12108, opened 2026-04-30
Add automatic backup when saving Doxyfile

StackAndPointer:
When saving a Doxyfile configuration, automatically create a .bak backup of the existing file before overwriting it. This provides a safety net for users who want to revert to previous settings.


StackAndPointer:
I've discovered a serious issue: under language settings other than English, after reading the configuration file, switching page settings reverts to default values!!! I will fix it later.

PR #12109, opened 2026-04-30
fix: Under language settings other than English, after reading the configuration file, switching page settings will revert to default values.

StackAndPointer:
When switching the UI language, the Expert view recreates all input widgets with default values, causing user settings to be lost.

This fix:
- Saves all current configuration values before recreating widgets
- Restores the saved values to the newly created widgets
- Blocks signals during restoration to prevent false 'modified' state

This ensures that configuration settings are preserved across language changes without triggering unsaved changes indicator.

### DIFF
--- a/addon/doxywizard/expert.cpp
+++ b/addon/doxywizard/expert.cpp
@@ -37,6 +37,7 @@
 #include <QTextStream>
 #include <QFileInfo>
 #include <QRegularExpression>
+#include <QMap>
 
 #define SA(x) QString::fromLatin1(x)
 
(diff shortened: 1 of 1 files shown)


paule32:
- you don't need to re-create the components.
- create a list, nd save the ptr's there with a objectName() - Qt5 comes with this:  **setObjectName("text")**
- you can find the objects per:  **widget = self.findChildren(ComponentClass)**
- this would be safer - instead keep safed memory

StackAndPointer:
> * you don't need to re-create the components.
> * create a list, nd save the ptr's there with a objectName() - Qt5 comes with this:  **setObjectName("text")**
> * you can find the objects per:  **widget = self.findChildren(ComponentClass)**
> * this would be safer - instead keep safed memory

Since part of the text is read from XML, it needs to be recreated.

paule32:
I integrate DoxyGen into my current Project, and I don't need to re-create the Components.
I thought, that DoxyGen(Wizard) was programming with Qt5 ?

Here is a screen shot of my implementation:

image

StackAndPointer:
> I integrate DoxyGen into my current Project, and I don't need to re-create the Components.
> I thought, that DoxyGen(Wizard) was programming with Qt5 ?
> 
> Here is a screen shot of my implementation:
> 
> image
> 

Doxywizard can be built with Qt5 or Qt6. The Wizard uses the approach you mentioned, but in the Expert, the components have to be recreated because translation switching relies on reading an external XML file.

PR #12110, opened 2026-04-30
Added validation to verify whether the file is a valid Doxyfile.

StackAndPointer:
When opening a file, validate that it is a valid Doxygen configuration file by checking if the first non-empty line starts with '# Doxyfile'. If not, show a warning dialog to inform the user.

- Use case-sensitive matching for '# Doxyfile' header
- Add translations for error messages in all supported languages

### DIFF
--- a/addon/doxywizard/doxywizard.cpp
+++ b/addon/doxywizard/doxywizard.cpp
@@ -446,45 +446,47 @@ void MainWindow::updateConfigFileName(const QString &fileName)
 
 void MainWindow::loadConfigFromFile(const QString & fileName)
 {
-  // Check if the file is a valid Doxyfile by looking for "# Doxyfile" in the first line
   QFile f(fileName);
-  if (f.open(QIODevice::ReadOnly | QIODevice::Text))
-  {
-    QString firstLine;
-    bool foundDoxyfileHeader = false;
-    
-    // Read lines until we find a non-empty, non-comment line or find the header
-    while (!f.atEnd())
-    {
-      QString line = QString::fromUtf8(f.readLine()).trimmed();
-      if (line.isEmpty())
-        continue;
-      firstLine = line;
-      break;
-    }
-    f.close();
-    
-    // Check if the first non-empty line starts with "# Doxyfile"
-    if (firstLine.startsWith(QString::fromLatin1("# Doxyfile"), Qt::CaseSensitive))
-    {
-      foundDoxyfileHeader = true;
-    }
-    
-    if (!foundDoxyfileHeader)
-    {
-      QMessageBox::warning(this,
-          tr("Invalid configuration file"),
-          tr("The file '%1' does not appear to be a valid Doxygen configuration file.\n\n"
-             "A valid Doxyfile should start with '# Doxyfile' on the first line.")
-          .arg(fileName));
-      return;
-    }
-  }
-  else
+  if (!f.open(QIODevice::ReadOnly | QIODevice::Text))
   {
     QMessageBox::warning(this,
         tr("Error opening file"),
         tr("Cannot open the file %1 for reading.").arg(fileName));
+    return;
+  }
+
+  bool isValidDoxyfile = false;
+  QRegularExpression configPattern(QString::fromLatin1("^[A-Z_][A-Z0-9_]*\\s*(\\+?=)"));
+
+  while (!f.atEnd())
+  {
+    QString line = QString::fromUtf8(f.readLine()).trimmed();
+
+    if (line.isEmpty())
+      continue;
+
+    if (line.startsWith(QString::fromLatin1("# Doxyfile"), Qt::CaseSensitive))
+    {
+      isValidDoxyfile = true;
+      break;
+    }
+
+    if (configPattern.match(line).hasMatch())
+    {
+      isValidDoxyfile = true;
+      break;
+    }
+  }
+  f.close();
+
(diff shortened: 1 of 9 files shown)


albert-github:
> When opening a file, validate that it is a valid Doxygen configuration file by checking if the first non-empty line starts with '# Doxyfile'

This is not true as the following:
```
QUIET = YES
```
is a valid doxygen settings file.

StackAndPointer:
@albert-github How is it now?

albert-github:
There is nowhere a requirement that when a line like `# Doxyfile`, e.g `doxygen -x Doxyfile` generates output with as first line `# Difference with default Doxyfile ...` In principle any comment line, empty line tagline or include line would do.

StackAndPointer:
@albert-github  How is it now?

albert-github:
Well when there are on startup incorrect items in the provided Doxyfile these are directly mentioned on the "run tab" page so the user has already its feedback.
When giving the name of a non-existing file this doesn't result in  an error but the doxywizard assumes that the file will be the name where the settings can be written into.

So I'm not sure at all that this PR adds something (@doxygen what is your opinion?)

StackAndPointer:
> Well when there are on startup incorrect items in the provided Doxyfile these are directly mentioned on the "run tab" page so the user has already its feedback. When giving the name of a non-existing file this doesn't result i an error but the doxywizard assumes that the file will be the name where the settings can be written into.
> 
> So I'm not sure at all that this PR adds something (@doxygen what is your opinion?)

When opening any file, modifying settings, and saving it, the original file will be completely destroyed.

albert-github:
> When opening any file, modifying settings, and saving it, the original file will be completely destroyed.

That is true, but the doxywizard creates a `.bak` file, but when saving the 2nd time the `.bak` file (so the original file) is really overwritten.
Maybe here something can be build to prevent the removal of the original file (though this will probably be difficult e.g. starting doxywizard a second time with the same Doxyfile (this is not the updated one from the first invocation) and saving it again.

doxygen:
@StackAndPointer Sorry for hijacking this PR, but I'm trying to make a new release and see that for the official binaries the language switching in Doxywizard doesn't work. Just nothing happens. See artifacts here https://github.com/doxygen/doxygen-release/actions/runs/25162953817 If you have any idea, please let me know ASAP.

StackAndPointer:
> @StackAndPointer Sorry for hijacking this PR, but I'm trying to make a new release and see that for the official binaries the language switching in Doxywizard doesn't work. Just nothing happens. See artifacts here https://github.com/doxygen/doxygen-release/actions/runs/25162953817 If you have any idea, please let me know ASAP.

@doxygen 
bd3d6712d8995cbaa368433b8a8eb38d
I've identified the issue: the .qm resource files are not being correctly embedded. The .qrc file that was previously removed was intended to embed the .qm resource files into the executable.
After adding the .qm resource files as shown in the image, version 1.17.0 functions properly.

doxygen:
@StackAndPointer Decided to revert the whole multilingual feature. It is just not production ready as it is.

doxygen:
I think we should drop the dynamic selection of the language, as this makes everything unnecessarily complicated. 

Instead, the first time the user starts the application a dialog should be presented to select the language. This language is then used each time the application starts. 

If the user later wants to change the language, a "change language..." menu option can be used. This will show a dialog to select the different language, with a cancel (no change) and quit option. Quit will close the application and when restarted, the application will start with the newly selected language.

This way the correct language only has to be applied once at startup. 

All .qm and .xml files needed for the translation should be in resource files embedded in the executable. No dynamic loading from disk is allowed.

StackAndPointer:
> I think we should drop the dynamic selection of the language, as this makes everything unnecessarily complicated.
> 
> Instead, the first time the user starts the application a dialog should be presented to select the language. This language is then used each time the application starts.
> 
> If the user later wants to change the language, a "change language..." menu option can be used. This will show a dialog to select the different language, with a cancel (no change) and quit option. Quit will close the application and when restarted, the application will start with the newly selected language.
> 
> This way the correct language only has to be applied once at startup.
> 
> All .qm and .xml files needed for the translation should be in resource files embedded in the executable. No dynamic loading from disk is allowed.

@doxygen However, this doesn't seem necessary; the embedding issue is not occurring because of this.
I just woke up, and I'm now planning to configure static linking for Qt6, then verify and fix the .qm resource embedding issue.

doxygen:
@StackAndPointer I also noticed a regression that the list of recently used file was not longer properly stored/restored. This is what made me revert the change. Just too many issues were introduced with this change. 

I find it also a bit strange to start up in English by default and then have the user navigate the English menus to find a way to switch to their native language. This is not how for instance the setup process of for instance an OS, phone, or TV works. There the user is always first prompted to select the language before continuing in their language assuming the get lost if they have to start with English. If we can use the default languages of the OS then this is also fine of course (but a bit hard to test for me).

I don't expect user to constantly switch language, so if we can keep things simple by not offering dynamic switching then nothing is lost in my opinion.

StackAndPointer:
@doxygen Now it already selects the language based on the system language setting when you open it for the first time.

StackAndPointer:
What should we do next?

paule32:
@StackAndPointer 
I suggest that the default Language is English at start-up.
And when Application is up, the user can switch the Language per Menue-Item:
In my Project I have such things:

image

PR #12112, opened 2026-04-30
Fix translation resource embedding for static Qt builds

StackAndPointer:
- Changed QFileInfo::exists() to QFile::exists() in qmFilePath() to properly detect Qt resource files (QFileInfo does not work with the ':/' resource path prefix)
- Added conditional check in CMakeLists.txt to only generate translations.qrc when QM files are available
- This fixes language switching in release builds with static Qt linking

@doxygen

### DIFF
--- a/addon/doxywizard/CMakeLists.txt
+++ b/addon/doxywizard/CMakeLists.txt
@@ -185,16 +185,20 @@ else()
   endif()
 endif()
 
-set(TRANSLATIONS_QRC_CONTENT "<!DOCTYPE RCC>\n<RCC version=\"1.0\">\n    <qresource prefix=\"/translations\">\n")
-foreach(qm_file ${doxywizard_QM_FILES})
-  get_filename_component(qm_name ${qm_file} NAME)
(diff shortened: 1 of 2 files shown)


StackAndPointer:
Is this fix correct?

doxygen:
@StackAndPointer Let's hope this does the trick. 

There is also the qtbase_xx.qm that needs to be embedded in the .qrc file instead of loaded from disk, but I already ignore it for now if it cannot be found.

doxygen:
@StackAndPointer Unfortunately that didn't help, see https://github.com/doxygen/doxygen-release/actions/runs/25178068538 still the switching doesn't have any effect. When switch to German only "switchLanguage(de)" is printed.

doxygen:
@StackAndPointer Now retrying with https://github.com/doxygen/doxygen/commit/2c22f1a5857c17fb6bbff108e4a0c7d21019b136

doxygen:
@StackAndPointer Decided to revert the whole feature. It is just not production ready as it is.

albert-github:
@doxygen 

- Do you plan to reintroduce this work again? or how to proceed?
- small note the src/translations (with files is still present)

PR #12115, opened 2026-05-01
Attempt to fix the embedding issue of translation resources in static Qt builds again

StackAndPointer:
Qt6 changes:
- Use qt_add_translations() which automatically handles .ts -> .qm -> resource embedding
- Set RESOURCE_PREFIX to '/translations' to match code expectations
- Set QM_OUTPUT_DIRECTORY to ensure consistent output location
- Remove QM_FILES_OUTPUT_VARIABLE which was disabling auto-embedding

Qt5 changes:
- Use ALL_TRANSLATION_FILES to compute .qm paths at configure time
- Add OBJECT_DEPENDS to ensure .qm files exist before resource compilation
- Remove conditional check on doxywizard_QM_FILES which could fail

@doxygen Could you take a look at this fix? My device isn't capable of supporting static linking for Qt, so I can’t verify it myself.

### DIFF
--- a/addon/doxywizard/CMakeLists.txt
+++ b/addon/doxywizard/CMakeLists.txt
@@ -172,32 +172,33 @@ set(ALL_TRANSLATION_FILES
 
 find_package(Qt6LinguistTools QUIET)
 if(Qt6LinguistTools_FOUND)
-  set_source_files_properties(${ALL_TRANSLATION_FILES} PROPERTIES OUTPUT_LOCATION ${CMAKE_CURRENT_BINARY_DIR})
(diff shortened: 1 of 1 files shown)


doxygen:
@StackAndPointer Please check the latest master. I've re-introduced the translation support again, but in a more static way, i.e. translations are applied at startup only and switching requires a restart. All resources should now be embedded, and the list of available translations is determined from the stored translation files, rather than from a hardcoded list in the code or Makefile. I still need to update the documentation.

doxygen:
@StackAndPointer this have been solved differently in the meantime.

StackAndPointer:
> @StackAndPointer this have been solved differently in the meantime.

OK

Issue #12113, opened 2026-04-30
Doxygen generates ill-formed XML when \p is at the end of a line


jwakely:
**Describe the bug**

Doxygen generates XML that is not well-formed:

```
$ xmllint --noout xml/algo_8h.xml 
xml/algo_8h.xml:92: parser error : Opening and ending tag mismatch: emphasis line 92 and computeroutput
ts of the output sequence. Evaluates <computeroutput><emphasis></computeroutput>
                                                                               ^
xml/algo_8h.xml:92: parser error : Opening and ending tag mismatch: computeroutput line 92 and emphasis
eroutput><emphasis></computeroutput>(result+N)=binary_op(*(first1+N),</emphasis>
                                                                               ^
```

**To Reproduce**

Given this C++ file:

```c++
  /**
   *  @brief Perform an operation on corresponding elements of two sequences.
   *  @ingroup mutating_algorithms
   *  @param  first1     An input iterator.
   *  @param  last1      An input iterator.
   *  @param  first2     An input iterator.
   *  @param  result     An output iterator.
   *  @param  binary_op  A binary operator.
   *  @return   An output iterator equal to @p result+(last-first).
   *
   *  Applies the operator to the corresponding elements in the two
   *  input ranges and assigns the results to successive elements of the
   *  output sequence.
   *  Evaluates @p
   *  *(result+N)=binary_op(*(first1+N),*(first2+N)) for each
   *  @c N in the range @p [0,last1-first1).
   *
   *  @p binary_op must not alter either of its arguments.
  */
  template<typename InputIterator1, typename InputIterator2,
           typename OutputIterator, typename BinaryOperation>
    OutputIterator
    transform(InputIterator1 first1, InputIterator1 last1,
              InputIterator2 first2, OutputIterator result,
              BinaryOperation binary_op)
    {
      return result;
    }
```

and this Doxyfile:

```
INPUT = algo.h
GENERATE_XML = true
```

the output contains:

```xml
Applies the operator to the corresponding elements in the two input ranges and assigns the results to successive elements of the output sequence. Evaluates <computeroutput><emphasis></computeroutput>(result+N)=binary_op(*(first1+N),</emphasis>(first2+N)) for each <computeroutput>N</computeroutput> in the range <computeroutput></computeroutput>[0,last1-first1).</para>
```

**Expected behavior**

Well-formed XML.

**Version**

1.14.0

Same result with self-built version from Git:
[...]

PR #12122, opened 2026-05-05
issue #12113 Doxygen generates ill-formed XML when \p is at the end of a line

albert-github:
In case of an unclosed tag at the end of a comment block this was already handled but due to the adding of an artificial tag (In this case a pair `<em>` / `</em>`  for the markdown handling of `*...*`) the matching of the tags was artificially corrupted. The later should not be the case.
Now correct output is generated although a warning still is given (for the `</em>` tag but this is correct as here the usage of the markdown replacement is not correct).


doxygen:
@albert-github I've changed the logic a little bit (effect should be the same but easier to understand), see https://github.com/doxygen/doxygen/commit/10f75f518816dacf14b4a49a6eb31c1a91375eec

By accident, I made the changes as part of the merge commit, so now GitHub doesn't recognize that the PR is merged. Please check and close manually if you agree with the changes.

doxygen:
Commit https://github.com/doxygen/doxygen/commit/081bfa315d247c1438f1041f5260dc97e4984fad fixes invalid XML for this case 
```C++
  /** @file */
  /**
   *  <b>
   *  Evaluates @p
   *  *result,*
   *
   *  @p binary_op
   *  </b>
   */
  int x;
```

albert-github:
Didn't see problem.

PR #12126, opened 2026-05-06
Recognize _Nullable and _Nonnull in C prototypes

petris:
It has become common to use _Nullable and _Nonnull to indicate if a pointer argument can be or must not be NULL, see for example manual pages of GNU libc.

Extend the rendering of direction in param command to show this information if it's part of the prototype.

In the XML output, render it as "nonnull" attribute.

### DIFF
--- a/src/docbookvisitor.cpp
+++ b/src/docbookvisitor.cpp
@@ -1435,6 +1435,8 @@ DB_VIS_C
       {
         m_t << "in,out";
       }
+      if (pl.isNullable()) m_t << ",null";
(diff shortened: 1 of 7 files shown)


albert-github:
It looks a bit like this is a compiler specific addition as I could not find any reference to it in the latest available drafts of The C Standard nor the c++ Standard.

To check the results of your work it would be good:


Furthermore:
- looking in the "visitors" I see here `!null` and I'm not sure this is the best representation when in would be better to have the original terms
- it was implemented for a number of "visitors" but there are more and probably it should be implemented there as well.

PR #12163, opened 2026-05-17
Create possibility to dump Doxyfile  from doxywizard

albert-github:
- Create possibility to dump Doxyfile  from doxywizard (with local language settings)
- exit doxywizard after dump or Doxyfile dump

PR #12165, opened 2026-05-17
Combination of `--language` and switching language in doxywizard

albert-github:
In case the `--language` is specified when starting the doxywizard  and switching in the doxywizard to a new language does not take effect as on restart the `--language` kicks in again. We need a new list with the possible arguments and use this at restart.


albert-github:
Note: we don't need to copy `--dump` as this will in future (#12163) be exiting the doxywizard.

PR #12167, opened 2026-05-18
Move "generator" part to the logical part in `config....xml`

albert-github:
Moving the translated values for "generator" part to the `config_....xml` files instead of the `*.ts` files


doxygen:
@albert-github This file has a new conflict. I think you know best how to merge it as you also made the other change.

PR #12168, opened 2026-05-18
Make doxygen output serveable with strict CSP

lws-team:
We have used doxygen for a long time at libwebsockets.org... the fact it produces inline styles / scripts has meant although the html part of the docs were fine, on our strict-CSP server the browser threw away all the inline styles and scripts for security reasons, disabling features like search.

The first of the two patches revisits all the generated and static inline content and migrates it to live in a JS or CSS file, so it is compatible with a strict CSP.

The second patch solves some XSS-able gaps that were left... although now you can use it with a strict CSP, that's less critical than it would be before.

The results can be seen at

https://libwebsockets.org/lws-api-doc-main/html/index.html

which is served with a strict CSP but now has functional doxygen content.


lws-team:
I updated the patches so they pass ctest... the doxygen output at the link is also updated with the output from the updated patches.

doxygen:
@lws-team Looks like there is a regression regarding "Additional inherited member" sections.

You can see it when running Doxygen on its own sources `make docs_internal` in the build directory and point the browser to for instance `doxygen_docs/html/df/d59/class_config_enum.html`. The section header with title `Public Member Functions inherited from ConfigOption` should expand when clicked by doesn't do this anymore.

templates/html/dynsections.js
```diff
@@ -215,3 +215,62 @@ let dynsection = {
 };
 
 /* @license-end */
```
doxygen: this marker needs to move to the end of the file

templates/html/menu.js
```diff
@@ -567,3 +567,22 @@ function initMenu(relPath,treeview) {
 }
 
 /* @license-end */
```
doxygen: this marker needs to move to the end of the file

templates/html/search.js
```diff
@@ -701,8 +706,40 @@ function init_search() {
...
+    searchField.addEventListener("focus", function() { searchBox.OnSearchFieldFocus(true); });
+    searchField.addEventListener("blur", function() { searchBox.OnSearchFieldFocus(false); });
+    searchField.addEventListener("keyup", function(e) { searchBox.OnSearchFieldChange(e); });
+  }
+
   const name = Cookie.readSetting(SEARCH_COOKIE_NAME,0);
   const id = searchBox.GetSelectionIdByName(name);
   searchBox.OnSelectItem(id);
 }
 /* @license-end */
```
doxygen: this marker needs to move to the end of the file

lws-team:
Thanks for the comments... I have updated the same two patches with a v3 that has both the fix for the inheritance toggle (tested by rebuilding the doxygen docs and viewing it on the filesystem in firefox), and put the EOF comments back where they belong.

doxygen:
@lws-team Thanks, looks like this indeed fixes the problems. 
Meanwhile, I found another regression: the navigation panel on the left (`div` with `id` "side-nav") is no longer resizable. The panel on the right (`div` with `id` "page-nav") still is resizable.

lws-team:
Thanks.  I again updated the patched in-place and that bug should also be solved.

doxygen:
Thanks for fixing that. 

I now switched to testing with `HTML_FILE_EXTENSION = .xhtml` (more strict version of HTML) and see that annotated.xhtml now has double `class` attributes for a `tr` tag.

```
XML Parsing Error: duplicate attribute
Location: file:///Users/dimitri/doxygen.build/doxygen_docs/html/annotated.xhtml
Line Number 166, Column 31:
<tr id="row_0_0_" class="odd" class="hidden"><td class="entry"><span class="spacer">&#160;</span><span class="spacer">&#160;</span><span class="icona"><span class="icon">C</span></span><a class="el" href="d4/d19/structanonymous__namespace_02layout_8cpp_03_1_1_element_callbacks.xhtml" target="_self">ElementCallbacks</a></td><td class="desc"></td></tr>
------------------------------^
```

lws-team:
Thanks, those extra classes should be merged in the last force-push.

doxygen:
@lws-team Looks like this broke the expand/collapse arrow functionality in `doxygen_docs/html/annotated.html` and `doxygen_docs/html/files.html` for Doxygen's internal documentation with `HTML_FILE_EXTENSION = .html`

lws-team:
Thanks for finding these.  That also should be fixed in the last force push, I tested it in firefox + ctest.

doxygen:
@lws-team ~~This fixes the problem, but seems to also introduce a new regression. When I look at the inheritance diagram for `doxygen_docs/html/d5/dcc/class_definition.html` (running `make docs_internal`), it places the diagram just outside of the canvas view. You can drag the diagram back into the view by click+move the mouse, but it should be shown correctly in the first place.~~

Nevermind, it was a caching issue. After clearing the output directory and regenerating the images it works as expected.

PR #12173, opened 2026-05-21
Possibility to blend out documentation in doxywizard

albert-github:
Create the possibility to blend out the documentation in the righthand panel of the doxywizard so it is clearer where the different settings are.

### DIFF
--- a/addon/doxywizard/doxywizard.cpp
+++ b/addon/doxywizard/doxywizard.cpp
@@ -51,6 +51,7 @@
 
 // globally accessible variables
 bool DoxygenWizard::debugFlag = false;
+bool DoxygenWizard::hideDocumentation = false;
 QString DoxygenWizard::langCode;
 static QList<QString> newArgs;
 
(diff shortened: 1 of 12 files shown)


doxygen:
@albert-github The problem I see with this change is that if one searches for something, the matches in the documentation are not visible when the documentation is hidden. So I think the flag should only have an effect when in "show all" mode (empty search field).

albert-github:
When searches are used the searches might not be visible, when searching for e.g. `EXTRACT` or `WARN` one will get some matches on the field names. 
I think that in the "search" mode the result might still be usable as just the fields are visible without the interfering documentation.

An alternative might be that in the "search" mode with "hide documentation" set only the fieldnames are searched.

doxygen:
@albert-github Ok, let's keep it as you made it. I've simplified things a bit because there is no real need to hold the hideDocumentation flag as a global (we can use QSettings+the value of the checkbox to keep track).

PR #12195, opened 2026-06-06
Incorrect display of images in the treeview(2)

albert-github:
This is a similar problem as solved with https://github.com/doxygen/doxygen/pull/12187, but now as complicating factor to have an `align` attribute with the `img` tag. Now it does not only influence the treeview in the left panel but also the treeview in the outline panel (right)  and the the table of contents. The example is now like:
```
# Vega-Altair <a href="https://altair-viz.github.io/"><img align="right" src="https://altair-viz.github.io/_static/altair-logo-light.png" height="50"></img></a>

## sect

## sect

### sub sect

### sub sect
```

Example: [example.tar.gz](https://github.com/user-attachments/files/28661746/example.tar.gz)

(Found by Fossies for the Altair package).

### DIFF
--- a/templates/html/doxygen.css
+++ b/templates/html/doxygen.css
@@ -593,6 +593,10 @@ img.formulaInl, img.inline {
 	vertical-align: middle;
 }
 
+div.toc img.inline {
+        float: none;
+        vertical-align: top
+}
 div.center {
 	text-align: center;
 	margin-top: 0px;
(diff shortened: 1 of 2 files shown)


doxygen:
@albert-github I've also limited the maximum height of the images in the indices so they do not stick out.

albert-github:
@doxygen, would indeed be good that the images are limited to a certain height.
Pity there are that much sizes involved (and that they are not parameterized).

I also retested the #12187 (the similar problem, but with vim) and here the image is not that well readable anymore after the limitation, but I think that when a user provides an image he should see to it that it displays well (and he can always tweak his own css file).

PR #12227, opened 2026-06-23
Reformulate conditional inclusions in configuration file

albert-github:
Reformulate conditional inclusions in configuration file from a "don't include" to an "do include" version


src/configgen.py
```diff
@@ -144,12 +145,22 @@ def addValues(var, node):
                 print("  %s->addValue(\"%s\");" % (var, name))
 
 
+def getFilter(node):
```
doxygen: I think it would be cleaner to pass down the the mode as a parameter rather than as a global

addon/doxywizard/expert.cpp
```diff
@@ -107,6 +107,14 @@ static void translateEnumDescription(QDomElement &valueElem,const QDomElement &t
   }
 }
 
+static bool getFilter(QDomElement docsVal)
```
doxygen: It think it would be clearer to pass the filter value as a parameter rather than hardcoding 'doxywizard' inside the filter.

src/configgen.py
albert-github: It does however mean that the mode argument will be dragged through the entire call tree.

addon/doxywizard/expert.cpp
albert-github: It does however mean that the mode argument will be dragged through the entire call tree.

Issue #12249, opened 2026-07-17
HTML mobile view: expanding and closing the hamburger menu creates a gap at the bottom.


quantum-leaps:
**Describe the bug**
With the following HTML settings:

DISABLE_INDEX          = NO
GENERATE_TREEVIEW      = YES
FULL_SIDEBAR           = NO

HTML mobile view (narrow screen): expanding and closing the hamburger menu creates a gap at the bottom.

**Screenshots**
<img width="1758" height="727" alt="Image" src="https://github.com/user-attachments/assets/8133c270-4d52-4d7b-a43a-bc6a62c3b8b9" />

**To Reproduce**
Completely reproducible.

**Expected behavior**
Closing the mobile menu restores the webpage height to the state before opening the menu.

**Version**
Doxygen 1.17.0

**Stack trace**
N/A

**Additional context**
N/A

albert-github:
It can easily be shown by using the page https://doxygen.github.io/doxygen-docs/dd/da0/todo.html
(I see it with Firefox, Google Chrome and Edge)

albert-github:
I've just pushed a proposed patch, pull request #12250

quantum-leaps:
The proposed patch ([PR #1250](https://github.com/doxygen/doxygen/pull/12250) works for me in the configuration shown in the original issue report. I hope it will be included in the next official Doxygen release.
Thank you!

albert-github:
- Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

doxygen:
This issue was previously marked 'fixed but not released',
which means it should be fixed in Doxygen version 1.18.0.
Please verify if this is indeed the case. Reopen the
issue if you think it is not fixed, and please include any additional information
that you think can be relevant (preferably in the form of a self-contained example).

PR #12250, opened 2026-07-18
issue #12249 HTML mobile view: expanding and closing the hamburger menu creates a gap at the bottom.

albert-github:
- add missing resize call
- remove delay when "slideUp" as it gave a strange jump of the bottom breadcrumb

### DIFF
--- a/templates/html/menu.js
+++ b/templates/html/menu.js
@@ -27,6 +27,7 @@ function initMenu(relPath,treeview) {
   const SHOW_DELAY = 250;  // 250ms delay before showing
   const HIDE_DELAY = 500;  // 500ms delay before hiding
   const SLIDE_DELAY = 250; // 250ms slide up/down delay
+  const SLIDE_DELAY_MAIN = 0; // 0ms slide up/down delay for main menu
   const WHEEL_STEP = 30;   // 30 pixel per mouse wheel tick
   const ARROW_STEP = 5;    // 5 pixel when hovering arrow up/down
   const ARROW_POLL_INTERVAL = 20; // 20ms per arrow up/down check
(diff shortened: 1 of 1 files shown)


templates/html/menu.js
```diff
@@ -138,8 +139,9 @@ function initMenu(relPath,treeview) {
           initResizableIfExists();
         });
       } else {
-        slideUp(mainMenu, SLIDE_DELAY, () => {
+        slideUp(mainMenu, SLIDE_DELAY_MAIN, () => {
```
doxygen, line 142: only adding the `initResizableIfExists()` seems enough. I don't think there is a need for `SLIDE_DELAY_MAIN`

albert-github: For the problem this is correct, but the result gives a nasty jump, in my opinion. as written in the comment:
> remove delay when "slideUp" as it gave a strange jump of the bottom breadcrumb

doxygen: ok, but a `slideUp` of 0 duration doesn't make much sense. Better remove the `slideUp` call then, and place the actions that were inside the `slideUp` body directly in the `else` statement.

albert-github: That is correct but I left it in this way as in that case no code duplication is necessary and is, probably, better understandable.
Feel free to change as you think is best.

doxygen:
@albert-github Decided to keep the animation (result isn't so bad), see https://github.com/doxygen/doxygen/commit/721d44116805e8ebe45dc83330b82a05fec1dd49

Somehow Github doesn't see it as a merge, so I'll close it manually.

albert-github:
At the moment GitHub looks like to have a bit of a latency problem, hence probably that it, yet, doesn't see the closing.

image

PR #12291, opened 2026-08-16
Compile warning in dstring

albert-github:
In `dstring.cpp` we get the warning:
```
D:\a\doxygen\doxygen\src\dstring.cpp(72): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
```

Replaced loop with a correct `size_t` loop.

### DIFF
--- a/src/dstring.cpp
+++ b/src/dstring.cpp
@@ -69,7 +69,11 @@ size_t DString::rfind_insensitive( const char *str, size_t index) const
   else if (index>len) return DString::npos; // bad index
   else if (index+slen>len) index = len-slen; // str would be too long
   const char *pos = data()+index;
-  for (size_t i=index+1; i--;) if (dstrnicmp(pos--,str,slen)==0) return i;
+  for (size_t i=index+1; i>0; ) 
+  {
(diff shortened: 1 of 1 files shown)


src/dstring.cpp
```diff
@@ -69,7 +69,7 @@ size_t DString::rfind_insensitive( const char *str, size_t index) const
   else if (index>len) return DString::npos; // bad index
   else if (index+slen>len) index = len-slen; // str would be too long
   const char *pos = data()+index;
-  for (int i=index; i>=0; i--) if (dstrnicmp(pos--,str,slen)==0) return i;
+  for (size_t i=index+1; i--;) if (dstrnicmp(pos--,str,slen)==0) return i;
```
doxygen: This construct is quite subtle. Copilot suggested to use:
```c++
for (size_t i = index+1; i>0; ) {
    --i;
    /* use i */
}
```
to not have to rely on the post-decrement-in-condition trick.

albert-github: It is indeed quite subtle.
Feel free to change it.

PR #12294, opened 2026-08-17
Compilation warnings regarding `size_t`

albert-github:
- a column number cannot be negative, so better to use `size_t
- correcting some other compilation warnings regarding size_t / int conversions

### DIFF
--- a/src/memberdef.cpp
+++ b/src/memberdef.cpp
@@ -509,7 +509,7 @@ class MemberDefImpl final : public DefinitionMixin<MemberDefMutable>
     const MemberDef *m_categoryRelation = nullptr;
     DString m_declFileName;
     int m_declLine = -1;
-    size_t m_declColumn = -1;
+    size_t m_declColumn = 1;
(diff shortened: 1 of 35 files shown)


src/memberdef.cpp
```diff
@@ -509,12 +509,12 @@ class MemberDefImpl final : public DefinitionMixin<MemberDefMutable>
     const MemberDef *m_categoryRelation = nullptr;
     DString m_declFileName;
     int m_declLine = -1;
-    int m_declColumn = -1;
+    size_t m_declColumn = -1;
```
doxygen: -1 is not a valid default for size_t

albert-github:
See some more problems during compilation, looking into then (commentscan, htmlgen, latexgen, searchindex_js, util)

albert-github:
@doxygen,

Looks like a longer standing problem, maybe you have an idea?

albert-github:
Looks like to happen with a.o. Ubuntu Latest GCC Debug (Arm)

PR #12311, opened 2026-08-27
XML output for function pointers is no longer a valid C declaration

gkiagia:
In 1.18.0 the XML output for function pointers (both `typedef`s and struct/class members) changed shape. `<definition>` is no longer valid C, and `<type>` + `<name>` + `<argsstring>` no longer concatenate into a valid declaration, which they did in every earlier release.

`git bisect` points at 772ac869 ("Refactoring: Better align interface of QCString with std::string"), whose commit message notes `int s.find -> size_t s.find`, so this looks like an unintended side effect of that migration rather than a deliberate format change.

## Minimal example

`src/test.h`:

```c
/*! \file test.h */

/*! \brief A function pointer typedef. */
typedef int (*MyCallback) (int a, int b);

/*! \brief A struct with a function pointer member. */
struct MyIface {
  /*! \brief a member function pointer */
  int (*do_it) (struct MyIface *self, int x);
};
```

`Doxyfile`:

```
PROJECT_NAME = min
INPUT = src
OUTPUT_DIRECTORY = out
GENERATE_HTML = NO
GENERATE_LATEX = NO
GENERATE_XML = YES
QUIET = YES
```

## Expected (1.17.0 and earlier)

```xml
<type>int(*</type>
<definition>typedef int(* MyCallback) (int a, int b)</definition>
<argsstring>)(int a, int b)</argsstring>
<name>MyCallback</name>
```

`type + " " + name + argsstring` → `int(* MyCallback) (int a, int b)` — valid C.

## Actual (1.18.0)

```xml
<type>int(*)</type>
<definition>typedef int(*) MyCallback(int a, int b)</definition>
<argsstring>(int a, int b)</argsstring>
<name>MyCallback</name>
```

`type + " " + name + argsstring` → `int(*) MyCallback(int a, int b)` — not valid C. `<definition>` on its own is not valid C either.

The struct member `MyIface::do_it` changes identically.

## Root cause

`src/doxygen.cpp`, the function-pointer correction in `addVariable()`.

1.17.0, where `QCString::find()` returns `int` and `-1` when not found:

```c
int ai = type.find('[', i);
if (ai > i)                          // no '[' → -1 > i → false
{
  ...                                //   → correctly skipped
}
else if (type.find(')', i) != -1)    // function ptr, not "int (*bla)[10]"
{
  type = type.left(type.length()-1); // "int(*)" → "int(*"
  args.prepend(") ");                // "(a,b)"  → ")(a,b)"
}
```

1.18.0, where `DString::find()` returns `size_t` and `npos` (`src/dstring.h:244`):

```c
size_t ii = i;
size_t ai = type.find('[', ii);
if (ai > ii)                         // no '[' → npos > ii → TRUE
{
  args.prepend(type.mid(ai));        // mid(npos) → "" → no-op
  type = type.left(ai);              // left(npos) → whole string → no-op
}
else if (...)                        // never reached
```

The `find()` return type was migrated but the `ai > i` comparison was not. When the type contains no `[`, `npos > ii` is true, so the "function pointer array" branch is taken instead — and both of its statements are silently no-ops, because `DString::mid()` returns empty for `index > size` and `DString::left(npos)` returns the whole string. The correction never runs and the scanner's raw `int(*)` / `(a, b)` reach the output.

## Impact

Reported downstream in WirePlumber as https://gitlab.freedesktop.org/pipewire/wireplumber/-/issues/995: the Sphinx C domain rejects the declarations Breathe builds from this XML and the affected typedefs and vfunc members drop out of the C domain index, losing all their cross-references.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -3320,7 +3320,7 @@ static void addVariable(const Entry *root,int isFuncPtr=-1)
     {
       size_t ii = i;
       size_t ai = type.find('[',ii);
-      if (ai>ii) // function pointer array
+      if (ai!=DString::npos && ai>ii) // function pointer array
       {
         args.prepend(type.mid(ai));
         type=type.left(ai);
--- a/testing/123/123__function__pointer_8c.xml
+++ b/testing/123/123__function__pointer_8c.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<doxygen xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="compound.xsd" version="" xml:lang="en-US">
+  <compounddef id="123__function__pointer_8c" kind="file" language="C++">
+    <compoundname>123_function_pointer.c</compoundname>
+    <innerclass refid="struct_my_iface" prot="public">MyIface</innerclass>
+    <sectiondef kind="typedef">
+      <memberdef kind="typedef" id="123__function__pointer_8c_1af03e798d107bf0e5d3800db992ae5680" prot="public" static="no">
+        <type>int(*</type>
+        <definition>typedef int(* MyCallback) (int a, int b)</definition>
+        <argsstring>)(int a, int b)</argsstring>
+        <name>MyCallback</name>
+        <briefdescription>
+          <para>A function pointer typedef. </para>
+        </briefdescription>
+        <detaileddescription>
+        </detaileddescription>
+        <inbodydescription>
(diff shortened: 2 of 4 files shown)


albert-github:
The test is nowadays indeed not correct and should indeed contain the test on `npos`. The problem also shows in the HTML output.

As a side note I don't like the way doxygen splits the "type" and "argsstring" here in respect to the round brackets:
```
<type>int(*</type>

<argsstring>)(int a, int b)</argsstring>
```

gkiagia:
It looks strange indeed. The "buggy" XML actually looks better, but...

PR #12321, opened 2026-09-01
doxygen.cpp: fix parsing of typedef of C function pointers (1.8.0 regression)

rouault:
Commit 772ac869946245992ee79d6245d88b95812cb4cd ("Refactoring: Better align interface of QCString with std::string") broke parsing of typedef of C function pointers
like ``typedef void(* foo) (int bar)`` where the definition field after that commit became ``typedef void(*) foo(int bar)`` instead of the expected ``typedef void(* foo) (int bar)``

The reason is that the change of the type of the ``ai`` variable from ``int`` to ``size_t`` causes a test like ``if (ai>ii)`` to be true when ai==npos, which is not expected.

While debugging to spot the offending site, I found other similar patterns, so is likely there might be other bugs due to that commit lurking.

This commit adds a regression test for that use case.

### DIFF
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -3345,7 +3345,7 @@ static void addVariable(const Entry *root,int isFuncPtr=-1)
     {
       size_t ii = i;
       size_t ai = type.find('[',ii);
-      if (ai>ii) // function pointer array
+      if (ai!=QCString::npos && ai>ii) // function pointer array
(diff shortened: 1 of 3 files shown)


rouault:
closing. I see this has already been fixed in master per 2a611fda57f8fa54ca5f2d38bf5418f691f30018

PR #12331, opened 2026-09-06
Add optional LaTeX listings backend for Markdown code fences

jreybert:
## Summary

Adds an opt-in LaTeX backend for ordinary Markdown fenced code blocks using the LaTeX `listings` package.

- Adds `LATEX_USE_LISTINGS`, `LATEX_LISTINGS_STYLE`, and `LATEX_LISTINGS_NUMBERED`.
- Supports fence language, `#id`, `caption=`, and `numbers=left|right|none` attributes.
- Provides the bundled `DoxyCode` listings style, chapter-based listing captions, and copy-safe line numbers using `accsupp`.
- Documents custom listings languages and preserves existing handling for tables, diagrams, snippets, and generated source listings.
- Adds regression coverage for the opt-in backend, default backend behavior, and bundled style.

## Why

The standard `DoxyCode` backend remains the default. The optional backend lets users use native `listings` styling and language support for Markdown fences while keeping Doxygen captions and cross-references consistent.

## Example

[playground.md](https://github.com/user-attachments/files/31890837/playground.md)

Enable the feature with:

```text
LATEX_USE_LISTINGS   = YES
LATEX_LISTINGS_STYLE = DoxyCode
```

Example fence:

````markdown
```{.cpp #parser caption="Parser entry point" numbers=none}
int parse();
```
````

## Validation

- Focused tests: `999_latex_listings`, `1000_latex_default_code`, and `1001_latex_listings_default_style` all pass.
- Full CTest suite: 122/123 pass. The unrelated existing `058_strong_enum` test differs only in XML source-column values (`1` versus `5`).

## Attribution

This change was implemented with assistance from an AI coding agent and reviewed interactively by the contributor.

### DIFF
--- a/doc/markdown.dox
+++ b/doc/markdown.dox
@@ -461,6 +461,20 @@ int func(int a,int b) { return a*b; }
 The dot is optional, the curly braces are optional when the that language name begins with an
 alphabetical character and further characters are alphanumerical characters or a plus sign.
 
+Following the Pandoc fenced-code attribute syntax, an attribute block can also contain
+an identifier and a caption. The identifier is converted to the corresponding Doxygen
+\ref cmdanchor "\\anchor" command and the caption is shown after the code block:
+
+    ~~~~~~~~~~~~~~~{.c #multiply-example caption="Multiplication function"}
+    int multiply(int a,int b) { return a*b; }
+    ~~~~~~~~~~~~~~~
+
+The language, identifier, and caption are independent. Use a class beginning with a
+dot for the language, an identifier beginning with c # for the anchor, and a quoted
+c caption value. This syntax is currently supported for ordinary fenced code blocks;
+fences for diagrams such as c dot, c msc, c plantuml, and c mermaid retain their
+existing specialized handling.
+
 Another way to denote fenced code blocks is to use 3 or more backticks (```):
 
 ~~~~~~{.unparsed}
--- a/doc/markdown.dox
+++ b/doc/markdown.dox
@@ -463,7 +463,9 @@ alphabetical character and further characters are alphanumerical characters or a
 
 Following the Pandoc fenced-code attribute syntax, an attribute block can also contain
 an identifier and a caption. The identifier is converted to the corresponding Doxygen
-\ref cmdanchor "\\anchor" command and the caption is shown after the code block:
+\ref cmdanchor "\\anchor" command and the caption is shown after the code block.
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Comment by jreybert** [comment]
I would like to share a few personal thoughts on the current LaTeX/PDF output and how it presents Doxygen’s capabilities to new users.

Doxygen’s LaTeX/PDF support feels somewhat overlooked compared with its HTML output. The default HTML output already looks very professional without any customization, while still offering extensive options for further styling.
By contrast, the default PDF output (especially the cover page) looks quite minimal. Of course, it is possible, and even recommended, to customize it with a dedicated LaTeX template. However, given the variety of LaTeX environments, ensuring compatibility across all setups is understandably challenging.

I think this may obscure Doxygen’s potential for producing professional documentation. Its extended Markdown support is excellent, and the LaTeX/PDF output can genuinely achieve very high-quality results. Yet the default rendering may prevent new users from realizing what is possible.

The project could benefit from a slightly more polished default cover page and header/footer, to better showcase the level of quality that can be achieved. It might also be worthwhile to add a dedicated documentation page on customizing the LaTeX/PDF output. What do you think?

**Comment by albert-github** [comment]
Regarding your remarks in https://github.com/doxygen/doxygen/pull/12331#issuecomment-5570692592 this does not belong to this proposed PR and thus should not be here it should be in the discussion section (https://github.com/doxygen/doxygen/discussions) so other users might see it as well.

**Comment by albert-github** [comment]
I didn't looks at the validity / desirability of this PR but looked at the warning from the tests:
```
The unrelated existing 058_strong_enum test differs only in XML source-column values (1 versus 5).
```
This might not be unrelated as before we did not observe this and it should be checked as most likely this is from this proposed PR.
Might be caused by missing to insert:
```
lineCount(yytext,yyleng);
```
in doctokenizer.l at the changed rule around line 762.

**Comment by albert-github** [comment]
Thinking about the problem of the test 58 it might be that there is a bit of a hiccup on your side regarding the versions used /synchronization against master as with the commit:
```
Commit: be4fba0ec8a09898231e88d252f94ab06b1da5cc [be4fba0]
Parents: 325a07c408

Date: Thursday, August 20, 2026 11:01:33 AM

Incorrect column number in XML in case of `<tab>`
```
there was a correction in test 58 where column numbers went from `1` to `5`.

The problem indicated in https://github.com/doxygen/doxygen/pull/12331#issuecomment-5571701853 (`lineCount`) is a valid problem.

Issue #12329, opened 2026-09-06
warning: multiple use of section label 'deprecated'


wismill:
**Describe the bug**
Doxygen warns:

```
warning: multiple use of section label 'deprecated'
```

while it should automatically handle duplicated sections.

**To Reproduce**
Create a markdown document with multiple `### Deprecated` section and set `GENERATE_DEPRECATEDLIST = YES` in the configuration file.

See:
- https://github.com/xkbcommon/libxkbcommon/blob/dc602de4d68f53c825c2c9af66bd5f53135964e2/NEWS.md?plain=1
- https://github.com/xkbcommon/libxkbcommon/blob/dc602de4d68f53c825c2c9af66bd5f53135964e2/doc/Doxyfile.in

**Expected behavior**
No warning: duplicated sections are handled automatically.

**Version**
1.19.0 6d2c004cb2fe03b2aa88b7f62e69522e54aaeacd

**Additional context**

It may be due to the special treatment of `GENERATE_DEPRECATEDLIST = YES`, because I have a lot of other repeated section title in the changelog that do not trigger the warning.

albert-github:
When running with the indicated 2 files I don't get a warning, but that is logical as the file `NEWS.md` is excluded, but even when adjusting the doxygen settings file so that the file is processed I don't see message. Also with a small test I created: [example.tar.gz](https://github.com/user-attachments/files/31907335/example.tar.gz)

I don't see the problem.

So:
- Can you please attach a, small, self contained example (source+configuration file in a, compressed, tar or zip file!) that allows us to reproduce the problem?
   Please don't add external links as they might not be persistent (also references to GitHub repositories are considered non persistent).

wismill:
Thanks for tackling this. See this small reproducer based on your test: [example.tar.gz](https://github.com/user-attachments/files/31907778/example.tar.gz).

- Fixed typo
- Added `@deprecated` to trigger the bug.

albert-github:
I've just pushed a proposed patch, pull request #12334

albert-github:
- Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

wismill:
Thank you for this fix and all the others!

PR #12334, opened 2026-09-07
issue #12329 warning: multiple use of section label 'deprecated'

albert-github:
Give the internal `deprecated` list a unique name (same automatically for other generated xref lists, `todo`, `bug`, `test`).

### DIFF
--- a/src/commentscan.l
+++ b/src/commentscan.l
@@ -4442,15 +4442,17 @@ static void addRequirementId(yyscan_t yyscanner)
 //-----------------------------------------------------------------
 
 static void addXRefItemToEntry(Entry *current, const DString &docs,
-                        const DString &listName,const DString &itemTitle,
+                        const DString &_listName,const DString &itemTitle,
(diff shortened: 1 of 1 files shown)


src/commentscan.l
```diff
@@ -4442,15 +4442,17 @@ static void addRequirementId(yyscan_t yyscanner)
 //-----------------------------------------------------------------
 
 static void addXRefItemToEntry(Entry *current, const DString &docs,
-                        const DString &listName,const DString &itemTitle,
+                        const DString &_listName,const DString &itemTitle,
```
doxygen: while technically correct, it is usually discouraged to start identifiers with an `_` (identifiers starting with `__` are reserved, as are identifiers starting with `_` followed by a capital letter). 
See also https://www.geeksforgeeks.org/cpp/what-are-the-rules-for-using-an-underscore-in-a-cpp-identifier/

Suggest to use `baseName` or `listName_`

albert-github: updated to `baseName`

Issue #12339, opened 2026-09-09
Broken @todo styling in HTML output


wismill:
**Describe the bug**
After #12329 `@todo` styling does not work anymore in HTML.

**To Reproduce**

Use a `@todo` entry.

**Expected behavior**

The TODO entry is styled to highlight it.

**Version**

b4d03e16a3c00c1b1017c56134d87b68d2b68253

**Additional context**

It seems to be due to the corresponding `<dl>` having `class="Doxygen_List_todo"` instead of `class="todo"`.

albert-github:
I've just pushed a proposed patch, pull request #12341

albert-github:
- Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

PR #12341, opened 2026-09-10
issue #12339 Broken `@todo` styling in HTML output

albert-github:
In #12329 a problem was noted with pre defined list names and sections, this was fixed by #12334 though a small part had been forgotten.

### DIFF
--- a/src/htmldocvisitor.cpp
+++ b/src/htmldocvisitor.cpp
@@ -2073,17 +2073,19 @@ void HtmlDocVisitor::operator()(const DocXRefItem &x)
 
   forceEndParagraph(x);
   bool anonymousEnum = x.file()=="@";
+  DString key = x.key();
+  if (key.startsWith("Doxygen_List_")) key=key.mid(13);
   if (!anonymousEnum)
   {
     DString fn = x.file();
     addHtmlExtensionIfMissing(fn);
-    m_t << "<dl class=\"" << x.key() << "\"><dt><b><a class=\"el\" href=\""
+    m_t << "<dl class=\"" << key << "\"><dt><b><a class=\"el\" href=\""
         << x.relPath() << fn
         << "#" << x.anchor() << "\">";
   }
   else
   {
-    m_t << "<dl class=\"" << x.key() << "\"><dt><b>";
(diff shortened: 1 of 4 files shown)


src/htmldocvisitor.cpp
```diff
@@ -2073,17 +2073,19 @@ void HtmlDocVisitor::operator()(const DocXRefItem &x)
 
   forceEndParagraph(x);
   bool anonymousEnum = x.file()=="@";
+  DString key = x.key();
+  if (key.startsWith("Doxygen_List_")) key=key.mid(13);
```
wismill: Maybe make `Doxygen_List_` a constant and that can be used everywhere:

```cpp
inline constexpr char Doxygen_List_Prefix[] = "Doxygen_List_";
```

In particular here:

```suggestion
  if (key.startsWith(Doxygen_List_Prefix)) key=key.mid(sizeof(Doxygen_List_Prefix) - 1);
```

Disclaimer: I code in C, so pardon my bad C++.

wismill (changes requested):

src/htmldocvisitor.cpp
albert-github: Might be a good idea as at that moment no typos can occur.
@doxygen should this be implemented (maybe as an element of the class `Doxygen` or of a new class `Doxygen_Constant_Strings`)

doxygen: @albert-github There is already a key.stripPrefix that can be used to strip a prefix from a string and avoid the hardcoded size or name repetition. If the prefix is used more often a constant is indeed a good idea.

wismill: > If the prefix is used more often a constant is indeed a good idea.

It is used in 2 *distant* places: here and in the identifier attribution. Using the same constant in both place ensure they will never *never* get out-of-sync, i.e. fix the current bug for good.

albert-github: Actually there are 3 places where the `Doxygen_List_` is used:
- commentscan.l
- htmldocvisitor.cpp
- reflist.cpp

I've updated the proposed PR

Issue #12346, opened 2026-09-14
DOT_IMAGE_FORMAT=svg:cairo fails since 1.17.0


boyoncho:
**Describe the bug**

`DOT_IMAGE_FORMAT = svg:cairo` works with Doxygen 1.16.1, but fails with 1.17.0 and 1.18.0.

Example error:

```text
error: Failed to rename /workspace/build-1.17.0/html/classBase__inherit__graph.dot.svg:cairo to /workspace/build-1.17.0/html/classBase__inherit__graph.svg:cairo!
```

All versions were tested in the same Ubuntu 26.04 Docker environment with:

```text
dot - graphviz version 14.1.2 (0)
```

With Doxygen 1.16.1, the expected file is generated:

```text
classBase__inherit__graph.svg
```

With Doxygen 1.17.0 and 1.18.0, Graphviz actually generates:

```text
classBase__inherit__graph.dot.cairo.svg
```

while Doxygen tries to rename:

```text
classBase__inherit__graph.dot.svg:cairo
```

This appears to be a regression related to the batched `dot` processing introduced in 1.17.0. The full `svg:cairo` Graphviz format specification is being treated as a filename suffix.

**To Reproduce**

A complete Docker-based reproduction is attached.

[reproduce.tar.gz](https://github.com/user-attachments/files/32179205/reproduce.tar.gz)

Run:

```bash
./reproduce.sh
```

It builds Doxygen 1.16.1, 1.17.0, and 1.18.0 from the official source tarballs and compares their output.

Relevant Doxyfile setting:

```text
DOT_IMAGE_FORMAT = svg:cairo
```

**Expected behavior**

`svg:cairo` should produce normal `.svg` files as it does with Doxygen 1.16.1.

**Version**

* Works: 1.16.1
* Fails: 1.17.0, 1.18.0
* Graphviz: 14.1.2
* Ubuntu: 26.04

albert-github:
I've just pushed a proposed patch, pull request #12347

albert-github:
- Code has been integrated in master on GitHub (please don't close the issue as this will be done at the moment of an official release).

PR #12347, opened 2026-09-14
issue #12346 `DOT_IMAGE_FORMAT=svg:cairo` fails since 1.17.0

albert-github:
As noted in the issue parts of the dot logic are rewritten and due to this the `DOT_IMAGE_FORMAT` with renderer aren't working properly anymore as a file with extension `renderer.format` is created automatically while in the past a file with an explicit extension `format` was created.


src/util.cpp
```diff
@@ -4963,9 +4963,23 @@ bool mainPageHasTitle()
...
+{
+  size_t i= format.find(':'); // strip renderer part when using e.g. 'png:cairo:gd' as format
+  return i==DString::npos ? format : format.left(i);
+}
+
+DString getDotImageExtensionGenerated(const DString &format)
+{
+  StringVector splitExt = split(format.str(),":");
+  DString extGen;
+  for (unsigned i = splitExt.size(); i-- > 0; )
```
doxygen: please use `size_t` instead of `unsigned` and move the `i--;` to the start of the body.

albert-github: I did a cut and past from stack overflow and  hence it slipped in...

Issue #12351, opened 2026-09-16
Add support to PAPER_TYPE for A3 paper.


ProjectHSI:
This issue suggests to add "a3" to the list of supported paper geometry/type.

The existing output looks already good on "a3" with the geometry property manually edited to be "a3", at least from memory. Thus, for larger projects/projects with larger code variable/type names, this proves the beneficial space for long names & graphs (granted, unsure if it would be practical to print).

albert-github:
I indeed doubt whether or not it is practical to print, bind and read though that is up to the user.

You can already accomplish the desired behaving by creating one's own header file by mean of `doxygen -w latex ...` and edit here the section:
```
  \geometry{%
    $papertype,%
    top=2.5cm,%
    bottom=2.5cm,%
    left=2.5cm,%
    right=2.5cm%
  }
```
to e.g.:
```
  \geometry{%
    a3paper,%
    top=2.5cm,%
    bottom=2.5cm,%
    left=2.5cm,%
    right=2.5cm%
  }
```
and don't set the `PAPER_TYPE` or just set it to `a4` and mention the thus created header file in `LATEX_HEADER`.
(All a bit mean, I admit.)

> The existing output looks already good on "a3" with the geometry property manually edited to be "a3"

Did you, besides changing`a4paper`  to  `a3paper`  change anything else?


@doxygen what should we do, as a side note the `PAPER_TYPE` is also used in `rtfgen.cpp`?

ProjectHSI:
> I indeed doubt whether or not it is practical to print, bind and read though that is up to the user.
> 
> You can already accomplish the desired behaving by creating one's own header file by mean of `doxygen -w latex ...` and edit here the section:
> 
> ```
>   \geometry{%
>     $papertype,%
>     top=2.5cm,%
>     bottom=2.5cm,%
>     left=2.5cm,%
>     right=2.5cm%
>   }
> ```
> 
> to e.g.:
> 
> ```
>   \geometry{%
>     a3paper,%
>     top=2.5cm,%
>     bottom=2.5cm,%
>     left=2.5cm,%
>     right=2.5cm%
>   }
> ```
> 
> and don't set the `PAPER_TYPE` or just set it to `a4` and mention the thus created header file in `LATEX_HEADER`. (All a bit mean, I admit.)
> 
> > The existing output looks already good on "a3" with the geometry property manually edited to be "a3"
> 
> Did you, besides changing`a4paper` to `a3paper` change anything else?
> 
[...]

PR #12355, opened 2026-09-18
issue #12351 Add support to PAPER_TYPE for A3 paper.

albert-github:
Add the `A3` `PAPER_TYPE`.

(Note: #12352 has to be integrated beforehand as otherwise we still jump back to the `A4` default).


doxygen:
@albert-github Is this a sensible paper type for Doxygen output, being double the size of an A4?

albert-github:
With the issue I wrote:
> I indeed doubt whether or not it is practical to print, bind and read though that is up to the user.

So here I already expressed my doubts and personally I don't think it is something nice to read but e.g. for poster like sessions it might be useful

Issue #12226, opened 2026-06-23
Anonymous union's `@brief` and documentation shown instead of member's in `Data Fields` table and `Field Documentation`


parmi93:
When a struct contains an anonymous `union` with a named member (e.g., `} handle;`), Doxygen has two issues:

1. The `Data Fields` table shows the `@brief` of the anonymous `union` type instead of the `@brief` of the named member (defined via trailing `/**< ... */` comment).
2. The `Field Documentation` section shows the full documentation of the anonymous `union` type instead of just the documentation of the named member `handle`.

Both issues do not happen with named union types (e.g., `union my_union_t { ... } union_member`), where the member's own brief and documentation are shown correctly.

```c
/**
 * @file temp.h
 *
 * @defgroup my_group
 * @{
 */

/**
 * @brief Configuration struct for the STM32L4xx I2C porting layer.
 */
typedef struct
{
    /**
     * @brief Union holding the I2C handle, which can be either thread-safe or a simple STM32-HAL
     * handle.
     *
     * @note The active member pointer is stored internally for the entire program lifetime.
     *
     * @warning The behaviour is not defined in the following cases:
     * - The active member does not point to a valid, initialized handle when this union is passed
     *   to `esc_stse_bus_init()`.
     */
    union
    {
        /** @brief Thread-safe STM32-HAL I2C handle. Must be initialized by the user before being
         * passed to `esc_stse_bus_init()`.
         *
         * Defined in the **[STM32L4xx HAL Thread-Safe
         * wrapper](https://gitlab.eseye.net/things-development/smartconnect2/middlewares/stm32l4xx_hal-thread-safe)**
         * middleware. Available only when
         * `ESC_STSE_STM32L4XX_HAL_ENABLE_SUPPORT_FOR_THREAD_SAFE_I2C` is defined.
         */
        esc_stm32l4xx_hal_thread_safe_i2c_S *stm32_hal_thread_safe;

        /** @brief Non-thread-safe STM32-HAL I2C handle. Must be initialized by the user before
         * being passed to `esc_stse_bus_init()`.
         *
         * Defined in the [STM32L4xx-HAL
         * driver](https://github.com/STMicroelectronics/stm32l4xx-hal-driver).
         */
        I2C_HandleTypeDef *stm32_hal;
    } handle; /**< @brief I2C handle; the active member of this `union` depends on the value of
               * `is_thread_safe`. */

[...]

PR #12361, opened 2026-09-21
Fix \ref to file-scope statics when no file scope is known (#12360)

### LINKED ISSUE #12360: Regression in 1.18: \ref cannot resolve file-scope static variables (m_fileScope is null on the \ref path)
### Describe the bug

Since 1.18.0, `\ref` / `@ref` can no longer resolve a reference to a **file-scope `static` variable**, even when the reference appears in the very same file as the definition. Doxygen 1.17.0 resolves it fine.

This appears to be a side effect of e1bff0176b285f38434af6b9c303c53651bd8b4f ("Prevent wrong cross references between static functions defined in multiple files"), which is intentional in spirit but over-applies when the lookup has no file scope.

### Expected behavior

`@ref MY_STATIC_CONST` resolves to the documented file-scope static, as it did in 1.17.0.

### Minimal reproducer

`a.h`:

```cpp
/// @brief A documented static constant.
static const int STATIC_CONST = 1;

/// @brief A documented non-static constant.
const int NONSTATIC_CONST = 1;

/// @brief A documented function.
void myFunction();

/// @brief ref to static: @ref STATIC_CONST here.
void caseA();

/// @brief ref to nonstatic: @ref NONSTATIC_CONST here.
void caseC();

/// @brief ref to function: @ref myFunction here.
void caseD();
```

`Doxyfile`:

```
PROJECT_NAME    = ref2
INPUT           = a.h
GENERATE_HTML   = NO
GENERATE_LATEX  = NO
EXTRACT_ALL     = YES
EXTRACT_STATIC  = YES
QUIET           = YES
```

Results:

* **1.17.0** — no warnings; all three references resolve.
* **1.18.0** — `a.h:10: warning: unable to resolve reference to 'STATIC_CONST' for \ref command`

Only the `static` case regresses; the non-static variable and the function still resolve. Adding an explicit file scope (`@ref a.h::STATIC_CONST`) does not help.

### Analysis

e1bff01 added a filter in `SymbolResolver::Private::getResolvedSymbolRec()`:

```cpp
if (emd &&
    emd->isStatic() &&
    emd->getFileDef()!=m_fileScope)
{
  // skip lookup for static members that are not in the current file scope
  return true;
}
```

The `\ref` path never sets a file scope, so `m_fileScope` is `nullptr` and the condition is true for *every* static, not just those from other files:

* `resolveLink()` (`src/util.cpp`) calls `resolveRef(scName,lr,TRUE,resContext,&md,lang)`, leaving the `currentFile` parameter at its default `nullptr`.
* `getDefs()` only propagates it when set: `if (input.currentFile) resolver.setFileScope(input.currentFile);`

So with no file-scope context the new check degrades from "skip statics from other files" to "skip all statics".

### Suggested fix

Only apply the filter when a file scope is actually known, preserving 1.17 behaviour when it is not:

```cpp
if (emd &&
    emd->isStatic() &&
    m_fileScope &&                       // only filter when the file scope is known
    emd->getFileDef()!=m_fileScope)
```

I am happy to open a PR with this change plus a regression test.

### Version

1.18.0 (conda-forge `doxygen-1.18.0`), Linux x86_64. Confirmed working on 1.17.0 with an otherwise identical configuration.

nocnokneo:
Fixes #12360.

### Problem

Since 1.18.0, `\ref` cannot resolve a reference to a file-scope `static` variable, even when the reference sits in the same file as the definition:

```cpp
/// @brief A documented static constant.
static const int MY_STATIC_CONST = 1;

/// @brief Refers to @ref MY_STATIC_CONST here.
void myFunction();
```

```
warning: unable to resolve reference to 'MY_STATIC_CONST' for \ref command
```

1.17.0 resolves this. Non-static variables and functions are unaffected.

### Cause

e1bff01 added a filter in `SymbolResolver::Private::getResolvedSymbolRec()` that skips static members defined in a different file:

```cpp
if (emd &&
    emd->isStatic() &&              // a static function or variable
    emd->getClassDef()==nullptr &&  // not a class member
    emd->getFileDef()!=m_fileScope) // defined in a different file
```

The `\ref` path never establishes a file scope, so `m_fileScope` is `nullptr` there:

* `resolveLink()` calls `resolveRef(scName,lr,TRUE,resContext,&md,lang)`, leaving `currentFile` at its default `nullptr`.
* `getDefs()` only forwards it when set: `if (input.currentFile) resolver.setFileScope(input.currentFile);`

With `m_fileScope == nullptr` the comparison is true for every static, so the check degrades from "skip statics from other files" to "skip all statics".

### Fix

Apply the filter only when the file scope is actually known. Cross-file protection is unchanged wherever a file scope is set (the code linkification paths, which call `setFileScope()`); where it is not, behaviour returns to 1.17.

### Testing

* New test `125_static_ref.cpp` covers `\ref` to a file-scope static. It fails on master with `unable to resolve reference to 'MY_STATIC_CONST'` and passes with this change.
* Full suite: 124/125 pass. The only failure is `012_cite.dox`, which needs `bibtex` (not installed in my environment) and fails identically without this change.
* Checked against a real project that regressed on 1.18: unresolved `\ref` warnings drop from 19 to 3. The remaining 3 are all references to a member of an anonymous union, which is a separate issue (#12226).


albert-github:
I have a small doubt about this fix as to broad as when we have also:
```
/// The class
class cls
{
public:
/// @brief A documented static constant.
static const int cls_STATIC_CONST = 1;

/// @brief A documented non-static constant.
const int cls_NONSTATIC_CONST = 1;

/// @brief A documented function.
void cls_myFunction();

/// @brief ref to static: @ref cls_STATIC_CONST and @ref STATIC_CONST here.
void cls_caseA();

/// @brief ref to nonstatic: @ref cls_NONSTATIC_CONST and @ref NONSTATIC_CONST here.
void cls_caseC();

/// @brief ref to function: @ref cls_myFunction here.
void cls_caseD();
};
```
the `STATIC_CONST` in the class is linked and this should in my opinion not be the case, though looks like this was broken in 1.17.0 as well as there is was linked!

Example: [example.tar.gz](https://github.com/user-attachments/files/32462576/example.tar.gz)

nocnokneo:
Good catch — you are right, and your example shows it clearly. I have pushed a different fix.

Your `bb.h` case is exactly what e1bff01 set out to block: `@ref STATIC_CONST` inside `cls` resolves to the static in `aa.h`, a different file. Running your example three ways confirms my original patch gave that protection up:

| build | `cls_caseA`'s `@ref STATIC_CONST` |
| --- | --- |
| 1.17.0 | resolves to `aa_8h_...` (the cross-file link you object to) |
| master | unresolved — e1bff01 working as intended |
| my first patch | resolves to `aa_8h_...` again — protection lost |

So guarding on `m_fileScope` was too broad, as you suspected: it disabled the filter on the `\ref` path entirely instead of making it work there.

The new version leaves your filter in `symbolresolver.cpp` untouched and fixes the actual gap — the `\ref` path had no file scope to compare against. `resolveLink()` now takes the file the documentation block came from and forwards it to `resolveRef()`, which is what `handleLinkedWord()` already does for autolinks. With that context present the existing check does the right thing on its own:

* `@ref` to a static in its own file resolves again (#12360)
* `@ref` to a static in another file stays unresolved — your `bb.h` case still warns, as you want

On your example the only remaining warning is the cross-file one.

Test 125 now covers both directions, and is deliberately sensitive to each failure mode:

* on master it fails, because the same-file reference does not resolve
* with the filter merely relaxed it also fails, because the cross-file reference then resolves

Full test suite passes (the only failure here is `012_cite.dox`, which needs `bibtex`, absent in my environment, and fails identically without this change).

src/docnode.cpp
```diff
@@ -86,6 +86,14 @@ static const StringUnorderedSet g_plantumlEngine {
...
 // replaces { with < and } with > and also
 // replaces &gt; with < and &gt; with > within string s
+// The file the documentation block being parsed came from, so that symbol
+// lookup can tell a file-scope static in this file from one in another file.
+static const FileDef *currentFileDef(const DocParser *parser)
+{
+  bool ambig = false;
+  return Doxygen::inputNameLinkedMap->findFileDef(parser->context.fileName,ambig);
+}
+
```
albert-github: The first 2 lines of the comment belong to the `unescapeCRef` function and not to this new function.
An old issue here is already that the line:
```
// replaces &gt; with < and &gt; with > within string s
```
should read:
```
// replaces &lt; with < and &gt; with > within string s
```

src/util.cpp
```diff
@@ -2438,7 +2439,7 @@ bool resolveLink(/* in */ const DString &scName,
   else // probably a member reference
   {
     const MemberDef *md = nullptr;
-    bool res = resolveRef(scName,lr,true,resContext,&md,lang);
+    bool res = resolveRef(scName,lr,true,resContext,&md,lang,false,currentFile);
```
albert-github: The default setting for `lookForSpecializations` is here `true`, why is there `false` here now?

src/docnode.cpp
nocnokneo: You are right, my helper was inserted between that comment and the function it describes. Moved it above the comment so `unescapeCRef` keeps its own description, and fixed the `&gt;` that should have been `&lt;` while touching those lines.

src/util.cpp
nocnokneo: That was an accident on my part, thanks. The original call relied on the default, so passing `false` silently changed behaviour; it now passes `true` to match. Re-ran the suite afterwards: 124/125, and the resolution results are unchanged (same-file `@ref` resolves, your cross-file case still does not).

nocnokneo:
@albert-github I've updated this to properly handle \ref to file-scoped statics _ony in the same file_. PTAL

